PR #10, opened 2023-10-31
C++ SD v1.5 pipeline

yangsu2022:


yury-gorbachev:
Could we move this to different path?
@Wovchena , what would be correct format? ImageGeneration/StableDiffusion1_5?

Wovchena:
Following the style proposed in https://github.com/openvinotoolkit/openvino.genai/pull/9 it should be `image_generation/stable_diffusion_1_5/cpp`

yangsu2022:
updated with "image_generation/stable_diffusion_1_5/cpp" and use new trapezoidal implementation without Boost

wenjiew:
@xipingyan to take a look at the code and provide comments if there is any. Thanks!

xipingyan:
It is a good example. It contains model convert, compress, independent C++ reference codes.

PR #19, opened 2023-11-21
[DO_NOT_MERGE] Indirect cache transformation

slyalin:
Transformation of a model to introduce the indirect cache which avoids permutation of the cache tensors when beam search is used. Instead, kv-tensors are permuted just before ScaledDotProductAttention operation, which should be a target for further plugin optimizations. Assumption is that the plugins will apply Gather in a smart way.

Transformation is enabled by new --indirect_cache key and compatible with stateless and stateful models (can be applied in combination with --make_stateful). The key is enabled only when --fuse_cache_reorder is set and --num_beams > 1.

A new input/output pair appears: beam_table_input/beam_table_output that is hidden as a state when --make_stateful is applied. Necessary operations for beam_table update is fused into the model regardless of state-less/full property.

### DIFF
--- a/llm_bench/python/benchmark.py
+++ b/llm_bench/python/benchmark.py
@@ -430,6 +430,14 @@ def get_argprser():
         help='Fuse ops related to cache reordering to the model, applied only when num_beams > 1',
     )
     parser.add_argument(
+        '--indirect_cache',
+        action='store_true',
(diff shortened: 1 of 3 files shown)


eaidova:
@slyalin Is this PR still applicable or can be closed?

PR #22, opened 2023-11-22
use statswith to match model name

wgzintel:


peterchen-intel (approved):

peterchen-intel:
@eaidova 
This change is to avoid our engineers whose file path has "yi" substring.

eaidova (approved):

PR #45, opened 2023-11-30
[Stateful]set correct batch for stateful chatglm2/3

zhangYiIntel:


peterchen-intel (approved):

slyalin (approved):

eaidova (approved):

llm_bench/python/utils/ov_utils.py
```diff
@@ -211,9 +211,10 @@ def _reorder_cache_stub(self, past_key_values, beam_idx):
             input = ov_model.input(kv_name_pair[0])
             shape = input.get_partial_shape()
 
-            # suppose 0-th dimension is a batch
+            # By default, batch is the 0-th but chatglm uses 1-st dimension as batch
             # TODO: Deduce from a model via ordinal reshape
-            shape[0] = batch_size * num_attention_heads * num_beams
+            batch_idx = 1 if hf_model.config.model_type == 'chatglm' else 0
```
slyalin, line 216: We need similar fix in case when `fuse_cache_reorder` is used. It is necessary fix run beam-search on stateful models for this case. Please similarly change 0 to 1 in this line: https://github.com/openvinotoolkit/openvino.genai/blob/16e7d148d41eee8ebb30b4463e23fd25f8f54a14/llm_bench/python/utils/ov_utils.py#L176. @zhangYiIntel, could you provide another PR to do that?

zhangYiIntel: #47 is created for beam-search case.

PR #51, opened 2023-12-01
Use revision for qwen

eaidova:


llm_bench/python/convert.py
```diff
@@ -1408,8 +1408,9 @@ class Baichaun2OpenVINOConfig(TextDecoderOnnxConfig):
 def convert_qwen(args):
     config = AutoConfig.from_pretrained(args.model_id, trust_remote_code=True)
     cuda, post_init = patch_gptq(config)
+    revision = "c02ede58c0ab0045f5e4788c35842bec6a7baa0a" if post_init is not None else "2abd8e5777bb4ce9c8ab4be7dbbd0fe4526db78d"
```
peterchen-intel, line 1411: Is it better to give an warning message?

eaidova: I do not understand which warning do you mean?

PR #52, opened 2023-12-01
Move stateful model preparation to optimum-intel

slyalin:
- Stateful model preparation code is moved to optimum-intel (https://github.com/huggingface/optimum-intel/pull/486).
- Bettertransformer is applied together with stateful transformation on optimum-intel side.
- state initializer that we discussed with CPU/GPU teams also implemented on optimim-intel side and activated by this PR.
- Provide a single command line argument: --stateful to activate stateful transformation. It combines three earlier used keys: --make_stateful, --bettertransformer and --fuse_cache_reorder. When applied, --stateful leverages capabilities from optimum-intel to transform model in stateful form.

A new key --stateful can be applied with both convert.py and benchmark.py. If it is applied with convert.py, the produced model saved in IR is stateful and with bettertransformer applied to produce SDPAs. If --stateful is applied with benchmark.py enables conversion models to stateful form immediately before compilation even if source IR is stateless. Applying this key in benchmark.py will not introduce SDPAs because we cannot apply bettertransfomrer on already converted IRs.

To get most optimized models you should apply --stateful when converting model with convert.py. Use --stateful with benchmark.py for debugging purposes when SDPA is not critical for the result and you cannot re-convert the model with --stateful.


llm_bench/python/utils/ov_utils.py
```diff
@@ -153,83 +251,98 @@ def greedy_search(input_port):
...
+            def _reorder_cache_stub(self, past_key_values, beam_idx):
+                # TODO: Apply it differently based on model type
+                self.next_beam_idx = np.array(beam_idx)  # save beam_idx to be used as an input in the next iteration
+                return past_key_values
 
-            input.get_node().set_partial_shape(shape)
+            hf_model.use_cache_as_state = False
+            hf_model._reorder_cache = types.MethodType(_reorder_cache_stub, hf_model)
+            hf_model.forward = types.MethodType(forward_simplified, hf_model)  # need custom forward to set beam_idx input to OV model
+            hf_model.next_beam_idx = np.zeros([num_beams * batch_size], dtype=int)  # initial value for beam_idx is all zeros
```
usstq, line 300: this default value has flaw,  for example when num_beams=4 and batch_size=2, next_beam_idx should be `[0,0,0,0,1,1,1,1]`  instead of all zeros

slyalin:
@AlexKoff88, this is PR that refers to a potential PR for optimum-intel: https://github.com/slyalin/optimum-intel/tree/stateful. We still need to migrate some stuff from llm_bench to that branch (part of the code that is intended for optimum-intel still resides in this PR).

peterchen-intel:
https://github.com/openvinotoolkit/openvino.genai/blob/master/llm_bench/python/requirements.txt#L10C30-L10C30 Update the optimal-intel commit ID here.

slyalin:
> https://github.com/openvinotoolkit/openvino.genai/blob/master/llm_bench/python/requirements.txt#L10C30-L10C30 Update the optimal-intel commit ID here.

@peterchen-intel , it is already a part of this PR: https://github.com/openvinotoolkit/openvino.genai/pull/52/files#diff-d681fd3dff6a7436cb98eea5996a541e278ddcb902e047dc9504418d951b0f56R10. Is this the same place? Actually two files are affected: requirements and requirements-dev.

llm_bench/python/requirements.txt
```diff
@@ -7,7 +7,7 @@ torch
 transformers>=4.33.0
 diffusers>=0.22.0
 optimum>=1.14.0,<1.15.0
-git+https://github.com/huggingface/optimum-intel.git@f248835b16ce4ec054d6d4d629dff4213fe94157
+git+https://github.com/slyalin/optimum-intel.git@stateful
```
peterchen-intel, line 10: Is there a PR to merge this temporary branch to optimum-intel?

slyalin: https://github.com/huggingface/optimum-intel/pull/486

.github/workflows/llm_bench-python.yml
```diff
@@ -56,6 +55,19 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Test stateful
+        run: |
+          python -m pip install -r llm_bench/python/requirements.txt optimum==1.14
+          python -m pip uninstall --yes openvino
+          python -m pip install openvino-nightly==2023.3.0.dev20231206
```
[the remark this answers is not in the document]
slyalin: Removed strict version in this PR.

slyalin:
@eaidova, as this branch should be backward compatible with vanilla llm_bench usage (new functionality is enabled only if you provide --stateful), please merge it to master if there are no concerns after code review.

llm_bench/python/utils/model_utils.py
```diff
@@ -49,6 +50,16 @@ def set_default_param_for_ov_config(ov_config):
         ov_config['NUM_STREAMS'] = '1'
 
 
+def add_stateful_model_arguments(parser: argparse.ArgumentParser):
+    parser.add_argument(
+        '--stateful',
```
eaidova, line 55: should we name it --make_stateful for backward compatibility?

.github/workflows/llm_bench-python.yml
```diff
@@ -56,6 +55,19 @@ jobs:
...
+  stateful:
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Test stateful
+        run: |
+          python -m pip install -r llm_bench/python/requirements.txt optimum==1.14
```
eaidova, line 69: why optimum 1.14?

llm_bench/python/utils/model_utils.py
slyalin: Validation will apply necessary changes, I wouldn't build any backward compatibility story for experimental keys here. It is named stateful for simplicity and alignment with optimum-intel.

llm_bench/python/utils/model_utils.py
```diff
@@ -49,6 +50,16 @@ def set_default_param_for_ov_config(ov_config):
...
+        action='store_true',
+        default=None,
```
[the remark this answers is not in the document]
slyalin: No, it will return `None`, if you don't set it. In our case it means "use default". In convert.py it will activate stateless. In benchmark.py will take model as-is, which allows running both stateful and stateless models in benchmark.py without changing and requiring a particular mode.

eaidova (approved):

llm_bench/python/convert.py
```diff
@@ -175,8 +177,8 @@ def convert_optimum_causallm_base(model, args):
     )
     if "decoder_with_past_model" in models_and_onnx_configs:
         models_and_onnx_configs = {"model": models_and_onnx_configs["decoder_with_past_model"]}
-    if args.bettertransformer:
```
eaidova, line 178: now it is not possible to use only bettertransformer without stateful?

slyalin: Correct.

PR #64, opened 2023-12-11
[ChatGLM2/3]Simplify the attention_mask to make stateful model correct

zhangYiIntel:
The original attention mask generation may have an extra connection between past key & attention_mask If node. This connection is not necessary since to generate attention_mask, the only need is the Shape of past-key-value. The past-key-value itself is not needed.


zhangYiIntel:
@eaidova Hi could you help review this change ? Thanks!

eaidova:
@zhangYiIntel please fix code style, control sample accuracy looks good, so modification, can be merged

eaidova (approved):

PR #65, opened 2023-12-11
Add method to find use case from config.json in addition to model name

helena-intel:
This makes finding use cases more robust for cases where:
- the model directory does not start with the model type name. For example, something-t5 instead of t5. 
- the model is a variant on a supported model type. For example: google/flan-ul2 which has a t5 model type.

This makes it more robust to use the benchmark script with an already existing directory of models that do not follow the internal model structure and naming convention. It also prevents having to hardcode model types for all variants (e.g. google/flan-ul2 which is based on t5).

It's an additional method that runs only if the existing filename based method fails.

### DIFF
--- a/llm_bench/python/utils/model_utils.py
+++ b/llm_bench/python/utils/model_utils.py
@@ -66,17 +66,12 @@ def analyze_args(args):
     model_args['torch_compile_backend'] = args.torch_compile_backend
     model_args['convert_tokenizer'] = args.convert_tokenizer
 
-    model_path = args.model
     model_framework = args.framework
-    path = os.path.normpath(model_path)
-    model_names = path.split(os.sep)
-    model_path = Path(model_path)
+    model_path = Path(args.model)
     if not model_path.exists():
         raise RuntimeError(f'==Failure FOUND==: Incorrect model path:{model_path}')
-    if model_framework == 'ov':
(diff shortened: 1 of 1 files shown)


llm_bench/python/utils/model_utils.py
```diff
@@ -95,13 +90,31 @@ def analyze_args(args):
...
+    model_names = path.split(os.sep)
+    for model_name in reversed(model_names):
         for case, model_ids in USE_CASES.items():
             for model_id in model_ids:
                 if model_name.lower().startswith(model_id):
-                    log.info(f'==SUCCESS FOUND==: use_case: {case}, model_name: {model_name}')
+                    log.info(f'==SUCCESS FOUND==: use_case: {case}, model_type: {model_name}')
                     return case, model_name
+
+    # 2. try to get use_case from model config
```
peterchen-intel, line 104: Shall we take this as first choice?

eaidova (approved):

llm_bench/python/utils/model_utils.py
eaidova: I think we can consider that later, there is some possible issues with model versioning, if they stored with remote code and share the same model_type (e.g. we have different classes for chatglm and chatglm2/3, model_type in config the same

PR #91, opened 2023-12-20
Try to print git commit id

wgzintel:


llm_bench/python/benchmark.py
```diff
@@ -40,6 +41,10 @@
 stable_diffusion_hook = StableDiffusionHook()
 
 
+def get_git_revision_hash() -> str:
+    return subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('ascii').strip()
```
peterchen-intel, line 45: When git is not found, or git rev-parse HEAD returns nothing or error, we should do nothing here.
Can we use git hooks to save a version.txt under top folder, and print version.txt here if it exists?

peterchen-intel (approved):

eaidova (approved):

eaidova:
@wgzintel  @peterchen-intel  maybe it is also will be good to introduce versioning into llm_bench? I think it may be more robust way for checking code version instead of commit id (version also can include commit id, as example nncf version use that).  Current solution may detect wrong repo and commit if user run benchmarking from another directory, but which is also under git

peterchen-intel:
@eaidova Do you mean define a file like https://github.com/openvinotoolkit/nncf/blob/release_v280/nncf/version.py#L12C1-L12C12
and adopt https://github.com/openvinotoolkit/nncf/blob/release_v280/setup.py#L77 ?

eaidova:
@peterchen-intel , yes, I think something similar will be fine, we have no setup, so I think we can combine them in one function get_version, and use it for printing llm_bench version

PR #93, opened 2023-12-20
llm_bench add position_ids for Qwen

Wovchena:


llm_bench/python/convert.py
```diff
@@ -1119,8 +1119,6 @@ def convert_aquilachat(args):
     'ldm': convert_ldm_super_res,
     'mpt': convert_mpt,
     'replit': convert_mpt,
-    'chatglm2': convert_causal_lm,
```
eaidova, line 1122: please return back!, it was specially added before chatglm for correct matching

Wovchena: Returned

eaidova (approved):

PR #95, opened 2023-12-21
fix aquilachat and codegen2 after rebasing

eaidova:


llm_bench/python/convert.py
```diff
@@ -1076,15 +1076,15 @@ def convert_codegen2(args):
     config = AutoConfig.from_pretrained(args.model_id, trust_remote_code=True)
     if config.model_type == "codegen":
         config.model_type = "codegen2"
-
+    print(config)
```
eaidova, line 1079:
```suggestion
    print(config)
```

PR #113, opened 2023-12-25
causal_lm: set state max batch

Wovchena:
Ticket 128111


text_generation/causal_lm/cpp/beam_search_causal_lm.cpp
```diff
@@ -22,6 +22,14 @@ std::string detokenize(ov::InferRequest& detokenizer, const std::vector<int64_t>
...
     return detokenizer.get_output_tensor().data<std::string>()[0];
 }
+
+void set_state_max_batch(std::vector<ov::VariableState>&& states, size_t batch_size) {
+    for (ov::VariableState& state : states) {
+        ov::Shape shape = state.get_state().get_shape();
+        shape.front() = batch_size;
+        state.set_state(ov::Tensor{ov::element::f32, shape});
+    }
+}
```
slyalin, line 32: Does it mean that reshape doesn't work for the model?

Is this a workaround for broken dynamic batch size?

Wovchena: It's not a full reshape. It's an implementation of suggested workaround to reshape states only, to fix broken dynamic batch size.

You can see in the tests status, that it didn't work out.

PR #116, opened 2023-12-29
Simplify output log

wgzintel:
[simply_log_output.txt](https://github.com/openvinotoolkit/openvino.genai/files/13816865/simply_log_output.txt)
[num_beams_2_log.txt](https://github.com/openvinotoolkit/openvino.genai/files/13816870/num_beams_2_log.txt)


peterchen-intel:
@wgzintel Please keep MD5 output even it is same as warm-up.

peterchen-intel:
@wgzintel Let's update the logs for inference metrics:
Now: [ INFO ] [warm-up] First token infer latency: 134.15 ms/token, other tokens infer latency: 28.17 ms/token, len of tokens: 512
Expect: [ INFO ] [warm-up] First infer latency: 134.15 ms/infer, other infers latency: 28.17 ms/infer, inference count: 512
Please give an output log with --num_beams=2

llm_bench/python/utils/metrics_print.py
```diff
@@ -46,15 +44,29 @@ def print_metrics(
...
-            f"other tokens infer latency: {iter_data['other_tokens_infer_avg_latency']:.2f} ms/token, len of tokens: {len(tms_infer)}",
+            f"[{iter_str}] First infer latency: {iter_data['first_token_infer_latency']:.2f} ms/infer, "
+            f"other infers latency: {iter_data['other_tokens_infer_avg_latency']:.2f} ms/infer, inference count: {len(tms_infer)}",
         )
     if stable_diffusion is not None:
         print_stable_diffusion_infer_latency(iter_str, iter_data, stable_diffusion)
+    output_str = ''
     if max_rss_mem != '' and max_rss_mem > -1:
-        log.info(f'[{iter_str}] max rss memory cost:\n{max_rss_mem}')
+        output_str += 'Max rss memory cost: {:.2f}, '.format(max_rss_mem)
```
peterchen-intel, line 54:
```suggestion
        output_str += 'Max rss memory cost: {:.2f}MBytes, '.format(max_rss_mem)
```

llm_bench/python/utils/metrics_print.py
```diff
@@ -46,15 +44,29 @@ def print_metrics(
...
     if max_shared_mem != '' and max_shared_mem > -1:
-        log.info(f'[{iter_str}] max shared memory cost:\n{max_shared_mem}')
+        output_str += 'max shared memory cost: {:.2f}'.format(max_shared_mem)
```
peterchen-intel, line 56:
```suggestion
        output_str += 'max shared memory cost: {:.2f}MBytes'.format(max_shared_mem)
```

peterchen-intel (approved):

eaidova:
@wgzintel  @peterchen-intel what do you think about ability to save model output into file? for log sequence len, it may be more convinient for comparing output text between runs having them in the file instead of parsing in command line. I think we can make flag for that or stay some limit for sequence len starting form which we will automatically save output result in file

eaidova (approved):

peterchen-intel:
> @wgzintel @peterchen-intel what do you think about ability to save model output into file? for long sequence len, it may be more convinient for comparing output text between runs having them in the file instead of parsing in command line. I think we can make flag for that with destination path or stay some limit for sequence len starting form which we will automatically save output result in file or even do it always

How about add a --output_dir option, if it is provided, then the generated text and images will save to that directory. with <model name>_<precision>_p<prompt index>_<iteration>.<txt/jpeg/png>

peterchen-intel:
Follow up CVS-129192

PR #120, opened 2024-01-03
[GHA] Add label to PRs

peterchen-intel:
Add llm_bench label to PRs changing files under llm_bench/ folder, update .github/label_config.yml to add new adding label rules/configurations


peterchen-intel:
Verified with https://github.com/openvinotoolkit/openvino.genai/pull/121
Use pull_request_target event to avoid permission issue.

eaidova (approved):

PR #124, opened 2024-01-03
Call reset in the end of text generation in causal_lm sample

Wovchena:
Ticket 128656


text_generation/causal_lm/cpp/beam_search_causal_lm.cpp
```diff
@@ -76,6 +76,11 @@ int main(int argc, char* argv[]) try {
             std::cout << beam.score << ": " << detokenize(detokenizer, beam.tokens) << '\n';
         }
     }
+    // While it is not required to call reset function for InferRequest object if only one sequence is processed, it is
+    // useful for education purpose. In case if the user really going to process multiple sequences, it is required to
+    // call reset function.
+    // Note that this is not required in this particular sample scenario (but harmless anyway)
```
slyalin, line 82:
```suggestion
    // Model is stateful which means that context (kv-cache) which belongs to a particular text sequence is accumulated inside the model during the generation loop above.
    // This context should be reset before processing the next text sequence.
    // While it is not required to reset context in this sample as only one sequence is processed, it is called for education purposes:
```

Wovchena: Applied

ilya-lavrenov (approved):

PR #156, opened 2024-01-17
move optimum intel back on original repo

eaidova:


llm_bench/python/requirements.txt
```diff
@@ -7,7 +7,7 @@ torch
 transformers>=4.33.0
 diffusers>=0.22.0
 optimum>=1.14.0,<1.15.0
```
eaidova, line 9:
```suggestion
optimum>=1.14.0,<1.17.0
```

PR #182, opened 2024-01-24
Finalize releases/2023/3

Wovchena:


thirdparty/openvino_contrib
Wovchena, line 1: FYI, this isn't the last openvino_contrib commit, but the one which is going to be used for publication

ilya-lavrenov (approved):

PR #188, opened 2024-01-25
Describe stateful model

Wovchena:


slyalin:
@Wovchena, please take the source or a picture from @itikhono that was used in https://github.com/openvinotoolkit/openvino/pull/22410. Adopt it to LLMs text generation by adding dedicated inputs and outputs. Prepare both stateless and stateful variants just to illustrate the difference between two sorts of the model.

Use something like this for stateless model:
image

And similar to this for stateful models with beam_idx:
image

slyalin:
Probably these pictures can be inserted as they are in this PR.

text_generation/causal_lm/cpp/README.md
```diff
@@ -4,6 +4,26 @@ These examples showcase inference of text-generation Large Language Models (LLMs
 
 ## How it works
 
+### Stateful LLM
+
+The examples infer a stateful model. The stateful model keeps internal state which is updated after every inference. That state is used as additional implicit model inputs. It is opposed to stateless model which requires passing state tensors explicitly. A more detailed introduction to stateful models can be found in https://github.com/openvinotoolkit/openvino/pull/22410. TODO: update link
+
+The intermediate inference results are usually stored in KV (key value) cache in case of LLMs. This cache is represented by the corresponding inputs and outputs. Stateless model approach requires manipulating these outputs to use them as inputs for the next inference. Stateful models remove that burden and keep KV cache internally. Although the cache is still accessible with state API.
```
slyalin, line 11:
```suggestion
The intermediate inference results are usually stored in KV (key/value) cache in case of LLMs. This cache is represented by the corresponding inputs and outputs in model implemented originally in DL framework (e.g. PyTorch models from HuggingFace). Stateless model approach requires manipulating these outputs to use them as inputs for the next inference. Stateful models remove that burden and keep KV cache internally. Although the cache is still accessible with state API.
```

Also fuse with previous paragraph to avoid an extra LLM-independent education on stateful models. Start with LLMs specific first to give the domain context, and then go to stateful terms and compare it with stateless.

Try to borrow a few sentences from https://docs.openvino.ai/2023.3/gen_ai_guide.html

ilya-lavrenov (approved):

text_generation/causal_lm/cpp/README.md
```diff
@@ -4,6 +4,26 @@ These examples showcase inference of text-generation Large Language Models (LLMs
...
+
+The images below represents stateless and stateful LLM pipelines. The model has 4 inputs:
+1. `input_ids` contains the next selected token
+2. `attention_mask` is filled with `1`
+3. `position_ids` encodes a position of currently generating token in the sequence
+4. `beam_idx` selects beams
+
+There's also an arbitrary number of implicit inputs grouped into KV cache state. The model has 1 output `logits` describing the predicted distribution over the next tokens and implicit outputs grouped into KV cache state.
```
slyalin, line 23: Please do not use `implicit inputs`, you have already described everything about stateful model, and this part is called `state`.

text_generation/causal_lm/cpp/README.md
```diff
@@ -4,6 +4,26 @@ These examples showcase inference of text-generation Large Language Models (LLMs
...
+
+To provide a possibility to implement beam search without accessing model internal state, a stateful LLM converted with `optimum-intel` or [llm_bench](../../../llm_bench/python/) introduces additional 1-dimentional `beam_idx` input. `beam_idx` must contain indices of elements in a batch which are supposed to be selected and evolve during next beam search iteration. Suppose there are two running beams. To proceed generating both beams at the next iteration, `beam_idx` values must be `[0, 1]`, pointing to batch elements `0` and `1`. To drop the last beam and split the other beam in two, `beam_idx` must be set to `[0, 0]`, this results in utilizing only the part of KV cache corresponding to zeroth element in the batch. The process of selecting appropriate entries in cache is called Cache Reorder.
```
slyalin, line 15: Provide a picture that illustrates how beam_idx fork sequence and removes sequence. Do it in a separate PR please.

text_generation/causal_lm/cpp/README.md
Wovchena: Done

text_generation/causal_lm/cpp/README.md
Wovchena: Removed

PR #190, opened 2024-01-26
Fix memory issue in TextStreamer

sammysun0711:
This PR aim to fix potential memory issue in TextStreamer used in `greedy_causal_lm.cpp`. 

Here is problem description:
https://github.com/openvinotoolkit/openvino.genai/blob/c3c3173715a578ab2a49ecd58de1f45c48bdcfd8/text_generation/causal_lm/cpp/greedy_causal_lm.cpp#L30-L45

During the text generation, for each text_streamer{std::move(detokenizer)} call, token_cache, text.size() and print_len will increased by 1. 

However when `\n` generated, token_cache will be clear, print_len=0 inside if condition, but print_len will be override with `text.size()` again (which is some value >0, depends on generated token_cache, say 120).
https://github.com/openvinotoolkit/openvino.genai/blob/c3c3173715a578ab2a49ecd58de1f45c48bdcfd8/text_generation/causal_lm/cpp/greedy_causal_lm.cpp#L44
Then for next token, we have len(token_cache)=1, text.size()=1, but print_length=120, then program hang in following line because invalid memory access:
https://github.com/openvinotoolkit/openvino.genai/blob/c3c3173715a578ab2a49ecd58de1f45c48bdcfd8/text_generation/causal_lm/cpp/greedy_causal_lm.cpp#L43

This PR add return in if condition to avoid unexpected print_len assignment.


sammysun0711:
@Wovchena, could you please have a check?

Wovchena:
Thank you. Please retarget to `releases/2023/3`. I will merge `releases/2023/3` into `master` later. Or you can submit a separate PR for `releases/2023/3`

sammysun0711:
Re-target to releases/2023/3

.github/dependabot.yml
Wovchena, line 1: You shouldn't bring commits from master

Wovchena (changes requested):

Wovchena:
I'll merge https://github.com/openvinotoolkit/openvino.genai/pull/201 instead

PR #202, opened 2024-01-31
fix image file conflict when there are multi instance

tiger100256-hu:
while running stable diffusion model and there are multi instance at same time, will cause created image file conflict


tiger100256-hu:
fixed by https://github.com/openvinotoolkit/openvino.genai/pull/238, close this PR

PR #222, opened 2024-02-13
image generation windows ci tests

likholat:


image_generation/stable_diffusion_1_5/cpp/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ if(POLICY CMP0135)
     cmake_policy(SET CMP0135 NEW)
 endif()
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
```
Wovchena, line 12: Do I understand correctly that C++17 fails on windows but C++20 passes?

likholat: Yes

Wovchena (approved):

PR #226, opened 2024-02-14
Added int4 configs when num channels non divisible by default group size

ljaljushkin:
3 models has the number of channels non-divisible by default group size=128 and weight compression fails with error:
`nncf.errors.ValidationError: Channel size 8640 should be divisible by size of group 128 `
Found int4 configs within 0.2 similarity drop on WWB.

ref ticket: 132035


ljaljushkin:
@skuros @eaidova

skuros (approved):

eaidova (approved):

PR #239, opened 2024-02-22
add option --interleave

wgzintel:
if have option --interleave, input prompts are processed in interleave manner, else input prompts are processed in subsequent manner.

### DIFF
--- a/llm_bench/python/benchmark.py
+++ b/llm_bench/python/benchmark.py
@@ -472,7 +472,7 @@ def get_argprser():
         '--interleave',
         action='store_true',
         help='if the value is True, input prompts are processed in interleave manner'
-        'if the value is False, input prompts are processed in subsequent manner'
+        'if the value is False (default), input prompts are processed in subsequent manner'
     )
     utils.model_utils.add_stateful_model_arguments(parser)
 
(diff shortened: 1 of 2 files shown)


llm_bench/python/benchmark.py
```diff
@@ -453,6 +468,12 @@ def get_argprser():
     parser.add_argument(
         '--convert_tokenizer', action='store_true', help='Convert tokenizer to OpenVINO format'
     )
+    parser.add_argument(
```
peterchen-intel, line 471: Shall we add a default=False?

wgzintel: We add action='store_true' means if ‘--interleave‘ is given on the command line, its value is True, otherwise it is False. 
Reference --convert_tokenizer.

peterchen-intel (approved):

llm_bench/python/benchmark.py
```diff
@@ -453,6 +468,12 @@ def get_argprser():
...
+        '--interleave',
+        action='store_true',
+        help='if the value is True, input prompts are processed in interleave manner'
+        'if the value is False, input prompts are processed in subsequent manner'
```
peterchen-intel, line 475:
```suggestion
        'if the value is False (default), input prompts are processed in subsequent manner'
```

peterchen-intel (approved):

PR #253, opened 2024-02-28
Added prototype of data aware compression.

andreyanufr:
Data-aware compression with updated configurations.


eaidova:
@andreyanufr generally looks good, could you please also consider adding command line options for enabling data-aware quantization from user side or it will be handled separately?

eaidova (approved):

PR #256, opened 2024-02-29
Fix lcm scheduler randn_function 

yangsu2022:
**Issue:**
The C++ LCM pipeline get worse generation with C++ std lib, while SD pipeline has tolerated generation.

**Analysis:**
- The pipeline got same value of the randn_function output "noise', while the value of 3 torch_noise_step files are different. 
- In the python pipeline, the torch.randn() is used with default global generator, which is set via torch.manual_seed().
- LCM has the same initial random number generator(RNG) function with SD, so the root cause should not be initial RNG or std::mt19937 of RNG, but the RNG in the randn_function of LCM scheduler
- To fix it, here set the generator and normal distribution as the member variables in the class lcmscheduler instead of creating them repeatedly in each step function.

**Results:** 
Now the reproducible generation result with seed 42 is at right (left: read noise, mid: old not read noise, right: now not read noise)
`Prompt "beautiful pink unicorn"`
image
`Prompt "cyberpunk cityscape like Tokyo New York  with tall buildings at dusk golden hour cinematic lighting"`
image


image_generation/lcm_dreamshaper_v7/cpp/cpp_random.bmp
Wovchena, line 1: Either remove it or add a link from README describing it

Wovchena (dismissed):

image_generation/lcm_dreamshaper_v7/cpp/cpp_random.bmp
yangsu2022: [README](https://github.com/openvinotoolkit/openvino.genai/blob/master/image_generation/lcm_dreamshaper_v7/cpp/README.md?plain=1#L101) has already the image link with description.
`* Generate image with soulcard lora and C++ generated latent and noise ./stable_diffusion -r -l path/to/soulcard.safetensors`
`image`
Here I just replace the image with new one.

Wovchena: You didn't replace it. You add a new one

yangsu2022: When see "BIN +0 Bytes (100%)" in the "Files changed" section of GitHub pull request (PR), it indicates that an image file has been modified but the actual size of the file remains the same? 
My local folder just has one image_generation/lcm_dreamshaper_v7/cpp/cpp_random.bmp.

yangsu2022:
The image generation would be used in the next week's CVPR, please help to review, thanks.

ilya-lavrenov (approved):

Issue #218, opened 2024-02-12
Migrate DEFAULT 4-bit weight quantization of LLMs to optimum configs


AlexKoff88:
- As we merged support of default 4-bit configs in Optimum-Intel, we need to adjust llm.bench accordingly.
- All the custom model configs should be disabled in llm.bench as well, but we can keep the ability to customize the set of Optimum configs with in llm.bench

PR #283, opened 2024-03-05
migrate on new optimum-intel version

eaidova:
- [x] rely on optimum configs for export
- [x] rely on optimum configs for weights compression - https://github.com/openvinotoolkit/openvino.genai/issues/218


llm_bench/python/requirements.txt
```diff
@@ -1,14 +1,13 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
-wheel
 numpy
-openvino>=2023.3.0
+openvino>=2023.1.0
```
eaidova, line 3:
```suggestion
openvino>=2023.3.0
```

Issue #142, opened 2024-01-11
Add speculative decoding example


AlexKoff88:
Speculative decoding is a very popular method to speed up text generation significantly. Moreover, it is already being adopted by industry. 
Can you have such an example for text generation? For example, it could be Llama2-7B + TinyLlama.

pavel-esir:
ticket for this issue is assigned for me. @andrei-kochin please assign this issue to me as well?

PR #288, opened 2024-03-06
Add speculative decoding

pavel-esir:
CVS-132865
Closes https://github.com/openvinotoolkit/openvino.genai/issues/142


text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+    lm.set_tensor("input_ids", input_ids);
+    lm.set_tensor("attention_mask", attention_mask);
+    ov::Tensor position_ids = lm.get_tensor("position_ids");
+    position_ids.set_shape(input_ids.get_shape());
+    std::iota(position_ids.data<int64_t>(), position_ids.data<int64_t>() + position_ids.get_size(), 0);
+    init_key_values(lm, draft_kv_size);
+    lm.infer();
+
+    // target
+    ov::InferRequest lm_target = core.compile_model(
```
Wovchena, line 238: Using TinyLlama-1.1B-Chat-v1.0 for both draft and target also results in divergence. I wasn't able to identify the problem, but maybe it will be easier to debug it this way

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+    cerr << text << " | " << ss.str() << endl;
+}
+
+string tokens_to_string(ov::InferRequest& detokenizer, std::vector<int64_t> tokens) {
+    return detokenize(detokenizer, tokens);
+}
+
+int main(int argc, char* argv[]) try {
+    int draft_kv_size = 22;
+    int target_kv_size = 32;
```
ilya-lavrenov, line 197: we need to extract if from model, take number of inputs, subtract 3 (input_ids, position_ids and mask) and divide by 2

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+    auto old_seq_len = shape[2];
+    auto size_2 = shape[3];
+    
+    // todo: assume batch is always 1, if different one more cycle is needed
+    // copy elements from the old to the new tensor
+    auto new_tensor = ov::Tensor{ov::element::f32, {BATCH_SIZE, size_1, new_seq_len, size_2}};
+    auto new_tensor_data = new_tensor.data<float>();
+    for (int i = 0; i < size_1; i++) {
+        for (int j = 0; j < new_seq_len; j++) {
+            for (int k = 0; k < size_2; k++) {
```
ilya-lavrenov, line 152: we can use std::memcpy instead of inner loop

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+    
+    std::vector<float> probabilities;
+    for (int i = 0; i < size; i++) {
+        double probability = exp(logits_data[i] / temperature) / sum_exp;
+        probabilities.push_back(probability);
+    }
+    return probabilities;
+}
+
+int random_sample(const ov::Tensor& logits, float temperature) {
```
ilya-lavrenov, line 42: is it going to be used as next improvement?

pavel-esir: it will not be used, i will remove it

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: sure, memcpy should be better. Also thought about that

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: sure. This was a draft solution. I will generalize for ready for review version

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+    for (int i = 0; i << tensor.get_shape().back(); i++) {
+        res_vector.emplace_back(tensor.data<int64_t>()[i]);
+    }
+    return res_vector;
+}
+
+ov::Tensor append_element(ov::Tensor tensor_val, int64_t element) {
+    auto vec = convert_to_vector(tensor_val);
+    vec.emplace_back(element);
+    return ov::Tensor{ov::element::i64, {BATCH_SIZE, tensor_val.get_shape().back()}, vec.data()};
```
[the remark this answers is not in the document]
pavel-esir: yes, you are right. I figured that out by myself, this code i just leftover. I will remove it

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+        std::string{argv[1]} + "/openvino_tokenizer.xml", "CPU").create_infer_request();
+    auto [input_ids, attention_mask] = tokenize(tokenizer, argv[3]);
+    ov::InferRequest detokenizer = core.compile_model(
+        std::string{argv[1]} + "/openvino_detokenizer.xml", "CPU").create_infer_request();
+    // todo: for debug reason add one more detokenizer since the original one is moved to TextStreamer
+    ov::InferRequest detokenizer_2 = core.compile_model(
+        std::string{argv[1]} + "/openvino_detokenizer.xml", "CPU").create_infer_request();
+    
+    // draft model
+    ov::InferRequest lm = core.compile_model(
```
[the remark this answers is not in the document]
pavel-esir: sure, `main_model` sounds more adequate than `target_model`

text_generation/causal_lm/cpp/speculative_decoding.cpp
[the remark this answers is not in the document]
pavel-esir: removed this obsolete code

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,379 @@
...
+        ss_2 << "past_key_values." << i << ".key";
+        request.set_tensor(ss_2.str(), request.get_tensor(ss_1.str()));
+        ss_1.str("");
+        ss_2.str("");
+
+        ss_1 << "present." << i << ".value";
+        ss_2 << "past_key_values." << i << ".value";
+        request.set_tensor(ss_2.str(), request.get_tensor(ss_1.str()));
+        ss_1.str("");
+        ss_2.str("");
```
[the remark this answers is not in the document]
pavel-esir: applied, now i use port numbers instead of names

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: done. I use `inputs().size()`/`outputs().size()` instead of hardcoded values.

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: done. In this place I will also add one more cycle for batched, just to be sure.

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: removed

text_generation/causal_lm/cpp/speculative_decoding.cpp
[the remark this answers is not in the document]
pavel-esir: done, renamed

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: i find where divergence occurs, and fixed. Now speculative sampling and greedy causal return exactly the same values with 1.5x performance increase.

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+            logits = draft_model.get_tensor("logits").data<float>();
+            int64_t arg_max_token = std::max_element(logits, logits + vocab_size) - logits;
+            out_token = arg_max_token;
+            accum_draft_tokens.emplace_back(arg_max_token);
+        }
+        auto res_str_draft = tokens_to_string(detokenizer_2, accum_draft_tokens);
+
+        // main
+        main_input_ids.set_shape({BATCH_SIZE, K});
+        // main network will give also K out tokens
```
AlexKoff88, line 229:
```suggestion
        // main model will give also K out tokens
```

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+        // feed the same first token to the main network and do not give the last token generated by draft
```
AlexKoff88, line 230:
```suggestion
        // feed the same first token to the main model and skip the last token generated by draft
```

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+    ov::InferRequest draft_model = core.compile_model(std::string{argv[1]} + "/openvino_model.xml", "CPU").create_infer_request();
+
+    draft_model.set_tensor("input_ids", input_ids);
+    draft_model.set_tensor("attention_mask", attention_mask);
+    ov::Tensor position_ids = draft_model.get_tensor("position_ids");
+    position_ids.set_shape(input_ids.get_shape());
+    std::iota(position_ids.data<int64_t>(), position_ids.data<int64_t>() + position_ids.get_size(), 0);
+    init_key_values(draft_model);
+    draft_model.infer();
+
```
AlexKoff88, line 146: I also propose running the cpp linter for the code and autotuning the style.

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+    // Compile models
+    ov::Core core;
+    core.add_extension(OPENVINO_TOKENIZERS_PATH);  // OPENVINO_TOKENIZERS_PATH is defined in CMakeLists.txt
+    // tokenizer and detokenizer work on CPU only
+    ov::InferRequest tokenizer = core.compile_model(
+        std::string{argv[1]} + "/openvino_tokenizer.xml", "CPU").create_infer_request();
+    auto [input_ids, attention_mask] = tokenize(tokenizer, argv[3]);
+    ov::InferRequest detokenizer = core.compile_model(
+        std::string{argv[1]} + "/openvino_detokenizer.xml", "CPU").create_infer_request();
+    // todo: for debug reason add one more detokenizer since the original one is moved to TextStreamer
```
ilya-lavrenov, line 132: can we avoid `std::move` and keep a single detokenizer?

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+    // forwards present key/values from Results to Parameters with past key/values
+    // the first inputs are: input_ids, attention_masks, position_ids, other ones are past key/values.
+    // the first output is logits, others are present key.value
+    auto num_inputs = request.get_compiled_model().inputs().size();
+    for (size_t idx = 3; idx < num_inputs; ++idx) {
+        request.set_input_tensor(idx, request.get_output_tensor(idx - 2));
+    }
+}
+
+void init_key_values(ov::InferRequest request) {
```
ilya-lavrenov, line 71: do we need to set initial shapes? why are shapes in IR?

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+ov::Tensor trimm_tensor(ov::Tensor tensor, uint64_t new_seq_len) {
+    // copy elements from the old to the new tensor
+    // tensor shape = [batch, size_1, seq_len, size_2]
+    auto old_tensor_data = tensor.data<float>();
+    auto shape = tensor.get_shape();
+    size_t size_1 = shape[1];
+    size_t old_seq_len = shape[2];
```
ilya-lavrenov, line 86: let's also try to support cases when seq_len is first dimension

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+    init_key_values(draft_model);
+    draft_model.infer();
+
+    // main
+    ov::InferRequest main_model = core.compile_model(
+    std::string{argv[2]} + "/openvino_model.xml", "CPU").create_infer_request();
+
+    auto main_input_ids = main_model.get_tensor("input_ids");
+    main_input_ids.set_shape(input_ids.get_shape());
+    for (int i = 0; i < input_ids.get_shape()[1]; i++) {
```
ilya-lavrenov, line 153: let's introduce `seq_len_dim_idx` instead of hardcoded `1`

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: on batch axis and seq_len in the IR shapes are -1, if we don't initialize them with concrete values we will and up having error.
```
Shape inference input shapes {{0,4,0,64},{1,4,6,64}}
Argument shapes are inconsistent;
```

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: Done. I was unable to verify on chatglm since it failed on intel-optimum during conversion. But it should work

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: done, removed a couple of warnings.

text_generation/causal_lm/cpp/speculative_decoding.cpp
```diff
@@ -0,0 +1,300 @@
...
+    // draft model
+    ov::InferRequest draft_model = core.compile_model(std::string{argv[1]} + "/openvino_model.xml", "CPU").create_infer_request();
+
+    draft_model.set_tensor("input_ids", input_ids);
+    draft_model.set_tensor("attention_mask", attention_mask);
+    ov::Tensor position_ids = draft_model.get_tensor("position_ids");
+    position_ids.set_shape(input_ids.get_shape());
+    std::iota(position_ids.data<int64_t>(), position_ids.data<int64_t>() + position_ids.get_size(), 0);
+    init_key_values(draft_model);
+    draft_model.infer();
```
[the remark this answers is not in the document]
pavel-esir: Totally agree. Provided more detailed description in ready for review version. Please take a look

text_generation/causal_lm/cpp/speculative_decoding.cpp
pavel-esir: i got rid of such constructions, and now i call `...get_shape()[1]` only once, no need to introduce new variable

text_generation/causal_lm/cpp/speculative_decoding.cpp
ilya-lavrenov: you can use llm_bench to convert the model

text_generation/causal_lm/cpp/README.md
```diff
@@ -36,6 +36,18 @@ The program loads a tokenizer, a detokenizer and a model (`.xml` and `.bin`) to
 
 The program loads a tokenizer, a detokenizer and a model (`.xml` and `.bin`) to OpenVINO. A prompt is tokenized and passed to the model. The model predicts a distribution over the next tokens and group beam search samples from that distribution to explore possible sequesnses. The result is converted to chars and printed.
 
+### speculative_sampling_lm
```
ilya-lavrenov, line 39: we can also provide a reference to https://huggingface.co/blog/assisted-generation as reference article

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+    // tokenizer and detokenizer work on CPU only
+    ov::InferRequest tokenizer = core.compile_model(
+        std::string{argv[1]} + "/openvino_tokenizer.xml", "CPU").create_infer_request();
+    auto [draft_input_ids, draft_attention_mask] = tokenize(tokenizer, argv[3]);
+    ov::InferRequest detokenizer = core.compile_model(
+        std::string{argv[1]} + "/openvino_detokenizer.xml", "CPU").create_infer_request();
+    TextStreamer text_streamer{std::move(detokenizer)};
+
+    // draft model
+    ov::InferRequest draft_model = core.compile_model(std::string{argv[1]} + "/openvino_model.xml", "CPU").create_infer_request();
```
ilya-lavrenov, line 150: @AlexKoff88 in HG it's called assistant model (see https://huggingface.co/blog/assisted-generation#greedy-decoding-with-assisted-generation)
is `draft_model` our invention or common name?

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+        draft_tokens.clear();
+        first_token = next_token;
+    }
+    text_streamer.end();
+    // Model is stateful which means that context (kv-cache) which belongs to a particular
+    // text sequence is accumulated inside the model during the generation loop above.
+    // This context should be reset before processing the next text sequence.
+    // While it is not required to reset context in this sample as only one sequence is processed,
+    // it is called for education purposes:
+    draft_model.reset_state();
```
ilya-lavrenov, line 287: looks like it's not required, because we don't work with stateful model

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <openvino/openvino.hpp>
+#include <cmath>
+#include <random>
+
+constexpr size_t BATCH_SIZE = 1;
+
+// sequence length axis in key/values tensors, for most cases [BATCH_SIZE, size_1, seq_len, size_2], 
+// threfore usually SEQ_LEN_AXIS = 2
```
ilya-lavrenov, line 11:
```suggestion
// therefore usually SEQ_LEN_AXIS = 2
```

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+ov::Tensor trimm_tensor(ov::Tensor& tensor, uint64_t seq_len_axis, uint64_t new_seq_len) {
+    // Copy elements from the old to a new tensor and return it.
+    // It's assumed that key/values tensor has a shape [BATCH_SIZE, size_1, seq_len, size_2] or [seq_len, ...],
+    // It that's not the case for your model please implement your own trim method.
+    OPENVINO_ASSERT(seq_len_axis == 2 || seq_len_axis == 0, "Cannot trim key/values with sequence length axis = ", seq_len_axis);
+    
+    auto old_tensor_data = tensor.data<float>();
+    auto shape = tensor.get_shape();
+    size_t size_1 = shape[1];
```
ilya-lavrenov, line 93:
```suggestion
    size_t num_kv_heads = shape[1];
```

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+    size_t old_seq_len = shape[2];
+    size_t size_2 = shape[3];
```
ilya-lavrenov, line 95:
```suggestion
    size_t head_size = shape[3];
```

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    // main model
+    ov::InferRequest main_model = core.compile_model(std::string{argv[2]} + "/openvino_model.xml", "CPU").create_infer_request();
+
+    // Input tensors for the main model should not be mixed with draft.
+    // Do not feed the same draft_postion_ids to the main, but copy input_ids from the draft_input_ids
+    auto input_ids = main_model.get_tensor("input_ids");
+    input_ids.set_shape(draft_input_ids.get_shape());
+    for (int i = 0; i < seq_len; ++i)
+        input_ids.data<int64_t>()[i] = draft_input_ids.data<int64_t>()[i];
```
ilya-lavrenov, line 168: `memcpy` instead of `for` loop

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+
+    auto position_ids = main_model.get_tensor("position_ids");
+    position_ids.set_shape(draft_input_ids.get_shape());
+    std::iota(position_ids.data<int64_t>(), position_ids.data<int64_t>() + position_ids.get_size(), 0);
+    
+    // To coollect kv-cache for the <PROMPT> and to get the next token run the very first infer request
+    init_key_values(draft_model);
+    init_key_values(main_model);
+    draft_model.infer();
+    main_model.infer();
```
ilya-lavrenov, line 182: do we need to infer main model before main loop?
I suppose first main_model infer can be done for prompt + K  draft tokens

E.g. we can have do / while cycle instead of iteration before loop + do / while

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: in the original paper it's called draft model, so it's not our invention it's a common name
https://arxiv.org/pdf/2302.01318.pdf

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,294 @@
...
+            std::fill_n(draft_attention_mask.data<int64_t>(), draft_attention_mask.get_size(), 1);
+            draft_position_ids.data<int64_t>()[0] = int64_t(draft_attention_mask.get_size() - 1);
+
+            forward_key_values(draft_model);  // forward KV cache from Result outputs to Parameter inputs
+            draft_model.infer();
+
+            auto draft_logits = draft_model.get_tensor("logits").data<float>();
+            int64_t arg_max_token = std::max_element(draft_logits, draft_logits + vocab_size) - draft_logits;
+            next_token = arg_max_token;
+            draft_tokens.emplace_back(arg_max_token);
```
ilya-lavrenov, line 236: should we handle `next_token != SPECIAL_EOS_TOKEN && seq_len < max_sequence_length` conditon here?

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: removed

text_generation/causal_lm/cpp/greedy_causal_lm.cpp
```diff
@@ -91,7 +91,11 @@ int main(int argc, char* argv[]) try {
...
     // There's no way to extract special token values from the detokenizer for now
     constexpr int64_t SPECIAL_EOS_TOKEN = 2;
-    while (out_token != SPECIAL_EOS_TOKEN) {
+    
+    auto seq_len = input_ids.get_size();
+    int max_sequence_length = 100;
+    while (out_token != SPECIAL_EOS_TOKEN && seq_len < max_sequence_length) {
+        ++seq_len;
         lm.get_tensor("input_ids").data<int64_t>()[0] = out_token;
         lm.get_tensor("attention_mask").set_shape({BATCH_SIZE, lm.get_tensor("attention_mask").get_shape().at(1) + 1});
```
ilya-lavrenov, line 100:
```suggestion
        lm.get_tensor("attention_mask").set_shape({BATCH_SIZE, seq_len});
```
and below similar for `position_ids` and `attention_mask` filling

pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,290 @@
...
+    }
+}
+
+void init_key_values(ov::InferRequest request) {
+    auto num_inputs = request.get_compiled_model().inputs().size();
+    for (size_t idx = 3; idx < num_inputs; ++idx) {
+        auto kv_tensor = request.get_input_tensor(idx);
+        ov::Shape tensor_shape = kv_tensor.get_shape();
+        // tensor_shape[2] = 0;
+        // kv_tensor.set_shape(tensor_shape);
```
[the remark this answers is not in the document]
pavel-esir: done

pavel-esir:
Communicated with @ilya-lavrenov. I will edit KV updating for stateful model since stateful is the default optimum-intel scenario

pavel-esir:
stateful version is ready, no drop in performance is observed. Stateful speculative version is still ~1.5x faster than greedy decoding

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,272 @@
...
+        print_len = 0;
+    }
+};
+}
+
+ov::Tensor trimm_tensor(const ov::Tensor& tensor, uint64_t seq_len_axis, uint64_t new_seq_len) {
+    // Copy elements from the old to a new tensor and return it.
+    // It's assumed that key/values tensor has a shape [BATCH_SIZE, num_kv_heads, seq_len, head_size] or [seq_len, ...],
+    // It that's not the case for your model please implement your own trim method.
+    OPENVINO_ASSERT(seq_len_axis == 2 || seq_len_axis == 0, "Cannot trim key/values with sequence length axis = ", seq_len_axis);
```
ilya-lavrenov, line 70: looks like `seq_len_axis == 0` is also not supported, because `update_kv_cache` performs this trimming. 
maybe we can handle both cases in `trimm_tensor` as it was some time ago?

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,272 @@
...
+    
+    auto old_tensor_data = tensor.data<float>();
+    auto shape = tensor.get_shape();
+    size_t num_kv_heads = shape[1];
+    size_t old_seq_len = shape[2];
+    size_t head_size = shape[3];
+    
+    // if new_seq_len equal to old one no need to copy tensor, return as is
+    if (old_seq_len == new_seq_len)
+        return tensor;
```
ilya-lavrenov, line 80: let's also assert that `new_seq_len <= old_seq_len`

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,272 @@
...
+    main_model.get_tensor("beam_idx").data<int32_t>()[0] = 0;
+
+    // To coollect kv-cache for the <PROMPT> and to get the next token run the very first infer request
+    draft_model.infer();
+    main_model.infer();
+
+    size_t vocab_size = draft_model.get_tensor("logits").get_shape().back();
+    OPENVINO_ASSERT(vocab_size == main_model.get_tensor("logits").get_shape().back(), "vocab size should be the same for the both models");
+       
+    // logits shape is [BATCH_SIZE, input_ids.size, vocab_size]
```
ilya-lavrenov, line 174:
```suggestion
    // logits shape is [BATCH_SIZE, seq_len, vocab_size]
```

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
ilya-lavrenov: I don't see conditions inside this loop..

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,272 @@
...
+            draft_model.infer();
+
+            auto draft_logits = draft_model.get_tensor("logits").data<float>();
+            int64_t arg_max_token = std::max_element(draft_logits, draft_logits + vocab_size) - draft_logits;
+            next_token = arg_max_token;
+            draft_tokens.emplace_back(arg_max_token);
+        }
+
+        // For the main network, K tokens will be fed at once in a single infer request.
+        input_ids.set_shape({BATCH_SIZE, K});
```
ilya-lavrenov, line 225: looks like by this time we actually processed `K + 1` tokens (one before loop, K inside while loop)

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,272 @@
...
+        data_logits = logits.data<float>();  // [BATCH_SIZE, K, vocab_size]
+        size_t disagree_idx = K - 1;
+        // Iterate through the predicted tokens from the main model and compare them with draft predictions.
+        // In the worst-case scenario (disagreement at the beginning), iter will increase by 1.
+        // In the best-case scenario, all elements match, and K predicted tokens will be taken.
+        for (size_t i = 0; i < K; i++) {
+            auto start = data_logits + vocab_size * i;
+            auto stop = data_logits + vocab_size * (i + 1);
+            next_token = std::max_element(start, stop) - start;
+            text_streamer.put(next_token);
```
ilya-lavrenov, line 248: should we `put` after `if` below? because current token can be `next_token != draft_tokens[i] `

text_generation/causal_lm/cpp/README.md
pavel-esir: cited this link and original papers

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: done

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: on the draft model yes, when we get here for the first time`K+1` tokens have been processed, but on the target only one.   

But anyways, number of tokens processed do not affect this number, it's defined only by number of tokens we want to process for the future. So what is your concern here?

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: `put` before `if` is the correct way, even if the `next_token != draft_tokens[i]` they both are generated with the same input token and `next_token` contains target model prediction and regardless what is coming next it's always correct. 

If `next_token != draft_tokens[i]` then the token coming after `next_token` will be incorrect. Probably name next is confusing and `out_token` would be better, what do you think?

pavel-esir: renamed to `out_token`

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
pavel-esir: I added such condition below for the target model. For draft model check `SPECIAL_EOS_TOKEN` is not necessary, because it can be incorrect, we should check that on target. 

We could've check for `max_sequence_length` here in draft and save a few infer requests on the draft at the end. But draft infer requests are lightweight. So I opted to add this condition also to the target so that code looks cleaner

text_generation/causal_lm/cpp/README.md
```diff
@@ -36,6 +36,20 @@ The program loads a tokenizer, a detokenizer and a model (`.xml` and `.bin`) to
...
+
+This approach reduces the need for multiple infer requests to the main model, enhancing performance. For instance, in more predictable parts of text generation, the draft model can, in best-case scenarios, generate the next K tokens that exactly match the target. In tha caste the are validated in a single inference request to the main model (which is bigger, more accurate but slower) instead of running K subsequent requests. More details can be found in the original paper https://arxiv.org/pdf/2211.17192.pdf, https://arxiv.org/pdf/2302.01318.pdf
+
+Important note: models should belong to the same familiy and have same tokenizers, and they both should be converted with `--disable-stateful`, e.g.:
+
+```sh
+python3 ../../../llm_bench/python/convert.py --model_id TinyLlama/TinyLlama-1.1B-Chat-v1.0 --output_dir ./TinyLlama-1.1B-Chat-v1.0/ --disable-stateful --precision FP16
```
ilya-lavrenov, line 50:
```suggestion
python3 ../../../llm_bench/python/convert.py --model_id TinyLlama/TinyLlama-1.1B-Chat-v1.0 --output_dir ./TinyLlama-1.1B-Chat-v1.0/ --precision FP16
```

text_generation/causal_lm/cpp/speculative_decoding_lm.cpp
```diff
@@ -0,0 +1,265 @@
...
+    uint64_t seq_len = draft_input_ids.get_shape()[1];
+
+    // main model
+    ov::InferRequest main_model = core.compile_model(std::string{argv[2]} + "/openvino_model.xml", "CPU").create_infer_request();
+
+    // Input tensors for the main model should not be mixed with draft.
+    // Do not feed the same draft_postion_ids to the main, but copy input_ids from the draft_input_ids
+    auto input_ids = main_model.get_tensor("input_ids");
+    input_ids.set_shape(draft_input_ids.get_shape());
+    memcpy(input_ids.data<int64_t>(), draft_input_ids.data<int64_t>(), sizeof(int64_t) * seq_len);
```
ilya-lavrenov, line 146:
```suggestion
    draft_input_ids.copy_to(input_ids);
```

ilya-lavrenov (approved):

.github/workflows/causal_lm_cpp.yml
```diff
@@ -212,6 +212,41 @@ jobs:
...
+          python-version: 3.8
+      - name: Install OpenVINO
+        run: |
+          mkdir ./ov/
+          curl https://storage.openvinotoolkit.org/repositories/openvino/packages/2023.3/linux/l_openvino_toolkit_ubuntu20_2023.3.0.13775.ceeafaf64f3_x86_64.tgz | tar --directory ./ov/ --strip-components 1 -xz
+          sudo ./ov/install_dependencies/install_openvino_dependencies.sh
+      - name: Download, convert and build
+        run: |
+          source ./ov/setupvars.sh
+          python -m pip install --upgrade-strategy eager "optimum>=1.14" -r ./llm_bench/python/requirements.txt "transformers<4.38" ./thirdparty/openvino_tokenizers/[transformers] --extra-index-url https://download.pytorch.org/whl/cpu
```
AlexKoff88, line 232: I believe this test should take a lot of time. We could reduce it by changing models if it is really a bottleneck. @pavel-esir , can you please comment on the execution time?

ilya-lavrenov: image

pavel-esir: i agree that these models are a bit heavy, i tried to replace them with lighter combinations e.g. TinyLlama & Llama-7b but Llama cannot be downloaded without access token so it's unavailable at pre-commit.

I also tried with TinnyLlama and OpenLLaMA but surprisingly they have different tokenizers. So the current combination is the lightest i was able to make work, unfortunately there are no lightest Dollies to use as a draft.

Download & conversion of both models take ~10 m, run ~2 m, it's around twice of the time of usual tests here ~5. But even without speculative testing we also have a test that runs 14 m cpp-beam_search_causal_lm-windows.

So it's comparable and not critical i think

AlexKoff88: Thanks, I am fine with it then

AlexKoff88 (approved):

PR #289, opened 2024-03-07
Force to generate "inference count" tokens

wgzintel:
CVS-133717
1. if has option -ic, output token size is same as infer count
2. if without option -ic, output token size is generated by default according to the model.
3. remove the default output limit of 512 tokens.
4. if set env LOGLEVEL=DEBUG, will print latency of all tokens.

examples:
set env LOGLEVEL=DEBUG
[bloomz-560m-without-ic.txt](https://github.com/openvinotoolkit/openvino.genai/files/15245407/bloomz-560m-without-ic.txt)
[bloomz-560m-ic-1024.txt](https://github.com/openvinotoolkit/openvino.genai/files/15245409/bloomz-560m-ic-1024.txt)
[llama-2-7b-chat-without-ic-.txt](https://github.com/openvinotoolkit/openvino.genai/files/15245412/llama-2-7b-chat-without-ic-.txt)
[llama-2-7b-chat-without-ic-.txt](https://github.com/openvinotoolkit/openvino.genai/files/15245415/llama-2-7b-chat-without-ic-.txt)


peterchen-intel:
@eaidova Looks like the input size has been reset at the position which is end token if min_new_tokens is set to infer_count, it gives a "fake" 2nd token latency. Will investigate it.

eaidova:
@peterchen-intel, could you please investigate issue from your team side, I have no capacity for that and this not related to model conversion, but rather to API and benchmarking code

eaidova:
@wgzintel  @peterchen-intel, please try to remove eos_token_id from generation (it should prevent early stopping due to end of sequence reached)

```
model.generation_config.eos_token_id = None
model.config.eos_token_id = None
model.generate(..., eos_token_id=None)
```

peterchen-intel:
> @eaidova Force to generate 1024 tokens and get even better 2nd token latency (other tokens latency). I suspect the input buffer has been reset if an end token is generated but is forced to continue. Same Input token size: 6 [using_end_token_stopping.txt](https://github.com/openvinotoolkit/openvino.genai/files/14518379/using_end_token_stopping.txt) First token latency: 174.74 ms/token, other tokens latency: 65.44 ms/token, len of tokens: 4 [using_ic_to_generate_tokens(greedy_search).txt](https://github.com/openvinotoolkit/openvino.genai/files/14518380/using_ic_to_generate_tokens.greedy_search.txt) First token latency: 169.89 ms/token, other tokens latency: 33.90 ms/token, len of tokens: 1024

@eaidova @helena-intel My suspicion is wrong, the reason of longer output tokens getting better **average** 2nd token latency is: In warm-up iteration, the 2nd token has significant longer latency than "other 2nd tokens"(3rd and later tokens).
    [ DEBUG ] [0]0.2908
    [ DEBUG ] [1]0.1538
    [ DEBUG ] [2]0.0373
    …
    [ DEBUG ] [1023]0.0390

peterchen-intel (approved):

peterchen-intel (approved):

PR #294, opened 2024-03-08
verified dolly-v2-3b

Vishwa44:
Verified dolly-v2-3b with [greedy_causal_lm.cpp](https://github.com/openvinotoolkit/openvino.genai/blob/master/text_generation/causal_lm/cpp/greedy_causal_lm.cpp) and [beam_search_causal_lm.cpp](https://github.com/openvinotoolkit/openvino.genai/blob/master/text_generation/causal_lm/cpp/beam_search_causal_lm.cpp)
Added it to [supported models list](https://github.com/openvinotoolkit/openvino.genai/tree/master/text_generation/causal_lm/cpp#supported-models)


pavel-esir:
FYI @Vishwa44, right after i merged i figured out that styling should be corrected. Done it here https://github.com/openvinotoolkit/openvino.genai/pull/296

PR #296, opened 2024-03-11
update Readme for dolly-v2-3b

pavel-esir:
corrected style in https://github.com/openvinotoolkit/openvino.genai/pull/294

FYI @Vishwa44


ilya-lavrenov:
Looks like current state is also OK, is not it?
image

pavel-esir:
> Looks like current state is also OK, is not it?

I just thought that specifying concrete version of `dolly-v2-3b` would be better, because there are also other variations `dolly-v2-12b`, `dolly-v2-7b`. Despite this should we still keep it simply `Dolly`?

Vishwa44:
I can work on verifying other Dolly models, and updated the readme accordingly

pavel-esir:
> I can work on verifying other Dolly models, and updated the readme accordingly

Thanks @Vishwa44 i created issue for that https://github.com/openvinotoolkit/openvino.genai/issues/297, but could you please firstly add test for v2.3b as @p-wysocki asked?


@ilya-lavrenov what do you think? approve this changes or we will leave it just Dolly

ilya-lavrenov (approved):

Issue #270, opened 2024-03-01
[Good First Issue]: Verify zephyr-7b-beta with GenAI text_generation


p-wysocki:
### Context

This task regards enabling tests for **zephyr-7b-beta**. You can find more details under openvino_notebooks [LLM chatbot README.md](https://github.com/openvinotoolkit/openvino_notebooks/tree/main/notebooks/254-llm-chatbot).

Please ask general questions in the main issue at https://github.com/openvinotoolkit/openvino.genai/issues/259

### What needs to be done?

Described in the main Discussion issue at: https://github.com/openvinotoolkit/openvino.genai/issues/259

### Example Pull Requests

Described in the main Discussion issue at: https://github.com/openvinotoolkit/openvino.genai/issues/259

### Resources

- [Contribution guide - start here!](https://github.com/openvinotoolkit/openvino/blob/master/CONTRIBUTING.md)
- [Intel DevHub Discord channel](https://discord.gg/7pVRxUwdWG) - engage in discussions, ask questions and talk to OpenVINO developers

### Contact points

Described in the main Discussion issue at: https://github.com/openvinotoolkit/openvino.genai/issues/259

Godwin-T:
.take

p-wysocki:
Hello @Godwin-T, are you still working on this? Is there anything we could help you with?

tushar-thoriya:
Hello @Godwin-T what is the status ? I want to work on this issue.

Godwin-T:
Yes i am still working on this.

p-wysocki:
The PR has been merged, we must have forgotten to close the ticket. :) Thank you for your contribution @Godwin-T!

PR #305, opened 2024-03-14
Verify zephyr-7b-beta with GenAI text_generation #270

Godwin-T:
Added Zephry 7B to list of supported models

Verify zephyr-7b-beta with GenAI text_generation #270


pavel-esir (approved):
LGTM @Godwin-T thanks for contribution!

PR #307, opened 2024-03-14
Align beam_search_causal_lm with new transformers version

as-suvorov:


text_generation/causal_lm/cpp/group_beam_searcher.hpp
```diff
@@ -173,7 +181,7 @@ struct GroupBeamSearcher {
                 continue;
             }
             std::vector<Beam> candidates;
-            candidates.reserve(2 * parameters.group_size);
+            candidates.reserve(parameters.group_size * 2 * parameters.group_size);
```
ilya-lavrenov, line 184: do we need so much candidates?

as-suvorov: I wanted to discuss this with @Wovchena. I just fixed the reservation as our implementation actually evaluates `parameters.group_size * 2 * parameters.group_size` candidates. As far as I understand HF implementation evaluates only `parameters.group_size * 2` for each group. I've tried to limit this size to match HF implementation but results didn't match HF so I rolled it back.
I could investigate this further.

ilya-lavrenov: But later we break by `add_count == 2 * parameters.group_size` anyway..

as-suvorov: We do it for each beam in the group:
[`for (const Beam& beam : group->ongoing) {`](https://github.com/openvinotoolkit/openvino.genai/pull/307/files#diff-d3a3a74d1700bb0c99bc8ff7dd7fb1ab300cc20d41011ade6290b91ceaa2d5d7R185)

image
image

ilya-lavrenov: I think we can merge with current change, but later ask @Wovchena why do we add so many candidates

ilya-lavrenov (approved):

text_generation/causal_lm/cpp/group_beam_searcher.hpp
Wovchena: `transformers` implementation evaluates every possible combination of beam and its continuation within the group relying on tensor broadcasting. That gives `group_size*vocab_size` candidates. After that they accept `group_size * 2` best candidates. My implementation optimizes that. In the worst case all `2 * parameters.group_size` candidates are taken from a single beam while all other beams from the previous iteration are discarded. it's impossible to tell which beams are going to be discarded, so the algorithm keeps `2 * parameters.group_size` for every beam which results in `parameters.group_size * 2 * parameters.group_size` candidates,

Good catch!

Issue #254, opened 2024-02-28
Llama-2-7b-hf convert "Export model to OpenVINO directly failed" "Model will be exported to ONNX"


oldmikeyang:
python convert.py --model_id /home/llm/disk/llm/meta-llama/Llama-2-7b-hf --output_dir /home/llm/disk/llm/meta-llama/Llama-2-7b-hf-openvino --precision FP32
INFO:nncf:NNCF initialized successfully. Supported frameworks detected: torch, onnx, openvino
/home/llm/miniconda3/envs/openvino/lib/python3.9/site-packages/diffusers/utils/outputs.py:63: UserWarning: torch.utils._pytree._register_pytree_node is deprecated. Please use torch.utils._pytree.register_pytree_node instead.
  torch.utils._pytree._register_pytree_node(
/home/llm/miniconda3/envs/openvino/lib/python3.9/site-packages/diffusers/utils/outputs.py:63: UserWarning: torch.utils._pytree._register_pytree_node is deprecated. Please use torch.utils._pytree.register_pytree_node instead.
  torch.utils._pytree._register_pytree_node(
[ INFO ] openvino runtime version: 2023.3.0-13775-ceeafaf64f3-releases/2023/3
Loading checkpoint shards: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:03<00:00,  1.50s/it]
Using the export variant default. Available variants are:
    - default: The default ONNX variant.
Using framework PyTorch: 2.2.1+cu121
Overriding 1 configuration item(s)
        - use_cache -> True
/home/llm/miniconda3/envs/openvino/lib/python3.9/site-packages/transformers/modeling_utils.py:4193: FutureWarning: `_is_quantized_training_enabled` is going to be deprecated in transformers 4.39.0. Please use `model.hf_quantizer.is_trainable` instead
  warnings.warn(
The cos_cached attribute will be removed in 4.40. Bear in mind that its contents changed in v4.38. Use the forward method of RoPE from now on instead.
The sin_cached attribute will be removed in 4.40. Bear in mind that its contents changed in v4.38. Use the forward method of RoPE from now on instead.
[...]

PR #308, opened 2024-03-14
Fix llama2 conversion issues "Conversion is failed for: aten::mul"

aahouzi:
## Type of Change

- The PR fixes an issue mentioned in #254. There was a fix proposed in #246, but this was only applied the cpp part and not the llm_bench/python. As we speak, with current code version, the llama2 conversion fails with same error when using llm_bench/python.

- This PR introduces same fix as #246 for llm_bench/python too + fixes some missing documentation in README and typos

## Description

- Same as above

## Expected Behavior & Potential Risk

- N/A

## How has this PR been tested?

- N/A

## Dependency Change?

- N/A

### DIFF
--- a/llm_bench/python/README.md
+++ b/llm_bench/python/README.md
@@ -14,6 +14,8 @@ source python-env/bin/activate
 pip install update --upgrade
 pip install -r requirements.txt
 ```
+> Note: For llama models, ensure to have transformers<4.38.
+
 ### 2. Convert a model to OpenVINO IR
    
 The conversion script for preparing benchmarking models,
--- a/llm_bench/python/README.md
+++ b/llm_bench/python/README.md
@@ -88,6 +88,8 @@ Parameters:
 * `-p` - interactive prompt text
 * `-pf` - path of JSONL file including interactive prompts
 * `-n` - number of benchmarking iterations, if the value greater 0, will exclude the first iteration. (default=0)
+* `-ic` - limit the output token size (default 512) of text_gen and code_gen models.
+
 
 ``` bash
 python ./benchmark.py -h # for more information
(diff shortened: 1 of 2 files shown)


eaidova:
@aahouzi  hello, I'm not think it is good place for proposing this fix, because according to my knowledge, our master branch is aligned with openvino master and it has llama supported with transformers>4.38 as well. You can check it using openvino nightly package. It will block for us validation other models that was added in transformers 4.38 (stablelm, qwen2, gemma) Can it be moved closer to text generation demo requirements instead? or you can switch to branch for 2024.0 release, where llm_bench has such requirements fix for compatibility with public 2024.0.0 release

aahouzi:
@eaidova Sure, does it work now ?

aahouzi:
@eaidova it should be fine now ?

eaidova:
> @eaidova it should be fine now ?

changes that fixes llama with transformers>4.38 was merged 3 weeks ago, they are not part of openvino 2024.0, but will be in 2024.1, it is possible to convert and run such model using nightly builds for 2024.1 from here:
c++ package https://storage.openvinotoolkit.org/repositories/openvino/packages/nightly/
wheels https://storage.openvinotoolkit.org/wheels/nightly/openvino/

aahouzi:
@eaidova so for a user who is using your repo for the first time, how can they know they need to install nightly builds without asking you directly ? They'll be stuck with this bug for quite some time. It's a matter of user experience of our software. At least keep the note, and when you release version 2024.1, you can remove it.

eaidova:
@aahouzi as you can see here there is already explicit specification transformers version in text_generation pipeline documentation:
https://github.com/openvinotoolkit/openvino.genai/tree/master/text_generation/causal_lm/cpp#install-openvino

I do not see reasons forcing that for other subprojects that are not directly related to this demo code

aahouzi:
@eaidova I don't get it, I introduced commits that removed my change to the requirements.txt file. I'm not forcing anything, it's just a small note under the install section for people that wants to run llama benchmark with Python, and not have their conversion fail, that's all.

eaidova (approved):

PR #309, opened 2024-03-15
Avoid pip freeze warning

peterchen-intel:
`
Skipping line in requirement file [openvino.genai/llm_bench/python/requirements.txt] because it's not clear what it would install: git+https://github.com/huggingface/optimum-intel.git@552de65a9c5f7fa1a2f0ce6859ebdeedaeaabe53                                                                                                                 
  (add #egg=PackageName to the URL to avoid this warning)   
Skipping line in requirement file [openvino.genai/llm_bench/python/requirements.txt] because it's not clear what it would install: git+https://github.com/openvinotoolkit/nncf.git                                                                                                                                                               
  (add #egg=PackageName to the URL to avoid this warning)   
`


eaidova (approved):

peterchen-intel:
Still need to figure out how pip freeze works for optimum-intel and nncf which are with commit IDs.

PR #314, opened 2024-03-20
add submodule update into sd installation instructions

eaidova:
installing dependency as 
pip install ../../../thirdparty/openvino_tokenizers[transformers] may be confusing for users and lead to error as there is no any mentions about submodule initialization before it in comparison with text_generation demos https://github.com/openvinotoolkit/openvino.genai/tree/master/text_generation/causal_lm/cpp#install-openvino

### DIFF
--- a/image_generation/stable_diffusion_1_5/cpp/README.md
+++ b/image_generation/stable_diffusion_1_5/cpp/README.md
@@ -23,7 +23,6 @@ conda install openvino c-compiler cxx-compiler make
 
 1. Install dependencies to import models from HuggingFace:
 ```shell
-
 git submodule update --init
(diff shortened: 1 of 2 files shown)


image_generation/stable_diffusion_1_5/cpp/README.md
```diff
@@ -23,6 +23,8 @@ conda install openvino c-compiler cxx-compiler make
 
 1. Install dependencies to import models from HuggingFace:
 ```shell
+
+git submodule update --init
```
eaidova, line 27:
```suggestion
git submodule update --init
```

ilya-lavrenov (approved):

PR #353, opened 2024-04-09
Read EOS token from model runtime information for speculative_decoding_lm

anzr299:
Extension to issue #277, Added the functionality to read EOS token from model runtime information in the speculative_decoding_lm.


anzr299:
@ilya-lavrenov this is the PR corresponding to the change we discussed.

anzr299:
The readme does not contain proper instruction for running speculative_decoding_lm. in the example, it runs the "Llama-2-7b-chat-hf" model as the main model directory; 
`/build/Release/speculative_decoding_lm ./TinyLlama-1.1B-Chat-v1.0/pytorch/dldt/FP16/ ./Llama-2-7b-chat-hf/pytorch/dldt/FP16/ "Why is the Sun yellow?"`
The previous steps do not mention any installation of this model and it also requires extra permission to use this model(model at meta-llama/Llama-2-70b-chat-hf)

ilya-lavrenov (approved):

pavel-esir:
> The previous steps do not mention any installation of this model and it also requires extra permission to use this model(meta-llama/Llama-2-70b-chat-hf)

Thanks for noticing that. I will update Readme in the next following PRs

PR #361, opened 2024-04-11
Add min_new_tokens when infer_count is set

helena-intel:
When specifying infer_count, it may be expected that exactly that number of tokens is generated, but that is not currently the case. The number of tokens generated depends on model, prompt, and max_new_tokens. This PR adds min_new_tokens to `.generate()` so that the expected number of tokens is generated always. This does mean that model output can be repetitive at the end - but when infer_count is set, the goal of benchmarking is to get performance with the specified number of tokens, not to get the best output for a particular prompt. When infer_count is not set, min_new_tokens also is not changed.

CVS-133717

### DIFF
--- a/llm_bench/python/benchmark.py
+++ b/llm_bench/python/benchmark.py
@@ -106,7 +106,12 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
     if (args['mem_consumption'] == 1 and num == 0) or args['mem_consumption'] == 2:
         mem_consumption.start_collect_memory_consumption()
     start = time.perf_counter()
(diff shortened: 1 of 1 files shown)


peterchen-intel:
There is already a PR https://github.com/openvinotoolkit/openvino.genai/pull/289, but looks like it will provide a 'fake' and better performance for 2nd token latency (average of other tokens latency except 1st token) https://github.com/openvinotoolkit/openvino.genai/pull/289#issuecomment-1989746433

peterchen-intel:
Will be implemented in https://github.com/openvinotoolkit/openvino.genai/pull/289

PR #377, opened 2024-04-23
Update NNCF defaults for several models

KodiaqQ:
#### Updated NNCF defaults for:

- zephyr-7b-beta
- llama-7b
- stable-zephyr-3b-dpo
- baichuan2-7b-chat
- mistral-7b-v0.1

Based on the experiments from 135227


llm_bench/python/utils/nncf_utils.py
```diff
@@ -59,4 +60,8 @@ def get_compressed_path(output_dir: str, base_precision, option: str):
     "bloomz-560m": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 64, "ratio": 0.8,
                     "dataset": {"name": "wikitext,wikitext-2-v1,train[:1000],text", "awq": True}},
     "mixtral-8x7b-v0.1": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 128, "ratio": 0.8},
+    "baichuan2-7b-chat": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 128, "ratio": 0.8,
+                          "dataset": {"name": "wikitext,wikitext-2-v1,train[:1000],text", "awq": True}},
+    "mistral-7b-v0.1": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 128, "ratio": 0.9},
+    "llama-7b": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 128, "ratio": 0.7},
```
ljaljushkin, line 66: image
Could you elaborate the algorithm or threshold for selecting configuration? 
The green ones are accepted and the one with the highest error is selected?

KodiaqQ: All as you described.
The threshold of 5% drop was selected as the target. The highest drop was chosen to achieve the most balanced model (accurate & performed).

ljaljushkin: Why did you choose ratio=0.7 ? I don't see that configuration in the table. 
why not asym, gs=128, ratio=1 AWQ?

ljaljushkin (changes requested):

llm_bench/python/utils/nncf_utils.py
KodiaqQ: The assumption of the setting ratio=0.7 is because the experiment results (Xeon 8180) show that ratio=0.8 drops accuracy more than 5%.
Asymmetric configuration drops performance as I know and that is the reason for choosing the best symmetric configuration.

KodiaqQ:
@eaidova, could you merge it, please?

ljaljushkin (approved):

PR #378, opened 2024-04-23
Add prompt lookup decoding sample

sammysun0711:
Add prompt lookup decoding sample based on https://github.com/apoorvumang/prompt-lookup-decoding
- Initial test with llama2 on CPU works


sammysun0711:
Close it since https://github.com/openvinotoolkit/openvino.genai/pull/379 merged

PR #389, opened 2024-04-26
Throw exception when MD5 is not aligned instead of warning

wgzintel:


peterchen-intel:
Need to wait for the fixing of this CVS-140095

yeonbok:
Hello, from the gpu plugin's perspective, we do not guarantee first and second inference's bit exactness. It is because we have two types of kernels : 1) shape agnostic kernels that are executed first and build more optimized static kernels. 2) Once optimized static kernels are ready later, they are executed. However these kernels have different configuration (e.g.,, tile size) which causes compuation result different because the order of multipy & add is different.  Please refer to the issue [138890](https://jira.devtools.intel.com/browse/CVS-138890).

peterchen-intel:
@wgzintel  Let's always compare to the previous one instead of comparing to warm-up.

peterchen-intel:
> Hello, from the gpu plugin's perspective, we do not guarantee first and second inference's bit exactness. It is because we have two types of kernels : 1) shape agnostic kernels that are executed first and build more optimized static kernels. 2) Once optimized static kernels are ready later, they are executed. However these kernels have different configuration (e.g.,, tile size) which causes compuation result different because the order of multipy & add is different. Please refer to the issue [138890](https://jira.devtools.intel.com/browse/CVS-138890).

@yeonbok  Can you please share the link for the documentation to explain this difference?

peterchen-intel (approved):

PR #397, opened 2024-04-28
Update openvino to 2024.1

wgzintel:

PR #401, opened 2024-04-29
Update OV and NNCF versions

peterchen-intel:
To the counterpart releases for this branch.


peterchen-intel:
Duplicated https://github.com/openvinotoolkit/openvino.genai/pull/397

PR #435, opened 2024-05-10
Set trust remote code for custom model

wgzintel:
1. set trust_remote_code=true for 'baichuan', 'qwen', 'chatglm' to permit loading of these models.
2. enable Phi2 and starcoder models.


llm_bench/python/utils/ov_utils.py
```diff
@@ -143,7 +143,12 @@ def create_text_gen_model(model_path, device, **kwargs):
     else:
         remote_code = False
         try:
-            model_config = AutoConfig.from_pretrained(model_path)
+            remote_models = ['baichuan', 'qwen', 'chatglm']
```
eaidova, line 146: I think we need to differentiate qwen and qwen1.5 models here, because the second one is not using remote code

eaidova (approved):

llm_bench/python/utils/ov_utils.py
peterchen-intel: @eaidova Since these models throw exception in model_class.from_pretrained() instead of AutoConfig.from_pretrained(), we are thinking to catch the exception of model_class.from_pretrained() and call both AutoConfig.from_pretrained()/model_class.from_pretrained() again with trust_remote_code=True.  Any preference from your side?

eaidova: from my side no, only possibly it may be time consuming as leads to loading model twice, but as we speak only for loading IR and model already downloaded, it should not be issue

eaidova: could you please check with these changes?
https://github.com/openvinotoolkit/openvino.genai/pull/440
explicit setting trust_remote_code=False for config prevent waiting user input and allow capturing exception as expected

wgzintel: @eaidova  Merged [into](https://github.com/openvinotoolkit/openvino.genai/pull/440) this PR, please take a look.

eaidova (approved):

peterchen-intel (approved):

PR #448, opened 2024-05-16
add google/gemma-2b-it

pavel-esir:


pavel-esir:
Access to model google/gemma-2b-it is restricted. You must be authenticated to access it.

PR #449, opened 2024-05-16
output no hook data warning when it is text gen model

wgzintel:


peterchen-intel (approved):

eaidova:
@peterchen-intel @wgzintel  could you please explain? What this warning messaged mean?

benchmarking text generation models is base usacase for llm bench, so this warning appears in majority cases and lead to users confusion

wgzintel:
@eaidova  If the user uses a transformer version lower than 4.40, a warning will be output to remind the user that there is no latency printout for the 1st token and 2nd tokens in greedy search and beam search.

eaidova:
> @eaidova If the user uses a transformer version lower than 4.40, a warning will be output to remind the user that there is no latency printout for the 1st token and 2nd tokens in greedy search and beam search.

can we update requirements.txt to use transformers>=4.40 by default for avoid this or enable hook for old transformers as well?
this message is not clear from user side. It is better to rephrase that detailed latency reports is not collected as well as it is hard to understand from code side how it is connected with tokenization time, so possibly better to change condition to list of inference time empty

wgzintel:
> > @eaidova If the user uses a transformer version lower than 4.40, a warning will be output to remind the user that there is no latency printout for the 1st token and 2nd tokens in greedy search and beam search.
> 
> can we update requirements.txt to use transformers>=4.40 by default for avoid this or enable hook for old transformers as well? this message is not clear from user side. It is better to rephrase that detailed latency reports is not collected as well as it is hard to understand from code side how it is connected with tokenization time, so possibly better to change condition to list of inference time empty

OK, will do it.

PR #457, opened 2024-05-22
Force to generate more tokens

wgzintel:


llm_bench/python/benchmark.py
```diff
@@ -34,8 +34,7 @@
 DEFAULT_SUPER_RESOLUTION_STEPS = 50
 DEFAULT_SUPER_RESOLUTION_WIDTH = 128
 DEFAULT_SUPER_RESOLUTION_HEIGHT = 128
-DEFAULT_OUTPUT_TOKEN_SIZE = 512
```
peterchen-intel, line 37: Default is 512 if infer_count is not set, take infer_count if it is set no matter how large the number.

wgzintel: Done

PR #334, opened 2024-03-30
Generate pipeline

pavel-esir:
LLM return logits with probabilities of each token, these probabilities can be converted to tokens/words with different technics: greedy decoding, beam search decoding, random sampling, etc.

This requires writing user unfriendly post-processing even for the simplest scenario of greedy decoding. In order to make live easier we we combined all decoding scenarios into a single function call, where the decoding method and parameters are specified by arguments.

In this PR we provide a user friendly API for text generation inspired by `generate` method from HuggingFace transformers library. 

- [x] enable calling tokenizers/detokenizers from LLMPipeline
- [ ] add callback for streaming mode - done partially, need to improve
- [x] rewritten samples with the current approach: [causal_lm/cpp/generate_pipeline/generate_sample.cpp#L73-L83](https://github.com/pavel-esir/openvino.genai/blob/generate_pipeline/text_generation/causal_lm/cpp/generate_pipeline/generate_sample.cpp#L73-L83)
- [x] Multibatch greedy decoding
- [ ] Speculative decoding
- [ ] Grouped Beam Search decoding: ready for batch 1, need to rebase multibatch support after merging https://github.com/openvinotoolkit/openvino.genai/pull/349
- [x] Random sampling

Example 1: Greedy search generation
```
LLMPipeline pipe(model_path, device);

// Will try to load config from generation_config.json.
// but if not found default velues for gready search will be used
GenerationConfig config = pipe.generation_config();

cout << pipe(prompt, config.max_new_tokens(20));
```

Example 2: TextStreaming mode
```
LLMPipeline pipe(model_path, device);

GenerationConfig config = pipe.generation_config();

auto text_streamer = TextStreamer{pipe};
auto text_streamer_callback = [&text_streamer](std::vector<int64_t>&& tokens, LLMPipeline& pipe){
    text_streamer.put(tokens[0]);
};

pipe(prompt, config.max_new_tokens(20).set_callback(text_streamer_callback));
text_streamer.end();
```

CVS-132907 CVS-137920

PR #459, opened 2024-05-22
Jinja2Cpp Licensing

jacekpawlak:
Added licensing file to Jinja2Cpp as it's statically linked


pavel-esir (approved):

licensing/third-party-programs.txt
Wovchena, line 1: Master doesn't have Jinja2Cpp yet. Jinja2Cpp's license is copied to root third-party-programs.txt already in https://github.com/openvinotoolkit/openvino.genai/pull/334, where Jinja2Cpp's itself is added

jacekpawlak: I did not see that. In that case I'm closing this PR. Thanks for clarification

PR #462, opened 2024-05-28
grammar corrector support in WWB

ljaljushkin:
This PR introduces support for `AutoForSeq2SeqLM` models in WWB. 
Previously, WWB only supported `AutoForCasualLM`, assuming that the `generate` method copies the prompt to the output. 
But AutoForSeq2SeqLM generates output differently: there is no copy of the prompt, and it directly generates output.

The fix was checked on the [example](https://gist.github.com/ljaljushkin/5a489a27cd0020ddbd42ea7ae54be688). It evaluates grammar correction with Seq2Seq model using WWB.


ljaljushkin:
@AlexKoff88 @andreyanufr

AlexKoff88:
@eaidova, please merge

PR #473, opened 2024-06-05
Fix chat template, add test for chat scenario

olpipi:
Fix chat template, add test for chat scenario

### DIFF
--- a/.github/workflows/causal_lm_cpp.yml
+++ b/.github/workflows/causal_lm_cpp.yml
@@ -579,4 +579,8 @@ jobs:
           output.close()
           "
           diff pred.txt ref.txt
-          echo "Chat sample" passed
+          echo "Chat sample cpp" passed
+          export PYTHONPATH=./build/:$PYTHONPATH
+          timeout 30s ./samples/python/chat_sample/chat_sample.py ./TinyLlama-1.1B-Chat-v1.0/ < input.txt > ./pred2.txt
+          diff pred2.txt ref.txt
+          echo "Chat sample python" passed
--- a/.github/workflows/causal_lm_cpp.yml
+++ b/.github/workflows/causal_lm_cpp.yml
@@ -526,7 +526,7 @@ jobs:
           | diff ./pred_greedy.txt -
 
   cpp-chat_sample-ubuntu:
-    runs-on: ubuntu-20.04-16-cores
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v4
         with:
@@ -551,11 +551,9 @@ jobs:
         run: |
           source ./ov/setupvars.sh
           printf 'What is 2 + 2?\nWhat is the previous answer?\nAdd 1 to it.\nSubtract 5 from it.\nWhy is the sun yellow?\nWhat was my first question?\nStop!\n' > ./input.txt
-          timeout 30s cat input.txt | ./build/samples/cpp/chat_sample/chat_sample ./TinyLlama-1.1B-Chat-v1.0/ > ./pred.txt
+          timeout 30s ./build/samples/cpp/chat_sample/chat_sample ./TinyLlama-1.1B-Chat-v1.0/ < input.txt > ./pred.txt
           python -c "
           from transformers import LlamaTokenizer, AutoModelForCausalLM
-          with open('pred.txt', 'r') as file:
-              predictions = file.read()
           model_id = 'TinyLlama/TinyLlama-1.1B-Chat-v1.0'
           tokenizer = LlamaTokenizer.from_pretrained(model_id)
           model = AutoModelForCausalLM.from_pretrained(model_id)
@@ -566,16 +564,19 @@ jobs:
               return {'role': 'assistant', 'content': answer}
           chat_history = []
           chat_prompt = ''
+          output = open('ref.txt', 'w')
           for prompt in prompts:
+              output.write('question:\n')
               chat_history.append(gen_prompt(prompt))
               chat_prompt = tokenizer.apply_chat_template(chat_history, tokenize=False, add_generation_prompt=True)
               tokenized = tokenizer(chat_prompt, return_tensors='pt')
               answer = model.generate(**tokenized, max_length=1000, do_sample=False)
               answer_str = tokenizer.decode(answer[0, tokenized['input_ids'].numel():], skip_special_tokens=True)
               chat_history.append(gen_answer(answer_str))
-              idx = predictions.find(answer_str)
-              if -1 == idx:
-                  raise RuntimeError(f'Missing "{answer_str=}" from predictions')
-              predictions = predictions[:idx] + predictions[idx + len(answer_str):]
+              output.write(answer_str)
+              output.write('\n----------\n')
+          output.write('question:\n')
+          output.close()
           "
+          diff pred.txt ref.txt
           echo "Chat sample" passed
--- a/.github/workflows/causal_lm_cpp.yml
+++ b/.github/workflows/causal_lm_cpp.yml
@@ -524,3 +524,58 @@ jobs:
           && export PYTHONPATH=./build/:$PYTHONPATH
           && timeout 50s samples/python/greedy_causal_lm/greedy_causal_lm.py ./redpajama-3b-chat/ "Alan Turing was a"
           | diff ./pred_greedy.txt -
+
+  cpp-chat_sample-ubuntu:
+    runs-on: ubuntu-20.04-16-cores
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: actions/setup-python@v4
+        with:
+          python-version: 3.8
+      - name: Install OpenVINO
+        run: |
+          mkdir ./ov/
+          curl ${{ env.l_ov_link }} | tar --directory ./ov/ --strip-components 1 -xz
+          sudo ./ov/install_dependencies/install_openvino_dependencies.sh
+      - name: Download, convert and build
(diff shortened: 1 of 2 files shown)


.github/workflows/causal_lm_cpp.yml
```diff
@@ -524,3 +524,58 @@ jobs:
           && export PYTHONPATH=./build/:$PYTHONPATH
           && timeout 50s samples/python/greedy_causal_lm/greedy_causal_lm.py ./redpajama-3b-chat/ "Alan Turing was a"
           | diff ./pred_greedy.txt -
+
+  cpp-chat_sample-ubuntu:
+    runs-on: ubuntu-20.04-16-cores
```
Wovchena, line 529:
```suggestion
    runs-on: ubuntu-20.04
```
Tiny llama usually fits the cheaper runner.

.github/workflows/causal_lm_cpp.yml
```diff
@@ -524,3 +524,58 @@ jobs:
...
+          python -m pip install --upgrade-strategy eager -r ./samples/requirements.txt --pre --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
+          python -m pip install ./thirdparty/openvino_tokenizers/[transformers] --pre --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
+          optimum-cli export openvino --trust-remote-code --weight-format fp16 --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 TinyLlama-1.1B-Chat-v1.0
+          cmake -DCMAKE_BUILD_TYPE=Release -S ./ -B ./build/
+          cmake --build ./build/ --config Release -j
+      - name: Compare
+        run: |
+          source ./ov/setupvars.sh
+          printf 'What is 2 + 2?\nWhat is the previous answer?\nAdd 1 to it.\nSubtract 5 from it.\nWhy is the sun yellow?\nWhat was my first question?\nStop!\n' > ./input.txt
+          timeout 30s cat input.txt | ./build/samples/cpp/chat_sample/chat_sample ./TinyLlama-1.1B-Chat-v1.0/ > ./pred.txt
```
Wovchena, line 554:
```suggestion
          timeout 30s ./build/samples/cpp/chat_sample/chat_sample ./TinyLlama-1.1B-Chat-v1.0/ < input.txt > ./pred.txt
```

Wovchena, line 554: Please also verify that python sample output is identical to pred.txt. See `diff` calls above.

.github/workflows/causal_lm_cpp.yml
```diff
@@ -524,3 +524,58 @@ jobs:
...
+          chat_history = []
+          chat_prompt = ''
+          for prompt in prompts:
+              chat_history.append(gen_prompt(prompt))
+              chat_prompt = tokenizer.apply_chat_template(chat_history, tokenize=False, add_generation_prompt=True)
+              tokenized = tokenizer(chat_prompt, return_tensors='pt')
+              answer = model.generate(**tokenized, max_length=1000, do_sample=False)
+              answer_str = tokenizer.decode(answer[0, tokenized['input_ids'].numel():], skip_special_tokens=True)
+              chat_history.append(gen_answer(answer_str))
+              idx = predictions.find(answer_str)
```
Wovchena, line 576: beamsearch calls `find()` because the beams could be reordered. You can strengthen the test by comparing the strings directly.

.github/workflows/causal_lm_cpp.yml
```diff
@@ -524,3 +524,58 @@ jobs:
...
+              if -1 == idx:
+                  raise RuntimeError(f'Missing "{answer_str=}" from predictions')
+              predictions = predictions[:idx] + predictions[idx + len(answer_str):]
+          "
+          echo "Chat sample" passed
```
Wovchena, line 581:
```suggestion
          echo "Chat sample" passed

```

src/cpp/src/llm_pipeline.cpp
```diff
@@ -14,6 +14,24 @@
...
+    auto first_size = fisrt.input_ids.get_size();
+    auto second_size = second.input_ids.get_size();
+    ov::Shape new_shape{1, first_size - second_size};
+
+    ov::Tensor new_input_ids(ov::element::i64, new_shape);
+    auto data_ptr = fisrt.input_ids.data<int64_t>();
+    std::copy(data_ptr + second_size, data_ptr + first_size, new_input_ids.data<int64_t>());
+
+    ov::Tensor new_attention_mask(ov::element::i64, new_shape);
+    std::fill_n(new_attention_mask.data<int64_t>(), new_shape[1], 1);
```
Wovchena, line 29: can you return view tensors to `fisrt` instead of copy

src/cpp/src/llm_pipeline.cpp
```diff
@@ -101,12 +119,17 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
                 constexpr bool add_generation_prompt = true;
                 auto new_templated_chat_history  = m_tokenizer.apply_chat_template(m_history, add_generation_prompt);
-                
-                prompt = new_templated_chat_history.substr(m_templated_chat_history.size());
+                auto new_chat_tokens = m_tokenizer.encode(new_templated_chat_history);
+                if (m_is_cache_empty) {
+                    encoded_input = new_chat_tokens;
+                } else {
+                    auto prev_chat_tokens = m_tokenizer.encode(m_templated_chat_history);
+                    encoded_input = subtract_chat_tokenized_inputs(new_chat_tokens, prev_chat_tokens);
```
Wovchena, line 127: Why is the history stored as `std::string` instead of encoded tensors?
@pavel-esir, maybe you can explain?

Wovchena (changes requested):

src/cpp/src/llm_pipeline.cpp
olpipi: I wrote this code originally. 
std::string is used because after tokenization on next step we will get the same tokens we have in kv-cache. 
Encoded tensors don't contain <eos> token at the end by default. And kv-cache contains it. So we have add it manually.

.github/workflows/causal_lm_cpp.yml
olpipi: done

.github/workflows/causal_lm_cpp.yml
olpipi: done

.github/workflows/causal_lm_cpp.yml
olpipi: done

.github/workflows/causal_lm_cpp.yml
olpipi: is there any difference?

.github/workflows/causal_lm_cpp.yml
olpipi: done

src/cpp/src/llm_pipeline.cpp
olpipi: original tensor with real memory will be dead before this view tensors is used

src/cpp/src/llm_pipeline.cpp
Wovchena: Please add the explanation to the code comments.

.github/workflows/causal_lm_cpp.yml
Wovchena: One extra line. You applied it.

src/cpp/src/llm_pipeline.cpp
Wovchena: My expectation was that such tensor prolongs the lifetime of the original tensor. This seems to be the truth because the underluing RoiTensor stores the original tensor in `make_tensor()`.
```C++
Tensor::Tensor(const Tensor& owner, const Coordinate& begin, const Coordinate& end)
    : _impl{make_tensor(owner._impl, begin, end)},
      _so{owner._so} {}
```
Isn't that correct?

.github/workflows/causal_lm_cpp.yml
Wovchena: I mean that sample: https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/chat_sample/chat_sample.py

src/cpp/src/llm_pipeline.cpp
olpipi: I thought you're talking about another view tensor Tensor(const ov::Output<const ov::Node>& port, void* host_ptr...)
Let me try this one with Coordinate

.github/workflows/causal_lm_cpp.yml
olpipi: done

pavel-esir (approved):
Lgtm. As in understand it's left only to add and explanation to comments. Approve in advance

src/cpp/src/llm_pipeline.cpp
olpipi: done

Wovchena (approved):

olpipi:
@Wovchena please enabled auto-merge again. I updated comment once more. Unfortunately, I don't have permission to do it myself

Wovchena:
build_jenkins

Wovchena (approved):

src/cpp/src/llm_pipeline.cpp
olpipi: CVS-146098 - ROI tensor issue

PR #489, opened 2024-06-11
Don't recreate LLMPipeline

Wovchena:


Wovchena:
Here's oneliner for windows cmd to reproduce by running multiple times `cmd /c "for /l %i in (1, 1, 9999) do @echo starting %i && pytest tests\python_tests\test_generate_api.py::test_decoding --sw || exit %errorolevel%"`

PR #491, opened 2024-06-12
Integrate static shape LLM execution pipeline

TolyaTalamanov:
# Overview

Since static shape pipeline is quite different from what is done for CPU/GPU plugins, it makes sense to separate into different implementation not to mix with existing one. 

Moved all common things into `LLMPipelineImplBase` with intent to inherit static shape implementation from it.

## Static pipeline approach details
- Both prefill / kvcache models are reshaped to static size at the pipeline initialization stage (`StaticLLMPipeline` ctor) and has hardcoded size of `1024`. Due to this, only `1024` tokens in total (`max_tokens`) can be handled.
- Chat conversation mode is not supported. KV-cache is reset on every "generate" call.
- Only `greedy` search decoding is handled for now
- Only batch size 1 is supported

## Examples

```
ov::genai::LLMPipeline pipe(model_path, "NPU");
ov::genai::GenerationConfig config = pipe.get_generation_config();
pipe.generate("Why is the Sun yellow?", config)
``` 
or with `streamer`
```
std::function<bool(std::string)> streamer = [](std::string word) { std::cout << word << std::flush; return false; }; 
...
pipe.generate("Why is the Sun yellow?", config, streamer)
```
or on raw `input_ids`:
```
auto tokenizer = pipe.get_tokenizer();
auto encoded = tokenizer.encode(prompt);
pipe.generate(encoded.input_ids, config, streamer);
```

**Note:** Batched input isn't supported:
```
std::vector<std::string> texts = { "table is made of", "Alan Turing was a" };
pipe.generate(texts config, streamer); // Throw: Currently only batch size=1 is supported
```


src/cpp/src/llm_pipeline.cpp
```diff
@@ -16,6 +16,8 @@
 #include "utils.hpp"
 #include "text_callback_streamer.hpp"
 
+#include "openvino/opsets/opset13.hpp"
```
TolyaTalamanov, line 19: Requirement for `add_slices_to_kvcache_inputs` transformation

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
 }  // namespace genai
 }  // namespace ov
 
+namespace {
```
TolyaTalamanov, line 397: Utils for `NPU` pipeline, perhaps the entire pipeline implementation with utils should be hidden in separate file

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+
+ov::Core get_core() {
+    static ov::Core core;
+    return core;
+}
+
+std::shared_ptr<ov::Model> add_slices_to_kvcache_inputs(const std::shared_ptr<ov::Model>& model) {
```
TolyaTalamanov, line 404: @slyalin is there any profit to modify the existing model rather than creating a new one?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+        new_params.push_back(new_param);
+    }
+    return std::make_shared<ov::Model>(model->get_results(), ov::SinkVector{}, new_params);
+}
+
+void reshape_to_static(std::shared_ptr<ov::Model> model,
+                       const uint32_t input_size,
+                       const uint32_t kvcache_size) {
+    std::map<std::string, ov::PartialShape> new_shapes;
+    for (auto input : model->inputs()) {
```
TolyaTalamanov, line 448: `beam_idx` isn't handled there, as it's not present in models for NPU

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    ) override;
+
+    EncodedResults generate(
+        const EncodedInputs& inputs,
+        OptionalGenerationConfig generation_config,
+        StreamerVariant streamer
+    ) override;
+
+    void start_chat() override {
+        OPENVINO_THROW("Currently chat conversation mode isn't supported on NPU");
```
TolyaTalamanov, line 508: for NPU device

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    };
+    void finish_chat() override {
+        OPENVINO_THROW("Currently chat conversation mode isn't supported on NPU");
```
TolyaTalamanov, line 511: same

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+
+NPULLMPipelineImpl::NPULLMPipelineImpl(
+    const std::filesystem::path& path,
+    const ov::AnyMap& config
+) : LLMPipelineImplBase(path.string(),
+                        from_config_json_if_exists(path),
+                        chat_template_from_tokenizer_json_if_exists(path)) {
+    auto kvcache_model = get_core().read_model(path / "openvino_model.xml");
+    // TODO: Add this point, there must be transformation applied to expose KV-cache from the model
+    // ov::pass::ExposeKVCacheFromModel().run_on_model(kvcache_model);
```
TolyaTalamanov, line 536: @slyalin FYI

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    auto prefill_model = kvcache_model->clone();
+    prefill_model->set_friendly_name(kvcache_model->get_friendly_name() + "_prefill");
+
+    // FIXME: There must be better logic than just hardcoded values
```
TolyaTalamanov, line 540: I think it should be enough as first step...

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    OptionalGenerationConfig generation_config,
+    StreamerVariant streamer
+) {
+    // FIXME: Most of this method is copy-paste from LLMPipelineImpl
+    ov::Tensor input_ids;
+    ov::Tensor attention_mask;
+
+    if (auto data = std::get_if<ov::Tensor>(&inputs)) {
+        input_ids = *data;
+        // FIXME: As NPULLMpipelineImpl doesn't use search algorithms (e.g greedy_search)
```
TolyaTalamanov, line 628: btw, how it's handled for CPU/GPU?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    if (auto streamer_obj = std::get_if<std::monostate>(&streamer)) {
+        streamer_ptr = nullptr;
+    } else if (auto streamer_obj = std::get_if<std::shared_ptr<StreamerBase>>(&streamer)) {
+        streamer_ptr = *streamer_obj;
+    } else if (auto callback = std::get_if<std::function<bool(std::string)>>(&streamer)) {
+        streamer_ptr = std::make_shared<TextCallbackStreamer>(m_tokenizer, *callback);
+    }
+
+    auto batch_size = input_ids.get_shape().at(0);
+    if ((batch_size != 1 || !(config.is_greedy_decoding() || config.is_multinomial())) && streamer_ptr) {
```
TolyaTalamanov, line 652: Batch size > 1 isn't even supported for NPU pipeline

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    }
+
+    // NB: Start of NPU-friendly execution part
+    ov::genai::EncodedResults results;
+
+    // NB: Only batch=1 is supported now
+    results.scores.resize(1u);
+    results.tokens.resize(1u);
+
+    auto prompt_len = input_ids.get_size();
```
TolyaTalamanov, line 664: What if `prompt_len` is bigger than prefill model may accept?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +394,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    auto* attention_mask_data = m_kvcache_request.get_tensor("attention_mask").data<int64_t>();
+
+    const size_t max_tokens = config.get_max_new_tokens(prompt_len);
+    for (int i = 0; i < max_tokens - 1; ++i) {
+        input_ids_data[0] = last_token;
+        position_ids_data[0] = m_kvcache_desc.num_stored_tokens;
+        attention_mask_data[m_kvcache_desc.total_size - m_kvcache_desc.num_stored_tokens - 1] = 1u;
+
+        m_kvcache_request.infer();
+        m_kvcache_desc.num_stored_tokens += 1;
```
TolyaTalamanov, line 720: Need to check if KV-cache isn't full...

src/cpp/src/llm_pipeline.cpp
```diff
@@ -359,37 +757,41 @@ ov::genai::LLMPipeline::LLMPipeline(
...
 
 ov::genai::LLMPipeline::LLMPipeline(
-    const std::string& path, 
-    const std::string& device, 
+    const std::string& path,
+    const std::string& device,
     const ov::AnyMap& config
 ) {
-    m_pimpl = make_unique<LLMPipelineImpl>(std::filesystem::path(path), device, config);
+    if (device == "NPU") {
```
TolyaTalamanov, line 779: Same should be done for another ctor

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Looks it doesn't
```
Exception from src/inference/src/cpp/infer_request.cpp:74:
Exception from src/inference/src/cpp/infer_request.cpp:66:
Exception from src/plugins/intel_cpu/src/infer_request.cpp:342:
Failed to set empty tensor for port!
```

TolyaTalamanov: UPD, it's handled in release branch, fixed it in master as well

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Put exception

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Handled

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Fixed

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Done

src/cpp/src/llm_pipeline.cpp
```diff
@@ -65,17 +67,46 @@ ov::genai::OptionalGenerationConfig get_config_from_map(const ov::AnyMap& config
         return std::nullopt;
 }
 
+void remove_special_tokens(ov::genai::TokenizedInputs& tokenized_input,
+                           const std::string& original_text,
+                           const std::vector<std::string> special_tokens) {
```
TolyaTalamanov, line 72: ref is missed

TolyaTalamanov: Added

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +400,357 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
 }  // namespace genai
 }  // namespace ov
 
+namespace {
+
+ov::Core get_core() {
```
TolyaTalamanov, line 405: Looks it's not even needed as now used only in `ctor`

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +400,357 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+
+        m_kvcache_request.infer();
+        m_kvcache_desc.num_stored_tokens += 1;
+
+        last_token = utils::argmax(m_kvcache_request.get_tensor("logits"), 0);
+        results.tokens[0].push_back(last_token);
+        results.scores[0] = 0u;
+
+        // NB: KV-cache is full, further generation is impossible
+        if (m_kvcache_desc.num_stored_tokens == m_kvcache_desc.total_size) {
```
TolyaTalamanov, line 731: It should be below `streamer` otherwise it won't print the last generated token

TolyaTalamanov: Done

src/cpp/src/llm_pipeline.cpp
```diff
@@ -65,17 +67,46 @@ ov::genai::OptionalGenerationConfig get_config_from_map(const ov::AnyMap& config
         return std::nullopt;
 }
 
+void remove_special_tokens(ov::genai::TokenizedInputs& tokenized_input,
```
pavel-esir, line 70: luckily this bug is solved in openvino_tokenizers, so `remove_special_tokens` can be safely dropped.

src/cpp/src/llm_pipeline.cpp
```diff
@@ -90,21 +121,103 @@ ov::genai::EncodedResults multinominal_decoding(
...
+    virtual EncodedResults generate(
+        const EncodedInputs& inputs,
+        OptionalGenerationConfig generation_config,
+        StreamerVariant streamer
+    ) = 0;
+
+    virtual void start_chat() = 0;
+    virtual void finish_chat() = 0;
+
+    std::string apply_chat_template(const std::vector<std::pair<std::string, std::string>>& prompts) const;
```
pavel-esir, line 145: In the latest version we moved `apply_chat_template` to `ov::genai::Tokenizer` to make it similar to HF. Unfortunately the latest version is in release and not cherry-picked to release.

pavel-esir (commented):
llm_pipeline part mainly looks good. Only one thing, the latest changes are not ported from release to master yet. I think, it would be better if this PR is merged after https://github.com/openvinotoolkit/openvino.genai/pull/529

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: That makes things easier, I'll rebase once changes are available, thank you!

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +400,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    reshape_to_static(kvcache_model, 1u, max_kvcache_size);
+    // (4) Add slices to kvcache model
+    kvcache_model = add_slices_to_kvcache_inputs(kvcache_model);
+    // (5) Compiled both models for NPUW
+    std::map<std::string, std::string> cfg = { {"NPU_USE_NPUW", "YES"},
+                                               {"NPU_COMPILATION_MODE_PARAMS", "compute-layers-with-higher-precision=Sqrt,Power,ReduceMean,Add" },
+                                               {"NPUW_FOLD", "YES"},
+                                               {"NPUW_DCOFF_TYPE", "f16"},
+                                               {"NPUW_DCOFF_SCALE", "YES"},
+                                               {"NPUW_ONLINE_PIPELINE", "NONE"} };
```
TolyaTalamanov, line 577: FIXME must not be there!

TolyaTalamanov:
@ilya-lavrenov, @Wovchena Could you have a look, please?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +400,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+
+    OPENVINO_ASSERT(std::holds_alternative<std::string>(inputs));
+    const auto& text = std::get<std::string>(inputs);
+    auto tokenized_input = m_tokenizer.encode(std::get<std::string>(inputs));
+    remove_special_tokens(tokenized_input, text, {m_tokenizer.get_eos_token(), m_tokenizer.get_bos_token()});
+    auto encoded_results = generate(tokenized_input, config, streamer);
+    return {m_tokenizer.decode(encoded_results.tokens), encoded_results.scores};
+}
+
+EncodedResults NPULLMPipelineImpl::generate(
```
ilya-lavrenov, line 617: can we move it to a dedicated file?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +400,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+                                               {"NPU_COMPILATION_MODE_PARAMS", "compute-layers-with-higher-precision=Sqrt,Power,ReduceMean,Add" },
+                                               {"NPUW_FOLD", "YES"},
+                                               {"NPUW_DCOFF_TYPE", "f16"},
+                                               {"NPUW_DCOFF_SCALE", "YES"},
+                                               {"NPUW_ONLINE_PIPELINE", "NONE"} };
+    ov::AnyMap properties{cfg.begin(), cfg.end()};
+    m_prefill_request = core.compile_model(prefill_model, "NPU", properties).create_infer_request();
+    m_kvcache_request = core.compile_model(kvcache_model, "NPU", properties).create_infer_request();
+    // (6) Initialize tensors
+    prepareForNewConversation();
```
ilya-lavrenov, line 582:
```suggestion
    prepare_for_new_conversation();
```
can we use snake case?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -339,11 +400,354 @@ std::pair<std::string, Any> generation_config(const GenerationConfig& config) {
...
+    const size_t max_tokens = config.get_max_new_tokens(prompt_len);
+    for (int i = 0; i < max_tokens - 1; ++i) {
+        input_ids_data[0] = last_token;
+        position_ids_data[0] = m_kvcache_desc.num_stored_tokens;
+        attention_mask_data[m_kvcache_desc.total_size - m_kvcache_desc.num_stored_tokens - 1] = 1u;
+
+        m_kvcache_request.infer();
+        m_kvcache_desc.num_stored_tokens += 1;
+
+        last_token = utils::argmax(m_kvcache_request.get_tensor("logits"), 0);
```
ilya-lavrenov, line 722: what's about other sampling parameters?

src/cpp/src/llm_pipeline.cpp
ilya-lavrenov: does it mean batch is not supported by NPU? 
Currently, we use `beam_idx` to drop sequences from batch, which have been finished. 
Or you can support it with padding even for finished sequences?

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: yea, you'are right, will change

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Only batch=1 is supported, there is a special check for that in code. In general, I believe we still need to handle `beam_idx` in reshape function to comply with openvino models (not ours hand-crafted) but just set the shape of `[1]`.

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Ideally we should, to keep it isolated. Most probably we will need additional `llm_pipeline_impl_base.hpp` (not to define `LLMPipelineImplBase` in public header) and `npu_pipeline_impl.hpp` with their corresponding `.cpp` files

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: well, I thing that's it for `greedy` search which is the only one that currently supported for NPU pipeline.

I think the rest of decoding search algos can be also supported but require a little refactoring to re-use it in NPU pipeline. Didn't touch to minimize changes...

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: yep, removed it, thx!

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Also removed, even less changes now!

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Removed

src/cpp/src/llm_pipeline.cpp
```diff
@@ -293,15 +678,23 @@ ov::genai::LLMPipeline::LLMPipeline(
...
 
 ov::genai::LLMPipeline::LLMPipeline(
-    const std::string& path, 
-    const std::string& device, 
+    const std::string& path,
+    const std::string& device,
     const ov::AnyMap& config
 ) {
-    m_pimpl = make_unique<LLMPipelineImpl>(std::filesystem::path(path), device, config);
+    if (device == "NPU") {
```
Wovchena, line 693: Can you handle meta devices? For example, will `MULTI:NPU` work?

src/cpp/src/llm_pipeline.cpp
slyalin: Modifying an existing model is possible only if you have a full control over it. You shouldn't modify user-side models. If you cloned this model, then modifying in place is just simpler, no reason to create a new one.

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -82,6 +82,8 @@ class DecodedResults {
     }
 };
 
+class LLMPipelineImplBase;
```
ilya-lavrenov, line 85: please, move NPU impl to a dedicated files so that common refactoring should avoid changing NPU specifics

ilya-lavrenov (changes requested):

src/cpp/src/llm_pipeline.cpp
ilya-lavrenov: I suppose MULTI and LLM are not very compatible right now

src/cpp/src/llm_pipeline.cpp
ilya-lavrenov: agreed to have it as a requirement for merging

src/cpp/include/openvino/genai/llm_pipeline.hpp
ilya-lavrenov, line 85: I suppose we can move it to `LLMPipeline`'s private section

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: nah, there is not much sense as `NPUW` already proxy plugin

src/cpp/include/openvino/genai/llm_pipeline.hpp
TolyaTalamanov: Done

TolyaTalamanov: It's an interface now, in case it's part of `LLMPipeline` it should be public.

src/cpp/src/llm_pipeline.cpp
```diff
@@ -9,55 +9,20 @@
 #include <openvino/openvino.hpp>
 #include "openvino/genai/generation_config.hpp"
 #include "openvino/genai/llm_pipeline.hpp"
+#include "llm_pipeline_base.hpp"
+#include "llm_pipeline_npu.hpp"
 #include "utils.hpp"
 #include "text_callback_streamer.hpp"
 
-namespace {
```
TolyaTalamanov, line 15: Moved them to `utils` so far, to re-use for new pipeline

src/cpp/src/llm_pipeline_npu.cpp
```diff
@@ -0,0 +1,305 @@
...
+    // FIXME: There must be better logic than just hardcoded values
+    m_kvcache_desc = KVCacheDesc { 1024u, 0u };
+    const uint32_t max_prompt_size = m_kvcache_desc.total_size;
+    const uint32_t max_kvcache_size = m_kvcache_desc.total_size;
+    reshape_to_static(prefill_model, max_prompt_size, max_kvcache_size);
+    reshape_to_static(kvcache_model, 1u, max_kvcache_size);
+    // (5) Add slices to kvcache model
+    kvcache_model = add_slices_to_kvcache_inputs(kvcache_model);
+    // (6) Compiled both models for NPUW
+    std::map<std::string, std::string> cfg = { {"NPU_USE_NPUW", "YES"},
```
TolyaTalamanov, line 135: Must gone from there, consider extracting this information from `plugin_config` as temporal solution. Ideally it will be just `USE_NPUW=YES`

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Moved

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Got it, thanks!

src/cpp/src/llm_pipeline_npu.cpp
TolyaTalamanov: Done

src/cpp/src/llm_runner.hpp
ilya-lavrenov, line 1: not used?

src/cpp/src/llm_pipeline_base.hpp
```diff
@@ -0,0 +1,42 @@
...
+        StreamerVariant streamer
+    ) = 0;
+
+    virtual EncodedResults generate(
+        const EncodedInputs& inputs,
+        OptionalGenerationConfig generation_config,
+        StreamerVariant streamer
+    ) = 0;
+
+    virtual void start_chat() = 0;
```
ilya-lavrenov, line 32: as I see NPU pipeline does not implement this method. Is it OK ?
at least to throw exception that NPU does not support chat mode for now.

src/cpp/src/llm_pipeline.cpp
```diff
@@ -77,26 +42,23 @@ EncodedResults beam_search(
     GenerationConfig config
 );
 
-
-class LLMPipeline::LLMPipelineImpl {
+class LLMPipelineImpl final : public LLMPipelineImplBase {
```
ilya-lavrenov, line 45:
```suggestion
class StatefulLLMPipeline final : public LLMPipelineImplBase {
```

src/cpp/src/llm_pipeline_base.hpp
TolyaTalamanov: NPU does implement it, otherwise it won't be possible to instantiate `NPULLMPipelineImpl`:
https://github.com/TolyaTalamanov/openvino.genai/blob/at/integrate-npu-friendly-llm-execution/src/cpp/src/llm_pipeline_npu.hpp#L36-L41

src/cpp/src/llm_runner.hpp
TolyaTalamanov: oh jez, sorry!

src/cpp/src/llm_pipeline_npu.cpp
```diff
@@ -0,0 +1,310 @@
...
+) {
+    GenerationConfig config = (generation_config.has_value()) ? *generation_config : m_generation_config;
+    if (std::holds_alternative<std::vector<std::string>>(inputs)) {
+        OPENVINO_THROW("Currently only batch size=1 is supported for NPU device");
+    }
+
+    OPENVINO_ASSERT(std::holds_alternative<std::string>(inputs));
+    const auto& text = std::get<std::string>(inputs);
+    auto tokenized_input = m_tokenizer.encode(text);
+    auto encoded_results = generate(tokenized_input, config, streamer);
```
ilya-lavrenov, line 180: chat template is not handled here

TolyaTalamanov: Done

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -139,49 +140,51 @@ NPULLMPipelineImpl::NPULLMPipelineImpl(
...
     if (std::holds_alternative<std::vector<std::string>>(inputs)) {
-        OPENVINO_THROW("Currently only batch size=1 is supported for NPU device");
+        OPENVINO_THROW("Currently only batch size=1 is supported");
     }
 
     OPENVINO_ASSERT(std::holds_alternative<std::string>(inputs));
-    const auto& text = std::get<std::string>(inputs);
-    auto tokenized_input = m_tokenizer.encode(text);
+    auto prompt = std::get<std::string>(inputs);
+    prompt = m_tokenizer.apply_chat_template({ {{"role", "user"}, {"content", prompt}} }, true /* add_generation_prompt */);
```
ilya-lavrenov, line 181: what if chat template is not available?

TolyaTalamanov: For `StatefulLLMPipeline` it's applied unconditionally under the `is_chat_conversion`. For `StaticLLMPipeline`, chat converstation isn't supported at all, so just applied w/o any checks

TolyaTalamanov: is there any API from tokenizer to check it?

ilya-lavrenov: @pavel-esir

pavel-esir: If `is_chat_conversation` is not yet supported for `StaticLLMPipeline` then apply_chat_template should not be called inside generate. If user wishes to have a chat he/she should apply template on his prompt outside of generate, like this
image

ilya-lavrenov:
build_jenkins

src/cpp/src/llm_pipeline.cpp
```diff
@@ -9,55 +9,20 @@
...
-
 namespace ov {
 namespace genai {
 
 ov::genai::EncodedResults greedy_decoding(
-    ov::InferRequest& model_runner, 
-    ov::Tensor prompts, 
-    ov::Tensor attention_mask, 
-    const GenerationConfig sampling_params, 
-    const std::shared_ptr<StreamerBase> streamer,
```
Wovchena, line 60: Just for your future contributions (don't revert in this PR). If you want to fix trailing white-spaces, you should do it everywhere and add a check that no new spaces are added to precommit.

Inplace fixes increase the diff size. At the same time there's nothing preventing the same person who originally added white-spaces to add them again. Which would increase the diff size for that person's PR as well.

src/cpp/src/llm_pipeline.cpp
```diff
@@ -77,36 +42,32 @@ EncodedResults beam_search(
     GenerationConfig config
 );
 
-
-class LLMPipeline::LLMPipelineImpl {
+class StatefulLLMPipeline final : public LLMPipelineImplBase {
```
Wovchena, line 45: I was always curios, why would someone prohibit inheritance by adding `final`. I'm OK with that, but can you explain the reasoning? Especially for a class which isn't declared in a header.

Wovchena (approved):

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Wasn't intentional, just my editor :)

pavel-esir (approved):

dmatveev:
Oh wow, it is merged!

PR #509, opened 2024-06-14
fix greedy hook for new transformers

eaidova:


eaidova:
@wgzintel @peterchen-intel  FIY, _gready_seatch method was deprecated and in transformers 4.41 it was not used directly in model anymore, _sample method used instead

llm_bench/python/benchmark.py
```diff
@@ -103,7 +103,13 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
...
         mem_consumption.start_collect_memory_consumption()
     start = time.perf_counter()
-    result = model.generate(**input_data, max_new_tokens=int(max_output_token_size), num_beams=args['num_beams'], use_cache=True)
+    result = model.generate(
+        **input_data, 
+        generation_config=GenerationConfig(), 
+        max_new_tokens=int(max_output_token_size), 
+        num_beams=args['num_beams'], 
+        use_cache=True
+    )
```
eaidova, line 112:
```suggestion
    result = model.generate(
        **input_data,
        generation_config=GenerationConfig(),
        max_new_tokens=int(max_output_token_size),
        num_beams=args['num_beams'],
        use_cache=True
    )
```

PR #513, opened 2024-06-14
Unskip pytests

as-suvorov:


tests/python_tests/test_generate_api.py
```diff
@@ -228,18 +228,14 @@ def test_ov_tensors(model_descr, inputs):
     dict(max_new_tokens=20, num_beam_groups=3, num_beams=15, diversity_penalty=1.0)
 ]
 batched_prompts = [
-    ['table is made of', 'They sky is blue because', 'Difference between Jupiter and Mars is that'],
+    ['table is made', 'They sky is blue because', 'Difference between Jupiter and Mars is that'],
```
as-suvorov, line 231: The `table is made of` prompt diverge genai vs optimum implementation on the last token. The score of correct and incorrect beams is exactly the same. It's the same for both genai and optimum implementations. But the order of candidates beams is different. I investigated the option to align them, maybe to apply sorting by the token id. But `pytorch` `top_k` implementation is not deterministic in terms of handling duplicates: https://github.com/pytorch/pytorch/issues/27542. `torch.Tensor([1, 2, 5, 4, 5]).topk(1, largest=True, sorted=True)` might return different order of indexes, depending on device or tensor data. We could try to copy `pytorch.top_k` implementation, but I'm not sure if it's good idea.

tests/python_tests/test_generate_api.py
```diff
@@ -711,32 +707,3 @@ def test_chat_1(model_descr, generation_config):
         print(f'ov_output: {chat_history_ov}')
     assert chat_history_ov == chat_history_hf
     pipe.generate('你好！ 你好嗎？', max_new_tokens=20)
-
-
-@pytest.mark.skip(reason="probably both models ov + hf doesn't fit to memory")
-@pytest.mark.precommit
-@pytest.mark.skipif(sys.platform.startswith("win"), reason="probably not enough space for this model on Win")
-def test_left_pad():
```
as-suvorov, line 719: Test is redundant as `pad_left` implementation tested implicitly on batched prompts.

Wovchena (approved):

Wovchena (approved):

PR #514, opened 2024-06-14
device config and external token/detok for genai llm bench

eaidova:


llm_bench/python/benchmark.py
```diff
@@ -12,7 +12,7 @@
 import utils.model_utils
 import torch
 import numpy as np
-from openvino.runtime import get_version
+from openvino.runtime import get_version, Tensor
```
eaidova, line 15:
```suggestion
from openvino.runtime import get_version
```

PR #522, opened 2024-06-18
Limit transformers to versions less than 4.40

wgzintel:


eaidova:
@wgzintel  @peterchen-intel  why?

If it is due to issue with callback, then it should be fixed in https://github.com/openvinotoolkit/openvino.genai/pull/509

peterchen-intel:
Test

peterchen-intel (approved):

PR #537, opened 2024-06-20
Add config to enable dynamic quantize

byungilm:
+ Config to set dynamic_quantization_group_size for ov

### DIFF
--- a/llm_bench/python/benchmark.py
+++ b/llm_bench/python/benchmark.py
@@ -674,6 +674,15 @@ def get_argprser():
         'if the value is False (default), input prompts are processed in interleave manner'
     )
     parser.add_argument('-od', '--output_dir', help='Save the input text and generated text, images to files')
+    parser.add_argument(
+        '-dq',
+        '--dynamic_quantization_group_size',
+        default=None,
+        required=False,
+        type=int,
+        help='Set group size of dynamic quantization. Dynamic quantization for GPU is enabled if this argument is equal to or bigger than 32.',
+    )
+
     utils.model_utils.add_stateful_model_arguments(parser)
     parser.add_argument("--genai", action="store_true")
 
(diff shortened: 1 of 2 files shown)


llm_bench/python/benchmark.py
```diff
@@ -674,6 +674,15 @@ def get_argprser():
         'if the value is False (default), input prompts are processed in interleave manner'
     )
     parser.add_argument('-od', '--output_dir', help='Save the input text and generated text, images to files')
+    parser.add_argument(
+        '-dq',
+        '--dynamic_quantization_group_size',
+        default=None,
+        required=False,
+        type=int,
+        help='Set group size of dynamic quantization. Dynamic quantization for GPU is enabled if this argument is equal to or bigger than 32.',
```
isanghao, line 683: "Set group size of FC dynamic quantization" would be enough

byungilm: Fixed

byungilm (commented):
Mingyu

llm_bench/python/benchmark.py
```diff
@@ -674,6 +674,15 @@ def get_argprser():
         'if the value is False (default), input prompts are processed in interleave manner'
     )
     parser.add_argument('-od', '--output_dir', help='Save the input text and generated text, images to files')
+    parser.add_argument(
```
wgzintel, line 677: Do we need to check the value of dynamic_quantization_group_size? Can this value be less than or equal to 0?

byungilm: This value can be equal to 0 which means disable dynamic quantizing.

byungilm:
Use '-lc config.json' 
The file config.json contains :
config =  {"DYNAMIC_QUANTIZATION_GROUP_SIZE":"64"}

PR #538, opened 2024-06-20
Enable win lib build

as-suvorov:


.github/workflows/causal_lm_cpp.yml
```diff
@@ -180,8 +186,9 @@ jobs:
                 predictions = predictions[:idx] + predictions[idx + len(ref):]
           "
           echo "Multi prompt" passed
+
   cpp-greedy_causal_lm-windows:
-    if: false
+    # if: false
```
Wovchena, line 191: > Warning: Python is not installed. Please install one of Python 3.8 - 3.12 (64-bit) from https://www.python.org/downloads/

Apparently `actions/setup-python@v4` installs Python in a way that doesn't let Windows to detect it. It's desireabale to keep executing `samples\python\greedy_causal_lm\greedy_causal_lm.py` without direct call to python to verify that the approach works. Maybe you can find a way to fix that. Alternatively, please, submit an issue to `actions/setup-python` and change the test to run via `python` call.

as-suvorov:
@Wovchena , Win build has been fixed prior to this PR. I enabled `casual_lm_cpp` and `genai_package` for win in this PR. There are still 3 failing tests in `genai_python_lib` for win, I propose to fix them in the next PR.

Wovchena (approved):

PR #541, opened 2024-06-21
Fix streamer

pavel-esir:
Couldn't find string on which the original issue is reproduced, therefore couldn't add test for this case.

But stumbled across one more problem with decoding utf-8 in streamer (similar to previous issue with utf-8 strings in results). Added case for the latest problem.

ticket CVS-144047

### DIFF
--- a/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
+++ b/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
@@ -37,7 +37,7 @@ struct TextStreamer {
     void put(int64_t token) {
         token_cache.push_back(token);
         std::string text = detokenize(detokenizer, token_cache);
-        if (!text.empty() && '\n' == text.back()) {
+        if (!text.empty() && '\n' == text.back() && text.size() > print_len) {
             // Flush the cache after the new line symbol
             std::cout << std::string_view{text.data() + print_len, text.size() - print_len};
             token_cache.clear();
--- a/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
+++ b/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
@@ -47,9 +47,9 @@ struct TextStreamer {
         if (text.size() >= 3 && text.compare(text.size() - 3, 3, "�") == 0) {
             // Don't print incomplete text
             return;
-        } else if (text.size() > print_len && std::any_of(text.begin() + print_len, text.end(), [](char c) { return c == ' '; })) {
+        } else if (text.size() > print_len) {
             // It is possible to have a shorter text after adding new token.
-            // Print to output only if text lengh is increaesed and if containt full words.
+            // Print to output only if text lengh is increaeseds.
             std::cout << std::string_view{text.data() + print_len, text.size() - print_len} << std::flush;
             print_len = text.size();
         }
@@ -57,7 +57,7 @@ struct TextStreamer {
 
     void end() {
         std::string text = detokenize(detokenizer, token_cache);
-        if (text.size() < print_len)
+        if (text.size() <= print_len)
             return ;
         std::cout << std::string_view{text.data() + print_len, text.size() - print_len} << '\n';
         token_cache.clear();
--- a/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
+++ b/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
@@ -47,13 +47,18 @@ struct TextStreamer {
         if (text.size() >= 3 && text.compare(text.size() - 3, 3, "�") == 0) {
             // Don't print incomplete text
             return;
+        } else if (text.size() > print_len && std::any_of(text.begin() + print_len, text.end(), [](char c) { return c == ' '; })) {
+            // It is possible to have a shorter text after adding new token.
+            // Print to output only if text lengh is increaesed and if containt full words.
+            std::cout << std::string_view{text.data() + print_len, text.size() - print_len} << std::flush;
+            print_len = text.size();
         }
-        std::cout << std::string_view{text.data() + print_len, text.size() - print_len} << std::flush;
-        print_len = text.size();
     }
 
     void end() {
         std::string text = detokenize(detokenizer, token_cache);
+        if (text.size() < print_len)
+            return ;
         std::cout << std::string_view{text.data() + print_len, text.size() - print_len} << '\n';
         token_cache.clear();
         print_len = 0;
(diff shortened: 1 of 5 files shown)


samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
```diff
@@ -47,13 +47,18 @@ struct TextStreamer {
         if (text.size() >= 3 && text.compare(text.size() - 3, 3, "�") == 0) {
             // Don't print incomplete text
             return;
+        } else if (text.size() > print_len && std::any_of(text.begin() + print_len, text.end(), [](char c) { return c == ' '; })) {
+            // It is possible to have a shorter text after adding new token.
+            // Print to output only if text lengh is increaesed and if containt full words.
```
Wovchena, line 52: While you can't add it to tests (I guess because permission is need to download llama3), can you describe an example in code comments. For example, requirement for the increasing sequence is clear (although example would be nice anyway), but why is space required?

samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
```diff
@@ -47,13 +47,18 @@ struct TextStreamer {
...
+            print_len = text.size();
         }
-        std::cout << std::string_view{text.data() + print_len, text.size() - print_len} << std::flush;
-        print_len = text.size();
     }
 
     void end() {
         std::string text = detokenize(detokenizer, token_cache);
+        if (text.size() < print_len)
+            return ;
```
Wovchena, line 61: 1. The inversion of `text.size() > print_len` from `put()` is `text.size() <= print_len`, not `text.size() < print_len`.
2. Why isn't space required here?

src/python/py_generate_pipeline.cpp
```diff
@@ -25,8 +25,12 @@ using ov::genai::StringInputs;
 using ov::genai::TokenizedInputs;
 using ov::genai::Tokenizer;
 
+// When StreamerVariant is used utf-8 decoding is done by pybind and can lead to exception on incomplete texts.
+// Therefore strings decoding should be handled with PyUnicode_DecodeUTF8(..., "replace") to not throw errors.
+using PyBindStreamerVariant = std::variant<std::function<py::object(py::object)>, std::shared_ptr<StreamerBase>, std::monostate>;
```
Wovchena, line 30: Why not
```suggestion
using PyBindStreamerVariant = std::variant<std::function<bool(std::string)>, std::shared_ptr<StreamerBase>, std::monostate>;
```
or stl's pybind analogs?

src/python/py_generate_pipeline.cpp
```diff
@@ -263,30 +267,54 @@ py::object call_common_generate(
...
+        // Wrap python streamer with manual utf-8 decoding. Do not rely
+        // on pybind automatic decoding since it raises exceptions on incomplete strings.
+        auto callback_wrapped = [&py_callback](std::string subword) -> bool {
+            auto py_str = PyUnicode_DecodeUTF8(subword.data(), subword.length(), "replace");
+            return py_callback(py::reinterpret_borrow<py::object>(py_str)).cast<bool>();
+        };
+        streamer = callback_wrapped;
+    },
+    [&streamer](std::shared_ptr<StreamerBase> streamer_cls){
+        streamer = streamer_cls;
```
Wovchena, line 289: It seems StreamerBase must also perform custom `PyUnicode_DecodeUTF8()`

Wovchena (changes requested):

samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
pavel-esir: I took already existing solution CVS-143742 As i understood we check for spaces to print only if there are complete words, but i think such strict check it's not necessary, not i removed it.

I don't face this error locally on the reported model and prompt (the code below run on master)
image

I was told that in this dataset 8 of 1000 prompts have this issue 
https://huggingface.co/datasets/anon8231489123/ShareGPT_Vicuna_unfiltered/blob/main/ShareGPT_V3_unfiltered_cleaned_split.json

I will cycle through prompts from there to reproduce issue and add example with problematic, and will add test at least to check it locally.

samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
pavel-esir: 1. thanks for noticing, fixed
2. spaces are not strictly required, removed them from above

src/python/py_generate_pipeline.cpp
pavel-esir: With you suggestion `PyBindStreamerVariant` would be exactly equal to `StreamerVariant`. But if we leave `StreamerVariant` then while casing c++ std::string -> py::object / py::string for `lambda x: print(x)` it will raise exceptions for incomplete words.

Therefore object for python streamer should accept some python type as argument and we cast our std::string to python type manually.

>or stl's pybind analogs?
agreed with this one, changed to `std::variant<std::function<bool(py::str)>` to be more specific

src/python/py_generate_pipeline.cpp
pavel-esir: StreamerBase works only with encoded token ids, therefore there are no strings and utf problems.

I user wished to write his own streamer, inherits StreamerBase and detokenizes with our tokenizer then we will handle that
https://github.com/openvinotoolkit/openvino.genai/blob/master/src/python/py_generate_pipeline.cpp#L423-L427

samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.cpp
Wovchena: I encourage you can ask why they used spaces in their fix

pavel-esir: i didn't ask yet, but i suspected they looked to HF
https://github.com/huggingface/transformers/blob/main/src/transformers/generation/streamers.py#L108-L112

pavel-esir: i found prompt on which issue is reproduced. This 
`'I have an interview about product speccing with the company Weekend Health. Give me an example of a question they might ask with regards about a new feature'` generates sequence which can become shorter when new tokens are added

image

Added test `test_callback_decoding_metallama` for that, but on precommit it's skipped. Also added unicode decoding test for tokenizer, but xfailed them until we enable ov::Tensor in pybind.

Wovchena (approved):

ilya-lavrenov (approved):

PR #549, opened 2024-06-24
improve chat template processing

pavel-esir:
**TODO**
- [x] - add chat template applying tests
- [x] - throw informative exception if Jinja2Cpp was not able to process template

ticket: CVS-143685


pavel-esir:
please do not review, PR is not ready yet

pavel-esir:
PR is ready for review

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -366,10 +348,12 @@ void update_position_ids(ov::Tensor&& position_ids, const ov::Tensor&& attention
...
 namespace genai {
 
-EncodedResults beam_search(ov::InferRequest& lm,
+std::pair<EncodedResults, std::vector<int32_t>> beam_search(ov::InferRequest& lm,
                            ov::Tensor input_ids,
                            ov::Tensor attention_mask,
-                           GenerationConfig config) {
+                           GenerationConfig config, 
+                           std::optional<ov::Tensor> position_ids = std::nullopt,
+                           std::optional<std::vector<int32_t>> selected_beam_idx = std::nullopt) {
```
ilya-lavrenov, line 356: do we assume that beam search and chat scenario are compatible?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -206,21 +209,53 @@ class LLMPipeline::LLMPipelineImpl {
...
+
+            ov::Tensor new_atten_mask =  ov::Tensor{ov::element::i64, {batch_size, kv_cache_len + prompt_len}};
+            std::copy(atten_mask_history.data<int64_t>(), atten_mask_history.data<int64_t>() + kv_cache_len,
+                    new_atten_mask.data<int64_t>());
+            std::copy(attention_mask.data<int64_t>(), attention_mask.data<int64_t>() + prompt_len,
+                    new_atten_mask.data<int64_t>() + kv_cache_len);
+            atten_mask = new_atten_mask;
+        } else {
+            atten_mask = attention_mask;
+        }
```
ilya-lavrenov, line 231: attention mask and position ids are specific for stateful and static model runners, while they are not required for PA-based.
I would move it to model runners / pipelines to avoid complexity in generic LLM pipeline impl.

But maybe it should be done during migration to CB rails
CC @Wovchena

src/cpp/src/tokenizer.cpp
```diff
@@ -345,9 +381,19 @@ class Tokenizer::TokenizerImpl {
...
             {"pad_token", m_pad_token},
+            {"system_message", system_message.empty() ? jinja2::EmptyValue() : jinja2::Value{system_message}},
             {"add_generation_prompt", add_generation_prompt},
         };
-        return tpl.RenderAsString(params).value();
+        std::string history_txt;
+        try {
+            history_txt = tpl.RenderAsString(params).value();
+        } catch (const std::exception& error) {
+            OPENVINO_THROW("chat_template is not supported please "
```
ilya-lavrenov, line 391: can we write that it's not supported for this specific model? to avoid confusion that it's not supported at all

E.g. we can also write that it's not supported as JinjaCpp limitation.

tests/python_tests/test_generate_api.py
```diff
@@ -742,8 +742,8 @@ def test_unicode_pybind_decoding_3():
 ]
 
 configs = [
-    dict(max_new_tokens=500),
-    # dict(num_beam_groups=3, num_beams=15, num_return_sequences=15, max_new_tokens=20, diversity_penalty=1.0)
+    # dict(max_new_tokens=500),
+    dict(num_beam_groups=3, num_beams=15, num_return_sequences=1, max_new_tokens=500, diversity_penalty=1.0)
```
ilya-lavrenov, line 746: should we enable both configs? 
is 500 required for func tests? can we have lower value to speed up tests execution?

tests/python_tests/test_generate_api.py
```diff
@@ -795,7 +794,144 @@ def test_chat_1(model_descr, generation_config):
...
+    {'role': 'user', 'content': 'What is the previous answer?'},
+    {'role': 'assistant', 'content': 'The previous answer was: 1 + 1 = 2'},
+    {'role': 'user', 'content': 'Why is the sun yellow?'},
+    {'role': 'assistant', 'content': 'Because it emits yeloow light.'},
+    {'role': 'user', 'content': 'What was my first question?'},
+    {'role': 'assistant', 'content': "I don't remember such difficult questions.\n Ask something different"},
+]
+@pytest.mark.precommit
+@pytest.mark.parametrize('chat_config', chat_configs)
+def test_apply_chat_template(model_tmp_path, chat_config):
```
ilya-lavrenov, line 913: can we split big `tests/python_tests/test_generate_api.py` into files?
E.g. chat template is worth to have its dedicated file.

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -452,9 +449,10 @@ EncodedResults beam_search(ov::InferRequest& lm,
         ) {
             results.scores.push_back(beam->get().score);
             results.tokens.push_back(std::move(beam->get().tokens));
+            selected_beams.push_back(beam->get().global_beam_idx);
         }
     }
-    return results;
+    return {results, selected_beams};
```
ilya-lavrenov, line 455: should we return a single element? E.g. beam idx with highest score.
I don't see how multiple idx could be used on successive iterations

src/cpp/src/group_beam_searcher.cpp
pavel-esir: it's possible to use with beam search as well. I couldn't verify it because of the problem of several calls of beam search, but it should work

tests/python_tests/test_generate_api.py
pavel-esir: I committed this by mistake when i was checking beam search. But at the moment beam search fails on several calls. At the end i will uncomment both of them.

I will decrease max_new_tokens for speed.

src/cpp/src/group_beam_searcher.cpp
ilya-lavrenov: ok, probably, I meant chat scenario with streamer and beam search.
Such scenario does not work, right?

src/cpp/src/llm_pipeline.cpp
```diff
@@ -206,21 +209,53 @@ class LLMPipeline::LLMPipelineImpl {
                         "(input_ids, attention_mask, position_ids, beam_idx) "
                         "but you have '" + std::to_string(num_inputs) + "' inputs");
 
+        
+        size_t kv_cache_len = 0;
+        ov::Tensor atten_mask;
```
Wovchena, line 214:
```suggestion
        ov::Tensor concatenated_attention_mask;
```

src/cpp/src/llm_pipeline.cpp
```diff
@@ -206,21 +209,53 @@ class LLMPipeline::LLMPipelineImpl {
...
-            result = ov::genai::greedy_decoding(m_model_runner, input_ids, attention_mask, 
-                                                config, streamer_ptr, 
-                                                is_chat_conversation, m_is_cache_empty);
+            result = ov::genai::greedy_decoding(m_model_runner, input_ids, atten_mask, 
+                                                config, streamer_ptr, position_ids);
         } else if (config.is_beam_search()) {
-            result = beam_search(m_model_runner, input_ids, attention_mask, config);
+            auto res = beam_search(m_model_runner, input_ids, atten_mask, config, position_ids, m_selected_beams);
+            result = res.first;
+            m_selected_beams = res.second;
```
Wovchena, line 248:
```suggestion
           std::tie(result, m_selected_beams) = beam_search(m_model_runner, input_ids, atten_mask, config, position_ids, m_selected_beams);
```

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -452,9 +449,10 @@ EncodedResults beam_search(ov::InferRequest& lm,
         ) {
             results.scores.push_back(beam->get().score);
             results.tokens.push_back(std::move(beam->get().tokens));
+            selected_beams.push_back(beam->get().global_beam_idx);
```
Wovchena, line 452: `selected_beam_idx.size()` is expected to be `1` because every beam is spawned from a single prompt. But `selected_beams` can be pushed to multiple times at this line. So the next `beam_search()` will fail at `OPENVINO_ASSERT((*selected_beam_idx).size() == 1);` for chat.

`selected_beam_idx` shouldn't be a vector. It's just `std::optional<int32_t>`.

Is the way to propagate an accepted by the user beam missing for chat? (https://github.com/openvinotoolkit/openvino.genai/pull/549/files#r1654606288 is probably the same question)

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -385,7 +369,21 @@ EncodedResults beam_search(ov::InferRequest& lm,
...
+    bool position_ids_available = position_ids.has_value();
+    if (position_ids_available)
+        lm.set_tensor("position_ids", *position_ids);
+
+    ov::Tensor beam_idx = lm.get_tensor("beam_idx");
+    beam_idx.set_shape({batch_size});
+    auto beam_data = beam_idx.data<int32_t>();
+    if (selected_beam_idx.has_value()) {
+        OPENVINO_ASSERT((*selected_beam_idx).size() == 1);
+        beam_data[0] = (*selected_beam_idx).at(0);
```
Wovchena, line 383: `OPENVINO_ASSERT(batch_size == 1, "continuation of generation is possible only for batch 1");` feels completely arbitrary when it's in `generate()`. Moving it there would explain that it's not possible because you can't fill the rest of `beam_data`.

Wovchena (changes requested):

src/cpp/src/llm_pipeline.cpp
Wovchena: There's no other way for now

src/cpp/src/tokenizer.cpp
```diff
@@ -345,9 +381,19 @@ class Tokenizer::TokenizerImpl {
             {"bos_token",  m_bos_token},
             {"eos_token", m_eos_token},
             {"pad_token", m_pad_token},
+            {"system_message", system_message.empty() ? jinja2::EmptyValue() : jinja2::Value{system_message}},
             {"add_generation_prompt", add_generation_prompt},
         };
-        return tpl.RenderAsString(params).value();
+        std::string history_txt;
```
pavel-esir, line 387:
```suggestion
```

src/cpp/src/group_beam_searcher.cpp
Wovchena: beam search fix was merged

src/cpp/src/llm_pipeline.cpp
pavel-esir: done

src/cpp/src/llm_pipeline.cpp
```diff
@@ -206,21 +209,53 @@ class LLMPipeline::LLMPipelineImpl {
...
+        } else {
+            atten_mask = attention_mask;
+        }
+
+        bool position_ids_available = (num_inputs == 4);
+        std::optional<ov::Tensor> position_ids = std::nullopt;
+        if (position_ids_available) {
+            position_ids = ov::Tensor{ov::element::i64, input_ids.get_shape()};
+            utils::initialize_position_ids(*position_ids, attention_mask, kv_cache_len);
+            // m_model_runner.set_tensor("position_ids", *position_ids);
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/llm_pipeline.cpp
pavel-esir: done

src/cpp/src/tokenizer.cpp
```diff
@@ -306,36 +306,72 @@ class Tokenizer::TokenizerImpl {
...
                                     bool add_generation_prompt, 
                                     const std::string& chat_template) const {
+        auto chat_tpl = chat_template.empty() ? m_chat_template : chat_template;
+        // Jinja2Cpp does not support slicing, e.g. [1:].
+        // In templates slicing is used typically in the header to find system prompt.
+        // If header containt that typical expression we update template and 
+        // extract system message manually from ChatHistory.
+        std::string header_with_slice = "{% if messages[0]['role'] == 'system' %}{% set loop_messages = messages[1:] %}{% set system_message = messages[0]['content'] %}";
+        std::string replacement_string = "{% if false %}{% set placeholder = false %}";
+        ChatHistory modified_history = history;
```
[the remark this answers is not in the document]

src/cpp/src/tokenizer.cpp
```diff
@@ -345,9 +381,19 @@ class Tokenizer::TokenizerImpl {
             {"bos_token",  m_bos_token},
             {"eos_token", m_eos_token},
             {"pad_token", m_pad_token},
+            {"system_message", system_message.empty() ? jinja2::EmptyValue() : jinja2::Value{system_message}},
             {"add_generation_prompt", add_generation_prompt},
         };
-        return tpl.RenderAsString(params).value();
+        std::string history_txt;
+        try {
+            history_txt = tpl.RenderAsString(params).value();
```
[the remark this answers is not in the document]
pavel-esir: accepted option with `return tpl.RenderAsString(params).value();`

src/cpp/src/group_beam_searcher.cpp
pavel-esir: I didn't quite understand. You mean it would be better to remove existing assert for batch_size == 1 from llm_pipeline and leave restriction here?

Anyways, i changed signature. Now only one selected beam idx (in case if num_return_sequences == 1) is returned and accepted, so this assert is removed.

src/cpp/src/llm_pipeline.cpp
```diff
@@ -314,14 +349,22 @@ ov::genai::Tokenizer ov::genai::LLMPipeline::get_tokenizer() {
...
         m_pimpl->m_is_cache_empty = true;
     }
 }
 
 void ov::genai::LLMPipeline::finish_chat() {
     m_pimpl->is_chat_conversation = false;
+    
+    (*m_pimpl->m_selected_beams).clear();
+    m_pimpl->m_selected_beams  = std::nullopt;
+
```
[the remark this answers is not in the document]
pavel-esir: done

tests/python_tests/test_generate_api.py
```diff
@@ -795,7 +794,144 @@ def test_chat_1(model_descr, generation_config):
...
+    openvino.save_model(ov_detokenizer, path / "openvino_detokenizer.xml")
+    
+    pipe = ov_genai.LLMPipeline(str(path), device, config={"ENABLE_MMAP": False})
+    pipe_with_kv_cache = ov_genai.LLMPipeline(str(path), device, config={"ENABLE_MMAP": False})
+
+    pipe_with_kv_cache.start_chat()
+    for question in quenstions:
+        chat_history_with_kv_cache.append({'role': 'user', 'content': question})
+        chat_history_ov.append({'role': 'user', 'content': question})
+        # pytest.set_trace()
```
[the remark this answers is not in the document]
pavel-esir: it was left by mistake. removed

tests/python_tests/test_generate_api.py
```diff
@@ -795,7 +794,144 @@ def test_chat_1(model_descr, generation_config):
...
+configs = [
+    dict(max_new_tokens=500),
+    # dict(num_beam_groups=2, num_beams=4, num_return_sequences=1, max_new_tokens=500, diversity_penalty=1.0)
+]
+@pytest.mark.parametrize("config", configs)
+@pytest.mark.parametrize("model_descr", chat_models_list())
+@pytest.mark.precommit
+@pytest.mark.skipif(sys.platform == "linux", reason="no space left on linux device for chat models")
+# Check that when history is stored in KV cache results are the same as when history stored in text.
+def test_chat_2(model_descr, config):
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/group_beam_searcher.cpp
pavel-esir: changed to return a single selected_beam

tests/python_tests/test_generate_api.py
pavel-esir: Done. Split to 2 files. Common utilities are moved to `ov_genai_test_utils`

src/cpp/src/group_beam_searcher.cpp
pavel-esir: >selected_beam_idx shouldn't be a vector. It's just std::optional<int32_t>.

done

src/cpp/src/group_beam_searcher.cpp
pavel-esir: I have an issue with beam search. Unfortunately it does not works for chat. Firstly, it returned output different that in HF. 

Now, when `selected_beam_idx` is a single int, it fails because of the classical `Eltwise shape infer input shapes dim index: 3 mismatch`. I am trying to root cause the issue

src/cpp/src/group_beam_searcher.cpp
Wovchena: > I didn't quite understand. You mean it would be better to remove existing assert for batch_size == 1 from llm_pipeline and leave restriction here?

Yes

src/cpp/src/group_beam_searcher.cpp
pavel-esir: fixed that issue

src/cpp/src/tokenizer.cpp

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,182 @@
...
+    if chat_history_ov != chat_history_hf:
+        print(f'hf_output: {chat_history_hf}')
+        print(f'ov_output: {chat_history_ov}')
+    assert chat_history_ov == chat_history_hf
+
+
+@pytest.mark.parametrize("generation_config", configs)
+@pytest.mark.parametrize("model_descr", get_chat_models_list())
+@pytest.mark.precommit
+# @pytest.mark.skipif(sys.platform == "linux", reason="no space left on linux device for chat models")
```
pavel-esir, line 79:
```suggestion
```

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,182 @@
...
```
pavel-esir, line 119:
```suggestion
```

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,182 @@
...
+    'What is the previous answer?',
+    'Why is the Sun yellow?',
+    'What was my first question?'
+]
+
+
+@pytest.mark.parametrize("generation_config", configs)
+@pytest.mark.parametrize("model_descr", get_chat_models_list())
+@pytest.mark.precommit
+# @pytest.mark.skipif(sys.platform == "linux", reason="no space left on linux device for chat models")
```
pavel-esir, line 38:
```suggestion
```

pavel-esir (commented):
removed leftover commented lines

src/cpp/src/group_beam_searcher.cpp
pavel-esir: Communicated offline. We have limitations that do not allow continuation for greedy as well
We drop batches which met eos 
https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/greedy_decoding.cpp#L128-L132
 Therefore assert is in the common `llm_pipeline.cpp`

src/cpp/src/llm_pipeline_base.hpp
Wovchena, line 1: Please, add a docstring that a sequence of generate() calls with beam_search and chat started selects the best scored beam as an accepted model response between generate() calls.

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,179 @@
...
+@pytest.mark.precommit
+def test_chat_compare_with_HF(model_descr, generation_config: Dict):
+    device = 'CPU'
+    chat_history_hf = []
+    chat_history_ov = []
+    chat_prompt = ''
+    model_id, path, tokenizer, model_opt, pipe = read_model(model_descr)
+
+    # HF in chat scenario does not add special tokens, but openvino tokenizer by default is converted with add_special_tokens=True.
+    # Since To match need to regenerate openvino_tokenizer/detokenizer.
```
Wovchena, line 46:
```suggestion
    # Need to regenerate openvino_tokenizer/detokenizer.
```
Same below

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,179 @@
...
+    ov_tokenizer, ov_detokenizer = openvino_tokenizers.convert_tokenizer(tokenizer, add_special_tokens=False, with_detokenizer=True)
+    openvino.save_model(ov_tokenizer, path / "openvino_tokenizer.xml")
+    openvino.save_model(ov_detokenizer, path / "openvino_detokenizer.xml")
```
Wovchena, line 49: Can overriding the tokenizers break the following tests? Same below

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,179 @@
...
+    if chat_history_ov != chat_history_hf:
+        print(f'hf_output: {chat_history_hf}')
+        print(f'ov_output: {chat_history_ov}')
+    assert chat_history_ov == chat_history_hf
+
+
+@pytest.mark.parametrize("generation_config", configs)
+@pytest.mark.parametrize("model_descr", get_chat_models_list())
+@pytest.mark.precommit
+def test_chat_compare_statefull_vs_text_history(model_descr, generation_config: Dict):
```
Wovchena, line 117: This seems to be redundant: if `chat_history_ov == chat_history_hf with add_generation_prompt=False` and `chat_history_ov == chat_history_hf with add_generation_prompt=True` then `chat_history_ov with add_generation_prompt=False` == `chat_history_ov with add_generation_prompt=False`.

Wovchena (changes requested):

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -0,0 +1,219 @@
...
+    return (
+        model_id,
+        path,
+        tokenizer,
+        opt_model,
+        ov_genai.LLMPipeline(str(path), device='CPU', config={"ENABLE_MMAP": False}),
+    )
+
+
+def stop_criteria_map():
```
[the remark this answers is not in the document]
pavel-esir: I don't like global vars in Python since anyone can modify them. Can we keep this dict in this method?

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -0,0 +1,219 @@
...
+        "HuggingFaceH4/zephyr-7b-gemma-sft-v0.1",
+        "MediaTek-Research/Breeze-7B-Instruct-v0_1",
+        "shanchen/llama3-8B-slerp-biomed-chat-chinese",
+        "MLP-KTLim/llama-3-Korean-Bllossom-8B",
+        "lucyknada/microsoft_WizardLM-2-7B",
+        "aloobun/CosmicBun-8B",
+        "codellama/CodeLlama-70b-Instruct-hf",
+        "gorilla-llm/gorilla-openfunctions-v2",
+        "BramVanroy/Llama-2-13b-chat-dutch"
+    ]
```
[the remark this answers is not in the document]
pavel-esir: done

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -0,0 +1,219 @@
...
+                                                                             add_special_tokens=True, 
+                                                                             with_detokenizer=True)
+        openvino.save_model(ov_tokenizer, path / "openvino_tokenizer.xml")
+        openvino.save_model(ov_detokenizer, path / "openvino_detokenizer.xml")
+        
+        # to store tokenizer config jsons with special tokens
+        tokenizer.save_pretrained(path)
+        
+        opt_model = OVModelForCausalLM.from_pretrained(model_id, export=True, trust_remote_code=True, 
+                                                       compile=False, device='CPU', load_in_8bit=False)
```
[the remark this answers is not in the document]
pavel-esir: I just moved this piece without changes. If model was already `saved_pretrained` no need to download it again. I suspect that even if we always do `OVModelForCausalLM.from_pretrained(model_id..` it will not download but use model form cache but still I see no harm in the existing code. Can we keep it?

src/cpp/src/llm_pipeline_base.hpp
```diff
@@ -29,7 +29,7 @@ class LLMPipelineImplBase {
         StreamerVariant streamer
     ) = 0;
 
-    virtual void start_chat() = 0;
+    virtual void start_chat(std::string system_message) = 0;
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/llm_pipeline_base.hpp
pavel-esir: done

src/cpp/src/llm_pipeline.cpp
```diff
@@ -199,41 +202,86 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+        ov::Tensor concatenated_attention_mask;
+        if (is_chat_conversation && !m_is_cache_empty) {
+            OPENVINO_ASSERT(batch_size == 1, "continuation of generation is possible only for batch 1");
+            // If history is saved in KV cache concatenate new attention_mask with the already existing.
+            // Between subsequent runs attention_mask should not be modified.
+            auto atten_mask_history = m_model_runner.get_tensor("attention_mask");
+            auto prompt_len = attention_mask.get_shape()[1];
+            kv_cache_len = atten_mask_history.get_shape()[1];
+
+            ov::Tensor new_atten_mask =  ov::Tensor{ov::element::i64, {batch_size, kv_cache_len + prompt_len}};
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -407,20 +402,22 @@ EncodedResults beam_search(ov::InferRequest& lm,
...
-    auto num_inputs = lm.get_compiled_model().inputs().size();
-    bool position_ids_available = num_inputs == 4;
     
     for (size_t length_count = 0; length_count < parameters.max_new_tokens; ++length_count) {
         lm.infer();
 
         std::tie(next_tokens, next_beams) = group_beam_searcher.select_next_tokens(lm.get_tensor("logits"));
-        if (next_tokens.empty()) {
+        if (next_tokens.empty() || length_count == parameters.max_new_tokens - 1) {
+            // Break the cycle right after we got next tokens but before masks are extended.
```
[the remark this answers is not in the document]
pavel-esir: removed

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -392,7 +375,19 @@ EncodedResults beam_search(ov::InferRequest& lm,
         prompts.push_back(std::vector<int64_t>{prompt_start, prompt_start + sequence_length});
     }
 
-    initialize_inputs(input_ids, attention_mask, lm);
+    lm.set_tensor("input_ids", input_ids);
+    lm.set_tensor("attention_mask", attention_mask);
+    bool position_ids_available = position_ids.has_value();
+    if (position_ids_available)
+        lm.set_tensor("position_ids", *position_ids);
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/greedy_decoding.cpp
```diff
@@ -12,56 +12,24 @@ EncodedResults greedy_decoding(
     ov::Tensor input_ids, 
     ov::Tensor attention_mask, 
     const ov::genai::GenerationConfig generation_config, 
-    const std::shared_ptr<StreamerBase> streamer, 
-    const bool is_chat_conversation,
-    const bool is_cache_empty
+    const std::shared_ptr<StreamerBase> streamer,
+    std::optional<ov::Tensor> position_ids = std::nullopt
```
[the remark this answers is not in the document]
pavel-esir: done

.github/workflows/genai_python_lib.yml
```diff
@@ -33,6 +33,7 @@ jobs:
       - run: source ./ov/setupvars.sh && PYTHONPATH=./build/:$PYTHONPATH python -m pytest ./tests/python_tests/test_generate_api.py -m precommit
       - run: source ./ov/setupvars.sh && python -m pip install . --verbose
       - run: python -m pytest ./tests/python_tests/test_generate_api.py -m precommit
+      - run: python -m pytest ./tests/python_tests/test_chat_generate_api.py -m precommit
```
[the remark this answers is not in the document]
pavel-esir: done

tests/python_tests/test_chat_generate_api.py
pavel-esir: There are some tests where HF does not match to genai in both `test_chat_compare_text_history_with_HF` `test_chat_compare_with_HF`. This can happen either generate was wrong, either we did something wrong with keeping/restoring history.

To ensure that we correctly kept history in KC cache and continued i added this test as well, it still can be greed while others are red.

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -462,9 +460,14 @@ EncodedResults beam_search(ov::InferRequest& lm,
         ) {
             results.scores.push_back(beam->get().score);
             results.tokens.push_back(std::move(beam->get().tokens));
+            selected_beams.push_back(beam->get().global_beam_idx);
         }
     }
-    return results;
+    std::optional<int32_t> selected_beam = std::nullopt;
+    if (batch_size == 1)
+        selected_beam = selected_beams[0];
```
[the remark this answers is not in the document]
pavel-esir: done

.github/workflows/genai_python_lib.yml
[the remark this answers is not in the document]
Wovchena: This covers continuous batching workflows added by @as-suvorov, please delete dedicated workflows testing python continuous batching part.

Why do you keep `-m precommit`? There's already `addopts = -m precommit` added by you in `pytest.ini`.

src/cpp/src/greedy_decoding.cpp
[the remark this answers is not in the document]
Wovchena: There are more of them

src/cpp/src/group_beam_searcher.cpp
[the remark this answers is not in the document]
Wovchena: Why did you skip
> And here you decide whether `selected_beam` or `std::nullopt` should be returned.

part? I thought it's needed to trigger `std::fill_n(beam_data, batch_size, 0);` appropriately.

tests/python_tests/ov_genai_test_utils.py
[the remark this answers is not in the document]
Wovchena: Well, yes. Why would you take away such a possibility? :) It offers an opportunity to hack a test if needed. Plus you seem to be fine having global vars in `test_chat_generate_api.py`.

src/cpp/src/llm_pipeline.cpp
```diff
@@ -199,41 +202,86 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
                         "(input_ids, attention_mask, position_ids, beam_idx) "
                         "but you have '" + std::to_string(num_inputs) + "' inputs");
 
+        
+        size_t kv_cache_len = 0;
+        ov::Tensor concatenated_attention_mask;
+        if (is_chat_conversation && !m_is_cache_empty) {
+            OPENVINO_ASSERT(batch_size == 1, "continuation of generation is possible only for batch 1");
+            // If history is saved in KV cache concatenate new attention_mask with the already existing.
```
Wovchena, line 210:
```suggestion
            // If history is saved in KV cache, concatenate new attention_mask with the already existing.
```

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -0,0 +1,219 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import pathlib
+import os
+import pytest
+import functools
+import openvino
+import openvino_tokenizers
+import openvino_genai as ov_genai
+from typing import Union, List, Dict, Tuple, Optional
```
Wovchena, line 11:
```suggestion
from typing import List, Tuple
```

tests/python_tests/test_chat_generate_api.py
```diff
@@ -0,0 +1,185 @@
+# Copyright (C) 2023-2024 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import openvino
+import openvino_tokenizers
+import openvino_genai as ov_genai
+import pytest
+from typing import Union, List, Dict, Tuple, Optional
+import sys
```
Wovchena, line 9:
```suggestion
from typing import Dict, Tuple
```

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -0,0 +1,219 @@
...
+
+    nightly_models = [
+        "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+        "meta-llama/Meta-Llama-3-8B-Instruct",
+        "meta-llama/Llama-2-7b-chat-hf",
+        # "google/gemma-2b-it",  # Cannot be downloaded without access token
+        # "google/gemma-7b-it",  # Cannot be downloaded without access token
+    ]
+
+    if pytest.run_marker == "precommit":
```
Wovchena, line 70: `python tests/python_tests/test_chat_generate_api.py -m precommit` results in
```
Traceback (most recent call last):
  File "C:\Users\vzlobin\r\g\tests\python_tests\test_chat_generate_api.py", line 33, in <module>
    @pytest.mark.parametrize("model_descr", get_chat_models_list())
                                            ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\vzlobin\r\g\tests\python_tests\ov_genai_test_utils.py", line 70, in get_chat_models_list
    if pytest.run_marker == "precommit":
       ^^^^^^^^^^^^^^^^^
AttributeError: module 'pytest' has no attribute 'run_marker'
```

How do I run it?

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -215,7 +215,20 @@ class OPENVINO_GENAI_EXPORTS LLMPipeline {
     GenerationConfig get_generation_config() const;
     void set_generation_config(const GenerationConfig& config);
 
-    void start_chat();
+
+    /**
+    * @brief start chat withh keeping history in kv cache.
```
Wovchena, line 220:
```suggestion
    * @brief start chat with keeping history in kv cache.
```

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -215,7 +215,20 @@ class OPENVINO_GENAI_EXPORTS LLMPipeline {
...
+    * Turns on keeping KV cache between generate calls and automatic applying of chat templates.
+    * In case if beam search is used KV cache is kept fot the generated sequence with maximal scores.
```
Wovchena, line 222:
```suggestion
    * In case if beam search is used, KV cache is kept fot the generated sequence with maximal scores.
```

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -407,24 +401,25 @@ EncodedResults beam_search(ov::InferRequest& lm,
...
     
-    for (size_t length_count = 0; length_count < parameters.max_new_tokens; ++length_count) {
+    for (size_t length_count = 0; ; ++length_count) {
         lm.infer();
 
         std::tie(next_tokens, next_beams) = group_beam_searcher.select_next_tokens(lm.get_tensor("logits"));
-        if (next_tokens.empty()) {
+        if (next_tokens.empty() || length_count == parameters.max_new_tokens - 1) {
+            // Break the cycle before masks are extended in update_attention_mask_with_beams.
+            // If generation will be contined attention_mask length should be equal to KV cache size.
```
Wovchena, line 411:
```suggestion
            // If generation is continued, attention_mask length should be equal to KV cache size.
```

Wovchena (changes requested):

tests/python_tests/ov_genai_test_utils.py
pavel-esir: It should've worked. For some reason `run_marker` was not set. 
https://github.com/pavel-esir/openvino.genai/blob/2d90fe1ea981c29d181a8f6894040b73b458448e/tests/python_tests/conftest.py#L17-L19

It works on my machine and seems to work on pre-commit
image

tests/python_tests/ov_genai_test_utils.py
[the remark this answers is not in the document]
pavel-esir: Fair enough :) Will do that

.github/workflows/genai_python_lib.yml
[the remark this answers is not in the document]
pavel-esir: Agreed, since `python -m pytest ./tests/python_tests/` covers them as well. Deleted separate jobs for continuous batching.

I used `-m precommit' by inertia :) Removed since it's not necessary

tests/python_tests/ov_genai_test_utils.py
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/greedy_decoding.cpp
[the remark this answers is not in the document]
pavel-esir: For position_ids every redundant nullopt was deleted. I found unnecessary default `nullopt` for `selected_beam_idx` and deleted them as well.

src/cpp/src/group_beam_searcher.cpp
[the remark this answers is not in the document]
pavel-esir: `to trigger std::fill_n(beam_data, batch_size, 0)` input argument is set on the very first generate call. At that poing we know for sure there are no selected beams and value is not set.

I just realized that in return `std::optional` is not necessary, we can always return beam idx for the highest score. If we are not in the chat\don't need to continue we just don't use that value no need to make it optional. `std::fill_n(be..` still will work fine, since on the very first call we set input arg `selected_beam_idx` to `nullopt`.

Wovchena (approved):

PR #577, opened 2024-07-04
Set default 4bit compression to INT4_ASYM

nikita-savelyevv:
This PR aligns default 4-bit weight compression parameters between optimum-intel and opengino.genai repositories. The default parameters are: `bits=4, sym=False, ratio=1.0, group_size=128`. This will be applied when there is no custom compression recipe for the given model id.

Corresponding PR to optimum-intel: https://github.com/huggingface/optimum-intel/pull/805


nikita-savelyevv:
@AlexKoff88 proposed to make the default compression mode asymmetric

eaidova (approved):

MaximProshin (approved):

PR #580, opened 2024-07-05
Support chat conversation for StaticLLMPipeline

TolyaTalamanov:
# Overview

Adding chat mode support for `StaticLLMPipeline`. 

The current implementation is naive - aggregates the entire chat conversation and pass as new prompt on every new `generate` call.


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -127,6 +126,7 @@ StaticLLMPipeline::StaticLLMPipeline(
     // (1) Read the template model - this will be kvcache model
     auto kvcache_model = core.read_model(path / "openvino_model.xml");
     // (2) TODO: Expose KV-cache input and output layers from kvcache model
```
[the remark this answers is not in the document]

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -245,22 +267,25 @@ EncodedResults StaticLLMPipeline::generate(
     ov::genai::EncodedResults results;
     // NB: Only batch=1 is supported now
     results.scores.resize(1u);
+    results.scores[0] = 0u;
```
TolyaTalamanov, line 270: Was missed previously

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -271,13 +296,13 @@ EncodedResults StaticLLMPipeline::generate(
     // NB: Now there are prompt_len tokens in KV-cache
     m_kvcache_desc.num_stored_tokens += prompt_len;
     int64_t last_token = utils::argmax(m_prefill_request.get_tensor("logits"), 0);
+    results.tokens[0].push_back(last_token);
```
TolyaTalamanov, line 299: also was missed

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -309,13 +334,12 @@ EncodedResults StaticLLMPipeline::generate(
 
         last_token = utils::argmax(m_kvcache_request.get_tensor("logits"), 0);
         results.tokens[0].push_back(last_token);
-        results.scores[0] = 0u;
```
TolyaTalamanov, line 312: has no sense in loop

samples/cpp/chat_sample/chat_sample.cpp
```diff
@@ -4,15 +4,14 @@
 #include "openvino/genai/llm_pipeline.hpp"
 
 int main(int argc, char* argv[]) try {
-    if (2 != argc) {
-        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR>");
+    if (2 != argc && 3 != argc) {
+        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> <DEVICE>");
```
slyalin, line 8: Show that `DEVICE` is optional

TolyaTalamanov: Sure, makes sense, done

samples/cpp/chat_sample/chat_sample.cpp
```diff
@@ -4,15 +4,14 @@
 #include "openvino/genai/llm_pipeline.hpp"
 
 int main(int argc, char* argv[]) try {
-    if (2 != argc) {
-        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR>");
+    if (2 != argc && 3 != argc) {
+        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> <DEVICE>(optional)");
```
Wovchena, line 8: Don't change that. The idea is to minimize the number of cmd args. You can extend the comment instead with NPU

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -77,18 +77,15 @@ void reshape_to_static(std::shared_ptr<ov::Model> model,
     model->reshape(new_shapes);
 }
 
-void fill_tensor(ov::Tensor tensor, int64_t fill_val) {
+void fill_tensor(ov::Tensor tensor, int64_t fill_val, int32_t offset = 0) {
```
Wovchena, line 80: Why not `size_t`. Or `ptrdiff_t` if you expect a negative offset, which doesn't make sense for `tensor_data + offset`? They are for pointer arithmetic.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -198,9 +206,23 @@ DecodedResults StaticLLMPipeline::generate(
...
+    if (m_is_chat_conversation) {
+        m_history.push_back({{"role", "user"}, {"content", prompt}});
+        constexpr bool add_generation_prompt = true;
+        prompt = m_tokenizer.apply_chat_template(m_history, add_generation_prompt);
+    }
+
+    auto tokenized_input = m_tokenizer.encode(prompt);
     auto encoded_results = generate(tokenized_input, config, streamer);
-    return {m_tokenizer.decode(encoded_results.tokens), encoded_results.scores};
+    DecodedResults decoded_results = {m_tokenizer.decode(encoded_results.tokens), encoded_results.scores};
```
Wovchena, line 219: Just to make it clear. Does that implementation runs the whole history on every generate()?

Wovchena (changes requested):

samples/cpp/chat_sample/chat_sample.cpp
TolyaTalamanov: Returned back anyway

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Yes, 
```
EncodedResults StaticLLMPipeline::generate(
    const EncodedInputs& inputs,
    OptionalGenerationConfig generation_config,
    StreamerVariant streamer)
```
knows nothing about chat conversation, it consumes input prompt up to 1024 tokens

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Overlooked it, thanks, fixed

src/cpp/src/llm_pipeline_static.cpp
dmatveev: Why only 1024? Shouldn't this be defined by the model? Also, when we turn the model to static shape form, what we use as a guidance to fix the input prompt/kvcache size?

TolyaTalamanov: > Shouldn't this be defined by the model?
It shouldn't model has dynamic shape

Wovchena (approved):

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -179,6 +176,17 @@ StaticLLMPipeline::StaticLLMPipeline(
 ) : StaticLLMPipeline(path, path.string(), device, config) {
 }
 
+void StaticLLMPipeline::start_chat(const std::string& system_message) {
+    if (system_message.empty()) {
+        m_history.push_back({{"role", "system"}, {"content", system_message}});
+    }
+    m_is_chat_conversation = true;
+};
+void StaticLLMPipeline::finish_chat() {
```
pavel-esir, line 185:
```suggestion

void StaticLLMPipeline::finish_chat() {
```

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -179,6 +176,17 @@ StaticLLMPipeline::StaticLLMPipeline(
 ) : StaticLLMPipeline(path, path.string(), device, config) {
 }
 
+void StaticLLMPipeline::start_chat(const std::string& system_message) {
+    if (system_message.empty()) {
```
pavel-esir, line 180: Looks like here is a typo, system message should be added to history only if not empty.

```suggestion
    if (!system_message.empty()) {
```

pavel-esir (approved):
I think there was a typo, but overall looks good. Approved

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: What is the difference?

pavel-esir: just formatting, for readability a blank line between function definitions

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -179,6 +176,18 @@ StaticLLMPipeline::StaticLLMPipeline(
 ) : StaticLLMPipeline(path, path.string(), device, config) {
 }
 
+void StaticLLMPipeline::start_chat(const std::string& system_message) {
```
Wovchena, line 179: I just noticed. After this is merged you may want to add `override` keyword

Wovchena: Never mind

PR #599, opened 2024-07-10
Remove restrictions on transformers

wgzintel:


peterchen-intel (approved):

peterchen-intel:
Remove the limitation since it is verified that we can work with transformers 4.42.3

eaidova (approved):

PR #602, opened 2024-07-10
[CI] OV local build

mryzhov:


samples/cpp/prompt_lookup_decoding_lm/CMakeLists.txt
```diff
@@ -16,6 +16,10 @@ set_target_properties(prompt_lookup_decoding_lm PROPERTIES
     INSTALL_RPATH_USE_LINK_PATH ON)
 target_compile_definitions(prompt_lookup_decoding_lm PRIVATE OPENVINO_TOKENIZERS_PATH="${OPENVINO_TOKENIZERS_PATH}")
 target_compile_features(prompt_lookup_decoding_lm PRIVATE cxx_std_17)
+if(APPLE)
+    # aligned deallocation is only available on macOS 10.14 or newer
+    target_compile_options(prompt_lookup_decoding_lm PRIVATE -fno-aligned-allocation)
```
Wovchena, line 21: Why is it needed?

If you state that it's only available on macOS 10.14 or newer in the comment, why isn't that reflected in implementation? I'm not sure, but I think older macOS isn't supported, in that case the comment is misleading because it gives perception that older macOS is supported.

Wovchena (dismissed):

samples/cpp/prompt_lookup_decoding_lm/CMakeLists.txt
mryzhov: We still support macOS>=12, see https://docs.openvino.ai/2024/about-openvino/release-notes-openvino/system-requirements.html

Wovchena: You shouldn't keep a comment about 10.14 if it's not supported because it's too old

mryzhov: I can remove, but the comment is still valid, because we set MACOSX_DEPLOYMENT_TARGET=10.12 on macOS x86

Wovchena: Where do you set it?

.github/workflows/linux.yml
```diff
@@ -0,0 +1,381 @@
...
+
+      - name: Upload openvino package
+        if: steps.openvino_download.outcome == 'success'
+        uses: actions/upload-artifact@v4
+        with:
+          name: openvino_package
+          path: openvino_package.tar.gz
+          if-no-files-found: 'error'
+
+  openvino_build:
```
ilya-lavrenov, line 53: is it possible to avoid OpenVINO build? And use only nightly, pre-releases or post-commits from main OV repo

.github/workflows/linux.yml
```diff
@@ -0,0 +1,381 @@
...
+          source ${OV_INSTALL_DIR}/setupvars.sh
+          cmake -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -S ./ -B ./build/
+          cmake --build ./build/ --config ${{ matrix.build-type }} --target package -j
+
+      - name: Test bindings
+        run: |
+          source ${OV_INSTALL_DIR}/setupvars.sh
+          python -m pip install ./thirdparty/openvino_tokenizers/[transformers] --find-links ${OV_INSTALL_DIR}/tools
+          python -m pip install --upgrade-strategy eager -r ./samples/requirements.txt --find-links ${OV_INSTALL_DIR}/tools
+          optimum-cli export openvino --trust-remote-code --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 TinyLlama-1.1B-Chat-v1.0
```
ilya-lavrenov, line 352: I don't see any tests running here.
Here we just convert model and don't run any tests

.github/workflows/linux.yml
```diff
@@ -0,0 +1,381 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: openvino_package
+          path: ${{ env.OV_INSTALL_DIR }}
+
+      - name: Extract OpenVINO packages
+        run: |
+          pushd ${OV_INSTALL_DIR}
+            tar -xzf openvino_package.tar.gz -C ${OV_INSTALL_DIR} --strip-components=1
+          popd
```
ilya-lavrenov, line 232: can we implement build for wheels similar to OpenVINO Tokenizers? 
In this case this build step will be much simpler

.github/workflows/linux.yml
```diff
@@ -0,0 +1,381 @@
...
+        run: |
+          source ${OV_INSTALL_DIR}/setupvars.sh
+          python -m pip install ./thirdparty/openvino_tokenizers/[transformers] -r ./tests/python_tests/requirements.txt --find-links ${OV_INSTALL_DIR}/tools --upgrade-strategy eager
+          python -m pytest ./tests/python_tests
+        env:
+          PYTHONPATH: "./build/:$PYTHONPATH"
+
+      - name: test bindings (wheel)
+        run: |
+          source ${OV_INSTALL_DIR}/setupvars.sh
```
ilya-lavrenov, line 271: do we need it here? I suppose we are testing via wheel

.github/workflows/linux.yml
mryzhov: yes, it is done like in tokenizers repo,

.github/workflows/linux.yml
mryzhov: it is the same as in tokenizers, we can switch if needed

samples/cpp/prompt_lookup_decoding_lm/CMakeLists.txt
mryzhov: removed this fix at all, set target 11.0

.github/workflows/linux.yml
```diff
@@ -0,0 +1,380 @@
...
+
+env:
+  PYTHON_VERSION: '3.8'
+  OV_BRANCH: 'master'
+  OV_TARBALL: ''
+
+permissions: read-all # Required by https://github.com/ossf/scorecard/blob/e23b8ad91fd6a64a0a971ca4fc0a4d1650725615/docs/checks.md#token-permissions
+
+jobs:
+  openvino_download:
```
ilya-lavrenov, line 24: let's define user friendly names with spaces, etc via `name: `

The comment is valid for all jobs in all pipelines.

.github/workflows/linux.yml
ilya-lavrenov: Please, fix the same for other pipelines

.github/workflows/linux.yml
ilya-lavrenov: Please, fix the same for other pipelines

.github/workflows/linux.yml
mryzhov: @ilya-lavrenov that is not a purpose of this PR, we could refactor the names separately, I'll add the names only for the new jobs in this PR.

.github/workflows/linux.yml
mryzhov: done

.github/workflows/linux.yml
mryzhov: added names

ilya-lavrenov:
@Wovchena I see that CB tests are explicitly run in causal_lm workflow. Is it still required? I suppose after merge CB and stateful, all tests are run on GenAI lib side

.github/workflows/linux.yml
```diff
@@ -260,16 +260,17 @@ jobs:
...
           source ${OV_INSTALL_DIR}/setupvars.sh
           python -m pip install ./thirdparty/openvino_tokenizers/[transformers] -r ./tests/python_tests/requirements.txt --find-links ${OV_INSTALL_DIR}/tools --upgrade-strategy eager
           python -m pytest ./tests/python_tests
         env:
           PYTHONPATH: "./build/:$PYTHONPATH"
 
-      - name: test bindings (wheel)
+      - name: Test bindings (wheel)
         run: |
+          source ${OV_INSTALL_DIR}/setupvars.sh
```
ilya-lavrenov, line 273: please, drop this line

mryzhov: we can't because we need to set OpenVINO env during the wheel build

Wovchena (changes requested):

.github/workflows/linux.yml
```diff
@@ -0,0 +1,382 @@
...
+          sudo apt-get install ccache
+
+      - name: Setup ccache
+        uses: actions/cache@v4
+        with:
+          save-always: true
+          path: ${{ env.CCACHE_DIR }}
+          key: ${{ runner.os }}-${{ runner.arch }}-ccache-genai-Release-${{ github.sha }}
+          restore-keys: |
+            ${{ runner.os }}-${{ runner.arch }}-ccache-genai-Release
```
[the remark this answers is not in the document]
mryzhov: to avoid using one more 3rd party dependency, will switch ov_build too

.github/workflows/linux.yml
```diff
@@ -0,0 +1,382 @@
...
+            -G "${{ env.CMAKE_GENERATOR }}" \
+            -DENABLE_CPPLINT=OFF \
+            -DENABLE_NCC_STYLE=OFF \
+            -DENABLE_TESTS=OFF \
+            -DENABLE_STRICT_DEPENDENCIES=OFF \
+            -DENABLE_SYSTEM_OPENCL=ON \
+            -DCMAKE_VERBOSE_MAKEFILE=ON \
+            -DCPACK_GENERATOR=TGZ \
+            -DENABLE_JS=OFF \
+            -DENABLE_SAMPLES=ON \
```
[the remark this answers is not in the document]
mryzhov: your pipelines use script to build samples, we can't disable samples due to it

.github/workflows/linux.yml
```diff
@@ -0,0 +1,382 @@
...
+    defaults:
+      run:
+        shell: bash
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Download OpenVINO build
+        id: openvino_download
+        run: |
+          wget ${{ env.OV_TARBALL}} --progress=bar:force:noscroll -O openvino_package.tar.gz
```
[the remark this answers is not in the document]
mryzhov: yes,  it is needed when you want to use nightly or release package, then just set the link to it here and openvino_build job will be skipped

.github/workflows/linux.yml
```diff
@@ -0,0 +1,382 @@
...
+          echo 'Acquire::Retries "10";' | sudo tee -a /etc/apt/apt.conf.d/80-retries > /dev/null
+          echo 'APT::Get::Assume-Yes "true";' | sudo tee -a /etc/apt/apt.conf.d/81-assume-yes > /dev/null
+          echo 'APT::Get::Fix-Broken "true";' | sudo tee -a /etc/apt/apt.conf.d/82-fix-broken > /dev/null
+          echo 'APT::Get::no-install-recommends "true";' | sudo tee -a /etc/apt/apt.conf.d/83-no-reсommends > /dev/null
+
+      - name: Install git
+        run: |
+          sudo apt-get update
+          sudo apt-get install --assume-yes --no-install-recommends git ca-certificates
+
```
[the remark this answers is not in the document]
mryzhov: it is needed for self-hosted runners,

.github/workflows/linux.yml
mryzhov: it builds and install it:
Building wheels for collected packages: openvino-genai
  Building wheel for openvino-genai (pyproject.toml): started
  Running command Building wheel for openvino-genai (pyproject.toml)
  Building wheel for openvino-genai (pyproject.toml): finished with status 'done'
  Created wheel for openvino-genai: filename=openvino_genai-2024.4.0.0-cp38-cp38-linux_x86_64.whl size=1[508](https://github.com/openvinotoolkit/openvino.genai/actions/runs/10196424457/job/28207653789#step:11:509)536 sha256=87e1f4b3bc1b54fd584275686aaebf6fffd7d7cc706db9cbe4f1851b06f0a1ca
  Stored in directory: /tmp/pip-ephem-wheel-cache-7qxk_eay/wheels/1b/b2/ae/cdcc204ed399c5f28832598137a73f7de722d4645acd4c8969
Successfully built openvino-genai
Installing collected packages: openvino-genai
Successfully installed openvino-genai-2024.4.0.0

.github/workflows/linux.yml
```diff
@@ -0,0 +1,382 @@
...
+      # Build
+      #
+
+      - name: Setup ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          max-size: "2000M"
+          # Should save cache only if run in the master branch of the base repo
+          # github.ref_name is 'ref/PR_#' in case of the PR, and 'branch_name' when executed on push
+          save: ${{ github.ref_name == 'master' && 'true' || 'false'  }}
```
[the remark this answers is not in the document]
mryzhov: then it should be done for each release branch, no need to keep this logic in master, moreover it is better to use released OV package instead like we do in tokenizers repo

.github/workflows/linux.yml
```diff
@@ -0,0 +1,382 @@
...
+      # Upload build artifacts and logs
+      #
+
+      - name: Upload openvino package
+        if: ${{ always() }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: openvino_package
+          path: ${{ env.BUILD_DIR }}/openvino_package.tar.gz
+          if-no-files-found: 'error'
```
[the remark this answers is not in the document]
mryzhov: if we remove the always then artifact will not be uploaded in case of any failures in the middle of the job.
Need to keep it as is

.github/workflows/linux.yml
Wovchena: That means the the tests must be run in a separate step

.github/workflows/linux.yml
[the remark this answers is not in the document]
Wovchena: Please add this as a code comment.

I noticed you resolved some of the conversations, but they are applicable to other two OS as well

.github/workflows/linux.yml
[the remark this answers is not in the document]
Wovchena: How can an artifact be considered valid if there was an error?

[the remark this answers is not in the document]
mryzhov: I didn't say it will be valid) But in case of error between build and upload_artifact steps we would like to save build artifacts anyway

.github/workflows/linux.yml
mryzhov: you are free to split steps as you want, lets do it separately

Wovchena (approved):

PR #603, opened 2024-07-10
Enable random sampling tests

as-suvorov:


tests/cpp/discrete_distribution.cpp
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) 2018-2024 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <gtest/gtest.h>
+
+#include <random>
+
+// std::mt19937 gives same output on different platforms
+// std::discrete_distribution is a platfrom specific
```
Wovchena, line 10:
```suggestion
// std::discrete_distribution is platform specific
```
`specific` is an adjective

tests/cpp/discrete_distribution.cpp
Wovchena, line 1: While this file is useful for problem description, it's pointless to test STL in GenAI. Attach it to the ticket and remove from here.

tests/python_tests/test_preemption.py
```diff
@@ -25,90 +37,233 @@ def get_beam_search_seq_len_300() -> GenerationConfig:
...
+    ),
+    (
+        {"num_kv_blocks": 100, "block_size": 32, "dynamic_split_fuse": True},
+        get_beam_search_seq_len_300(),
+    ),
+    (
+        {"num_kv_blocks": 100, "block_size": 32, "dynamic_split_fuse": False},
+        get_beam_search_seq_len_300(),
+    ),
+]
```
Wovchena, line 98: It's hard to extract the actual change. If you want to update the code style, you should do it in a separate PR. It seems that the whole PR is more dedicated to code style rather then tests enablement, so you can keep the change.

Wovchena (changes requested):

as-suvorov:
Opened https://github.com/openvinotoolkit/openvino.genai/pull/614 without code style changes

PR #604, opened 2024-07-10
Add ContinuousBatchingPipeline constructor similar to LLMPipeline

Wovchena:
That allows LLMPipeline to create ContinuousBatchingPipeline as a backend. There's also a constructor accepting ireq, which can be used if the model was already transformed appropriately for ContinuousBatchingPipeline. But it feels it's going to be misleading and it simpler just to throw if such constructor is called with ContinuousBatchingPipeline backend.


samples/cpp/continuous_batching_accuracy/continuous_batching_accuracy.cpp
```diff
@@ -78,7 +78,9 @@ int main(int argc, char* argv[]) try {
     // vLLM specific params
     scheduler_config.max_num_seqs = 2;
 
-    ov::genai::ContinuousBatchingPipeline pipe(models_path, scheduler_config);
+    // It's possible to construct a Tokenizer from a different path.
+    // If the Tokenizer isn't specified, it's loaded from the same folder.
+    ov::genai::ContinuousBatchingPipeline pipe(models_path, ov::genai::Tokenizer{models_path}, scheduler_config);
```
iefode, line 83: To use default arg for tokenizer?

Wovchena: I don't get the question.

I wanted to test that constructor. And I thought it can be kept for educational purposes.

mzegla (commented):
Doesn't python binding require adjustment to include the new constructor?

mzegla (approved):

iefode (approved):
LGTM

popovaan (approved):

PR #607, opened 2024-07-11
[Continuous batching] Replace standard max_element call with custom loop for greedy sampling

mzegla:
Searching for max element in a custom loop gives better performance than using std::max_element

### DIFF
--- a/src/cpp/src/sampler.hpp
+++ b/src/cpp/src/sampler.hpp
@@ -219,8 +219,13 @@ class Sampler {
     }
 
     Token _greedy_sample(const std::vector<Token>& logit_vector) const {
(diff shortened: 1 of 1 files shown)


src/cpp/src/sampler.hpp
```diff
@@ -219,8 +219,13 @@ class Sampler {
     }
 
     Token _greedy_sample(const std::vector<Token>& logit_vector) const {
-        auto out_token = std::max_element(logit_vector.begin(), logit_vector.end(), [](const Token& lhs, const Token& rhs) { return lhs.m_log_prob < rhs.m_log_prob; });
-        return *out_token;
+        Token max_token{0.0, 0};
```
Wovchena, line 222: I believe logits can be negative. -inf should be a good initial value.

Wovchena (approved):

PR #613, opened 2024-07-12
add generation time metrics

pavel-esir:
- Added performance metrics and updated Readme with description how to use them
- Added cpp and python sample for benchmarking

Sample to calculate and visualize performance metrics.
```
import tqdm
import pandas as pd
import matplotlib.pylab as pl

pipe = ov_genai.LLMPipeline('TinyLlama-1.1B-Chat-v1.0/')
config = ov_genai.GenerationConfig(max_new_tokens=15)
metrics_df = pd.DataFrame(columns=['batch_size', 'throughput', 'ttft', 'tpot', 'std_throughput', 'std_ttft', 'std_tpot'])

num_iter = 3
for batch_size in tqdm.tqdm([1, 2, 4, 16, 32, 64, 128]):
    prompts = ["The Sky is blue because"] * batch_size
    res = pipe.generate(prompts, config)
    metrics = res.perf_metrics
    
    for _ in range(num_iter - 1):
        res = pipe.generate(prompts, config)
        metrics += res.perf_metrics
    metrics_df = metrics_df._append({
        'throughput': metrics.get_throughput().mean, 'ttft': metrics.get_ttft().mean, 'tpot': metrics.get_tpot().mean,
        'std_throughput': metrics.get_throughput().std, 'std_ttft': metrics.get_ttft().std, 'std_tpot': metrics.get_tpot().std,
        'batch_size': batch_size, 
    }, ignore_index=True)

fig, axes = pl.subplots(nrows=3, ncols=1, figsize=(6, 8), sharex=True)

axes[0].plot(metrics_df['batch_size'], metrics_df['throughput'], '-o')
axes[1].plot(metrics_df['batch_size'], metrics_df['ttft'], '-o', )
axes[2].plot(metrics_df['batch_size'], metrics_df['tpot'], '-o')

axes[0].set_ylabel('Throughput'), axes[1].set_ylabel('TTFT'), axes[2].set_ylabel('TPOT')
axes[2].set_xlabel('Batch Size')
axes[0].grid(True), axes[1].grid(True), axes[2].grid(True)
pl.tight_layout()
```

image

ticket: CVS-132859


tests/python_tests/ov_genai_test_utils.py
```diff
@@ -81,6 +81,8 @@ def get_chat_templates():
     # but skips some models that currently are not processed correctly.
 
     skipped_models = {
+        "berkeley-nest/Starling-LM-7B-alpha", #  TODO: Need to enable and unskip, since it's preset in continious batching and has ~30 000 downloads.
```
pavel-esir, line 84: added by mistake. remove this

src/cpp/src/group_beam_searcher.cpp
```diff
@@ -362,14 +362,20 @@ std::pair<EncodedResults, int32_t> beam_search(ov::InferRequest& lm,
                            std::optional<int32_t> selected_beam_idx) {
     OPENVINO_ASSERT(config.num_beams % config.num_beam_groups == 0,
                     "number of beams should be divisible by number of groups");
-
-    // Initialize beam search
+    
     auto batch_size = input_ids.get_shape().at(0);
+    auto sequence_length = input_ids.get_shape().at(1);
+    
+    // Initialize time metric counters.
```
pavel-esir, line 369: remove

src/cpp/src/tokenizer.cpp
```diff
@@ -320,6 +320,8 @@ class Tokenizer::TokenizerImpl {
         
         // Replace what jinja2cpp doesn't support
         std::pair<std::string, std::string> replace_str_map[] = {
+            {"{-", "{"},
```
pavel-esir, line 323: remove

pavel-esir: remove

tests/python_tests/ov_genai_test_utils.py
pavel-esir: done

src/cpp/src/perf_metrics.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+
+} // namespace
+
+namespace ov {
+namespace genai {
+
+float PerfMetrics::get_duration_ms(std::chrono::steady_clock::duration duration) {
+    return std::chrono::duration_cast<std::chrono::milliseconds>(duration).count();
```
Wovchena, line 30: Don't cast duration until you really need to use the value (for printing). Return duration itself. That ensures the best accuracy. When you divide a duration, change it's representation to float. For example https://github.com/openvinotoolkit/openvino/blob/ffc135cb1240831411799bdb82ecac352c956f22/samples/cpp/benchmark/throughput_benchmark/main.cpp#L19. But your implementation needs an extra step: when mean is computed keep using source units (most likely nanoseconds, but it's unspecified, you can't rely on that) with float or double representation. Cast the duration to `Ms` or any other suitable unit to use `count()` and print.

pavel-esir: Done. Now durations are stored in microsecond `chrono::duration<float, std::ratio<1, 1000000>>` for better accuracy. If i store them in ms then tokenization/detokenization times can sometimes be 0, with microseconds.

I convert them only when mean/std are calculated.

pavel-esir:
now PR is final. @Wovchena please take a look

Wovchena:
It got a conflict

pavel-esir:
> It got a conflict

Resolved. Metrics match to llm_bench's numbers. Will open a separate PR to switch to native counters.
image

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
 }
 ```
 
+### Performance Metrics
+
+`ov.genai.PerfMetrics` (referred to as `PerfMetrics` for simplicity) is a structure that holds performance metrics for each generate call. `PerfMetrics` hold fields with mean and standard deviations for the following metrics:
```
Wovchena, line 201: There's
> `PerfMetrics` has

below. Like `PerfMetrics` is singular. But here
> `PerfMetrics` hold

`hold` doesn't have `s`, like if `PerfMetrics` were plural. Bot of the approaches seem to be valid, but you need to stick to one of them.

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
...
+print(f'tpot: {perf_metrics.mean_tpot:.2f}')
+```
+output:
+```sh
+mean_generate_duration: 76.28
+mean_ttft: 42.58
+mean_tpot 3.80
+```
+
+>**Note**: If the input prompt is just a string, the generate function will return only a string without perf_metrics. To obtain perf_metrics, provide the prompt as a list with at least one element or call generate with encoded inputs.
```
Wovchena, line 232:
```suggestion
>**Note**: If the input prompt is just a string, the generate function returns only a string without perf_metrics. To obtain perf_metrics, provide the prompt as a list with at least one element or call generate with encoded inputs.
```

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
...
+- `tpot`
+- `load_time`
+- `generate_duration`
+- `tokenization_duration`
+- `detokenization_duration`
+- `throughput`
+
+and:
+- `num_generated_tokens`
+- `num_input_tokens`
```
Wovchena, line 212: Please define each of the metrics. For example, when I first saw `tpot`, I though it was misspelled tput standing for throughput.

samples/cpp/benchmark_genai/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+    URL_HASH SHA256=523175f792eb0ff04f9e653c90746c12655f10cb70f1d5e6d6d9491420298a08)
+FetchContent_MakeAvailable(cxxopts)
+
+add_executable(benchmark_genai benchmark_genai.cpp)
+target_link_libraries(benchmark_genai PRIVATE openvino::genai cxxopts::cxxopts)
+set_target_properties(benchmark_genai PROPERTIES
+    COMPILE_PDB_NAME benchmark_genai
+    # Ensure out of box LC_RPATH on macOS with SIP
+    INSTALL_RPATH_USE_LINK_PATH ON)
+# target_compile_features(benchmark_genai PRIVATE cxx_std_11)
```
Wovchena, line 21:
```suggestion
```

samples/cpp/benchmark_genai/benchmark_genai.cpp
Wovchena, line 1: Looks like an extended greedy sample. One option is to merge it with chat_sample because it already has a loop over prompts. Is there a reason to have it as a separate sample?

Please, add tests: for the samples and functional

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
 }
 ```
 
+### Performance Metrics
```
Wovchena, line 199: When it gets merged, please, open another PR adding it to C++ and Python docstrings.

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
...
+output:
+```sh
+mean_generate_duration: 76.28
+mean_ttft: 42.58
+mean_tpot 3.80
+```
+
+>**Note**: If the input prompt is just a string, the generate function will return only a string without perf_metrics. To obtain perf_metrics, provide the prompt as a list with at least one element or call generate with encoded inputs.
+
+Several `perf_metrics` can be added with each other. In that case `raw_metrics` will be concatenated and mean/std values will be recalculated. This enhances benchmarking and accumulating statistics from several calls.
```
Wovchena, line 234:
```suggestion
Several `perf_metrics` can be added to each other. In that case `raw_metrics` are concatenated and mean/std values are recalculated. This accumulates statistics from several `generate()` calls.
```

src/cpp/src/greedy_decoding.cpp
```diff
@@ -1,7 +1,8 @@
 // Copyright (C) 2023-2024 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-#include "openvino/genai/llm_pipeline.hpp"
+#include "openvino/genai/perf_metrics.hpp"
+// #include "perf_counters.hpp"
```
Wovchena, line 5:
```suggestion
```

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
...
+```python
+import openvino_genai as ov_genai
+pipe = ov_genai.LLMPipeline(model_path, "CPU")
+res_1 = pipe.generate(["The Sun is yellow because"], max_new_tokens=20)
+res_2 = pipe.generate(["Why Sky is blue because"], max_new_tokens=20)
+perf_metrics = res_1.perf_metrics + res_2.perf_metrics
+
+print(f'generate_duration: {perf_metrics.mean_generate_duration:.2f}')
+print(f'ttft: {perf_metrics.mean_ttft:.2f}')
+print(f'tpot: {perf_metrics.mean_tpot:.2f}')
```
Wovchena, line 245: If you are going to provide snippets only for one language, it should be C++ because it's mor verbose and one of the GenAI features is the exitance of C++ interface.
The better option would be to cover both languages.
An even better option would be not to have them at all because there's no easy automatic way to validate the snippets. Refer to samples.

Wovchena (changes requested):

samples/python/benchmark_genai/benchmark_genai_automatic.py
```diff
@@ -0,0 +1,62 @@
...
+        for _ in range(num_warmup):
+            pipe.generate(prompt, config)
+        
+        res = pipe.generate(prompt, config)
+        metrics = res.metrics
+        for _ in range(num_iter - 1):
+            res = pipe.generate(prompt, config)
+            metrics += res.metrics
+        # pdb.set_trace()
+        metrics_df = metrics_df._append({
```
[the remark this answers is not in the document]
pavel-esir: this file was added accidentally by mistake :(

I will remove that

src/cpp/src/perf_metrics.cpp
Wovchena: You shouldn't store duration as any of the specified units. Store whatever `time_point_one - time_point_zero` returns. There's also a cast to int representation in constructors. For example
```C++
    m_pimpl = std::make_unique<StatefulLLMPipeline>(request, tokenizer, generation_config);
    auto stop_time = std::chrono::steady_clock::now();
    m_pimpl->m_load_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(stop_time - start_time).count();   
```
Which is bad in two ways: the first is described above, the second is that the used int representation in the cast. That's going to divide by 1000, because the default type is usually nanoseconds.

samples/python/benchmark_genai/benchmark_genai_automatic.py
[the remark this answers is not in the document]
pavel-esir: removed

src/README.md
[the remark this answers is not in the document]
pavel-esir: agreed

pavel-esir: agreed. Sticked to the plural

src/README.md
pavel-esir: agreed

samples/cpp/benchmark_genai/CMakeLists.txt
pavel-esir: removed

src/README.md
pavel-esir: agreed

src/cpp/src/greedy_decoding.cpp
pavel-esir: removed

src/README.md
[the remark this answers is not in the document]
pavel-esir: They are left because one corresponds to the number of tokens in the initial input promt and tokenization time, other to the number of generated tokens and detokenization time.

If we need them both we always can sum them. But if is do the sum on our side if someone want to separate them they will have to do extra efforts, e.g. trying to tokenize input prompt one again. I think it's better to leave separate

src/README.md
```diff
@@ -196,6 +196,55 @@ int main(int argc, char* argv[]) {
 }
 ```
 
+### Performance Metrics
+
+`ov.genai.PerfMetrics` (referred to as `PerfMetrics` for simplicity) is a structure that holds performance metrics for each generate call. `PerfMetrics` hold fields with mean and standard deviations for the following metrics:
+- `ttft`
+- `tpot`
+- `load_time`
```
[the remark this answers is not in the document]
pavel-esir: agreed, moved this metric below because it does not have mean/std. Added dimentions as welll

src/README.md
pavel-esir: defined metrics/dimensions

samples/cpp/benchmark_genai/benchmark_genai.cpp
pavel-esir: I just thought that greedy sample should be easy as possible and do not compilate it with other argument and outputs such as number of warmups, and so on.

Communicated personally. Main takeaways.
We already have a separate sample for each case, e.g. multinomial decoding despite the fact it differ just by generate params. Probably instead of adding more samples we should add features to the existing ones.

But still, benchmark_genai also allows benchmarking with multibatch inputs while chat has only batch 1. Therefore we decided to keep as a separate sample, at least in this PR

src/README.md
pavel-esir: Added to c++ as well.

```
An even better option would be not to have them at all because there's no easy automatic way to validate the snippets. Refer to samples.
```
probably it's possible to parse md extract snippets and compile them, but id doesn't seem trivial.

I referred to samples for more detailed usage.

src/cpp/src/perf_metrics.cpp
```diff
@@ -0,0 +1,118 @@
...
+    auto& right_gen_durations = right.raw_metrics.generate_durations;
+    
+    new_tok_durations.insert(new_tok_durations.end(), right_tok_durations.begin(), right_tok_durations.end());
+    new_detok_durations.insert(new_detok_durations.end(), right_detok_durations.begin(), right_detok_durations.end());
+    new_gen_durations.insert(new_gen_durations.end(), right_gen_durations.begin(), right_gen_durations.end());
+
+    res.num_generated_tokens = num_generated_tokens + right.num_generated_tokens;
+    res.num_input_tokens = num_generated_tokens + right.num_input_tokens;
+    res.load_time = load_time;
+    res.evaluate_statistics();
```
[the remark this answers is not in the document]
pavel-esir: Added getters. Now in order to get user should call `perf_metrics.get_tokenization_duration().mean` if statistics are fresh and already evaluated then it will return values, if they are not fresh getter will call `evaluate_statistics()`

Wovchena (approved):

src/README.md
pavel-esir: Done https://github.com/openvinotoolkit/openvino.genai/pull/713

PR #628, opened 2024-07-15
Enable random sampling CI

as-suvorov:


tests/python_tests/test_preemption.py
```diff
@@ -69,36 +94,73 @@ def test_preemption_with_multinomial(tmp_path, dynamic_split_fuse):
...
+            ],
+            [
+                "\nI don't have any knowledge on them. We are based out near Dubai so hopefully they will take care of us soon enough :) thanks though :",
+                "\nUAE is not one of the richest countries in Asia but definitely among those most corrupt nations because this corruption (and its own endemic practices) still",
+                "\nNope, I'm just going through my first semester there right now and it was nice to see some people who were doing well haha - we",
+                "\nIt's a country where your parents can never give you anything at all!  It also has an extremely low education system for many years... You",
+            ],
         ],
-    ])
+    ).get_ref_texts(),
```
Wovchena, line 158: You always call `get_ref_texts()` immediately after creating `RandomSamplingRefTexts`. Its instance isn't saved. That means it should be just a single function call. But in that particular case the simpler solution is to use dict:

```py
ref_texts={
        "linux": [...],
        "win32": [ ...],
        "darwin": [ ...],
    }[sys.platform]
```

Apply it to tests/python_tests/test_sampling.py as well.

Wovchena (changes requested):

tests/python_tests/test_preemption.py
Wovchena: Is there a reason not to use dictionary?

Wovchena: Never mind

Wovchena:
@mryzhov, FYI, workflow file was updated.

PR #632, opened 2024-07-16
update llm bench optimum commit to include patch for llama/gemma

eaidova:


peterchen-intel (approved):

peterchen-intel:
Close this since we have already updated to newer version.

PR #639, opened 2024-07-16
Prefix caching.

popovaan:
Implementation of prefix caching.

Ticket: CVS-138669


samples/cpp/continuous_batching_multiple_generation_sample/continuous_batching_multiple_generation_sample.cpp
```diff
@@ -0,0 +1,125 @@
...
+    const std::string models_path = "/home/panas/llm/models/opt-125m/";
+
+    // create dataset
+
+    std::vector<std::string> prompt_examples = {
+        "What is OpenVINO?\n",
+        "Explain this in more details?\n",
+        "What is your name?\n",
+        "Tell me something about Canada\n",
+        "What is OpenVINO?\n",
```
Wovchena, line 51: It seems that the only thing this sample features is prompts starting with the same prefix. Sample isn't need for that. You should just describe the applied optimization in other sample readme and code comments.

src/cpp/include/openvino/genai/scheduler_config.hpp
```diff
@@ -30,5 +30,8 @@ struct SchedulerConfig {
 
     // max number of scheduled sequences (you can think of it as "max batch size")
     std::size_t max_num_seqs = 256;
+
+    // enable caching of KV-blocks and reusage in next generations.
+    bool enable_prefix_caching = false;
```
Wovchena, line 35: Why is it disabled by default? Why would anyone want to configure that in the first place?

src/cpp/src/block_manager.hpp
```diff
@@ -6,20 +6,25 @@
...
 
 #include "sequence_group.hpp"
 
 namespace ov::genai {
 class KVCacheBlock {
     int m_ref_count;
     int m_index;
+    size_t m_hash;
+    size_t m_num_hashed_tokens;
+    time_t m_timestamp;
```
Wovchena, line 19: Why not `std::chrono`?

popovaan: No particular reason, I just need to store a timestamp in this field. Do you think it's better to use std::chrono?

src/cpp/include/openvino/genai/scheduler_config.hpp
popovaan: User might need to configure it because enabling prefix caching uses more RAM. All previously calculated KV-cache is kept in memory for future usages. KV-caches can be rewritten if we reached cache limit, but blocks are not actually released.

In case of disabling prefix caching we use only KV-cache required for batch calculation and when sequence is finished cache is released. So I think this parameter might be useful.

Regarding default value, we can set to true, it's not required to be false.

src/cpp/src/block_manager.hpp
Wovchena: chrono is better. It has more verbose interface and higher than 1 second precision.

src/cpp/include/openvino/genai/scheduler_config.hpp
Wovchena: Please add this explanation to docstrings.

Who are the beneficiaries of this feature? I hope this may help to pick the default value.

Wovchena: One approach to that question is when chat becomes efficient in CB, enable_prefix_caching must become true by default.

src/cpp/src/block_manager.hpp
popovaan: Replaced with chrono.

samples/cpp/continuous_batching_multiple_generation_sample/continuous_batching_multiple_generation_sample.cpp
popovaan: Removed the sample. 
Regarding description, I suppose we need to add it to CB chat readme when this PR is merged: https://github.com/openvinotoolkit/openvino.genai/pull/644

src/cpp/include/openvino/genai/scheduler_config.hpp
popovaan: Updated docstring. 
Regarding default value, I suppose we need to set it to true for chat scenario only. I don't think, that it is needed to be enabled by default.

iefode (approved):
LGTM

Wovchena (approved):

PR #644, opened 2024-07-18
Add CB naive chat

Wovchena:


Wovchena:
build_jenkins

src/cpp/src/continuous_batching_pipeline.cpp
```diff
@@ -238,25 +244,47 @@ class ContinuousBatchingPipeline::Impl {
...
             step();
+            if (streamer_ptr) {
+                std::unordered_map<uint64_t, GenerationOutput> token = generations.at(0).get()->back();
+                OPENVINO_ASSERT(1 == token.size());
+                OPENVINO_ASSERT(1 == token.begin()->second.generated_token_ids.size());
+                continue_generation = !streamer_ptr->put(token.begin()->second.generated_token_ids.at(0));
+            }
+        }
+        if (streamer_ptr) {
+            streamer_ptr->end();
```
mzegla, line 282: Why do we continue to execute this function when streamer is used? Output has already been read and pushed to streamer in above while loop.

Wovchena: That's just API. We document that end() is called in the end. One possible use case to let the streamer to flush the output.

andrei-kochin (approved):

PR #646, opened 2024-07-19
rm .github/ISSUE_TEMPLATE

Wovchena:
GenAI issues found by the commpunity tend to be crated using that template which isn't correct because they usually expect us to address them.


p-wysocki (commented):
Perhaps it would be easier to give better descriptions to the templates in the "New Issue" to point people to the correct template. If you think that deleting the GFI issue template altogether is the correct choice, I approve. I just hope it's not the end of GFIs from GenAI repo. :)

Wovchena:
Not the end. I was going to use one of the issues as a template for new GFI. I was trying to remember how issue assignment is implemented and just stumbled at the template because I forgot about your initial PR.

As for the general template, we don't have much to specify. It's not even reasonable to ask to provide a used version because they change rapidly and at the same time some of the problems are here for long time.

I need your GitHub button approval because Jenkins is unlikely to start its builds without the explicit approve.

p-wysocki (approved):

PR #648, opened 2024-07-19
Fix chat templates with slices, add tokenizer config for `mistralai/Mistral-7B-Instruct-v0.1`

yatarkan:


src/cpp/src/tokenizer.cpp
```diff
@@ -342,40 +343,32 @@ class Tokenizer::TokenizerImpl {
...
-        std::string replacement_string = "{% if false %}{% set placeholder = false %}";
-        
-        std::string system_message = "";
-        size_t pos = chat_tpl.find(header_with_slice);
-        if (pos != std::string::npos) {
-            chat_tpl.replace(pos, header_with_slice.length(), replacement_string);
-
-            if (!history.empty() && history[0].at("role") == "system") {
-                system_message = history[0].at("content");
-                history.erase(history.begin());
```
yatarkan, line 359: @pavel-esir Could you please review and confirm that setting of `system_message` is not needed anymore with new approach with `slice()` user-defined callable?

pavel-esir: That looks right.

Wovchena (approved):

PR #649, opened 2024-07-19
[Continuous batching] Late token vector initialization in sampling

mzegla:
Changes:
- Further split of greedy and multinomial paths - using original logits buffer in greedy and whenever possible in multinomial sampling. Sorted vector is created only when top_p or top_k filters need to be applied.
- Fixing issue with top_k filter being applied always when multinomial sampling is used unless it's explicitly set to 0. Now default value (which is max for size_t) will not trigger applying top_k filter. The filter will also not be applied if top_k is bigger than logits vector size.
- Skipping multinomial tests

### DIFF
--- a/src/cpp/src/logit_processor.hpp
+++ b/src/cpp/src/logit_processor.hpp
@@ -16,12 +16,38 @@ struct Token {
     Token() = default;
 };
 
+struct Logits {
+    float * m_data = nullptr;
+    size_t m_size;
+    // Late initialized for top_p or top_k transforms
+    std::vector<Token> m_vector;
+
+    Logits(float* data, size_t size): m_data(data), m_size(size) {}
+
+
+    void initialize_vector() {
+        OPENVINO_ASSERT(m_vector.size() == 0, "Logits vector already initialized");
+        m_vector.reserve(m_size);
+        for (size_t i = 0; i < m_size; i++)
+            m_vector.emplace_back(m_data[i], i);   
+    }
+
+    bool is_vector_initialized() const {
+        return m_vector.size() > 0;
+    }
+
+    void resize(size_t new_size) {
+        m_size = new_size;
+        m_vector.resize(new_size);
+    }
+};
+
 namespace LogitTransformers {
 using TokenIds = std::vector<int64_t>;
 
 class ILogitTransformer {
 public:
-    virtual void apply(std::vector<Token>& logits) = 0;
+    virtual void apply(Logits& logits) = 0;
 
     virtual bool is_applicable(size_t generated_tokens_cnt = 0) {
         return true;
@@ -32,11 +58,15 @@ class TopPFilter : public ILogitTransformer {
 public:
     TopPFilter(double top_p) : m_top_p(top_p) {}
 
-    void apply(std::vector<Token>& logits) override {
-        std::sort(logits.begin(), logits.end(), [](const Token& lhs, const Token& rhs) {return lhs.m_log_prob > rhs.m_log_prob; });
+    void apply(Logits& logits) override {
+        if (!logits.is_vector_initialized()) {
+            // Initialize and sort vector
+            logits.initialize_vector();
+            std::sort(logits.m_vector.begin(), logits.m_vector.end(), [](const Token& lhs, const Token& rhs) {return lhs.m_log_prob > rhs.m_log_prob; });
+        }
         float probability_sum = 0.0f;
         size_t nucleus_size = 0;
-        for (const auto& probability : logits) {
+        for (const auto& probability : logits.m_vector) {
             probability_sum += probability.m_log_prob;
             nucleus_size += 1;
             if (probability_sum > m_top_p) break;
@@ -52,10 +82,18 @@ class TopKFilter : public ILogitTransformer {
 public:
     TopKFilter(size_t top_k) : m_top_k(top_k) {}
 
-    void apply(std::vector<Token>& logits) override {
-        std::sort(logits.begin(), logits.end(), [](const Token& lhs, const Token& rhs) {return lhs.m_log_prob > rhs.m_log_prob; });
-        size_t top_k = logits.size() >= m_top_k ? m_top_k : logits.size();
-        logits.resize(top_k);
+    // If this transform is used along with top_p, it should be applied after it since top_p sorts entire vector and top_k does it only partially
+    void apply(Logits& logits) override {
+
+        if (m_top_k >= logits.m_size) 
+            return;
+        
+        if (!logits.is_vector_initialized()) {
+            // Initialize and partially sort vector
+            logits.initialize_vector();
+            std::partial_sort(logits.m_vector.begin(), logits.m_vector.begin() + m_top_k, logits.m_vector.end(), [](const Token& lhs, const Token& rhs) {return lhs.m_log_prob > rhs.m_log_prob; });
+        }
+        logits.resize(m_top_k);
     }
 
 protected:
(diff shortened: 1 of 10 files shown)


src/cpp/src/logit_processor.hpp
```diff
@@ -16,12 +16,35 @@ struct Token {
...
+        OPENVINO_ASSERT(m_vector.size() == 0, "Logits vector already initialized");
+        m_vector.reserve(m_size);
+        for (size_t i = 0; i < m_size; i++)
+            m_vector.emplace_back(m_data[i], i);
+        
+    }
+
+    void resize(size_t new_size) {
+        m_size = new_size;
+        m_vector.resize(new_size);
```
Wovchena, line 38: `m_size` is already accessible in `m_vector.size()`. I think you should remove `m_size`.

src/cpp/src/logit_processor.hpp
```diff
@@ -16,12 +16,35 @@ struct Token {
...
+    size_t m_size;
+    // Late initialized
+    std::vector<Token> m_vector;
+
+    Logits(float* data, size_t size): m_data(data), m_size(size) {}
+
+
+    void initialize_vector() {
+        OPENVINO_ASSERT(m_vector.size() == 0, "Logits vector already initialized");
+        m_vector.reserve(m_size);
```
Wovchena, line 30: Are there use cases when `initialize_vector()` isn't called. If there are none, you can avoid storing `m_data` and move `m_vector` filling to the constructor.

src/cpp/src/logit_processor.hpp
Wovchena, line 1: Should you retarget to release branch?

Wovchena (changes requested):

src/cpp/src/logit_processor.hpp
mzegla: Yes, `initialize_vector` is called only if `top_p` or `top_k` are specified (when we need sorting). So for example if only `temperature` and some penalties are defined, we use `m_data` and `m_size` and `m_vector` is unused.

src/cpp/src/logit_processor.hpp
mzegla: This PR is part of a bigger change and relies on commits that are currently on master. I was thinking that we would merge it to master first and I'll create PR to release branch with cherry-picked commits for entire sampler optimization change. What do you think?

Wovchena (approved):

tests/python_tests/test_preemption.py
```diff
@@ -109,12 +109,12 @@ def test_preemption_with_multinomial(tmp_path, dynamic_split_fuse):
...
         "linux": [
             [
-                "\nI've seen this expression used too many times without making sense.\nAs an AI engineer, and as a scientist, we should make everything easier"
+                " Buzzfeed  ESPN  CNBC  MSNBC  CBS\nFox News is on top of the list.\nIf a news station tries to afford real estate"
             ],
             [
-                " position of the Z-shaped groove?\n0.41\nWhat is the current position of the Z-shaped groove?\n0.11\n",
-                " status of all of this? I can't stop thinking about it.\nIt's been a while since I've seen it. I found it a",
-                " status of your blog? Do you accept feedback?\nYes, I’m happy to accept feedback at this time (I’m a"
+                " condition of the leg?\nIt's been quite a while since I've seen it, so I didn't really know if it was good or bad",
```
iefode, line 115: A bit strange refs..

src/cpp/src/sampler.hpp
```diff
@@ -294,17 +304,29 @@ SamplerOutput Sampler::sample(std::vector<SequenceGroup::Ptr> & sequence_groups,
                     running_sequence->append_token(sampled_token_id.m_index, sampled_token_id.m_log_prob);
                 };
                 for (size_t running_sequence_id = 0; running_sequence_id < num_running_sequences; ++running_sequence_id) {
+                    static ManualTimer timer1("sample::_get_logit_vector");
```
iefode, line 307: Debug code?

iefode (commented):
In general LGTM

src/cpp/src/sampler.hpp
mzegla: Yes, I added it to see what amount of time different operations take. 
Shall I remove all new timers from this PR?

tests/python_tests/test_preemption.py
mzegla: Multinomial tests for preemption fail on master, so perhaps there's something wrong going on with it and we get strange outputs.

Wovchena: test_preemption.py passed here https://github.com/openvinotoolkit/openvino.genai/actions/runs/10056169378/job/27794414983?pr=666. But I didn't do anything specific.

src/cpp/src/logit_processor.hpp
```diff
@@ -286,14 +329,15 @@ class LogitProcessor {
                 if (sampling_params.top_p != 1.0f) {
                     m_logit_transformers.emplace_back(new LogitTransformers::TopPFilter(sampling_params.top_p));
                 }
-                if (sampling_params.top_k > 0) {
+                // TODO: Uncommenting below condition requires changes in reference texts in tests
+                if (sampling_params.top_k > 0 /* && sampling_params.top_k < std::numeric_limits<size_t>::max() */) {
```
as-suvorov, line 333: I encountered an issue with exactly same line. Random sampling results seems to depend on logits order. TopK filter with default topK (`std::numeric_limits<size_t>::max()`) gives different results than removed filter.

mzegla: Exactly, currently default top_k results in TopK filter being applied always for multinomial sampling and logits always get sorted which impacts the output.

src/cpp/src/logit_processor.hpp
```diff
@@ -16,12 +16,38 @@ struct Token {
...
+
+
+    void initialize_vector() {
+        OPENVINO_ASSERT(m_vector.size() == 0, "Logits vector already initialized");
+        m_vector.reserve(m_size);
+        for (size_t i = 0; i < m_size; i++)
+            m_vector.emplace_back(m_data[i], i);   
+    }
+
+    bool vector_initialized() const {
```
iefode, line 35: Minor: `vector_initialized` -> `is_vector_initialized` (is more clear)

src/cpp/src/logit_processor.hpp
```diff
@@ -52,10 +82,26 @@ class TopKFilter : public ILogitTransformer {
...
-        std::sort(logits.begin(), logits.end(), [](const Token& lhs, const Token& rhs) {return lhs.m_log_prob > rhs.m_log_prob; });
-        size_t top_k = logits.size() >= m_top_k ? m_top_k : logits.size();
-        logits.resize(top_k);
+    // If this transform is used along with top_p, it should be applied after it since top_p sorts entire vector and top_k does it only partially
+    void apply(Logits& logits) override {
+        
+        /*
+        TODO: Uncommenting this section requires changes in reference texts in tests
+
+        if (m_top_k >= logits.m_size)
```
iefode, line 91: Please, uncomment and change refs in tests :)

src/cpp/src/logit_processor.hpp
iefode: Please uncomment and update a ref

mzegla: Agreed to do it in next PR

src/cpp/src/logit_processor.hpp
mzegla: Agreed to do it in next PR

ilya-lavrenov (approved):

PR #650, opened 2024-07-19
Update OpenVINO archive links in GenAI build docs

helena-intel:
- Update archive links from 2024.2 prerelease to 2024.3 prerelease
- Change `unzip` to `tar -xf` in Windows instructions because tar is included by default on Windows and unzip is not
- For Linux and macOS, add note with links to other archive versions

### DIFF
--- a/src/docs/BUILD.md
+++ b/src/docs/BUILD.md
@@ -16,13 +16,15 @@
     cd openvino.genai
     ```
 2. Download OpenVINO archive and install dependencies:
+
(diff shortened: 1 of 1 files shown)


helena-intel:
I see that links are already updated in master. I followed instructions from https://github.com/openvinotoolkit/openvino.genai/blob/master/src/README.md and that led to 2024.2 branch instructions with the old links. I'll create a PR to update that page.

PR #683, opened 2024-07-25
Bump product version to match openvino/tokenizers

wkobielx:
Openvino and tokenizers are already bumped on master:
https://github.com/openvinotoolkit/openvino_tokenizers/pull/194
https://github.com/openvinotoolkit/openvino/pull/25520


eaidova:
@Wovchena could you please take a look?

wkobielx:
Not sure, if also github workflows should not be updated

wkobielx:
Okay, I see this is already covered in https://github.com/openvinotoolkit/openvino.genai/pull/627

PR #699, opened 2024-07-29
Freeze torch and trochvision for llm_bench

Wovchena:


llm_bench/python/requirements.txt
```diff
@@ -6,7 +6,8 @@ openvino-tokenizers
 openvino_genai
 auto-gptq>=0.5.1 # for gptq
 pillow
-torch
+torch==2.3
+torchvision==0.18
```
peterchen-intel, line 10:
```suggestion
torch<2.4.0
torchvision<0.19.0
```
Will add requirements_2024.3.txt with all validated fixed versions like https://github.com/openvinotoolkit/openvino.genai/blob/releases/2024/2/llm_bench/python/requirements_2024.2.txt

Wovchena: Yeah, but I don't know how to force internal validation to use that file. It was simpler just to freeze here and trigger the validation from branch.

PR #705, opened 2024-07-30
StaticLLMPipeline - align u4 zero points

TolyaTalamanov:


dmatveev (approved):
I am fine with this if it works

PR #717, opened 2024-07-31
Add iterable streamer to multinomial sample

pavel-esir:


src/README.md
```diff
@@ -210,6 +210,70 @@ int main(int argc, char* argv[]) {
 }
 ```
 
+This Python example demonstrates custom detokenization with buferisation. The streamer receives 
+integer tokens corresponding to each word or subword, one by one. If tokens are decoded individually, 
+subwords will not be concatenated correctly, and the resulting text will lack appropriate spaces.
```
Wovchena, line 215: I believe the problem is not how strings are concatenated. The problems is that detokenize(tokenize(" a)) == a and tokenizer needs a larger context to avoid omitting spaces.

src/README.md
```diff
@@ -210,6 +210,70 @@ int main(int argc, char* argv[]) {
...
+        if len(text) > self.print_len:
+            word = text[self.print_len:]
+            self.process_word(word)
+            self.tokens_cache = []
+            self.print_len = 0
+
+pipe = ov_genai.LLMPipeline(model_path, "CPU")
+text_print_streamer = TextPrintStreamer(pipe.get_tokenizer())
+
+pipe.generate("The Sun is yellow because", max_new_tokens=15, streamer=text_print_streamer)
```
Wovchena, line 274: You can move to any sample. That way it becomes tested.

The request was to add iterable. Add the required methods to `TextPrintStreamer`

Wovchena (changes requested):

src/README.md
pavel-esir: by concatenation problems i meant exactly what you are saying. Because of `detokenize(tokenize(" a)) == a` words are concatenated without spaces.

Corrected to express this explicitly.

src/README.md
pavel-esir: Looks like we understood a little bit differently priorities. I understood that the main issue was not with the iterable streamer itself but with the fact that text output was missing spaces. Mainly this PR is adding bufferisation which fixes missing spaces problem in the iterable streamer @eaidova already had.

Anyways i added `__next__`, `__iter__` to make streamer iterable and moved it to multinomial sample.

pavel-esir:
@Wovchena @eaidova please take a look. I have a question, do we need to update cpp sample as well?

src/README.md
```diff
@@ -206,10 +204,35 @@ int main(int argc, char* argv[]) {
 
     std::string model_path = argv[1];
     ov::genai::LLMPipeline pipe(model_path, "CPU");
-    std::cout << pipe.generate("The Sun is yellow because", ov::genai::streamer(custom_streamer), ov::genai::max_new_tokens(200));
+    std::cout << pipe.generate("The Sun is yellow because", , ov::genai::max_new_tokens(15), ov::genai::streamer(custom_streamer));
```
pavel-esir, line 207:
```suggestion
    std::cout << pipe.generate("The Sun is yellow because", ov::genai::max_new_tokens(15), ov::genai::streamer(custom_streamer));
```

Wovchena (approved):

samples/python/multinomial_causal_lm/multinomial_causal_lm.py
```diff
@@ -4,11 +4,120 @@
...
+        self.tokens_cache.append(token_id)
+        text = self.tokenizer.decode(self.tokens_cache)
+
+        word = ''
+        if len(text) > self.print_len and '\n' == text[-1]:
+            # Flush the cache after the new line symbol.
+            word = text[self.print_len:]            
+            self.tokens_cache = []
+            self.print_len = 0
+        elif len(text) >= 3 and text[-3:] == "�":
```
eaidova, line 93: can ord of unicode symblol used instead
```suggestion
        elif len(text) >= 3 and text[-3:] == chr(65533):
```

eaidova (approved):

PR #722, opened 2024-08-01
Add test case for native pytorch model

wgzintel:


.github/workflows/llm_bench-python.yml
```diff
@@ -55,6 +55,11 @@ jobs:
       with:
         name: llm.bench_diff
         path: llm.bench_diff.diff
+    - name: Test tiny-random-qwen of pytorch model on Linux
+      run: |
+        export GIT_LFS_SKIP_SMUDGE=0
+        git clone https://huggingface.co/katuni4ka/tiny-random-qwen
```
peterchen-intel, line 61: Can we clone just  latest version? It may save a little time.

wgzintel: Updated

PR #735, opened 2024-08-02
Rename config key according to DLB folder name

nikita-savelyevv:
- In DLB model cache there is a model with `llama-2-7b-hf` folder name, but not `llama-2-7b`.
- Also, `llama-2-7b-chat` folder was renamed to `llama-2-7b-chat-hf`, same for `llama-2-13b-chat`. See ticket 147470 for details.


KodiaqQ (approved):
Since these changes

nikita-savelyevv:
> Since these changes

@KodiaqQ What was meant here?

skuros (approved):

PR #743, opened 2024-08-05
Add config for open-llama-3b-v2

nikita-savelyevv:
Add config for `open-llama-3b-v2` model because it is present in optimum-intel https://github.com/huggingface/optimum-intel/blame/main/optimum/intel/openvino/configuration.py#L80


llm_bench/python/utils/nncf_utils.py
```diff
@@ -58,6 +58,8 @@ def get_compressed_path(output_dir: str, base_precision, option: str):
     "chatglm2-6b": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 128, "ratio": 0.72},
     "qwen-7b-chat": {"mode": nncf.CompressWeightsMode.INT4_SYM, "group_size": 128, "ratio": 0.6},
     "open-llama-3b": {"mode": nncf.CompressWeightsMode.INT4_ASYM, "group_size": 64, "ratio": 1.0, "all_layers": True},
+    "open-llama-3b-v2": {"mode": nncf.CompressWeightsMode.INT4_ASYM, "group_size": 64, "ratio": 1.0,
+                         "all_layers": True},
```
KodiaqQ, line 62: Provide the numbers of the accuracy & performance after the update in comparison with current configuration.

KodiaqQ: Okay, I see. Why the `"all_layers": True` was added? 
It's not a default option and may lead to a huge accuracy drop.

nikita-savelyevv: According to @ljaljushkin , it is because the config for `openlm-research/open_llama_3b` is with `all_layers=True`. However, as I understand there is no justification for this particular parameter.

I've measured PPL for FP16 and compressed models with and without all layers. Here are the results.
| Precision | all_layers | PPL   |
|-----------|------------|-------|
| FP16      |            | 12.40 |
| INT4_ASYM | False      | 13.16 |
| INT4_ASYM | True       | 13.44 |

Is 5% the maximum acceptable PPL drop for compressed model? In such case even with `all_layers=False`, the PPL is too high.

AlexKoff88: Both configurations in the table above are out of the 0.05% of relative accuracy drop. We should use data-aware tuning probably or reduce the ratio.

nikita-savelyevv: @KodiaqQ Could you please include this model into your experiments on finding optimal compression parameters?

KodiaqQ: I can, but it wouldn't be the first priority.

nikita-savelyevv: Thank you. Updated config without all_layers as a temporary solution.

KodiaqQ (approved):

AlexKoff88 (approved):

eaidova (approved):

PR #762, opened 2024-08-11
Add USS metric

wgzintel:


llm_bench/python/utils/memory_profile.py
```diff
@@ -23,15 +25,29 @@ def collect_memory_consumption(self):
             self.g_event.wait()
             while True:
                 process = psutil.Process(os.getpid())
-                rss_mem_data = process.memory_info().rss / float(2**20)
                 try:
-                    shared_mem_data = process.memory_info().shared / float(2**20)
+                    rss_mem_data = process.memory_info().rss / float(2**20)
```
peterchen-intel, line 29: Let's keep memory_info() here, and read rss/shared/private from the saved variable, so that all the memory info are from "same shot".

llm_bench/python/utils/memory_profile.py
```diff
@@ -23,15 +25,29 @@ def collect_memory_consumption(self):
...
+                    if sys.platform.startswith('linux'):
+                        shared_mem_data = process.memory_info().shared / float(2**20)
+                        uss_mem_data = rss_mem_data - shared_mem_data
+                    elif sys.platform.startswith('win'):
+                        uss_mem_data = process.memory_info().private / float(2**20)
+                        shared_mem_data = (rss_mem_data - uss_mem_data) if rss_mem_data > uss_mem_data \
```
peterchen-intel, line 35: rss_mem_data >= uss_mem_data this should be always correct, do you find any no in this case?

llm_bench/python/utils/memory_profile.py
```diff
@@ -23,15 +25,29 @@ def collect_memory_consumption(self):
             self.g_event.wait()
             while True:
                 process = psutil.Process(os.getpid())
-                rss_mem_data = process.memory_info().rss / float(2**20)
                 try:
-                    shared_mem_data = process.memory_info().shared / float(2**20)
```
peterchen-intel, line 28: Can move "/ float(2**20)" to get_max_memory_consumption, so that we can save a little time.

peterchen-intel (approved):

Issue #709, opened 2024-07-31
 convert.py do NOT support Meta-Llama-3.1-8B-Instruct


openvino-book:
Run [convert.py](https://github.com/openvinotoolkit/openvino.genai/blob/master/llm_bench/python/convert.py) on Meta-Llama-3.1-8B-Instruct model, Error occur as below:
> ValueError: `rope_scaling` must be a dictionary with two fields, `type` and `factor`, got {'factor': 8.0, 'low_freq_factor': 1.0, 'high_freq_factor': 4.0, 'original_max_position_embeddings': 8192, 'rope_type': 'llama3'}

![1722391465029](https://github.com/user-attachments/assets/85787f84-7dbd-4665-b26e-226f2403ab99)

eaidova:
@openvino-book according to this issue on hf hub, you need to upgrade transformers for loading llama3.1 https://huggingface.co/meta-llama/Meta-Llama-3.1-8B-Instruct/discussions/15

openvino-book:
> @openvino-book according to this issue on hf hub, you need to upgrade transformers for loading llama3.1 https://huggingface.co/meta-llama/Meta-Llama-3.1-8B-Instruct/discussions/15

Thank you @eaidova , it works!
> pip install --upgrade transformers
the worked version of transformers is 4.43.3

![1722734952888](https://github.com/user-attachments/assets/aaadb413-ab14-4e56-bd84-c4e87e27b824)

openvino-book:
Hi, @eaidova , after update the transformers, the convert.py works well. 

However, the benchmark.py raise an error  as below when running the converted Llama3.1 IR model.

![1722736391336](https://github.com/user-attachments/assets/e7822026-a12c-4e84-b19f-937a333c1a7c)

openvino-book:
running on the CPU has the same error
![1722736622228](https://github.com/user-attachments/assets/cd2d2111-d6c7-483c-9e1d-37eaeeea43a9)

openvino-book:
Hi, @eaidova  very interesting, there are not openvino_tokenizer.xml & openvino_detokenizer.xml in the converted model D:\openvino.genai\llm_bench\python\llama31_ov\pytorch\dldt\FP16
![1722739251500](https://github.com/user-attachments/assets/9c5346f8-dcd8-4dd8-a4f8-fe8edc968379)

peterchen-intel:
[...]

PR #801, opened 2024-08-27
Add hook sample for new transformers

wgzintel:
fix the issue https://github.com/openvinotoolkit/openvino.genai/issues/709


peterchen-intel:
@eaidova https://github.com/openvinotoolkit/openvino.genai/pull/801/commits/33a34c219dcf9ae87fcce854ed8fe1999e36a475#diff-2f690f6027653d8401edfa436d44a4f6a379874c33f5d557f09d72ae6a1c2a3aR152-R155 
Since this .clone() causes up to 33% 2nd token latency regression on SPR for most weekly models (llama-3-8b, llama-2-7b, qwen2-7b, chatglm3-6b etc.). 
We have replaced  .clone() with reference as it is in the old version of _sample(), do you see any risk?

eaidova:
@peterchen-intel no, I do not see any risks for that, reason of addition clone is not applicable for us, but it makes uncomparable perf with original optimum and llm bench (as issue still will be visible for model without hook

llm_bench/python/llm_bench_utils/hook_sample_v43.py
```diff
@@ -0,0 +1,235 @@
...
+
+            # forward pass to get next token
+            tic_infer = time.perf_counter()
+            outputs = self(**model_inputs, return_dict=True)
+            hook_greedy.tm_infer_list.append(time.perf_counter() - tic_infer)
+
+            if synced_gpus and this_peer_finished:
+                continue  # don't waste resources running the code we don't need
+
+            # Clone is needed to avoid keeping a hanging ref to outputs.logits which may be very large for first iteration
```
slyalin, line 150: Please explain why clone started to be slow in this case. Does it mean that something goes wrong in another place and we are hiding the effect here?

llm_bench/python/llm_bench_utils/hook_sample_v43.py
```diff
@@ -0,0 +1,235 @@
...
+            # (the clone itself is always small)
+            #next_token_logits = outputs.logits[:, -1, :].clone()
+            
+            # clone causes regression compared with transformers V41.1
```
slyalin, line 154: Please explain why clone started to be slow in this case. Does it mean that something goes wrong in another place and we are hiding the effect here?

slyalin (dismissed):

llm_bench/python/llm_bench_utils/hook_sample_v43.py
peterchen-intel: We don't have answer now. Will try to profile it.

slyalin: @peterchen-intel, please confirm that while providing a workaround for the issue in this PR we are not providing a solution for the customers that are using vanilla hf/optimum?

peterchen-intel: From this perspective, we should reproduce the issue and report it to transformers.

slyalin: > ... and report it to transformers
Or to OpenVINO if it turns out that something bad happened to that output tensor a slice of which is being cloned here.

llm_bench/python/llm_bench_utils/hook_greedy_search.py
```diff
@@ -354,5 +357,9 @@ def get_time_infer_list(self):
     def new_forward(self, model):
         """Define a new greedy search function."""
         model._greedy_search = new_greedy_search.__get__(model, model.__class__)
-        model._sample = hook_sample.new_sample.__get__(model, model.__class__)
-            
+        max_sample_version = version.parse('4.43.0')
+        trans_version = version.parse(transformers.__version__)
```
peterchen-intel, line 361:
```suggestion
        trans_version = version.parse(transformers.__version__)
        logger.info(f"transformers version: {trans_version}")
```

llm_bench/python/llm_bench_utils/hook_sample_v43.py
peterchen-intel: @slyalin @eaidova  After the investigation. This .clone() makes the issue CVS-150542, CVS-145996 become worse, it is not the root cause of the regression. Removing it can mitigate the regression (the worse part), but not the fixing for original issue. The original issue is due to the "busy-wait spin" conflict between oneTBB (OV inference) and OMP (torch post processing like greedy search or beam search), the solution is to limit the torch threads for post processing or setting [OMP_WAIT_POLICY|https://gcc.gnu.org/onlinedocs/libgomp/OMP_005fWAIT_005fPOLICY.html]=PASSIVE of OMP to disable busy-wait spin. We will ask oneTBB to support similar feature to disable oneTBB busy-wait spin.

peterchen-intel:
> @wgzintel @peterchen-intel do you have plan to fix sample code (with clone/without clone)? Looks like initial issue still reproducible
@eaidova  Will be fixed without changing transformers _sample() code except adding time profiling.

llm_bench/python/benchmark.py
```diff
@@ -725,6 +725,9 @@ def get_argprser():
 def main():
     logging_kwargs = {"encoding": "utf-8"} if sys.version_info[1] > 8 else {}
     log.basicConfig(format='[ %(levelname)s ] %(message)s', level=os.environ.get("LOGLEVEL", log.INFO), stream=sys.stdout, **logging_kwargs)
+    env_omp = os.getenv('OMP_WAIT_POLICY')
+    if env_omp is None or env_omp != 'PASSIVE':
+        log.warning('The environment OMP_WAIT_POLICY is not set to PASSIVE')
```
peterchen-intel, line 730:
```suggestion
        log.warning('It is recommended to set the environment variable OMP_WAIT_POLICY to PASSIVE, so that OpenVINO inference can use all CPU resources without waiting.')
```

llm_bench/python/benchmark.py
```diff
@@ -739,6 +742,14 @@ def main():
...
         if model_args['config'].get('PREC_BF16') and model_args['config']['PREC_BF16'] is True:
             log.warning('[Warning] Param bf16/prec_bf16 only work for framework pt. It will be disabled.')
+        original_torch_thread_nums = torch.get_num_threads()
+        num_beams = model_args['num_beams']
+        if num_beams > 1:
+            torch.set_num_threads(int(original_torch_thread_nums / 2))
+        else:
+            torch.set_num_threads(1)
+        log.info(f'The num_beams is {num_beams}, update Torch thread num from {original_torch_thread_nums} to {torch.get_num_threads()}, '
+                 f'avoid to use the CPU cores for OpenVINO inference.')
```
peterchen-intel, line 752:
```suggestion
        log.info(f'The num_beams is {num_beams}, update Torch thread num from {original_torch_thread_nums} to {torch.get_num_threads()} '
                 f'and avoid to use the CPU cores for OpenVINO inference.')
```

peterchen-intel (approved):

peterchen-intel:
Documentation on OpenVINO https://github.com/openvinotoolkit/openvino/pull/26950

llm_bench/python/README.md
```diff
@@ -159,7 +159,15 @@ For example, `--load_config config.json` as following will result in streams.num
 ``` 
 `<NUMBER>` is the number of total physical cores in 2 sockets.
 
-## 6. Additional Resources
+## 6. CPU Threading
+
+OpenVINO uses [oneTBB](https://github.com/oneapi-src/oneTBB/) as default threading library, while Torch uses [OpenMP](https://www.openmp.org/). Both threading libraries have ['busy-wait spin'](https://gcc.gnu.org/onlinedocs/libgomp/GOMP_005fSPINCOUNT.html) by default. So in LLM pipeline, when inference on CPU with OpenVINO and postprocessing with Torch(For example: greedy search or beam search), there is threading overhead in the switching between inference(OpenVINO with oneTBB) and postprocessing (Torch with OpenMP).
```
peterchen-intel, line 164:
```suggestion
OpenVINO is by default bult with [oneTBB](https://github.com/oneapi-src/oneTBB/) threading library, while Torch uses [OpenMP](https://www.openmp.org/). Both threading libraries have ['busy-wait spin'](https://gcc.gnu.org/onlinedocs/libgomp/GOMP_005fSPINCOUNT.html) by default. When running LLM pipeline on CPU device, there is threading overhead in the switching between inference on CPU with OpenVINO (oneTBB) and postprocessing (For example: greedy search or beam search) with Torch (OpenMP).
```

llm_bench/python/README.md
```diff
@@ -159,7 +159,15 @@ For example, `--load_config config.json` as following will result in streams.num
...
+
+**Alternative solutions**
+1. Use --genai option which uses OpenVINO genai APIs instead of optimum-intel APIs. In this case postprocessing is executed with OpenVINO genai APIs.
```
peterchen-intel, line 167:
```suggestion
1. Use --genai option which uses OpenVINO genai API instead of optimum-intel API. In this case postprocessing is executed with OpenVINO genai API.
```

llm_bench/python/README.md
```diff
@@ -159,7 +159,15 @@ For example, `--load_config config.json` as following will result in streams.num
...
+2. Without --genai option which uses optimum-intel APIs, set environment variable [OMP_WAIT_POLICY](https://gcc.gnu.org/onlinedocs/libgomp/OMP_005fWAIT_005fPOLICY.html) to PASSIVE which will disable OpenMP 'busy-wait', and benchmark.py will also limit the Torch thread number to avoid using CPU cores which is in 'busy-wait' by OpenVINO inference.
```
peterchen-intel, line 168:
```suggestion
2. Without --genai option which uses optimum-intel API, set environment variable [OMP_WAIT_POLICY](https://gcc.gnu.org/onlinedocs/libgomp/OMP_005fWAIT_005fPOLICY.html) to PASSIVE which will disable OpenMP 'busy-wait', and benchmark.py will also limit the Torch thread number to avoid using CPU cores which is in 'busy-wait' by OpenVINO inference.
```

llm_bench/python/llm_bench_utils/hook_greedy_search.py
```diff
@@ -354,5 +357,9 @@ def get_time_infer_list(self):
...
+        if trans_version >= max_sample_version:
+            model._sample = hook_sample_v43.new_sample.__get__(model, model.__class__)
+        else:
+            model._sample = hook_sample.new_sample.__get__(model, model.__class__)
```
peterchen-intel, line 365:
```suggestion
        model._sample = hook_sample.new_sample.__get__(model, model.__class__)
        trans_version = version.parse(transformers.__version__)
        if trans_version >= version.parse('4.45.0'):
            model._sample = hook_sample_v45.new_sample.__get__(model, model.__class__)
        elseif trans_version >= version.parse('4.43.0'):
            model._sample = hook_sample_v43.new_sample.__get__(model, model.__class__)         
```
Please add hook_sample_v45.py
https://github.com/huggingface/transformers/blob/v4.44.2/src/transformers/generation/utils.py#L2896  was removed in 4.45.0 https://github.com/huggingface/transformers/blob/v4.45.0/src/transformers/generation/utils.py#L2925

PR #804, opened 2024-08-28
Added mutex.lock() during restoring of cache.

popovaan:
Added mutex.lock() during restoring of cache.

### DIFF
--- a/src/cpp/src/continuous_batching_pipeline.cpp
+++ b/src/cpp/src/continuous_batching_pipeline.cpp
@@ -29,7 +29,6 @@ class ContinuousBatchingPipeline::Impl {
     std::shared_ptr<CacheManager> m_cache_manager;
     std::shared_ptr<ModelRunner> m_model_runner;
     std::shared_ptr<Sampler> m_sampler;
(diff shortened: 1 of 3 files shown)


src/cpp/src/continuous_batching_pipeline.cpp
```diff
@@ -29,6 +29,7 @@ class ContinuousBatchingPipeline::Impl {
     std::shared_ptr<CacheManager> m_cache_manager;
     std::shared_ptr<ModelRunner> m_model_runner;
     std::shared_ptr<Sampler> m_sampler;
+    std::mutex m_mutex;
```
mzegla, line 32: Could you move it down below `m_awaiting_requests_mutex` and rename to something like `m_block_manager_mutex` so it's easier to track which resource it protects?

popovaan: Renamed to `m_cached_blocks_map_mutex` and moved to `block_manager`, as there is no need to have it in `continuous_batching_pipeline.hpp`.

ilya-lavrenov (approved):

PR #818, opened 2024-09-04
use perf metrics genai in llm_bench

eaidova:
Update the way how metrics are collected in the llm-bench

CVS-151502 
https://github.com/openvinotoolkit/openvino.genai/pull/830


llm_bench/python/benchmark.py
```diff
@@ -290,19 +287,20 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
         max_shared_mem=max_shared_mem_consumption,
         max_uss_mem=max_uss_mem_consumption,
         prompt_idx=prompt_index,
-        tokenization_time=(tok_encode_time, tok_decode_time)
+        tokenization_time=tokenization_time
+
```
eaidova, line 291:
```suggestion

```

PR #878, opened 2024-09-18
StaticLLMPipeline: Enable DQ

TolyaTalamanov:


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -33,6 +33,49 @@ void align_u4_zp_constants(const std::shared_ptr<ov::Model>& model) {
...
+    if (group_size == -1) {
+        // NB: For CW configuration NPUW DQ is enabled unconditionally
+        return true;
+    }
+    const uint32_t kTotalParams = calculate_trainable_parameters(model);
+    if (!is_prefill_model && kTotalParams < 2'000'000'000) {
+        // NB: For GQ configurations NPUW DQ is enabled only
+        // for models that have <2B parameters and only for generate model (so far)
+        return true;
+    }
```
dmatveev, line 67: What you get here is double memory consumption for the weights - DQ transformation for group quant will transpose most of those. If it is ok we can keep it, if it is not then probably we can remove this check, to enable this for CW models only.

There is a PR which adds DQ for GQ, but there's some known problems and no LoS when it is merged.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -33,6 +33,49 @@ void align_u4_zp_constants(const std::shared_ptr<ov::Model>& model) {
...
+    return false;
+}
+
+void enable_npuw_dq_if_allowed(ov::AnyMap& config,
+                               const std::shared_ptr<ov::Model>& model,
+                               const bool is_prefill_model = false) {
+    if (allow_to_enable_npuw_dq(model, is_prefill_model)) {
+        config["NPUW_DQ"] = "YES";
```
AlexKoff88, line 75: OpenVINO already provides an option for dynamic quantization which is called "DYNAMIC_QUANTIZATION_GROUP_SIZE". Details are here: https://docs.openvino.ai/2024/learn-openvino/llm_inference_guide/llm-inference-hf.html#enabling-openvino-runtime-optimizations.

It would be great to have an alignment here.

dmatveev: I believe there's some mess in the terminology used by various projects, let's assume ours mean "DeQuantization" - when it is done with some help from the plugin

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Enabled only for CW quantized models

dmatveev (approved):

TolyaTalamanov:
@Wovchena Hi! Could you have a look, please? Can it be merged?

dmatveev (changes requested):

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Nice catch, fixed

dmatveev (approved):

Wovchena (approved):

src/cpp/src/llm_pipeline_static.cpp
dmatveev: Reverted :)

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -163,37 +193,40 @@ void merge_config_with(ov::AnyMap& lhs, const ov::AnyMap& rhs) {
...
-ov::AnyMap get_default_prefill_config() {
-    std::map<std::string, std::string> config = {
+ov::AnyMap get_default_prefill_config(const std::shared_ptr<ov::Model>& model) {
+    ov::AnyMap config = {
         { "NPU_USE_NPUW", "YES" },
         { "NPUW_FOLD", "YES" },
         { "NPUW_DCOFF_TYPE", "f16" },
         { "NPUW_DCOFF_SCALE",  "YES" },
+        { "NPUW_WEIGHTS_BANK",  "shared" },
         { "NPUW_ONLINE_AVOID", "P:RMSNorm/NPU" }
```
dmatveev, line 203: `AVOID`s are NOT required when DQ is in place. So please remove `AVOID`s from the default config if you enable DQ.

dmatveev (changes requested):

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Done

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -52,6 +52,36 @@ void align_u4_zp_constants(const std::shared_ptr<ov::Model>& model) {
...
+        return std::make_optional(it->second);
+    }
+    return std::nullopt;
+}
+
+void enable_npuw_dq_if_allowed(ov::AnyMap& config,
+                               const std::shared_ptr<ov::Model>& model) {
+    if (allow_to_enable_npuw_dq(model)) {
+        config["NPUW_DQ"] = "YES";
+        pop_option(config, "NPUW_ONLINE_AVOID");
```
dmatveev, line 81: AVOIDs shouldn't be required in general, I'd recommend to remove this as part of the next config update (but in scope for 2024.5)

dmatveev (approved):

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Scheduled

PR #886, opened 2024-09-20
Fixed logprobs for greedy.

popovaan:
Applied LogSoftmax to logits returned by _greedy_sample().
Ticket: CVS-152274

### DIFF
--- a/src/cpp/src/logit_processor.hpp
+++ b/src/cpp/src/logit_processor.hpp
@@ -299,6 +299,20 @@ public:
     }
 };
 
+
+class LogFilter : public ILogitTransformer {
+public:
+    LogFilter() {}
+    
+    void apply(Logits& logits) override {
+        OPENVINO_ASSERT(logits.is_vector_initialized());
+        for (size_t i = 0; i < logits.m_size; i++) {
+            logits.m_vector[i].m_log_prob = std::log(logits.m_vector[i].m_log_prob);
+        }
+    }
+};
+
+
 } // namespace LogitTransformers
 
 class LogitProcessor {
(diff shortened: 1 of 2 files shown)


popovaan:
I checked multinomial and it returns logprobs in range in [0,1], expected (-inf, 0].
So maybe add logsoftmax to LogitTransformers? 
https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/logit_processor.hpp#L325-L355

dtrawins:
> I checked multinomial and it returns logprobs in range in [0,1], expected (-inf, 0]. So maybe add logsoftmax to LogitTransformers so it applied for both greedy and multinomial? https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/logit_processor.hpp#L325-L355

Indeed it would be the best to include log10 operation to the current prob values so the returned results would match the logprobs property name.

popovaan:
> > I checked multinomial and it returns logprobs in range in [0,1], expected (-inf, 0]. So maybe add logsoftmax to LogitTransformers so it applied for both greedy and multinomial? https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/logit_processor.hpp#L325-L355
> 
> Indeed it would be the best to include log10 operation to the current prob values so the returned results would match the logprobs property name.

Done, please check.

ilya-lavrenov (approved):

src/cpp/src/logit_processor.hpp
```diff
@@ -299,6 +299,20 @@ class PresencePenaltyTransform : public IPenaltyTransformer {
     }
 };
 
+
+class LogFilter : public ILogitTransformer {
+public:
+    LogFilter() {}
+    
+    void apply(Logits& logits) override {
+        OPENVINO_ASSERT(logits.is_vector_initialized());
```
mzegla, line 308: This is true only if we have top_p or top_k set. Without them we don't need to sort results hence new vector is not initialized. So it will not work if only temperature is provided. 
To make it work in both scenarios we need to change `logits.m_vector[i].m_log_prob` values **if vector is initialized** and 
`logits.m_data[i]` **if vector is not initialized**.

PR #887, opened 2024-09-20
Added support of HF and GenAI models into CLI

AlexKoff88:


AlexKoff88:
@sswierze, @sstrehlk

AlexKoff88:
@eaidova, does it seem to me or llm_bench test suite was turned off for some reason?

llm_bench/python/who_what_benchmark/requirements.txt
```diff
@@ -3,6 +3,8 @@ sentence-transformers>=2.2.2
 openvino>=2023.3.0
 openvino-telemetry>=2023.2.1
 optimum-intel>=1.14
+openvino-tokenizers>=2023.3.0
+openvino-genai>=2024.3.0
```
eaidova, line 7: I suppose it will not work if openvino, openvino-tokenizers and openvino-genai installed with different version. Possibly it is better to set the same lower bound for all 3 packages

eaidova (approved):

PR #889, opened 2024-09-23
Fixed regression in multinomial.

popovaan:
Fixed regression after https://github.com/openvinotoolkit/openvino.genai/pull/886.
Turned on multinomial tests without ref comparison.

### DIFF
--- a/src/cpp/src/logit_processor.hpp
+++ b/src/cpp/src/logit_processor.hpp
@@ -300,25 +300,6 @@ public:
 };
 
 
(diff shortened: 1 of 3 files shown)


mzegla (approved):

dtrawins (changes requested):
it works in greedy search but there are incorrect generations in multinomial algorithm

dtrawins (approved):

iefode (approved):
LGTM

PR #898, opened 2024-09-26
Add kvcache convert to fp16 for StaticLLMPipeline

eshiryae:


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -16,6 +17,24 @@
 
 namespace {
 
+std::shared_ptr<ov::Model> cvt_kvcache_to_fp16(const std::shared_ptr<ov::Model>& model) {
```
TolyaTalamanov, line 20: @dmatveev do we still need to stick to the approach w/o `PrePostProcessor` or it can be safely used?

TolyaTalamanov: The answer - yes

TolyaTalamanov (approved):
Great! LGTM 👍 

Thank you @eshiryae

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -16,6 +17,24 @@
...
+
+    for (auto tensor : model->inputs()) {
+        if (tensor.get_any_name().find("past_key") != std::string::npos) {
+            ppp.input(tensor.get_any_name()).tensor().set_element_type(ov::element::Type_t::f16);
+        }
+    }
+
+    for (auto tensor : model->outputs()) {
+        if (tensor.get_any_name().find("present") != std::string::npos) {
+            ppp.output(tensor.get_any_name()).tensor().set_element_type(ov::element::Type_t::f16);
```
Wovchena, line 31: Why do the cache need to be converted to `f16` instead of preserving the original precision?

TolyaTalamanov: As cache should be copied from device to host on every iteration, making it f16 (having less byte) makes it faster on NPU

Wovchena: I suspect a quantized model would have quantized cache which has fewer bytes compared to f16.

dmatveev:
@TolyaTalamanov y it is not merged yet

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: KV cache tensors cannot be compressed via static quantization, so that even for compressed model they will be in float precision

PR #900, opened 2024-09-26
[Continuous Batching] Introduce echo parameter

mzegla:


src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -78,7 +78,8 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
     size_t max_length = SIZE_MAX;
     bool ignore_eos = false;
     size_t min_new_tokens = 0;
-
+    bool echo = true; //TODO: switch to false
```
Wovchena, line 81: Echo usually repeats exactly after a user, but there's nothing to repeat after except for the probmpt. This looks more like debug logging. While adding a function for debugging is debatable in https://github.com/openvinotoolkit/openvino.genai/pull/677#discussion_r1722008733, your case is debug logging. It must be handled with compile time definition or env var

Wovchena (changes requested):

src/cpp/include/openvino/genai/generation_config.hpp
mzegla: I'd argue that it's not just debug logging. `echo` is a regular parameter in OpenAI completions API:
https://platform.openai.com/docs/api-reference/completions/create#completions-create-echo
It repeats the prompt, but also processes it - so it can be valuable to analyze the prompt. When `echo` is set, results for the prompt are sent along with the completion (unless we set max_new_tokens == 0, then only prompt is returned since there is no completion).

vLLM supports it as well:
https://github.com/vllm-project/vllm/pull/1504/files

Wovchena: Ok.

Please, add dockstrings and python bindings

mzegla: Sure, I'll still have to work on the overall correctness, then I'll add it + tests for this feature as well.

src/cpp/src/sequence_group.hpp
```diff
@@ -421,13 +424,13 @@ class SequenceGroup {
 
     // returns context length taking into account scheduled tokens
     size_t get_context_len() const {
-        OPENVINO_ASSERT(!has_finished());
+        // OPENVINO_ASSERT(!has_finished());
```
Wovchena, line 427: Remove or uncomment.

tests/python_tests/test_sampling.py
```diff
@@ -123,7 +123,7 @@ class PlatformsRefTexts(TypedDict, total=False):
     darwin: List[List[str]]
 
 
-def get_current_plarform_ref_texts(ref_texts: PlatformsRefTexts) -> List[List[str]]:
+def get_current_platform_ref_texts(ref_texts: PlatformsRefTexts) -> List[List[str]]:
```
Wovchena, line 126: What changed here?

tests/python_tests/test_sampling.py
```diff
@@ -317,6 +317,64 @@ def test_individual_generation_configs_random(tmp_path, test_struct: RandomSampl
     # Discrete_distribution impl depends on platform, model inference results may depend on CPU.
 
 
+@pytest.mark.precommit
+@pytest.mark.parametrize("sampling_config", [get_greedy(), get_beam_search(), get_multinomial_all_parameters()])
+@pytest.mark.parametrize("max_num_batched_tokens", [2, 4, 256])
+def test_echo_without_completion(tmp_path, sampling_config, max_num_batched_tokens):
+    generation_config = sampling_config
```
Wovchena, line 324: It python it's just reference assignment. You probably need `copy.deepcopy()`.

tests/python_tests/test_sampling.py
```diff
@@ -317,6 +317,64 @@ def test_individual_generation_configs_random(tmp_path, test_struct: RandomSampl
...
+
+    outputs = pipe.generate(["What is OpenVINO?"], generation_configs)
+    assert(len(outputs))
+    for output in outputs:
+        assert(len(output.m_generation_ids))
+        for sequence in output.m_generation_ids:
+            assert(sequence == "What is OpenVINO?")
+
+    del pipe
+    shutil.rmtree(model_path)
```
Wovchena, line 347: You've already requested `tmp_path` from pytest. Why clean in manually?

Wovchena (changes requested):

tests/python_tests/test_sampling.py
Wovchena: Never mind :)

tests/python_tests/test_sampling.py
mzegla: Even if the test is parametrized? It's called 9 times and I suppose every time we get new sampling_config so even if it's just a reference isn't it okay?

tests/python_tests/test_sampling.py
mzegla: I was inspired by: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/common.py#L339
I thought that even if I don't use high level methods from `common` I should follow their flow.

tests/python_tests/test_sampling.py
Wovchena: You shouldn't introduce one extra reference at all in that case.

Wovchena (approved):

PR #921, opened 2024-10-07
Echo check

mzegla:


mzegla:
Task continued in: https://github.com/openvinotoolkit/openvino.genai/pull/900

PR #945, opened 2024-10-11
Update optimum-intel

wgzintel:

PR #946, opened 2024-10-11
Limit optimum to 1.22

wkobielx:


llm_bench/python/who_what_benchmark/requirements.txt
```diff
@@ -2,7 +2,9 @@ transformers>=4.35.2
 sentence-transformers>=2.2.2
 openvino>=2024.3.0
 openvino-telemetry
-optimum-intel>=1.14
+git+https://github.com/huggingface/optimum-intel.git@f34bd61df89f57f61c282c02297980299981ee78#egg=optimum-intel
+#limit optimum to <= 1.22
+optimum<=1.22
```
andrei-kochin, line 7:
```suggestion
```

andrei-kochin (changes requested):
We should not limit optimum while instead we should take the latest optimum-intel same as in #945

wkobielx:
Closing in favor of #945

PR #956, opened 2024-10-14
Update Stable Diffusion models comparison

AlexKoff88:


llm_bench/python/who_what_benchmark/tests/test_cli_text.py
```diff
@@ -15,6 +15,7 @@
 
 
 def run_wwb(args):
+    print(" ".join(["wwb"] + args))
```
eaidova, line 18: is it deug print?

eaidova (approved):

PR #966, opened 2024-10-15
Optimize the code of benchmark.py

wgzintel:
1. Split the code in the benchmark into multiple files
2. The index of the prompt added in printing is represented by P[ ]

example:
[ INFO ] [1][P0] Input token size: 32, Output size: 128, Infer count: 128, Tokenization Time: 0.33ms, Detokenization Time: 0.26ms, Generation Time: 1.46s, Latency: 11.43 ms/token
[ INFO ] [1][P0] First token latency: 14.82 ms/token, other tokens latency: 11.38 ms/token, len of tokens: 128 * 1
[ INFO ] [1][P0] First infer latency: 14.11 ms/infer, other infers latency: 10.64 ms/infer, inference count: 128
[ INFO ] [1][P0] Result MD5:['bd3f7255c27ada5f0de7d9a576a01304']
[ INFO ] [1][P1] Input token size: 1024, Output size: 128, Infer count: 128, Tokenization Time: 2.55ms, Detokenization Time: 0.94ms, Generation Time: 1.59s, Latency: 12.39 ms/token
[ INFO ] [1][P1] First token latency: 111.04 ms/token, other tokens latency: 11.59 ms/token, len of tokens: 128 * 1
[ INFO ] [1][P1] First infer latency: 110.39 ms/infer, other infers latency: 10.96 ms/infer, inference count: 128
[ INFO ] [1][P1] Result MD5:['d64a1bcacc4ab9805c2160c260016499']
[ INFO ] <<< Warm-up iteration is excluded. >>>
[ INFO ] [Total] Iterations: 2
[ INFO ] [Average] P[0] Input token size: 32, 1st token lantency: 14.82 ms/token, 2nd token lantency: 11.38 ms/token, 2nd tokens throughput: 87.85 tokens/s
[ INFO ] [Average] P[1] Input token size: 1024, 1st token lantency: 111.04 ms/token, 2nd token lantency: 11.59 ms/token, 2nd tokens throughput: 86.26 tokens/s


llm_bench/python/llm_bench_utils/bench_classification.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
```
eaidova, line 1: do we really need that? We do not use llm bench for classification models benchmarking, maybe it is good time for cleanup?

llm_bench/python/llm_bench_utils/bench_ldm_sr.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8 -*-
```
eaidova, line 1: I do not fully understand logic for splitting benchmarks by files here.
it can be better if it will be strong logically enough and readable in file naming, so I suggest 2 variants how to better organize that:
* by modality (image (image generation, ldm super resolution) /tex (llms text generation) )/audio (in future whisper, tts models)/multimodality (in future for visiual-language models like llava))

by task (text-generation, image-generation, super_resolution, asr e.t.c.)

@wgzintel  @peterchen-intel  which from these notation will be better?

eaidova: also not sure that llm_bench_utils is right place for that, it contains a lot of different utilities, while benchmarking code is important part of benchmark and should be visible for users ( it is not good to mix it with other utilities, maybe better to have separated namespace for that (e.g. from use_cases/text_generation.py), possibly out of llm_benhc_utils

llm_bench/python/llm_bench_utils/bench_classification.py
wgzintel: cleanup.

llm_bench/python/llm_bench_utils/bench_ldm_sr.py
wgzintel: @eaidova  organized by task and move files to folder task. Could you take a look please?

eaidova (approved):

PR #969, opened 2024-10-15
StaticLLMPipeline: Update config

TolyaTalamanov:


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -212,36 +211,48 @@ void merge_config_with(ov::AnyMap& lhs, const ov::AnyMap& rhs) {
...
-    };
+struct NPUDesc {
+    bool support_max_mem_alloc_size;
+};
+
+ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model,
+                                     const std::optional<NPUDesc>& desc) {
+    auto config = get_baseline_common_config();
+    if (desc.has_value() && desc->support_max_mem_alloc_size) {
+        config.emplace("NPUW_PMM", "NO");
```
dmatveev, line 234: Disabled PMM can be in the base (common) config I believe

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -212,36 +211,48 @@ void merge_config_with(ov::AnyMap& lhs, const ov::AnyMap& rhs) {
...
-        { "NPUW_ONLINE_AVOID", "P:RMSNorm/NPU" }
-    };
+struct NPUDesc {
+    bool support_max_mem_alloc_size;
+};
+
+ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model,
+                                     const std::optional<NPUDesc>& desc) {
+    auto config = get_baseline_common_config();
+    if (desc.has_value() && desc->support_max_mem_alloc_size) {
```
dmatveev, line 233: Also, is this check enough? shouldn't you check for the actual value?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -212,36 +211,48 @@ void merge_config_with(ov::AnyMap& lhs, const ov::AnyMap& rhs) {
...
+        config.emplace("NPUW_PMM", "NO");
+        config.emplace("NPUW_FUNCALL_FOR_ALL", "YES");
```
dmatveev, line 235: Once FCFA is on, the PARALLEL_COMPILE can be ON for both models

dmatveev (commented):
okay-ish for now but, you now, this never ends

dmatveev (approved):
So for it and merge., Tomorrow there will be another one. :D

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -212,36 +211,48 @@ void merge_config_with(ov::AnyMap& lhs, const ov::AnyMap& rhs) {
     }
 }
 
-ov::AnyMap get_default_prefill_config(const std::shared_ptr<ov::Model>& model) {
+ov::AnyMap get_baseline_common_config() {
     ov::AnyMap config = {
-        { "NPU_USE_NPUW", "YES" },
+        { "NPU_COMPILATION_MODE_PARAMS", "compute-layers-with-higher-precision=Sqrt,Power,ReduceMean,Add_RMSNorm" },
```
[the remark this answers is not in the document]
TolyaTalamanov: Yes, it should be just `Add` perhaps

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Will take it into account for the next PR that enables FCFA, thanks!

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: DOne

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: This part is removed, will be added in next PRs

PR #986, opened 2024-10-16
Whisper pipeline: fix timestamps extract

as-suvorov:
This PR adds:
- [x] Fix distil whisper accuracy, match with HF
- [x] Fix en models accuracy with timestamps, match with HF
- [x] add cpp, python samples checks

- [ ] Long-form audio support with [parallel chunking](https://huggingface.co/blog/asr-chunking).
- [ ] update documentation
- [ ] Add batch
- [ ] Try to trim input_ids cache between chunks for long-form audio to match HF

Completed:
- [x] support translation
- [x] support timestamps
- [x] Long-form audio support with sequential chunking
- [x] support perf metrics
- [x] Fix stop criteria. Current approach stops on eos_token which is no speech token. But there could be more speech tokens further which are wrongly skipped now. Investigation result: expected behavior, matches with HF. Enable 'return_timestamps' to get full transcription with no speech in the middle of audio.

Current limitations:

Tickets: CVS-147994, CVS-146010


as-suvorov:
> I got Segmentation fault for `whisper_speech_recognition whisper-base/ ~/Downloads/how_are_you_doing_today.wav` https://storage.openvinotoolkit.org/models_contrib/speech/2021.2/librispeech_s5/how_are_you_doing_today.wav

Could you please check if it's reproducible on master for you?

Wovchena:
GenAi master works

sbalandi (approved):

Wovchena:
I still get Segmentation fault (core dumped) on the same command

Wovchena (approved):

PR #990, opened 2024-10-16
OV Core fix for text2image

likholat:
Ticket: CVS-155261


src/cpp/include/openvino/genai/text2image/txt2img_core.hpp
```diff
@@ -0,0 +1,16 @@
...
+
+#pragma once
+
+#include "openvino/runtime/core.hpp"
+
+namespace ov {
+namespace genai {
+namespace txt2img_core {
+
+ov::Core singleton_core();
```
Wovchena, line 12: Make it private. An attempt to keep it private is the reason we have this issue now, so stick to the original intention.

ilya-lavrenov (approved):

Wovchena (approved):

PR #1043, opened 2024-10-22
StaticLLMPipeline: Pad KV cache tensors with zeros

TolyaTalamanov:


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -680,6 +681,8 @@ EncodedResults StaticLLMPipeline::generate(
 
         const auto& input_name = kvcache_compiled.inputs()[kStartInputKVCacheLayers + i].get_any_name();
         auto kvcache_in_tensor = m_kvcache_request.get_tensor(input_name);
+        fill_tensor<ov::float16>(kvcache_in_tensor, 0);
```
TolyaTalamanov, line 684: Seems with the latest openvino it's crucial, otherwise lead to accuracy problems

dmatveev (approved):
Should we have the same in our tooling?

PR #1062, opened 2024-10-24
[WWB]: Improved README

AlexKoff88:


ilya-lavrenov (approved):

tools/who_what_benchmark/README.md
```diff
@@ -11,8 +16,46 @@
 * Validation of text-to-image pipelines. Computes similarity score between generated images:
     * Supports Diffusers library and Optimum-Intel via `Text2ImageEvaluator` class.
 
-The main idea is to compare similarity of text generation between baseline and optimized LLMs.
+### Installation
+Install WWB and its requirements from the source using `pip` or any other package manager. For example,
+
+* `python -m venv eval_env`
+* `source eval_env/bin/activate`
+* `pip install -r requirements.txt`
```
eaidova, line 24: does requirement.txt installation separately from tool is necessary?

tools/who_what_benchmark/README.md
```diff
@@ -11,8 +16,46 @@
...
+wwb --target-model phi-3-openvino --gt-data gt.csv --model-type text
+
+# Measure similarity metric for OpenVINO GenAI inference backend
+wwb --target-model phi-3-openvino --gt-data gt.csv --model-type text --genai
+```
 
+### Compare Text-to-image models (Diffusers)
+```sh
+# Collect ground truth from the baseline Hugging Face Transformer model 
+wwb --base-model microsoft/Phi-3-mini-4k-instruct --gt-data gt.csv --model-type text --hf
```
eaidova, line 46: looks like text-to-image examples missed and accidentally copied llm example for phi3 twice....

tools/who_what_benchmark/README.md
AlexKoff88: I think it is ok for newcomers. Runtime folks came and ask about the tool usage and it was helpful for them as they are not Python users. So I would keep this.

tools/who_what_benchmark/README.md
AlexKoff88: Sorry, didn't submit the latest version.

AlexKoff88: Fixed

tools/who_what_benchmark/README.md
```diff
@@ -11,8 +16,43 @@
...
+```
+
+### Compare Text-to-image models (Diffusers)
+```sh
+# Export FP16 model to OpenVINO
+optimum-cli export openvino -m SimianLuo/LCM_Dreamshaper_v7 --weight-format fp16 sd-lcm-fp16
+# Export model with 8-bit quantized weights to OpenVINO
+optimum-cli export openvino -m SimianLuo/LCM_Dreamshaper_v7 --weight-format int8 sd-lcm-int8
+# Collect the references
+wwb --base-model sd-lcm-fp16 --gt-data lcm_test/sd_xl.json --model-type text-to-image
```
eaidova, line 50: running with --gt-data argument slightly confusing me, as user who not yet looking into help yet (of course I read that, just imagine the situation), I may suppose that it is some file that required to be manually prepared or store dataset input entries. Maybe it make sense to a little bit extend comment that reference data will be saved in this file?

P.S. I also found that there is no mention json for text2image case support in helper string and ther is small typo here 

https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/wwb.py#L219

I defined and not exists -> If defined and not exists

eaidova: Also I already mentioned that in chat, if argument not provided, tool will file with not very clear error message

```
  File "/home/ea/work/py311/lib/python3.11/site-packages/whowhatbench/text2image_evaluator.py", line 59, in __init__

    self.gt_dir = os.path.dirname(gt_data)
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 152, in dirname
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

is it possible to handle case and ask to provide instead?

eaidova (approved):

tools/who_what_benchmark/README.md
AlexKoff88: ok, from what I got you are proposing:
- Extend the comment and make it clear why .json file is used
- Verify that the path specified by the argument exists

right?

eaidova: > Verify that the path specified by the argument exists

no, I mean not verify that path exists (in base line running it is expected that it may not exists), but verify that argument provided, the default argument is None now if user run command like:

```
wwb --base-model SimianLuo/LCM_Dreamshaper_v7 --model-type text-to-image
```
i

AlexKoff88: got you. Thanks for discovering this.

AlexKoff88: fixed both comments. Added meaningful messages instead of asserts.

tools/who_what_benchmark/README.md
```diff
@@ -11,8 +16,43 @@
 * Validation of text-to-image pipelines. Computes similarity score between generated images:
     * Supports Diffusers library and Optimum-Intel via `Text2ImageEvaluator` class.
 
-The main idea is to compare similarity of text generation between baseline and optimized LLMs.
+### Installation
+Install WWB and its requirements from the source using `pip` or any other package manager. For example,
+
+* `python -m venv eval_env`
+* `source eval_env/bin/activate`
+* `pip install -r requirements.txt`
```
andreyanufr, line 24: I think we need to add diffusers to requirements.txt

AlexKoff88: fixed

PR #1065, opened 2024-10-24
Crop Tokens, Not Text

apaniukov:
Custom tokenizers might not skip special tokens during detokenization, which may result in the wrong prompt cropping.

ChatGLM3 similarity score jumps from 0.89 to 1 after the fix.


AlexKoff88 (approved):

AlexKoff88:
Thanks, @apaniukov, for the investigation and fix.

AlexKoff88:
@apaniukov, I think we should consider adding something like `crop_prompt` to GenAI .generate() API? Or at least showcase to users how to use the generation API properly so that results will match what one can see in the framework.

PR #1077, opened 2024-10-25
StaticLLMPipeline: Update config

TolyaTalamanov:


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -254,13 +247,15 @@ ov::AnyMap get_baseline_common_config() {
 
 ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model) {
     auto config = get_baseline_common_config();
+    if (is_cw_compressed(model)) {
+        config.emplace("NPUW_SLICE_OUT", "YES");
+    }
```
dmatveev, line 252: SLICE_OUT can be enabled unconditionally, why only CW?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -271,8 +266,8 @@ ov::AnyMap get_default_prefill_config(const std::shared_ptr<ov::Model>& model,
         return { };
     }
     auto config = get_default_common_config(model);
-    if (get_option<std::string>(config, "NPUW_FUNCALL_FOR_ALL").value_or("NO") == "YES") {
-        config.emplace("NPUW_PARALLEL_COMPILE", "YES");
+    if (is_cw_compressed(model)) {
+        config.emplace("NPUW_DQ", "YES");
```
dmatveev, line 270: If we don't enable DQ (see `else`), I'd also recommend to disable PMM (`PMM : NO`)

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Thought there were regressions for this case, will change

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Done

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -260,37 +255,33 @@ ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model) {
...
-        config.emplace("NPU_DPU_GROUPS", npudesc->max_tiles);
+    if (is_cw_compressed(model)) {
+        config.emplace("NPUW_DQ", "YES");
+    } else {
+        config.emplace("NPUW_PMM", "NO");
+        if (npudesc.has_value() &&
+            npudesc->arch == "4000" &&
+            npudesc->max_tiles != -1) {
+            config.emplace("NPU_DPU_GROUPS", npudesc->max_tiles);
+        }
```
dmatveev, line 272: you set DPU groups here only if it is GQ.

In fact, this setting should be done regardless of the Q type.

dmatveev (changes requested):

src/cpp/src/llm_pipeline_static.cpp

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -260,19 +255,16 @@ ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model) {
...
-    enable_npuw_dq_if_allowed(config, model);
     return config;
 }
 
 ov::AnyMap get_default_prefill_config(const std::shared_ptr<ov::Model>& model,
                                       const std::optional<NPUDesc>& npudesc) {
-    // NB: Empty config for devices != NPU
-    if (!npudesc.has_value()) {
-        return { };
-    }
```
dmatveev, line 272: is it ok?

dmatveev (approved):
Let's do this!

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Decided not to fallback into CPU but user NPUW:CPU instead

PR #1079, opened 2024-10-25
Speculative decoding - improve synchronization, fix handle notification

mzegla:


src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -218,8 +214,6 @@ UpdateRequestResult
 ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::update_request(uint64_t request_id,
                                                                                          const GeneratedSequences& candidates,
                                                                                          bool is_update_logit_processor) {
-    _pull_awaiting_requests();
```
iefode, line 221: we need to stay this line here because in `draft_model` requests will be running, and in `main_model` they be awaiting...

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -20,7 +20,6 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::Contin
 
 void
 ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::finish_request(SequenceGroup::Ptr request) {
```
iefode, line 22: Minor comment: generations of draft model will be lost, so speculative decoding won't be so good here.. anyway, not a problem at all, because request synchronization and draft_model generation handling is added here: https://github.com/openvinotoolkit/openvino.genai/pull/1056/files

iefode (commented):
In general, LGTM. Please fix the comment about `_pull_awaiting_request`. It's correct and will fix the ci

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -197,7 +194,6 @@ UpdateRequestResult
 ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::init_request_by_candidate(
     uint64_t request_id,
     const GeneratedSequences& candidates) {
-    _pull_awaiting_requests();
```
iefode, line 200: we need to stay this line here because in `draft_model` requests will be running, and in `main_model` they be awaiting... in init
so, my suggestion is:
```
    if (m_requests.empty()) {
        _pull_awaiting_requests();
    }
```

mzegla:
Fixes applied in: https://github.com/openvinotoolkit/openvino.genai/commit/49a12d5368aefb0f56f744ee479466135befc7ef

PR #1086, opened 2024-10-28
[WWB]: Add support for VisualLanguage models

AlexKoff88:


ilya-lavrenov (approved):

AlexKoff88:
@eaidova, please take a look as well.

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -42,6 +44,8 @@ def __init__(self, model, model_dir, model_type):
         self.model = model
         self.model_type = model_type
 
+        if model_type == "text" or model_type == "visual-text":
+            self.config = AutoConfig.from_pretrained(model_dir, trust_remote_code=True)
         if model_type == "text":
             self.config = AutoConfig.from_pretrained(
                 model_dir, trust_remote_code=True)
```
eaidova, line 51: looks like at this moment you already have loaded config according to condition above and you load it twice here

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
```diff
@@ -0,0 +1,175 @@
...
+    MODEL_TYPE_TO_CLS_MAPPING
+from PIL import Image
+from tqdm import tqdm
+from transformers import set_seed
+
+from .registry import register_evaluator
+from .text_evaluator import TextEvaluator
+
+
+def get_pil_from_url(url):
```
eaidova, line 16: maybe useful to extend this functional to load local images too like https://github.com/huggingface/diffusers/blob/main/src/diffusers/utils/loading_utils.py#L13 in case if dataset may be changed in future and for optimizing loading time for vaidation

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -147,11 +156,60 @@ def load_text2image_model(
     return model
 
 
+def load_visual_text_genai_pipeline(model_dir, device="CPU", ov_config=None):
+    try:
+        import openvino_genai
+    except ImportError:
+        logger.error("Failed to import openvino_genai package. Please install it.")
```
eaidova, line 163: does it make sense to print original error message too here?

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
AlexKoff88: what if I just import this function from there?

tools/who_what_benchmark/whowhatbench/wwb.py
AlexKoff88: merge issue, fixed

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
eaidova: yes, I think it is possible, the only thing it will requires to have diffusers at the moment of VLM models evaluation only for images loading. I found similar helper in transformers too, so probably to use it https://github.com/huggingface/transformers/blob/eed11f34abe558c76590366a21316318e5f820ed/src/transformers/image_utils.py#L344

eaidova (approved):

tools/who_what_benchmark/whowhatbench/wwb.py
AlexKoff88: why not, added

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
AlexKoff88: replaced

PR #1089, opened 2024-10-29
Corrected performance data for --genai

wgzintel:


eaidova (approved):

peterchen-intel:
@eaidova Shall we consider to remove following line instead of change in benchmark.py APP? https://github.com/openvinotoolkit/openvino.genai/blob/fa324cfaca838591a86cecc14cb44253b7fb7632/src/cpp/src/perf_metrics.cpp#L115
m_durations should be the duration for one inference which may generate one or more (batch_size > 1) tokens.  In the case  batch_size>1, it means @batch_size tokens are generated together in @m_durations[i] time, not one token is generated in @m_durations[i]/@batch_size

eaidova:
> @eaidova Shall we consider to remove following line instead of change in benchmark.py APP?
> 
> https://github.com/openvinotoolkit/openvino.genai/blob/fa324cfaca838591a86cecc14cb44253b7fb7632/src/cpp/src/perf_metrics.cpp#L115
> 
> @m_durations[i] should be the duration for one inference which may generate one or more (batch_size > 1) tokens. In the case batch_size>1, it means @batch_size tokens are generated together in @m_durations[i] time, not one token is generated in @m_durations[i]/@batch_size time

@peterchen-intel  yes, for me it looks preferable

PR #1102, opened 2024-10-29
Whisper pipeline: reset decoder with past state for each chunk

as-suvorov:
* reset decoder_with_past state for each chunk
* set transformers>=4.43 as it's required for whisper export with `cache_position` input
* add streamer tests
* remove streamer from samples. Timestamps extraction can drop segments in case of audio cut of in the middle of the word. Or if `max_new_tokens` stops generation in the middle of segment. This could cause streamer output and generate function result mismatch. Fix to be discussed.


ilya-lavrenov (approved):

eaidova (approved):

as-suvorov:
@ilya-lavrenov I added tests to check if streamer output matches pipeline result. The distil-small.en model fails tests. One of the chunks (30 s window) has no ending timestamp token (same for HF). Not finished segment tokens excluded from the result. But were sent to the streamer right after argmax.
For the moment I added assert to forbid streamer with `return_timestamps=True` and thus long-form audio.
As a fix we can 'stream' the whole chunk tokens after timestamps extraction.

ilya-lavrenov (approved):

PR #1103, opened 2024-10-29
StaticWhisperPipeline change to work with optimum models

eshiryae:


src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -340,16 +542,57 @@ WhisperPipeline::StaticWhisperPipeline::StaticWhisperPipeline(const std::filesys
     auto decoder_model = core.read_model(models_path / "openvino_decoder_model.xml");
     auto decoder_with_past_model = core.read_model(models_path / "openvino_decoder_with_past_model.xml");
 
+    add_attention_mask_input_for_decoder(decoder_model);
```
as-suvorov, line 545: Does it make sense to move `add_attention_mask_input`, `reshape_to_static`, `preprocess_**` functions to a separate file under whisper folder?

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -340,16 +542,57 @@ WhisperPipeline::StaticWhisperPipeline::StaticWhisperPipeline(const std::filesys
...
 
     // TODO: There must be model reshape to eliminate dynamism!
-
-    m_models.encoder = core.compile_model(encoder_model, "NPU").create_infer_request();
-    m_models.decoder = core.compile_model(decoder_model, "NPU").create_infer_request();
-    m_models.decoder_with_past = core.compile_model(decoder_with_past_model, "NPU").create_infer_request();
+    size_t max_sequence_length = 128;
+
+    reshape_to_static_encoder(encoder_model);
+    reshape_to_static(decoder_model, 4, 4);  // What is 4 here??
```
as-suvorov, line 557: 4 is max number of input tokens for decoder

src/cpp/src/whisper_pipeline.cpp
```diff
@@ -44,12 +87,13 @@ class WhisperPipeline::WhisperPipelineStatefulImpl : public WhisperPipeline::Whi
...
             core.compile_model((models_path / "openvino_encoder_model.xml").string(), device, compile_properties)
-                .create_infer_request();
-        m_models.decoder =
-            core.compile_model((models_path / "openvino_decoder_model.xml").string(), device, compile_properties)
-                .create_infer_request();
+                               .create_infer_request();
+        m_models.decoder = core.compile_model((models_path / "openvino_decoder_model.xml").string(), device, compile_properties)
+                               .create_infer_request();
+        auto decoder_with_past_model = core.read_model(models_path / "openvino_decoder_with_past_model.xml");
+        add_attention_mask_input(decoder_with_past_model);
```
as-suvorov, line 94: Is it required to add attention mask for stateful pipeline?

src/cpp/src/whisper_pipeline_static.cpp
TolyaTalamanov: I don't think so, they're relevant only for static shape implementation

src/cpp/src/whisper_pipeline_static.cpp
TolyaTalamanov: This comment should be removed

src/cpp/src/whisper_pipeline.cpp
TolyaTalamanov: No, `stateful` (in fact `dynamic`) pipeline handles it by providing dynamic shape input. 

For static shape pipeline attention mask is needed to specify what part of static shape input should be used.

as-suvorov: Should it be removed then?

eshiryae: These changes are prototype from Sergey, need to revert all changes in [whisper.cpp](https://github.com/openvinotoolkit/openvino.genai/pull/1103/files#diff-9eb3f21f8f27be6218b861f0159dc50546537f1bef39eb1a4ecf9b626ec8918c) and [whisper_pipeline.cpp](https://github.com/openvinotoolkit/openvino.genai/pull/1103/files#diff-501fdc7aca2545fe1133079658cae9efe4ab262da7998a0ea2a811714312b76d)

src/cpp/src/whisper_pipeline_static.cpp
eshiryae: Done

src/cpp/src/whisper_pipeline.cpp
eshiryae: Done

src/cpp/src/whisper_pipeline.cpp
```diff
@@ -27,6 +36,40 @@ ov::genai::OptionalWhisperGenerationConfig get_config_from_map(const ov::AnyMap&
 }
 }  // namespace
 
+
+void add_attention_mask_input(std::shared_ptr<ov::Model> model) {
```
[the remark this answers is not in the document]
eshiryae: Fixed

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+            preprocessor.input("attention_mask").tensor().set_element_type(ov::element::Type_t::f16);
+            preprocessor.input("attention_mask").preprocess().convert_element_type();
+        } else if (tensor.get_any_name().find("encoder_hidden_states") != std::string::npos) {
+            preprocessor.input("encoder_hidden_states").tensor().set_element_type(ov::element::Type_t::f16);
+            preprocessor.input("encoder_hidden_states").preprocess().convert_element_type(ov::element::Type_t::f32); // ()
+        } else if (tensor.get_any_name().find("past_key_values") != std::string::npos) {
+            preprocessor.input(tensor.get_any_name()).tensor().set_element_type(ov::element::Type_t::f16);
+            preprocessor.input(tensor.get_any_name()).preprocess().convert_element_type();
+
+            // if (tensor.get_any_name().find(".value") != std::string::npos) {
```
as-suvorov, line 474: Seems redundant, do we need to keep it?

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+            //}
+        }
+    }
+
+    for (auto tensor : model->outputs()) {
+        if (tensor.get_any_name().find("present") != std::string::npos) {
+            preprocessor.output(tensor.get_any_name()).tensor().set_element_type(ov::element::Type_t::f16);
+            preprocessor.output(tensor.get_any_name()).postprocess().convert_element_type();
+
+            // if (tensor.get_any_name().find(".value") != std::string::npos) {
```
as-suvorov, line 489: Same here

as-suvorov (approved):

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -160,10 +171,9 @@ int64_t decode_with_past(ov::InferRequest& decoder_with_past,
                          const std::vector<int64_t>& generated_tokens) {
     // FIXME: Avoid this cast to i32. Why it's not i64 precision in model?
     decoder_with_past.get_tensor("input_ids").data<int32_t>()[0] = static_cast<int32_t>(input_id);
```
soumendukrg, line 173: Since optimum creates i64 dtype for input_ids, do we still need this cast? This cast was initially required to support NPU-friendly models.

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+        }
+        new_shapes.emplace(input_name, new_shape);
+    }
+    model->reshape(new_shapes);
+}
+
+void preprocess_encoder(std::shared_ptr<ov::Model> model) {
+    ov::preprocess::PrePostProcessor preprocessor(model);
+
+    preprocessor.input("input_features").tensor().set_element_type(ov::element::Type_t::f32);
```
TolyaTalamanov, line 450: it's already `f32`, isn't it?

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+            });
+        }
+    };
+
+    ov::pass::Manager pm;
+    pm.register_pass<AttentionMaskInput>(model);
+    pm.run_passes(model);
+}
+
+void reshape_to_static(std::shared_ptr<ov::Model> model, const uint32_t input_size, const uint32_t kvcache_size) {
```
TolyaTalamanov, line 400: Perhaps should be separate function for every model to avoid confusion

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+    for (auto tensor : model->outputs()) {
+        if (tensor.get_any_name().find("present") != std::string::npos) {
+            preprocessor.output(tensor.get_any_name()).tensor().set_element_type(ov::element::Type_t::f16);
+            preprocessor.output(tensor.get_any_name()).postprocess().convert_element_type();
+
+            // if (tensor.get_any_name().find(".value") != std::string::npos) {
+            //    preprocessor.output(tensor.get_any_name()).tensor().set_layout(ov::Layout("NCWH"));
+            //    preprocessor.output(tensor.get_any_name()).model().set_layout(ov::Layout("NCHW"));
+            //} else if (tensor.get_any_name().find(".key") != std::string::npos) {
+            //    preprocessor.output(tensor.get_any_name()).tensor().set_layout(ov::Layout("NCHW"));
```
TolyaTalamanov, line 493: Let's remove it with refactoring PR

TolyaTalamanov (approved):
Slight revision of preprocessing stuff perhaps still needed here, but it's not critical for now, let's merge it!

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+            });
+        }
+    };
+
+    ov::pass::Manager pm;
+    pm.register_pass<AttentionMaskInput>(model);
+    pm.run_passes(model);
+}
+
+void add_attention_mask_input(std::shared_ptr<ov::Model> model) {
```
soumendukrg, line 371: Why are there 2 add_attention_mask functions?

src/cpp/src/whisper_pipeline_static.cpp
soumendukrg: I agree. Having separate reshape_to_static for decoder and decoder with past will be helpful.

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+void reshape_to_static(std::shared_ptr<ov::Model> model, const uint32_t input_size, const uint32_t kvcache_size) {
+    std::map<std::string, ov::PartialShape> new_shapes;
+    for (auto input : model->inputs()) {
+        const auto& input_name = input.get_any_name();
+        ov::PartialShape new_shape;
+        if (input_name.find("input_ids") != std::string::npos) {
+            new_shape = ov::PartialShape({1, input_size});
+        } else if (input_name.find("attention_mask") != std::string::npos) {
+            new_shape = ov::PartialShape({1, kvcache_size + 1});
+        } else if (input_name.find("position_ids") != std::string::npos) {
```
soumendukrg, line 409: position_ids is now deprecated as inputs, replaced with cache_position. May be removed?

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+            new_shape = ov::PartialShape({1, input_size});
+        } else if (input_name.find("cache_position") != std::string::npos) {
+            new_shape = ov::PartialShape({1});
+        } else if (input_name.find("encoder_hidden_states") != std::string::npos) {
+            const auto& partial_shape = input.get_partial_shape();
+            new_shape = partial_shape;
+            new_shape[0] = 1;     // batch_dim
+            new_shape[1] = 1500;  // FIXME: is it got from encoder output{'last_hidden_state'}
```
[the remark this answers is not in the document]
soumendukrg: It is however a required input for decoder model and static shapes are 1 for batch and 1500 for encoder sequence length as you have added. However, even the last dimension is also dynamic and varies with model (checked optimum exported model using transformers v4.45.2 and optimum-intel v1.20.0.

Ideally, the encoder output 'last_hidden_state' dimension can be used to reshape the encoder_hidden_states input to decoder. This will be straightforward.

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -326,6 +338,181 @@ bool check_decoder_model_compatibility(const std::shared_ptr<ov::Model>& decoder
...
+
+void reshape_to_static(std::shared_ptr<ov::Model> model, const uint32_t input_size, const uint32_t kvcache_size) {
+    std::map<std::string, ov::PartialShape> new_shapes;
+    for (auto input : model->inputs()) {
+        const auto& input_name = input.get_any_name();
+        ov::PartialShape new_shape;
+        if (input_name.find("input_ids") != std::string::npos) {
+            new_shape = ov::PartialShape({1, input_size});
+        } else if (input_name.find("attention_mask") != std::string::npos) {
+            new_shape = ov::PartialShape({1, kvcache_size + 1});
```
soumendukrg, line 408: Since 448 is passed as kvcache_size, using kvcache_size + 1 creates attention_mask of size 1, 449 which is wrong as 448 is max supported by model. 

Use 1, kvcache_size for mask, as the past_key_values is later reshaped to kvcache_size - 1.

soumendukrg (changes requested):

src/cpp/src/whisper_pipeline_static.cpp
TolyaTalamanov: I believe it's the matter of naming. The real kv cache size is 449 in this case though

src/cpp/src/whisper_pipeline_static.cpp
TolyaTalamanov: I believe it's left from non optimum-cli models, should be removed ofc

src/cpp/src/whisper_pipeline_static.cpp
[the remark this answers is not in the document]
TolyaTalamanov: This function should be split on 3 (`encoder`, `decoder`, `decoder_with_past`) to avoid further confusion - let's do this clean up after main part is merged

src/cpp/src/whisper_pipeline_static.cpp
TolyaTalamanov: I believe it's attempt to align model generated on-the-fly to NPU-friendly one. Perhaps just leave it as `i64` is fine

Issue #259, opened 2024-03-01
[Discussion][Good First Issue]: Verify different LLMs work with text_generation


p-wysocki:
### Context

This is an effort to increase Large Language Models tests coverage in [OpenVINO GenAI](https://github.com/openvinotoolkit/openvino.genai). 

Working on this task will let you familiarize yourself with: 

- C++ inference workflow
- Different LLM models
- GitHub Actions extension
- Adding Python tests
- And more

If you would like to add a new model which there's not a task for, please let us know! We would love to get outside ideas.

### What needs to be done?

0. Verify a model license is compatible with OpenVINO.GenAI's license. The model must have `LICENSE` file or specify a license tag in its repo. For example, https://huggingface.co/TinyLlama/TinyLlama-1.1B-Chat-v1.0/tree/main doesn't have the file, but has the tag `License: apache-2.0`. Apache 2.0 and MIT licenses are compatible with OpenVINO.GenAI's license. For any other license, please, ask in the corresponding Good First Issue if the given license is compatible. If the model's repo doesn't have a license, contact the authors asking to add it by starting a new discussion at model's repo.
1. Try running an LLM specified in your specific task with every sample form https://github.com/openvinotoolkit/openvino.genai/tree/master/samples and tools from https://github.com/openvinotoolkit/openvino.genai/tree/master/tools/continuous_batching. Install GenAi following https://github.com/openvinotoolkit/openvino.genai/blob/master/src/docs/BUILD.md. Perform `Option 2 - setting paths to built OpenVINO GenAI artifacts manually` instead of cmake option because not every sample is installed yet.
2. If it doesn't fail and outputs a reasonable responce, the PR should be submitted to extend the [supported models list](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/docs/SUPPORTED_MODELS.md).
3. If the model is small enough to fit into the default GitHub Actions runner, add the model to tests [here](https://github.com/openvinotoolkit/openvino.genai/blob/84702501b688590457e268f4b7f9c2b0bc012c1b/.github/workflows/causal_lm_cpp.yml#L62).
4. If the comparison against python `transformers` fails it's worth adding a test for the model anyway, just without comparison against python, leave a comment in workflow code if this is the case. Since default runners are available for everyone, one can verify the test passes by opening a PR in their fork first.
[...]

PR #1119, opened 2024-10-31
Verify chatglm3 6b

Aniruddha521:
I proceed as mentioned in the task https://github.com/openvinotoolkit/openvino.genai/issues/259 with the following changes.
**1)** Extended the nightly_model in the file [openvino.genai/tests/python_tests/ov_genai_test_utils.py](https://github.com/openvinotoolkit/openvino.genai/blob/c86fd779d49998a7fa2d5f0f25b2964654d1be25/tests/python_tests/ov_genai_test_utils.py#L22)
```python
nightly_models = [
        "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
        "facebook/opt-125m",
        "microsoft/phi-1_5",
        "microsoft/phi-2",
        "THUDM/chatglm2-6b",
        "THUDM/chatglm3-6b", # no beam_search
        "Qwen/Qwen2-0.5B-Instruct",
        "Qwen/Qwen-7B-Chat",
        "Qwen/Qwen1.5-7B-Chat",
        "argilla/notus-7b-v1",
        "HuggingFaceH4/zephyr-7b-beta",
        "ikala/redpajama-3b-chat",
        "mistralai/Mistral-7B-v0.1",
```
**2)** Added model to

[openvino.genai/.github/workflows/causal_lm_cpp.yml](https://github.com/openvinotoolkit/openvino.genai/blob/84702501b688590457e268f4b7f9c2b0bc012c1b/.github/workflows/causal_lm_cpp.yml#L62)

Line 62 in [8470250](https://github.com/openvinotoolkit/openvino.genai/commit/84702501b688590457e268f4b7f9c2b0bc012c1b)
 run: | 

cpp-greedy_causal_lm-Chatglm3-6b and cpp-prompt_lookup_decoding_lm-ubuntu-Chatglm3-6b

**3)** Extended the [supported_model_list](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/docs/SUPPORTED_MODELS.md) and added note
>[!Note]
> The `beam_search_causal_lm` is not supported in the `ChatGLM3-6B model`.


.github/workflows/causal_lm_cpp.yml
```diff
@@ -535,6 +570,78 @@ jobs:
...
+
+          echo 'Code:```python
+          def add(a, b):
+              return a + b
+          ```
+          Question: Can you please add 2 and 3
+          A:' > ./prompt.txt
+
+          ./build/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm ./TinyLlama-1.1B-Chat-v1.0/ "$(<prompt.txt)" > predictions_prompt_lookup.txt
+          ./build/samples/cpp/greedy_causal_lm/greedy_causal_lm ./TinyLlama-1.1B-Chat-v1.0/ "$(<prompt.txt)" > predictions_greedy.txt
```
Wovchena, line 614: Test `chatglm3-6b` instead.

.github/workflows/causal_lm_cpp.yml
```diff
@@ -535,6 +570,78 @@ jobs:
...
+          python -c "
+          with open('predictions_greedy.txt', 'r') as f:
+              predicted_greedy = f.readline()
+          with open('predictions_prompt_lookup.txt', 'r') as f:
+              predicted_prompt_lookup = f.readline()
+          assert predicted_greedy == predicted_prompt_lookup
+          "
+          echo "Prompt lookup" passed
+      - name: run and compare (model with seq_length_axis = 1)
```
Wovchena, line 623: It seems this step is just a copy of another one. Remove it.

tests/python_tests/ov_genai_test_utils.py
Wovchena, line 1: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/docs/SUPPORTED_MODELS.md isn't updated. Please, extend the note below the table that `optimum-cli` requires `--task text-generation-with-past` for `THUDM/chatglm3-6b` and beam search isn't supported for that model.

Wovchena (changes requested):

.github/workflows/causal_lm_cpp.yml
Aniruddha521: That's a mistake thank you for pointing it out, in my updated commit I have removed `cpp-prompt_lookup_decoding_lm-ubuntu-Chatglm3-6b` and just added the necessary part in `cpp-prompt_lookup_decoding_lm-ubuntu`

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -25,6 +25,7 @@ def get_models_list():
         "microsoft/phi-1_5",
         "microsoft/phi-2",
         "THUDM/chatglm2-6b",
+        "THUDM/chatglm3-6b", # no beam_search
```
Wovchena, line 28: Does every test pass with `THUDM/chatglm3-6b`? If not, please, mark the failing tests to be skipped.

Wovchena: Skips must happen only for that particular model

.github/workflows/causal_lm_cpp.yml
```diff
@@ -274,6 +274,41 @@ jobs:
           && call .\ov\setupvars.bat
           && python samples\python\greedy_causal_lm\lora.py .\TinyLlama\TinyLlama-1.1B-intermediate-step-1431k-3T\ adapter_model.safetensors "How to create a table with two columns, one of them has type float, another one has type int?"
 
+  cpp-greedy_causal_lm-Chatglm3-6b:
+    runs-on: ubuntu-24.04
```
Wovchena, line 278:
```suggestion
    runs-on: ubuntu-20.04-4-cores
```

Aniruddha521:
@Wovchena  can you please guide me that what possibly gone wrong for `cpp-greedy_causal_lm-Chatglm3-6b`  and  `cpp-prompt_lookup_decoding_lm-ubuntu` although both **greedy_causal_lm** and **prompt_lookup_decoding_lm** works for chatglm3-6b in my local machine

.github/workflows/causal_lm_cpp.yml
```diff
@@ -274,6 +274,41 @@ jobs:
...
+          cmake --build ./build/ --config Release -j
+      - name: Download and convert and model
+        run: |
+          source ./ov/setupvars.sh
+          python -m pip install --upgrade-strategy eager -r ./samples/requirements.txt
+          python -m pip install ./thirdparty/openvino_tokenizers/[transformers] --pre --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
+          optimum-cli export openvino --trust-remote-code --weight-format fp16 --model THUDM/chatglm3-6b chatglm3-6b --task text-generation-with-past
+      - run: > 
+          . ./ov/setupvars.sh
+          && timeout 2m ./build/samples/cpp/greedy_causal_lm/greedy_causal_lm ./chatglm3-6b/ 69 | diff <(timeout 2m samples/python/greedy_causal_lm/greedy_causal_lm.py ./chatglm3-6b/ 69) -
```
Wovchena, line 307: > @Wovchena can you please guide me that what possibly gone wrong for `cpp-greedy_causal_lm-Chatglm3-6b` and `cpp-prompt_lookup_decoding_lm-ubuntu` although both **greedy_causal_lm** and **prompt_lookup_decoding_lm** works for chatglm3-6b in my local machine

The samples work but the comparison fails which means the samples produced different results. Try running the samples again, but pay attention to their output.

By the way, here's a newer and more verbose implementation of the same idea using `tee` and `timeout-minutes`. It is also split into multiple steps to make it clear which command fails. You can rewrite your tests in a similar way to simplify reasoning about what happened. https://github.com/openvinotoolkit/openvino.genai/blob/6165c479f2d99b91712bc105fe7eba9139952487/.github/workflows/causal_lm_cpp.yml#L776-L790

Aniruddha521:
@Wovchena

Aniruddha521:
@Wovchena 
Why every test is failing during build app process?

Wovchena:
I guess we broke it :) This is not your fault.

Wovchena:
The issue is resolved. You need to align the sub-module version on your branch with the `master` branch version.

Aniruddha521:
@Wovchena 
I use the following command but doesn't seems like it is working:
```bash
git pull --rebase origin verify-chatglm-6b
git add resolved-file
git rebase --continue
git push origin verify-chatglm-6b --force
```
can you suggest me anything which I may have done wrong or missed.
Additionally my forked repo is up to date with `openvinotoolkit:master`
please help me in this matter!

Wovchena:
You can checkout `thirparty/openvino_tokenizers` from `master` and push to you branch to align the tokenizers with `master`:
```
git checkout master thirdparty/openvino_tokenizers/
git add thirdparty/openvino_tokenizers/
git commit -m tokenizers
git push
```

Wovchena:
Apparently that was wrong master. Try `git checkout a0268cd5c5fe71ccbc4dc773b502075867c859fe thirdparty/openvino_tokenizers` instead of `git checkout master thirdparty/openvino_tokenizers/`

Aniruddha521:
@Wovchena 
why prompt_lookup and greedy is are generating different output for chatglm3?[here](https://github.com/Aniruddha521/openvino.genai/blob/verify_chatglm3-6b/.github/workflows/causal_lm_cpp.yml)
```python
- name: run and compare
        run: |
          source ./ov/setupvars.sh

          echo 'Code:```python
          def add(a, b):
              return a + b
          ```
          Question: Can you please add 2 and 3
          A:' > ./prompt.txt

          ./build/samples/cpp/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm ./chatglm3-6b/ "$(<prompt.txt)" > predictions_prompt_lookup.txt
          ./build/samples/cpp/greedy_causal_lm/greedy_causal_lm ./chatglm3-6b/ "$(<prompt.txt)" > predictions_greedy.txt
          diff predictions_prompt_lookup.txt predictions_greedy.txt
          python -c "
          with open('predictions_greedy.txt', 'r') as f:
              predicted_greedy = f.readline()
          with open('predictions_prompt_lookup.txt', 'r') as f:
              predicted_prompt_lookup = f.readline()
          print(predicted_greedy)
          print(predicted_prompt_lookup)
          "
          echo "Prompt lookup" passed
```
because as per the error raised I think it is due to the different output generated by greedy and prompt lookup.
But it is probably generating same results for Qwen model since it is passing the tests.

Wovchena:
I don't know. That is the point of this task to test if chatglm is fine. Apparently it's not. You can print the results to see them

Aniruddha521:
@Wovchena 
As I can see  `cpp-greedy_causal_lm-Chatglm3-6b` is failing because the output generated by python sample and c++ sample are different and whereas `cpp-prompt_lookup_decoding_lm-ubuntu` is failing because `prompt lookup` and `greedy` generated text are different.
In both the cases I am using `diff` command to compare the generated text which may be causing `assertion error` , so will I only need to remove comparison part or remove whole  `cpp-greedy_causal_lm-Chatglm3-6b`  and cpp-prompt_lookup_decoding_lm-ubuntu` .
Also don't know why 
1) `Linux (Ubuntu 20.04, Python 3.9) / OpenVINO genai extension (cmake + wheel) (pull_request) Failing after 27m `
2) `Windows (VS 2019, Python 3.11) / OpenVINO genai extension (cmake + wheel) (pull_request)`
3) `macOS (12, Python 3.9) / OpenVINO genai extension (cmake + wheel) (pull_request)`
 the above tests are failing?
I also proceed as you mentioned earlier
```bash
git checkout a0268cd5c5fe71ccbc4dc773b502075867c859fe thirdparty/openvino_tokenizers
git add thirdparty/openvino_tokenizers/
git commit --amend
git push
```

Wovchena:
You need to find out why C++ and Python greedy produce different outputs and fix it. Other model runs are aligned, that means the problem is not about the samples themselves.

This PR diff shows that you've modified `thirdparty/openvino_tokenizers`. Eliminate that diff. `master`'s version of `openvino_tokenizers` should be kept. May be this is the reason for `Linux (Ubuntu 20.04, Python 3.9) / OpenVINO genai extension (cmake + wheel) (pull_request)` failure.

Aniruddha521:
@Wovchena 
Why I am getting build error and error for `whisper_speech_recognition` in **Windows (VS 2019, Python 3.11)**?
Did I miss anything?

Wovchena:
Something got broken in one of dependencies. All new PRs fail the same way. This is not your fault.

Aniruddha521:
@Wovchena 
I think there in no need for any further edit? this latest commit has resolved all the required checks.
Correct me if I am wrong

Aniruddha521:
@Wovchena 
I think that my latest commit satisfy all the required conditions given in the issue. Since it has passed all the github workflow.
Please correct me if I'm wrong 😅

.github/workflows/causal_lm_cpp.yml
```diff
@@ -274,6 +274,48 @@ jobs:
...
+        run: |
+          . ./ov/setupvars.sh
+          echo "why sun is yellow?" > ./prompt.txt
+          echo '--------------------------------C++ Generated--------------------------------------------------'
+          ./build/samples/cpp/greedy_causal_lm/greedy_causal_lm ./chatglm3-6b/ "$(<prompt.txt)"| tee cpp.txt
+          echo '-------------------------------Python Generated------------------------------------------------'
+          samples/python/greedy_causal_lm/greedy_causal_lm.py ./chatglm3-6b/ "$(<prompt.txt)"| tee py.txt
+          echo '-----------------------------------------------------------------------------------------------'
+          diff cpp.txt py.txt
+          echo "Why sun is yellow?" passed
```
Wovchena, line 315:
```suggestion
```
It's the last command in this step and there are no other tests in this step anyway. So completing it already signals that the test passed

.github/workflows/causal_lm_cpp.yml
```diff
@@ -534,6 +578,31 @@ jobs:
...
+          echo '-----------------------------------------------------------------------------------------------'
+          python -c "
+          with open('predictions_greedy.txt', 'r') as f:
+              predicted_greedy = f.readline()
+          with open('predictions_prompt_lookup.txt', 'r') as f:
+              predicted_prompt_lookup = f.readline()
+          assert predicted_greedy == predicted_prompt_lookup
+          print('Passes')
+          "
+          echo "Prompt lookup" passed
```
Wovchena, line 605:
```suggestion
          diff predictions_greedy.txt predictions_prompt_lookup.txt
```

tests/python_tests/ov_genai_test_utils.py
Wovchena, line 28: Does every test pass with `THUDM/chatglm3-6b`? If not, please, mark the failing tests to be skipped. Skips must happen only for that particular model.

thirdparty/openvino_tokenizers
Wovchena, line 1: It's not very convenient in GitHub interface to verify if you upgrade or downgrade the submodule. Please revert this change. You are not supposed to change it anyway.

src/docs/SUPPORTED_MODELS.md
```diff
@@ -155,7 +155,9 @@ The pipeline can work with other similar topologies produced by `optimum-intel`
 4. `position_ids` (optional) encodes a position of currently generating token in the sequence and a single `logits` output.
 
 > [!NOTE]
-> Models should belong to the same family and have the same tokenizers.
+>* Models should belong to the same family and have the same tokenizers.
+>* `optimum-cli` requires the `--task text-generation-with-past` argument for the `THUDM/chatglm3-6b` model
+>* The **beam search** is not supported for `ChatGLM3-6B model`.
```
Wovchena, line 160:
```suggestion
>* The **beam search** is not supported for `THUDM/chatglm3-6b` model.
```

Wovchena (changes requested):

.github/workflows/causal_lm_cpp.yml
```diff
@@ -274,6 +274,48 @@ jobs:
           && call .\ov\setupvars.bat
           && python samples\python\greedy_causal_lm\lora.py .\TinyLlama\TinyLlama-1.1B-intermediate-step-1431k-3T\ adapter_model.safetensors "How to create a table with two columns, one of them has type float, another one has type int?"
 
+  cpp-greedy_causal_lm-Chatglm3-6b:
+    runs-on: ubuntu-20.04-16-cores
```
Wovchena, line 278:
```suggestion
    runs-on: ubuntu-20.04-4-cores
```

Maybe this or `ubuntu-20.04-8-cores` would be enough

.github/workflows/causal_lm_cpp.yml
Aniruddha521: I too tried with the same approach earlier but it end up giving error because `diff predictions_greedy.txt predictions_prompt_lookup.txt` is comparing the whole contain of the file which are not same as you can see
```bash
-------------------------------Prompt lookup Generated-----------------------------------------
Yes, I can add 2 and 3
B: The sum of 2 and 3 is 5

Answer: B

Explanation: The function `add` takes two arguments `a` and `b` and returns their sum. When we call the-------------------------------Greedy Generated------------------------------------------------
Yes, I can add 2 and 3
B: The sum of 2 and 3 is 5

Answer: B

Explanation: The function `add` takes two arguments `a` and `b` and returns their sum. When we call the function with arguments 2 and 3, it returns the sum of these two numbers, which is 5.
-----------------------------------------------------------------------------------------------
```
if I use `diff predictions_greedy.txt predictions_prompt_lookup.txt` command then the error will look like
```bash
6c6
< Explanation: The function `add` takes two arguments `a` and `b` and returns their sum. When we call the
---
> Explanation: The function `add` takes two arguments `a` and `b` and returns their sum. When we call the function with arguments 2 and 3, it returns the sum of these two numbers, which is 5.
```
Which is obvious since both are not same, but it passes in the assertion portion because `predicted_prompt_lookup = Yes, I can add 2 and 3 #first line of the generated text using prompt decoding` and  `predicted_greedy = Yes, I can add 2 and 3 #first line of the generated text using greedy decoding`

Any suggestion from your side?

tests/python_tests/ov_genai_test_utils.py
Aniruddha521: For for python tests(including beam search) the output is showing `Killed` may be due to resource constraints.

image
image

thirdparty/openvino_tokenizers
Aniruddha521: current commit id :- `3fe6677f27c7f5eb149b959d5cd80dff76e9924b` 
current commit message:- Update Regex For Clean Tokenization Spaces (#314)

* Update Regex For Clean Tokenization Spaces

* Fix Replace Pattern Rewrite

* Update Pass Rate

* Del Unused Regex

can you provide me the commit id or commit message I am supposed to revert to

src/docs/SUPPORTED_MODELS.md
Aniruddha521: Done

.github/workflows/causal_lm_cpp.yml
Wovchena: You have to come up with an explanation why the generated text is different and probably fix it.
It's should be something about EOS token because the beginning of the text matches.

tests/python_tests/ov_genai_test_utils.py
Wovchena: I refer to python tests, not python samples

.github/workflows/causal_lm_cpp.yml
Wovchena: @iefode said that she encountered similar issue. So maybe it will be fixed

thirdparty/openvino_tokenizers
Wovchena: You can check the correct commit on master branch. It's 3fe6677f27c7f5eb149b959d5cd80dff76e9924b
```
cd ./thirdparty/openvino_tokenizers/
git checkout 3fe6677f27c7f5eb149b959d5cd80dff76e9924b
cd ./../..
git add ./thirdparty/openvino_tokenizers/
git commit -m tokenizers
git push
```

Wovchena: I mean 306dcd8daec36bbc680c50c68de1e954f42b0ab8 :)

Aniruddha521: Done

.github/workflows/causal_lm_cpp.yml
Wovchena: Prompt lookup was updated in https://github.com/openvinotoolkit/openvino.genai/pull/1245. Can you try the new version

PR #1180, opened 2024-11-08
Fixed passing of generation config params to VLM generate.

popovaan:
- Fixed passing of generation config params to VLM generate().
- Updated generation config params params list in `update_config_from_kwargs()` method.

Ticket: CVS-157050


src/python/py_utils.cpp
```diff
@@ -283,4 +301,65 @@ ov::genai::OptionalGenerationConfig update_config_from_kwargs(const ov::genai::O
     return res_config;
 }
 
+    
+bool generation_config_param_to_property(std::string key, py::object value, ov::AnyMap& map) {
```
ilya-lavrenov, line 305: maybe we can reuse this function in `update_config_from_kwargs` ? E.g. convert params to AnyMap, then call GenerationConfig::update_generation_config

ilya-lavrenov:
Let' submit fix to 2024.5

src/python/py_utils.cpp
ilya-lavrenov: I'm ok to merge as is for releases/2024/5 and improve on master branch

src/python/py_vlm_pipeline.cpp
```diff
@@ -88,8 +88,9 @@ ov::AnyMap vlm_kwargs_to_any_map(const py::kwargs& kwargs, bool allow_compile_pr
         } else if (key == "streamer") {
```
ilya-lavrenov, line 88: as far as I can see all these params `images`, `generation_config`, `streamer` etc can now be passed to `VLMPipeline` ctor, which is the right thing. Those options are `generate` properties.

src/python/py_utils.cpp
popovaan: > can we convert kwargs to AnyMap and then call GenerationConfig::update_from_map ?

Yes, did it inside `update_config_from_kwargs()` method.

src/python/py_vlm_pipeline.cpp
ilya-lavrenov: > I suppose we also need to fix it for text2image?

yes

src/python/py_utils.cpp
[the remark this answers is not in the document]
ilya-lavrenov: we can merge fix for the original problem to 2024.5 , while here we plan to perform more refactoring / extra fixes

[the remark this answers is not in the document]
Wovchena: okay

src/python/py_utils.cpp
```diff
@@ -156,8 +259,17 @@ ov::Any py_object_to_any(const py::object& py_obj) {
         return py::cast<ov::Output<ov::Node>>(py_obj);
     } else if (py::isinstance<ov::genai::SchedulerConfig>(py_obj)) {
         return py::cast<ov::genai::SchedulerConfig>(py_obj);
-    } else if (py::isinstance<ov::genai::AdapterConfig>(py_obj)) { 
+    } else if (py::isinstance<ov::genai::AdapterConfig>(py_obj)) {
         return py::cast<ov::genai::AdapterConfig>(py_obj);
+    } else if (py::isinstance<ov::genai::GenerationConfig>(py_obj)) {
+        return py::cast<ov::genai::GenerationConfig>(py_obj);
```
ilya-lavrenov, line 265: should we also add Whisper and Image generation configs here?
users can write `generate("prompt", generation_config=image_generation_config)`

src/python/py_utils.cpp
```diff
@@ -117,10 +141,89 @@ ov::Any py_object_to_any(const py::object& py_obj) {
...
+            } else if (py::isinstance<py::int_>(it)) {
+                check_type(PY_TYPE::INT);
+            } else if (py::isinstance<py::float_>(it)) {
+                check_type(PY_TYPE::FLOAT);
+            } else if (py::isinstance<py::bool_>(it)) {
+                check_type(PY_TYPE::BOOL);
+            }
+        }
+
+        if (_set.empty())
```
ilya-lavrenov, line 211: do we have real cases for it? because returning empty ov::Any typically leads to exceptions in calling code, because such `any` cannot be casted to actual types

Maybe we need to have a set of properties, which can empty sets and hardcode their types here as well? The same question for empty `dict`

src/python/py_utils.cpp
```diff
@@ -117,10 +141,89 @@ ov::Any py_object_to_any(const py::object& py_obj) {
...
+        switch (detected_key_type) {
+        case PY_TYPE::STR:
+            switch (detected_value_type) {
+            case PY_TYPE::INT:
+                return _dict.cast<std::map<std::string, int64_t>>();
+            default:
+                OPENVINO_ASSERT(false, "Unsupported attribute type.");
+            }
+        default:
+            OPENVINO_ASSERT(false, "Unsupported attribute type.");
```
ilya-lavrenov, line 186:
```suggestion
            OPENVINO_THROW("Unsupported attribute type.");
```
is it possible to provide more detailed error message? E.g. property name at least

src/python/py_utils.cpp
```diff
@@ -117,10 +141,89 @@ ov::Any py_object_to_any(const py::object& py_obj) {
...
+                OPENVINO_THROW("Incorrect attribute. Mixed types in the dict are not allowed.");
+            };
+            // check key type
+            if (py::isinstance<py::str>(it.first)) {
+                check_type(PY_TYPE::STR, detected_key_type);
+            }
+
+            // check value type
+            if (py::isinstance<py::int_>(it.second)) {
+                check_type(PY_TYPE::INT, detected_value_type);
```
ilya-lavrenov, line 170: do we have issues with int / size_t / int64_t here as well? the same for `list` type

src/python/py_utils.cpp
```diff
@@ -11,6 +11,8 @@
 
 #include "tokenizers_path.hpp"
 #include "openvino/genai/llm_pipeline.hpp"
+#include "openvino/genai/visual_language/pipeline.hpp"
+#include "openvino/genai/image_generation/text2image_pipeline.hpp"
```
ilya-lavrenov, line 15:
```suggestion
#include "openvino/genai/image_generation/image_generation_config.hpp"
```

src/python/py_utils.cpp
popovaan: Yes, added.

src/python/py_utils.cpp
popovaan: I only found these two:
https://github.com/openvinotoolkit/openvino.genai/blob/6bb564474f4b4582f82c811ab0947a82577af5ff/tests/python_tests/test_llm_pipeline_static.py#L19

Hardcoded them and added raising of exceptions by default.

src/python/py_utils.cpp
popovaan: Yes, added property names.

src/python/py_utils.cpp
```diff
@@ -179,11 +305,16 @@ ov::AnyMap kwargs_to_any_map(const py::kwargs& kwargs) {
...
-        if (utils::py_object_is_any_map(value)) {
+        // we need to unpack only dictionaries, which are passed with "config" name,
+        // because there are dictionary properties that should not be unpacked
+        if (utils::py_object_is_any_map(value) && key == "config") {
             auto map = utils::py_object_to_any_map(value);
             params.insert(map.begin(), map.end());
         } else {
-            params[key] = utils::py_object_to_any(value);
+            if (py::isinstance<py::none>(value)) {
+                continue;
```
Wovchena, line 315: `generate("", eos_token_id=None)` probably means it should be set to `-1` instead of being ignored here. Maybe we should just report an error for `None`s?

Wovchena (changes requested):

tests/python_tests/test_generate_api.py
```diff
@@ -635,7 +635,8 @@ def test_valid_configs(model_tmp_path):
 
 invalid_py_configs = [
     dict(num_beam_groups=3, num_beams=15, do_sample=True),
-    dict(unexisting_key_name=True),  # no eos_token_id no max_new_tokens, no max_len
+    # TODO: Currently unexpected params do not cause exceptions. Need to implement it in c++ and return this test
+  #  dict(unexisting_key_name=True),  # no eos_token_id no max_new_tokens, no max_len
```
ilya-lavrenov, line 639: do you plan to fix it in next PRs?

src/python/py_utils.cpp
ilya-lavrenov: BTW, `py_object_to_any` can handle `py::none`. So, maybe we need to handle `None`s only there?

I'm Ok to throw exception, but alternatively we can allow `None` for properties with default meaningful ctors like `negative_prompt_3` (std::string) or `Adapters`, e.g.:
```python
    image = pipe.generate(prompt, adapters=None)
```
instead of:
```python
    image = pipe.generate(prompt, adapters=openvino_genai.AdapterConfig())
```
@slyalin do you have any preferences?

Wovchena: `None`s as default values arise when there's a need to handle unspecified value alongside with the predefined default value in function args. Genai is in different situation. If something is unspecified we simply don't use that arg at all. So the example `image = pipe.generate(prompt, adapters=None)` should just be `image = pipe.generate(prompt)` in my understanding. Would that work?

ilya-lavrenov: `image = pipe.generate(prompt, adapters=None)` implies "don't use adapters, which were set in Pipeline ctor". E.g. pipeline can be created with multiple LoRAs which are configurable in runtime (`generate(..)`) and `None` means don't it LoRA at all.

It's rather a syntax sugar, so probably throwing an exception is a simplier solution.

slyalin: Looks like throwing an exception in this case is a better solution because:
1. I agree that option=None is the closest to not mentioning option at all.
2. Having adapters=None to be an alternative to adapters=AdapterConfig() contradicts the previous point, and making options "do not change the adapters" and "set empty adapters" nearly undistinguishable because in the first case, we shouldn't mention adapters at all, and in the second case, we can pass adapters=None, that is hardly decodable from the user perspective.

src/python/py_utils.cpp
```diff
@@ -54,41 +57,67 @@ ov::AnyMap py_object_to_any_map(const py::object& py_obj) {
...
         }
     }
     return return_value;
 }
 
-ov::Any py_object_to_any(const py::object& py_obj) {
+ov::Any py_object_to_any(const py::object& py_obj, std::string property_name) {
     // Python types
-    py::object float_32_type = py::module_::import("numpy").attr("float32");
+    // TODO: Remove this after ov::Any is fixed to allow pass types, that can be casted to target type.
```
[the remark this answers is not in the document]
popovaan: Done.

src/python/py_utils.cpp
popovaan: Added throwing of exception.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -313,6 +313,9 @@ template <typename T>
 T pop_or_default(ov::AnyMap& config, const std::string& key, const T& default_value) {
     auto anyopt = pop_option(config, key);
     if (anyopt.has_value()) {
+        if (anyopt.value().empty()) {
+            return T{};
```
Wovchena, line 317: Why not `default_value`?
Can you explain, why it is needed now

I guess that's because you switched from exceptions to `return ov::Any();`, but why?

popovaan: Almost any property of type `list`, `set`, or `dict` can potentially be empty and we should be able to accept such values. So Ilya suggested to return `ov::Any()` in bindings and process empty `ov::Any()` in utils. 

Regarding usage of `default_value`, I tried it locally but tests, that use confing shown below failed, because it is not equivalent to creation of empty container. So I changed to `T{}`.
https://github.com/openvinotoolkit/openvino.genai/blob/6bb564474f4b4582f82c811ab0947a82577af5ff/tests/python_tests/test_llm_pipeline_static.py#L19-L20

src/python/py_utils.cpp
```diff
@@ -179,11 +296,16 @@ ov::AnyMap kwargs_to_any_map(const py::kwargs& kwargs) {
...
-        if (utils::py_object_is_any_map(value)) {
+        // we need to unpack only dictionaries, which are passed with "config" name,
+        // because there are dictionary properties that should not be unpacked
+        if (utils::py_object_is_any_map(value) && key == "config") {
             auto map = utils::py_object_to_any_map(value);
             params.insert(map.begin(), map.end());
         } else {
-            params[key] = utils::py_object_to_any(value);
+            if (py::isinstance<py::none>(value)) {
+                OPENVINO_THROW("Property \"" + key + "\" can't be None.");
```
ilya-lavrenov, line 306:
```suggestion
                OPENVINO_ASSERT(!py::isinstance<py::none>(value), "Property \"", key, "\" can't be None.");
```

src/cpp/src/llm_pipeline_static.cpp
ilya-lavrenov: > but why?

yes, we need somehow bypass information from python that container (list, map, set) is empty, but we don't know type of this container (e.g. vector<of what type?>), while in this function we know this type `T`

maybe it will safer to `return T{};` only for containers? because for regular type empty `ov::Any` is still not expected

src/python/py_tokenizer.cpp
```diff
@@ -30,10 +30,10 @@ void init_tokenizer(py::module_& m) {
         R"(openvino_genai.Tokenizer object is used to initialize Tokenizer
            if it's located in a different path than the main model.)")
 
-        .def(py::init([](const std::filesystem::path& tokenizer_path, const std::map<std::string, py::object>& properties) {
+        .def(py::init([](const std::filesystem::path& tokenizer_path, const py::kwargs& kwargs) {
```
ilya-lavrenov, line 33: maybe we need to keep `map` for BW compatibility and add warning similar to `LLMPipeline`?

https://github.com/openvinotoolkit/openvino.genai/blob/bef96dd4103adad9c9c0e33c181bcf16a9fb234b/src/python/py_llm_pipeline.cpp#L110-L116

src/cpp/src/llm_pipeline_static.cpp
popovaan: Done.

src/python/py_tokenizer.cpp
popovaan: Done.

ilya-lavrenov (approved):

PR #1197, opened 2024-11-11
Flux pipeline

likholat:
Tested for `FLUX.1-schnell/INT4`
Prompt `"A cat holding a sign that says hello world"`
Pipeline params:
```python
pipe.generate(
        args.prompt,
        width=512,
        height=512,
        guidance_scale = 0.0,
        num_inference_steps = 5,
        num_images_per_prompt = 1,
        generator=Generator(10)
    )
```
For `seed = 10`
image

For `seed = 100`
image


src/cpp/src/image_generation/diffusion_pipeline.hpp
```diff
@@ -25,6 +25,16 @@ void batch_copy(ov::Tensor src, ov::Tensor dst, size_t src_batch, size_t dst_bat
     ov::Tensor(src, src_start, src_end).copy_to(ov::Tensor(dst, dst_start, dst_end));
 }
 
+ov::Tensor tensor_batch_copy(const ov::Tensor input, const size_t num_images_per_prompt, size_t batch_size_multiplier) {
```
ilya-lavrenov, line 28: `batch_size_multiplier` is not used

maybe we can call this function `repeat` (like it's used in diffusers) and move to `numpy_utils` ?

src/cpp/src/image_generation/models/flux_transformer_2d_model.cpp
```diff
@@ -0,0 +1,124 @@
...
+    read_json_param(data, "in_channels", in_channels);
+
+    file.close();
+
+    // block_out_channels should be read from VAE encoder / decoder config to compute proper m_vae_scale_factor
+    std::filesystem::path vae_config_path = config_path.parent_path().parent_path() / "vae_decoder" / "config.json";
+    file.open(vae_config_path);
+    OPENVINO_ASSERT(file.is_open(), "Failed to open ", vae_config_path);
+    data = nlohmann::json::parse(file);
+    read_json_param(data, "block_out_channels", block_out_channels);
```
ilya-lavrenov, line 32: looks like `block_out_channels` is not used at all, because `m_vae_scale_factor` is computed via `get_vae_scale_factor`

src/cpp/src/image_generation/models/flux_transformer_2d_model.cpp
```diff
@@ -0,0 +1,124 @@
...
+                                                      int tokenizer_model_max_length) {
+    OPENVINO_ASSERT(m_model, "Model has been already compiled. Cannot reshape already compiled model");
+
+    // hidden_states=latent_model_input,
+    // timestep=timestep,
+    // encoder_hidden_states=prompt_embeds,
+    // pooled_projections=pooled_prompt_embeds,
+
+    std::cout << "batch_size " << batch_size << std::endl;
+    std::cout << "tokenizer_model_max_length " << tokenizer_model_max_length << std::endl;
```
ilya-lavrenov, line 66: let's drop debug prints

src/cpp/src/image_generation/models/t5_encoder_model.cpp
```diff
@@ -0,0 +1,98 @@
...
+const T5EncoderModel::Config& T5EncoderModel::get_config() const {
+    return m_config;
+}
+
+T5EncoderModel& T5EncoderModel::reshape(int batch_size) {
+    OPENVINO_ASSERT(m_model, "Model has been already compiled. Cannot reshape already compiled model");
+
+    ov::PartialShape input_shape = m_model->input(0).get_partial_shape();
+    input_shape[0] = batch_size;
+    input_shape[1] = m_config.max_sequence_length;
```
ilya-lavrenov, line 47: does it mean we need to pad tokenizer's output to `max_sequence_length` ?

src/cpp/src/image_generation/models/t5_encoder_model.cpp
```diff
@@ -0,0 +1,98 @@
...
+    m_model->reshape(idx_to_shape);
+
+    return *this;
+}
+
+T5EncoderModel& T5EncoderModel::compile(const std::string& device, const ov::AnyMap& properties) {
+    OPENVINO_ASSERT(m_model, "Model has been already compiled. Cannot re-compile already compiled model");
+    ov::Core core = utils::singleton_core();
+    ov::CompiledModel compiled_model;
+    std::optional<AdapterConfig> adapters;
```
ilya-lavrenov, line 58:
```suggestion
```

src/cpp/src/image_generation/models/t5_encoder_model.cpp
```diff
@@ -0,0 +1,98 @@
...
+    // release the original model
+    m_model.reset();
+
+    return *this;
+}
+
+ov::Tensor T5EncoderModel::infer(const std::string& pos_prompt) {
+    OPENVINO_ASSERT(m_request, "T5 encoder model must be compiled first. Cannot infer non-compiled model");
+
+    const int32_t pad_token_id = m_clip_tokenizer.get_pad_token_id();
```
ilya-lavrenov, line 70:
```suggestion
    const int32_t pad_token_id = m_tokenizer.get_pad_token_id();
```
as it's not Clip tokenizer https://huggingface.co/black-forest-labs/FLUX.1-dev/blob/main/tokenizer_2/tokenizer_config.json#L938

src/cpp/src/image_generation/models/t5_encoder_model.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    auto perform_tokenization = [&](const std::string& prompt, ov::Tensor input_ids) {
+        std::fill_n(input_ids.data<int32_t>(), input_ids.get_size(), pad_token_id);
```
ilya-lavrenov, line 73: I think we need to pad only in case of static shapes (if user called `reshape`)

Or even avoid reshaping to `max_sequence_len` of tokenizer. To decide about it, we probably need to compare performance w/o and w/ reshaping to upper bound.

src/cpp/src/image_generation/schedulers/euler_discrete.cpp
```diff
@@ -319,5 +319,13 @@ void EulerDiscreteScheduler::add_noise(ov::Tensor init_latent, std::shared_ptr<G
     }
 }
 
+void EulerDiscreteScheduler::set_timesteps_with_sigma(std::vector<float> sigma, float mu) {
+    OPENVINO_THROW("EulerDiscreteScheduler doesn't support set_timesteps_with_sigma");
+}
+
+float EulerDiscreteScheduler::calculate_shift(size_t image_seq_len) {
+    OPENVINO_THROW("EulerDiscreteScheduler doesn't support `calculate_shift` method");
```
ilya-lavrenov, line 327: maybe we can keep such methods which are supported only by selected schedulers, to be implemented in base class with default implementations throwing an exception.

src/cpp/src/image_generation/schedulers/flow_match_euler_discrete.cpp
```diff
@@ -150,5 +150,43 @@ void FlowMatchEulerDiscreteScheduler::add_noise(ov::Tensor init_latent, std::sha
...
+
+    // fill timesteps
+    for (size_t i = 0; i < m_sigmas.size(); ++i) {
+        m_timesteps.push_back(m_sigmas[i] * m_config.num_train_timesteps);
+    }
+    m_sigmas.push_back(0);
+    m_step_index = -1, m_begin_index = -1;
+}
+
+float  FlowMatchEulerDiscreteScheduler::calculate_shift(size_t image_seq_len) {
```
ilya-lavrenov, line 179:
```suggestion
float FlowMatchEulerDiscreteScheduler::calculate_shift(size_t image_seq_len) {
```

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -0,0 +1,382 @@
...
+    FluxPipeline(PipelineType pipeline_type,
+                 const CLIPTextModel& clip_text_model,
+                 const T5EncoderModel& t5_text_model,
+                 const FluxTransformer2DModel& transformer,
+                 const AutoencoderKL& vae)
+        : DiffusionPipeline(pipeline_type),
+          m_clip_text_encoder(std::make_shared<CLIPTextModel>(clip_text_model)),
+          m_t5_text_encoder(std::make_shared<T5EncoderModel>(t5_text_model)),
+          m_vae(std::make_shared<AutoencoderKL>(vae)),
+          m_transformer(std::make_shared<FluxTransformer2DModel>(transformer)) {}
```
ilya-lavrenov, line 210: let's initialize generation config here similarly to https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/image_generation/stable_diffusion_3_pipeline.hpp#L205

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -0,0 +1,382 @@
...
+
+private:
+    void initialize_generation_config(const std::string& class_name) override {
+        assert(m_transformer != nullptr);
+        assert(m_vae != nullptr);
+
+        const auto& transformer_config = m_transformer->get_config();
+        const size_t vae_scale_factor = m_vae->get_vae_scale_factor();
+
+        m_default_sample_size = 128;
```
ilya-lavrenov, line 336: maybe it should be a part of FluxTransformer2D::Config ?

src/cpp/include/openvino/genai/image_generation/flux_transformer_2d_model.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+namespace ov {
+namespace genai {
+
+class OPENVINO_GENAI_EXPORTS FluxTransformer2DModel {
+public:
+    struct Config {
+        size_t in_channels = 64;
+
+        std::vector<size_t> block_out_channels = { 128, 256, 512, 512 };
```
ilya-lavrenov, line 24: looks like later we need to add `guidance_embeds`, as other Flux models rely on this parameter https://huggingface.co/Freepik/flux.1-lite-8B-alpha/blob/main/transformer/config.json#L11 and embed guidance scale

likholat: Yes, but currently I don't have `guidance` input for IR

likholat:
@ilya-lavrenov please take a look

likholat:
For `max_sequence_length=11`
image

ilya-lavrenov (approved):

PR #1238, opened 2024-11-20
[WWB]: Added ability to compare results for previously collected outputs w/o models provided

AlexKoff88:


AlexKoff88:
@sstrehlk

eaidova:
@AlexKoff88  for me changes looks good. Small question, should this example in readme also be updated to use csv instead json?
https://github.com/openvinotoolkit/openvino.genai/tree/master/tools/who_what_benchmark#compare-text-to-image-models-diffusers

AlexKoff88:
> @AlexKoff88 for me changes looks good. Small question, should this example in readme also be updated to use csv instead json? https://github.com/openvinotoolkit/openvino.genai/tree/master/tools/who_what_benchmark#compare-text-to-image-models-diffusers

yes, I will revise README significantly in the follow-up PR.

PR #1240, opened 2024-11-20
Static llm pipeline: stateful model

AsyaPronina:
Related PRs:
- OpenVINO: *https://github.com/openvinotoolkit/openvino/pull/27651*
- OpenVINO Unroll SPDA: *https://github.com/openvinotoolkit/openvino/pull/27891*
- OpenVINO Prefill/Generate sections: *https://github.com/openvinotoolkit/openvino/pull/28154*
- OpenVINO LLMCompiledModel refactoring: https://github.com/openvinotoolkit/openvino/pull/28267

### DIFF
--- a/src/cpp/src/llm_pipeline.cpp
+++ b/src/cpp/src/llm_pipeline.cpp
@@ -20,13 +20,13 @@ namespace {
 
 /* 
 * NPU reads some properties from the config file, but when LLMPipeline is initialized
-* from the model_str and weights_tensor, there are not files. 
+* from the model_str and weights_tensor, there are no files.
 * In the later case ModelDesc is stored in properties.
 * This function pops ModelDescr from the the properties and returns a pair of updated properties and ModelDescr.
 */
-std::pair<ov::AnyMap, ov::genai::ModelConfigDesc> split_model_descr(const ov::AnyMap& properties) {
+std::pair<ov::AnyMap, ov::genai::static_llm::ModelConfigDesc> split_model_descr(const ov::AnyMap& properties) {
     ov::AnyMap main_properties = properties;
-    ov::genai::ModelConfigDesc model_descr;
+    ov::genai::static_llm::ModelConfigDesc model_descr;
 
     auto pop_property = [](ov::AnyMap& orig_propertis, const std::string& key, auto& value) {
         if (orig_propertis.find(key) != orig_propertis.end()) {
@@ -105,7 +105,7 @@ ov::genai::LLMPipeline::LLMPipeline(
         auto [plugin_config, scheduler_config] = utils::split_scheduler_config(properties);
         m_pimpl = std::make_unique<ContinuousBatchingAdapter>(models_path, tokenizer, scheduler_config, device, plugin_config);
     } else if (device == "NPU") {
-        m_pimpl = std::make_unique<StaticLLMPipeline>(models_path, tokenizer, device, properties);
+        m_pimpl = static_llm::LLMPipelineFactory::create(models_path, tokenizer, device, properties);
     } else {
         m_pimpl = std::make_unique<StatefulLLMPipeline>(models_path, tokenizer, device, properties);
     }
@@ -124,7 +124,7 @@ ov::genai::LLMPipeline::LLMPipeline(
         auto [device_properties, scheduler_config] = utils::split_scheduler_config(properties);
         m_pimpl = std::make_unique<ContinuousBatchingAdapter>(models_path, scheduler_config, device, device_properties);
     } else if (device == "NPU") {
-        m_pimpl = std::make_unique<StaticLLMPipeline>(models_path, device, properties);
+        m_pimpl = static_llm::LLMPipelineFactory::create(models_path, device, properties);
     } else {
         m_pimpl = std::make_unique<StatefulLLMPipeline>(models_path, device, properties);
     }
@@ -162,7 +162,7 @@ ov::genai::LLMPipeline::LLMPipeline(
         // This will convert from AnyMap to ModelDesc.
         auto [filtered_properties, model_descr] = split_model_descr(properties);
 
-        m_pimpl = std::make_unique<StaticLLMPipeline>(
+        m_pimpl = static_llm::LLMPipelineFactory::create(
             utils::singleton_core().read_model(model_str, weights_tensor), 
             model_descr,
             tokenizer,
--- a/src/cpp/src/llm_pipeline_static.cpp
+++ b/src/cpp/src/llm_pipeline_static.cpp
@@ -660,19 +660,19 @@ void stream_generated_tokens(std::shared_ptr<ov::genai::StreamerBase> streamer_p
     }
 }
 
-enum PipelineKind {
+enum StaticPipelineKind {
     STATEFUL,
     STATELESS
 };
-PipelineKind str_to_pipeline(const std::string& str) {
+StaticPipelineKind str_to_pipeline(const std::string& str) {
     if (str == "STATEFUL") {
-        return PipelineKind::STATEFUL;
+        return StaticPipelineKind::STATEFUL;
     }
     if (str == "STATELESS") {
-        return PipelineKind::STATELESS;
+        return StaticPipelineKind::STATELESS;
     }
-    OPENVINO_THROW("Unsupported \"PIPELINE\" provided: " +
-                   str + ". Please select either \"STATEFUL\" or \"STATELESS\".");
+    OPENVINO_THROW("Unsupported \"PIPELINE\" provided: ",
+                   str, ". Please select either \"STATEFUL\" or \"STATELESS\".");
 }
 } // anonymous namespace
 
@@ -705,20 +705,9 @@ StatefulLLMPipeline::StatefulLLMPipeline(
     const ov::AnyMap& properties,
     const ov::genai::GenerationConfig& generation_config
 ) : LLMPipelineImplBase(tokenizer, generation_config) {
-
-    bool use_blobs = false;
-    auto anyopt = get_option<bool>(properties, "USE_BLOBS");
-    if (anyopt.has_value()) {
-        use_blobs = *anyopt;
-    }
-    // Using model_str and weights_tensor with blobs is meaningless.
-    if (use_blobs) {
-        OPENVINO_THROW("Blobs cannot be used with model string and weights tensor");
-    }
-
     ov::AnyMap properties_copy = properties;
     auto compiled = setupAndCompileModel(model, model_desc, properties_copy);
-    m_request  = compiled->create_infer_request();
+    m_request = compiled->create_infer_request();
 }
 
 std::shared_ptr<ov::CompiledModel> StatefulLLMPipeline::setupAndCompileModel(
@@ -764,10 +753,7 @@ DecodedResults StatefulLLMPipeline::generate(
     GenerationConfig config = (generation_config.has_value()) ? *generation_config : m_generation_config;
     std::string prompt;
     if (auto input_vector = std::get_if<std::vector<std::string>>(&inputs)) {
-        if (input_vector->size() > 1u) {
-            OPENVINO_THROW("Currently only batch size=1 is supported");
-        }
-        OPENVINO_ASSERT(!input_vector->empty());
+        OPENVINO_ASSERT(input_vector->size() == 1u, "Currently only batch size=1 is supported");
         prompt = std::move(input_vector->front());
     } else {
         OPENVINO_ASSERT(std::holds_alternative<std::string>(inputs));
@@ -827,9 +813,7 @@ EncodedResults StatefulLLMPipeline::generate(
         attention_mask = data->attention_mask;
     }
 
-    if (input_ids.get_shape().at(0) > 1u) {
-        OPENVINO_THROW("Currently only batch size=1 is supported");
-    }
+    OPENVINO_ASSERT(input_ids.get_shape().at(0) == 1u, "Currently only batch size=1 is supported");
 
     GenerationConfig config = (generation_config.has_value()) ? *generation_config : m_generation_config;
     // If eos_token_id was not provided, take value from default m_generation_config
@@ -846,9 +830,7 @@ EncodedResults StatefulLLMPipeline::generate(
         streamer_ptr = std::make_shared<TextCallbackStreamer>(m_tokenizer, *callback);
     }
 
-    if (!config.is_greedy_decoding()) {
-        OPENVINO_THROW("Currently only greedy decoding is supported");
-    }
+    OPENVINO_ASSERT(config.is_greedy_decoding(), "Currently only greedy decoding is supported");
 
     ov::Shape prompts_shape = input_ids.get_shape();
     const size_t batch_size = prompts_shape[0];
@@ -1450,8 +1432,8 @@ LLMPipelineFactory::create(const std::filesystem::path& models_path,
                            const std::string& device,
                            const ov::AnyMap& config) {
     auto properties = config;
-    const auto pipeline_mode = str_to_pipeline(pop_or_default(properties, "PIPELINE", std::string("STATELESS")));
-    if (pipeline_mode == PipelineKind::STATEFUL) {
+    const auto pipeline_mode = str_to_pipeline(pop_or_default(properties, "STATIC_PIPELINE", std::string("STATELESS")));
+    if (pipeline_mode == StaticPipelineKind::STATEFUL) {
         return std::make_unique<ov::genai::static_llm::StatefulLLMPipeline>(models_path, tokenizer, device, properties);
     }
     return std::make_unique<ov::genai::static_llm::StatelessLLMPipeline>(models_path, tokenizer, device, properties);
@@ -1471,8 +1453,8 @@ std::unique_ptr<LLMPipelineImplBase> LLMPipelineFactory::create(const std::share
                                                                 const ov::AnyMap& properties,
                                                                 const ov::genai::GenerationConfig& generation_config) {
     auto properties_copy = properties;
-    const auto pipeline_mode = str_to_pipeline(pop_or_default(properties_copy, "PIPELINE", std::string("STATELESS")));
-    if (pipeline_mode == PipelineKind::STATEFUL) {
+    const auto pipeline_mode = str_to_pipeline(pop_or_default(properties_copy, "STATIC_PIPELINE", std::string("STATELESS")));
+    if (pipeline_mode == StaticPipelineKind::STATEFUL) {
         return std::make_unique<ov::genai::static_llm::StatefulLLMPipeline>(model,
                                                                             model_desc,
                                                                             tokenizer,
--- a/src/cpp/src/llm_pipeline_static.cpp
+++ b/src/cpp/src/llm_pipeline_static.cpp
@@ -595,9 +595,7 @@ void update_config(ov::AnyMap& config, const std::pair<std::string, ov::Any>& pa
 }
 
 void rename_key(ov::AnyMap& config, const std::string& old_key, const std::string& new_key) {
-    if (config.count(old_key) == 0) {
-        return;
-    } else {
+    if (config.count(old_key) != 0) {
         auto opt_value = pop_option(config, old_key);
         config[new_key] = opt_value.value();
     }
@@ -695,7 +693,7 @@ StatefulLLMPipeline::StatefulLLMPipeline(
     ov::AnyMap properties = config;
 
     auto compiled = setupAndCompileModel(model, model_desc, properties);
-    m_request  = compiled->create_infer_request();
+    m_request = compiled->create_infer_request();
 }
 
 
@@ -883,8 +881,8 @@ EncodedResults StatefulLLMPipeline::generate(
     int64_t input_ids_data = -1;
     int64_t position_ids_data = prompt_len - 1;
     std::vector<int64_t> attention_mask_data(prompt_len - 1, 1);
-    m_request.set_tensor("input_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, (void*)&input_ids_data));
-    m_request.set_tensor("position_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, (void*)&position_ids_data));
+    m_request.set_tensor("input_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1},  reinterpret_cast<void*>(&input_ids_data)));
+    m_request.set_tensor("position_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, reinterpret_cast<void*>(&position_ids_data)));
 
     const size_t max_tokens = config.get_max_new_tokens(prompt_len);
     for (int i = 0; i < max_tokens - 1; ++i) {
@@ -893,14 +891,11 @@ EncodedResults StatefulLLMPipeline::generate(
             break;
         }
 
+        // Just change the variables here, as pointers to them are already set to corresponding tensors
         input_ids_data = last_token;
         ++position_ids_data;
+        // However, attention_mask changes its shape on each iteration, it should be re-set explicitly
         attention_mask_data.push_back(1);
-
-        auto* input_ids_tnsr_data = m_request.get_tensor("input_ids").data<int64_t>();
-        input_ids_tnsr_data[0] = input_ids_data;
-        auto* position_ids_tnsr_data = m_request.get_tensor("position_ids").data<int64_t>();
-        position_ids_tnsr_data[0] = position_ids_data;
         m_request.set_tensor("attention_mask", ov::Tensor(ov::element::i64, ov::Shape{1,attention_mask_data.size()}, (void*)&attention_mask_data[0]));
 
         m_request.infer();
@@ -1455,8 +1450,13 @@ LLMPipelineFactory::create(const std::filesystem::path& models_path,
                            const std::string& device,
                            const ov::AnyMap& config) {
     auto properties = config;
+<<<<<<< HEAD
     const auto pipeline_mode = str_to_pipeline(pop_or_default(properties, "PIPELINE", std::string("STATELESS")));
     if (pipeline_mode == PipelineKind::STATEFUL) {
+=======
+    const auto pipeline_mode = str_to_pipeline(pop_or_default(properties, "NPU_PIPELINE", std::string("STATELESS")));
+    if (pipeline_mode == StaticPipelineType::STATEFUL) {
+>>>>>>> ac5d32b4 (Fixed review comments)
         return std::make_unique<ov::genai::static_llm::StatefulLLMPipeline>(models_path, tokenizer, device, properties);
     }
     return std::make_unique<ov::genai::static_llm::StatelessLLMPipeline>(models_path, tokenizer, device, properties);
--- a/src/cpp/src/llm_pipeline_static.cpp
+++ b/src/cpp/src/llm_pipeline_static.cpp
@@ -398,12 +398,12 @@ KVAxesPosition get_kv_axes(const std::string& model_type) {
     return axes;
 }
 
-ov::genai::ModelConfigDesc get_modeldesc_from_json(const std::filesystem::path& filepath) {
+ov::genai::static_llm::ModelConfigDesc get_modeldesc_from_json(const std::filesystem::path& filepath) {
     std::ifstream file(filepath);
     OPENVINO_ASSERT(file.is_open(), "Could not open file: ", filepath);
     nlohmann::json config_data = nlohmann::json::parse(file);
 
-    ov::genai::ModelConfigDesc desc;
+    ov::genai::static_llm::ModelConfigDesc desc;
     desc.type = config_data["model_type"].get<std::string>();
     // NB: In case _name_or_path field isn't presented in config.json
     if (config_data.contains("_name_or_path")) {
@@ -412,6 +412,16 @@ ov::genai::ModelConfigDesc get_modeldesc_from_json(const std::filesystem::path& 
     desc.num_key_value_heads = config_data.contains("num_key_value_heads")
         ? config_data["num_key_value_heads"].get<int>() : -1;
     return desc;
+}
+
+std::string model_desc_to_string(const ov::genai::static_llm::ModelConfigDesc& model_desc) {
+    std::map<std::string, std::string> model_desc_map;
+    model_desc_map["type"] = model_desc.type;
+    model_desc_map["name_or_path"] = model_desc.name_or_path;
+    model_desc_map["num_key_value_heads"] = std::to_string(model_desc.num_key_value_heads);
+    std::stringstream result;
+    ov::util::Write<std::map<std::string,std::string>>()(result, model_desc_map);
+    return result.str();
 }
 
 void reshape_to_static(std::shared_ptr<ov::Model> model,
@@ -588,6 +598,21 @@ std::optional<uint32_t> pop_int_and_cast(ov::AnyMap& config, const std::string& 
     return std::nullopt;
 }
 
+void update_config(ov::AnyMap& config, const std::pair<std::string, ov::Any>& pair) {
+    if (config.count(pair.first) == 0) {
+        config.insert(pair);
+    }
+}
+
+void rename_key(ov::AnyMap& config, const std::string& old_key, const std::string& new_key) {
+    if (config.count(old_key) == 0) {
+        return;
+    } else {
+        auto opt_value = pop_option(config, old_key);
+        config[new_key] = opt_value.value();
+    }
+}
+
 ov::Tensor make_tensor_slice(ov::Tensor tensor, size_t dim, size_t start_pos, size_t end_pos) {
     ov::Shape start_shape(std::vector<size_t>(tensor.get_shape().size(), 0u));
     start_shape[dim] = start_pos;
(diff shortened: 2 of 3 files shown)


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -803,8 +804,8 @@ EncodedResults SMStaticLLMPipeline::generate(
     }
 
     int64_t input_ids_data = -1;
-    int64_t position_ids_data = prompt_len -1;
-    std::vector<int64_t> attention_mask_data(1, prompt_len);
+    int64_t position_ids_data = prompt_len - 1;
+    std::vector<int64_t> attention_mask_data(prompt_len - 1, 1);
```
dmatveev, line 808: LOL, @TolyaTalamanov !!

samples/cpp/chat_sample/chat_sample.cpp
```diff
@@ -10,7 +10,7 @@ int main(int argc, char* argv[]) try {
     std::string prompt;
     std::string models_path = argv[1];
 
-    std::string device = "CPU";  // GPU, NPU can be used as well
+    std::string device = "NPU";  // GPU, NPU can be used as well
```
smirnov-alexey, line 13: I believe the default device should remain `CPU`

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -472,7 +480,7 @@ std::optional<NPUDesc> extract_npu_descriptor(ov::Core& core) {
 ov::AnyMap get_baseline_common_config() {
     ov::AnyMap config = {
         { "NPU_COMPILATION_MODE_PARAMS", "compute-layers-with-higher-precision=Sqrt,Power,ReduceMean,Add_RMSNorm" },
-        { "NPUW_DEVICES", "NPU" },
+        { "NPUW_DEVICES", "NPU,CPU" },
```
smirnov-alexey, line 483: Shouldn't be changed I believe

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +613,228 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+        }
+        OPENVINO_ASSERT(!input_vector->empty());
+        prompt = std::move(input_vector->front());
+    } else {
+        OPENVINO_ASSERT(std::holds_alternative<std::string>(inputs));
+        prompt = std::get<std::string>(inputs);
+    }
+
+    ov::genai::TokenizedInputs tokenized_input;
+    //if (m_is_chat_conversation) {
```
smirnov-alexey, line 669: Why is it commented out?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +613,228 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+
+    tokenized_input = m_tokenizer.encode(prompt);
+    auto encode_stop_time =  std::chrono::steady_clock::now();
+    auto encoded_results = generate(tokenized_input, config, streamer);
+
+    auto decode_start_time =  std::chrono::steady_clock::now();
+    DecodedResults decoded_results = {m_tokenizer.decode(encoded_results.tokens), encoded_results.scores};
+    auto decode_stop_time =  std::chrono::steady_clock::now();
+
+    //if (m_is_chat_conversation) {
```
smirnov-alexey, line 687: Why is it commented out?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +613,228 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    raw_counters.generate_durations.emplace_back(PerfMetrics::get_microsec(stop_time - start_time));
+    raw_counters.tokenization_durations.emplace_back(PerfMetrics::get_microsec(encode_stop_time - start_time));
+    raw_counters.detokenization_durations.emplace_back(PerfMetrics::get_microsec(decode_stop_time - decode_start_time));
+    decoded_results.perf_metrics.m_evaluated = false;
+    decoded_results.perf_metrics.evaluate_statistics(start_time);
+    return decoded_results;
+}
+
+template <typename T>
+void print_tensor(ov::Tensor t) {
```
smirnov-alexey, line 705: Not needed

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +613,228 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+void SMStaticLLMPipeline::finish_chat() {
+    std::cout << "[LOG_DEBUG] finish_chat not supported!" << std::endl;
+}
+
+std::unique_ptr<LLMPipelineImplBase>
+StaticLLMPipelineFactory::create(const std::filesystem::path& models_path,
+                                 const ov::genai::Tokenizer& tokenizer,
+                                 const std::string& device,
+                                 const ov::AnyMap& config) {
+    //return std::make_unique<StaticLLMPipeline>(models_path, tokenizer, device, config);
```
smirnov-alexey, line 827: Remove

samples/cpp/chat_sample/chat_sample.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: It should be implemented in future, added fixme, removed comments

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: It should be implemented in future, added fixme, removed comments

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Removed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.hpp
```diff
@@ -10,6 +10,45 @@
 namespace ov {
 namespace genai {
 
+struct StaticLLMPipelineFactory {
```
dmatveev, line 13: I think there could be a better namespace job done, clearly `statik::` could be a namespace here, so we'd get `statik::StatelessLLMPipeline` (the old class) and `statik::StatefulLLMPipeline` (the new one).

`statik::` is picked to avoid the clash with the keyword, it could be `Static::` too.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +621,211 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    ov::AnyMap properties = config;
+
+    const uint32_t kMaxPromptLen = pop_int_and_cast(properties, "MAX_PROMPT_LEN").value_or(1024u);
+    const uint32_t kMinResponseLen = pop_int_and_cast(properties, "MIN_RESPONSE_LEN").value_or(128u);
+    std::string generate_hint = pop_or_default<std::string>(properties, "GENERATE_HINT", "FAST_COMPILE");
+    ModelDesc model_desc = get_modeldesc_from_json(models_path / "config.json");
+
+    update_config(properties, {"NPU_USE_NPUW", "YES"});
+    update_config(properties, {"NPUW_LLM", "YES"});
+    update_config(properties, {"NPUW_LLM_MODEL_DESC", model_desc_to_string(model_desc)});
```
dmatveev, line 643: since it is C++, can we use a C++ structure directly here? Or the option is exposed as string only?

dmatveev: A downside - change in the option structure when it is a string will never break the build, but a change in the structure type will.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +621,211 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    const ov::AnyMap& config
+) : LLMPipelineImplBase(tokenizer,
+                        utils::from_config_json_if_exists(models_path)) {
+    ov::Core core;
+    auto model = core.read_model((models_path / "openvino_model.xml").string());
+
+    ov::AnyMap properties = config;
+
+    const uint32_t kMaxPromptLen = pop_int_and_cast(properties, "MAX_PROMPT_LEN").value_or(1024u);
+    const uint32_t kMinResponseLen = pop_int_and_cast(properties, "MIN_RESPONSE_LEN").value_or(128u);
```
dmatveev, line 637: see the above code, it was aligned to 64. Probably it makes sense to unify how these options are handled between the two classes (without overdesign)

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +621,211 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+
+    ov::genai::TokenizedInputs tokenized_input;
+    // FIXME: Support chat conversation.
+    tokenized_input = m_tokenizer.encode(prompt);
+    auto encode_stop_time =  std::chrono::steady_clock::now();
+    auto encoded_results = generate(tokenized_input, config, streamer);
+
+    auto decode_start_time =  std::chrono::steady_clock::now();
+    DecodedResults decoded_results = {m_tokenizer.decode(encoded_results.tokens), encoded_results.scores};
+    auto decode_stop_time =  std::chrono::steady_clock::now();
```
dmatveev, line 684: I'd highly recommend to use smt like https://github.com/openvinotoolkit/openvino/blob/master/src/plugins/intel_npu/src/plugin/npuw/perf.cpp#L9 - but it is clearly not for this PR (cc: @TolyaTalamanov )

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +621,211 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    int64_t position_ids_data = prompt_len - 1;
+    std::vector<int64_t> attention_mask_data(prompt_len - 1, 1);
+    const size_t max_tokens = config.get_max_new_tokens(prompt_len);
+    for (int i = 0; i < max_tokens - 1; ++i) {
+        input_ids_data = last_token;
+        ++position_ids_data;
+        attention_mask_data.push_back(1);
+
+        m_request.set_tensor("input_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, (void*)&input_ids_data));
+        m_request.set_tensor("position_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, (void*)&position_ids_data));
```
dmatveev, line 777: I don't see a point in having these two set all the time - the data could've been updated in-place for these tensors - for the future @TolyaTalamanov

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +621,211 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+        raw_perf_counters.m_batch_sizes.emplace_back(batch_size);
+        if (streamer_ptr && streamer_ptr->put(last_token)) {
+            break;
+        }
+
+        if (last_token == config.eos_token_id && !config.ignore_eos) {
+            break;
+        }
+
+        // TODO: How to check that KV-Cache is full?
```
dmatveev, line 795: Is this still an open? I believe the llm_infer_request reports it via throw now?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -605,6 +621,211 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    return;
+}
+
+std::unique_ptr<LLMPipelineImplBase>
+StaticLLMPipelineFactory::create(const std::filesystem::path& models_path,
+                                 const ov::genai::Tokenizer& tokenizer,
+                                 const std::string& device,
+                                 const ov::AnyMap& config) {
+    auto properties = config;
+    const auto use_sm_pipeline = pop_or_default(properties, "USE_SM_PIPELINE", false);
```
dmatveev, line 815: shouldn't it be `false` or `NO`? Or maybe it shouldn't a binary option either?

we also need to be careful about the option name choice here.. And tbh I don't have a good name here in mind.

It shouldn't be a public-looking option, that's for sure. Maybe it shouldn't be a configurable option at all but an env var, like we did for memory allocation - but that'd complicate testing in the existing environments.

`NPU_PIPELINE = STATEFUL` (as opposed to the today's `STATELESS`)?

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: We can, but it is not obvious where to define this structure, since OpenVINO NPUW code is not exposed as Public API. However, we can pass it as `map<std::string, std::string>`, if you think it is better. Current implementation via `std::string` ensures that `compiled_model.get_property("NPUW_LLM_MODEL_DESC")` will print something meaningful (but it is not a requirement).

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Yes, I did the alignment inside of the `LLMCompiledModel`  constructor in the NPUW. Do you think I need to remove it there and instead do alignment here? I did it in `LLMCompiledModel`, since I thought it might be of implementation detail..

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Yes, that is true, I think we discussed something like that already. And the conclusion was to check contracts of GenAI pipeline to understand what should be passed into the request. I don't remember the outcome, might be GenAI dynamic pipeline always pass some input, I need to check it.

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Yes, that is true! However, I don't know if this should be exception or just end of the generation stage with the warning

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Thank you! Will fix this

AsyaPronina: User can set `false` and `NO` , OpenVINO `ov::Any` can parse it to the `bool`.

src/cpp/src/llm_pipeline_static.hpp
AsyaPronina: Thank you!

src/cpp/src/llm_pipeline_static.hpp
```diff
@@ -10,6 +10,45 @@
...
+                                                       const ov::genai::Tokenizer& tokenizer,
+                                                       const std::string& device,
+                                                       const ov::AnyMap& config);
+
+    static std::unique_ptr<LLMPipelineImplBase> create(const std::filesystem::path& path,
+                                                       const std::string& device,
+                                                       const ov::AnyMap& config);
+};
+
+class SMStaticLLMPipeline : public LLMPipelineImplBase {
```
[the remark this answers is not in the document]
AsyaPronina: Thanks, sure!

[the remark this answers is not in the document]
AsyaPronina: Fixed!

src/cpp/src/llm_pipeline_static.hpp
AsyaPronina: Fixed!

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: What do you think?

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: What do you think?

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed!

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed!

src/cpp/src/llm_pipeline_static.hpp
AsyaPronina: @dmatveev @TolyaTalamanov , please help to disambiguate: we allow user to pass dynamic stateful OpenVINO model into our new pipeline, where we are hiding things like converting of model to static and making it stateless. Should we this way still name the pipeline as `static_llm::StatefulLLMPipeline`, as it still works with the static and stateless models inside? Or it can really be named as `Stateful` because now, `LLMCompiledModel` doesn't expose to user additional inputs and outputs, that correspond to states? (However, I don't know if by this logic the pipeline is still static)

src/cpp/src/llm_pipeline.cpp
```diff
@@ -678,13 +678,13 @@ class ContinuousBatchingAdapter final : public LLMPipelineImplBase {
 
 /* 
 * NPU reads some properties from the config file, but when LLMPipeline is initialized
-* from the model_str and weights_tensor, there are not files. 
+* from the model_str and weights_tensor, there are no files.
 * In the later case ModelDesc is stored in properties.
 * This function pops ModelDescr from the the properties and returns a pair of updated properties and ModelDescr.
 */
-std::pair<ov::AnyMap, ov::genai::ModelConfigDesc> split_model_descr(const ov::AnyMap& properties) {
+std::pair<ov::AnyMap, ov::genai::static_llm::ModelConfigDesc> split_model_descr(const ov::AnyMap& properties) {
```
TolyaTalamanov, line 685: Why additional namespace?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -412,6 +412,16 @@ ov::genai::ModelConfigDesc get_modeldesc_from_json(const std::filesystem::path&
     return desc;
 }
 
+std::string model_desc_to_string(const ov::genai::static_llm::ModelConfigDesc& model_desc) {
```
TolyaTalamanov, line 415: We should prevent spreading this stub...

We should avoid having `ModelConfigDesc` at all as it tmp solution to solve two problems:
1. Specify `seq_len` dim for KV-cache
2. Figure out if need to enable `V` tensors transpose

I'd rather prefer having `SEQ_LEN_DIM` and `OPTIMIZE_V_TENSORS` than passing such config as `string`.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -636,8 +661,232 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    const ov::genai::GenerationConfig& generation_config
+) : LLMPipelineImplBase(tokenizer, generation_config) {
+
+    bool use_blobs = false;
+    auto anyopt = get_option<bool>(properties, "USE_BLOBS");
+    if (anyopt.has_value()) {
+        use_blobs = *anyopt;
+    }
+    // Using model_str and weights_tesnor with blobs is meaningless.
+    OPENVINO_ASSERT(!use_blobs, "blobs cannot be used with model string and weights tensor");
```
TolyaTalamanov, line 698: Then probably it should be `exception` for the user, right?

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Since shape doesn't change, I believe we can create and set tensor once and then only update underlying data.

But `StatefulLLMPipeline` still calling `set_tensor` on every iteration:
https://github.com/openvinotoolkit/openvino.genai/blame/master/src/cpp/src/lm_encoding.cpp#L183-L185

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -636,8 +661,232 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    auto& metrics = results.perf_metrics;
+    metrics.num_input_tokens = batch_size * input_ids.get_shape().at(1);
+    metrics.load_time = this->m_load_time_ms;
+    metrics.raw_metrics.generate_durations.emplace_back(PerfMetrics::get_microsec(stop_time - start_time));
+    metrics.evaluate_statistics(start_time);
+    return results;
+}
+
+void StatefulLLMPipeline::start_chat(const std::string& system_message) {
+    // FIXME: Implement later
```
TolyaTalamanov, line 882: Then it should be an exception that currently chat isn't supported for ...

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -674,14 +923,14 @@ StaticLLMPipeline::StaticLLMPipeline(
     }
 };
 
-StaticLLMPipeline::StaticLLMPipeline(
+StatelessLLMPipeline::StatelessLLMPipeline(
```
TolyaTalamanov, line 926: Oh jez, to invasive, let's keep it `StaticLLMPipeline`, what's the problem with this?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -1118,5 +1366,53 @@ EncodedResults StaticLLMPipeline::generate(
...
 }
 
+std::unique_ptr<LLMPipelineImplBase>
+LLMPipelineFactory::create(const std::filesystem::path& models_path,
+                                 const ov::genai::Tokenizer& tokenizer,
+                                 const std::string& device,
+                                 const ov::AnyMap& config) {
+    auto properties = config;
+    const auto pipeline_mode = pop_or_default(properties, "NPU_PIPELINE", std::string("STATELESS"));
+    OPENVINO_ASSERT(pipeline_mode == "STATELESS" || pipeline_mode == "STATEFUL",
```
TolyaTalamanov, line 1376: I believe it should be an exception in this case.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -1118,5 +1366,53 @@ EncodedResults StaticLLMPipeline::generate(
     return results;
 }
 
+std::unique_ptr<LLMPipelineImplBase>
+LLMPipelineFactory::create(const std::filesystem::path& models_path,
+                                 const ov::genai::Tokenizer& tokenizer,
+                                 const std::string& device,
+                                 const ov::AnyMap& config) {
+    auto properties = config;
+    const auto pipeline_mode = pop_or_default(properties, "NPU_PIPELINE", std::string("STATELESS"));
```
TolyaTalamanov, line 1375: It's also a good practice to transform it to `enum`

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -1118,5 +1366,53 @@ EncodedResults StaticLLMPipeline::generate(
...
+                    "Only STATELESS and STATEFULL NPU_PIPELINE modes are supported!");
+    if (pipeline_mode == "STATEFUL") {
+        return std::make_unique<ov::genai::static_llm::StatefulLLMPipeline>(models_path, tokenizer, device, properties);
+    }
+    return std::make_unique<ov::genai::static_llm::StatelessLLMPipeline>(models_path, tokenizer, device, properties);
+}
+
+std::unique_ptr<LLMPipelineImplBase>
+LLMPipelineFactory::create(const std::filesystem::path& models_path,
+                                 const std::string& device,
```
TolyaTalamanov, line 1386: idents broken

src/cpp/src/llm_pipeline_static.hpp
```diff
@@ -9,23 +9,42 @@
...
-    StaticLLMPipeline(
+    StatefulLLMPipeline(
         const std::filesystem::path& path,
         const ov::genai::Tokenizer& tokenizer,
         const std::string& device,
         const ov::AnyMap& config
     );
 
-    StaticLLMPipeline(
+    StatefulLLMPipeline(
```
TolyaTalamanov, line 47: it's weird we have `static_llm::StatefulLLMPipeline`.

Maybe it's time we name it:
`npu_impl::StaticLLMPipeline` - old dual model pipeline
`npu_impl::StatefulLLMPipeline` - the new one with dynamic shapes and states

and whatever else...
`npu_impl::SingleModelLLMPipeline`
`npu_impl::StaticWhisperPipeline`


CC: @dmatveev

src/cpp/src/llm_pipeline_static.hpp
TolyaTalamanov: > I think there could be a better namespace job done, clearly `statik::` could be a namespace here, so we'd get `statik::StatelessLLMPipeline` (the old class) and `statik::StatefulLLMPipeline` (the new one).
> 
> `statik::` is picked to avoid the clash with the keyword, it could be `Static::` too.

Why not just `npu_impl`?

TolyaTalamanov: @AsyaPronina The new pipeline is definitely `stateful` as pipeline doesn't handle dynamism and states any longer. Nobody cares what is inside

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: ofc, I don't mind, should be done for `stateful` implementation as well

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Well, I believe on `GenAI` side we set KV caches size explicitly, don't we? So we know its size and how many tokens were processed. Shouldn't be a problem to check if cache is full

src/cpp/src/llm_pipeline_static.hpp
AsyaPronina: Discussed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Supported chat

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline.cpp
AsyaPronina: Discussed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: If it's not used any more, I believe it can be removed

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -586,6 +596,21 @@ std::optional<uint32_t> pop_int_and_cast(ov::AnyMap& config, const std::string&
...
 }
 
+void update_config(ov::AnyMap& config, const std::pair<std::string, ov::Any>& pair) {
+    if (config.count(pair.first) == 0) {
+        config.insert(pair);
+    }
+}
+
+void rename_key(ov::AnyMap& config, const std::string& old_key, const std::string& new_key) {
+    if (config.count(old_key) == 0) {
```
TolyaTalamanov, line 606: Seems more intuitive:
```
if (config.count(old_key) != 0) {
    ...
}
```

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -632,12 +657,290 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
     }
 }
 
+enum NpuPipeline {
```
TolyaTalamanov, line 660: How about these:

- `NPUPipelineType`
- `NPUPipelineKind`
- `NPUPipelineImpl`

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -632,12 +657,290 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+    const ov::AnyMap& config
+) : LLMPipelineImplBase(tokenizer,
+                        utils::from_config_json_if_exists(models_path)) {
+
+    auto model = genai::utils::singleton_core().read_model((models_path / "openvino_model.xml").string());
+    ModelConfigDesc model_desc = get_modeldesc_from_json(models_path / "config.json");
+    ov::AnyMap properties = config;
+
+    auto compiled = setupAndCompileModel(model, model_desc, properties);
+    m_request  = compiled->create_infer_request();
```
TolyaTalamanov, line 693: extra space?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -632,12 +657,290 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+            update_config(pipeline_config, {"NPUW_LLM_OPTIMIZE_V_TENSORS", true});
+    }
+
+    rename_key(pipeline_config, "PREFILL_CONFIG", "NPUW_LLM_PREFILL_CONFIG");
+    rename_key(pipeline_config, "GENERATE_CONFIG", "NPUW_LLM_GENERATE_CONFIG");
+  
+    return std::make_shared<ov::CompiledModel>(genai::utils::singleton_core().compile_model(model, "NPU", pipeline_config));
+}
+
+DecodedResults StatefulLLMPipeline::generate(
```
TolyaTalamanov, line 754: Isn't it the complete copy-paste from `StatelessLLMPipeline`, does it make sense to re-use?

TolyaTalamanov: I mean, maybe isolate into common util function (simple and w/o inheritance)

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -632,12 +657,290 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+
+    results.tokens[0].push_back(last_token);
+    if (streamer_ptr && streamer_ptr->put(last_token)) {
+        return results;
+    }
+
+    int64_t input_ids_data = -1;
+    int64_t position_ids_data = prompt_len - 1;
+    std::vector<int64_t> attention_mask_data(prompt_len - 1, 1);
+    m_request.set_tensor("input_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, (void*)&input_ids_data));
```
TolyaTalamanov, line 881: Wondering, can we set tensors once and then only update them?

```
m_request.set_tensor(input_ids);
while (;;) {
    m_request.get_tensor(input_ids) <- write there
}
```

TolyaTalamanov: Or even like this:

```
input_ids_data = -1;
m_request.set_tensor("input_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, (void*)&input_ids_data));

```

TolyaTalamanov, line 881: `reinterpret_cast`?

TolyaTalamanov: Won't work with attention mask though

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -660,7 +963,7 @@ StaticLLMPipeline::StaticLLMPipeline(
     const auto use_blobs = pop_or_default(properties, "USE_BLOBS", false);
     if (!use_blobs) {
         ModelConfigDesc model_desc = get_modeldesc_from_json(models_path / "config.json");
-        auto model = genai::utils::singleton_core().read_model(models_path / "openvino_model.xml", {}, properties);
```
TolyaTalamanov, line 663: @ilya-lavrenov, it was changed by you last time:
https://github.com/openvinotoolkit/openvino.genai/blame/master/src/cpp/src/llm_pipeline_static.cpp#L663

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -1118,5 +1421,49 @@ EncodedResults StaticLLMPipeline::generate(
...
+    auto properties_copy = properties;
+    const auto pipeline_mode = str_to_pipeline(pop_or_default(properties_copy, "NPU_PIPELINE", std::string("STATELESS")));
+    if (pipeline_mode == NpuPipeline::STATEFUL) {
+        return std::make_unique<ov::genai::static_llm::StatefulLLMPipeline>(model,
+                                                                            model_desc,
+                                                                            tokenizer,
+                                                                            device,
+                                                                            properties_copy,
+                                                                            generation_config);
+    }
```
TolyaTalamanov, line 1459: assert it's `STATELESS` just in case

src/cpp/src/llm_pipeline_static.hpp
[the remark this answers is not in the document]
dmatveev: Hope you read the next comment. :)

src/cpp/src/llm_pipeline_static.hpp
dmatveev: > Why not just npu_impl?

Previously there was nothing about NPU

src/cpp/src/llm_pipeline_static.cpp
dmatveev: > We can, but it is not obvious where to define this structure, since OpenVINO NPUW code is not exposed as Public API.

So you say our NPU parameters are not exported via headers?

src/cpp/src/llm_pipeline_static.cpp
dmatveev: WHY NPU AGAIN

dmatveev: Anyways, we do pass NPUW options directly in the pipeline code, so why not. Everything works but not `Npu` please.

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: `attention_mask` shape changes per iteration, for other tensors we can, but it will look like a bit shady, at least comment should be provided

AsyaPronina: why?

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: `StaticPipelineType`, etc

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Just more preferable than C-style casts because of several reasons

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Should the option `NPU_PIPELINE` be changed also to `STATIC_PIPELINE`?

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Don't need this, check is done in `str_to_pipeline`

src/cpp/src/llm_pipeline_static.cpp
ilya-lavrenov: yes, please, keep master's branch version of `read_model`

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -632,12 +655,287 @@ void copy_columns_by_row_chunks(const ov::Tensor& src, ov::Tensor& dst) {
...
+
+StatefulLLMPipeline::StatefulLLMPipeline(
+    const std::filesystem::path& models_path,
+    const ov::genai::Tokenizer& tokenizer,
+    const std::string&,
+    const ov::AnyMap& config
+) : LLMPipelineImplBase(tokenizer,
+                        utils::from_config_json_if_exists(models_path)) {
+
+    auto model = genai::utils::singleton_core().read_model((models_path / "openvino_model.xml").string());
```
ilya-lavrenov, line 686:
```suggestion
    auto model = genai::utils::singleton_core().read_model(models_path / "openvino_model.xml", {}, config);
```

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -647,12 +660,287 @@ void stream_generated_tokens(std::shared_ptr<ov::genai::StreamerBase> streamer_p
...
+    STATELESS
+};
+PipelineKind str_to_pipeline(const std::string& str) {
+    if (str == "STATEFUL") {
+        return PipelineKind::STATEFUL;
+    }
+    if (str == "STATELESS") {
+        return PipelineKind::STATELESS;
+    }
+    OPENVINO_THROW("Unsupported \"PIPELINE\" provided: " +
```
ilya-lavrenov, line 674:
```suggestion
    OPENVINO_THROW("Unsupported \"PIPELINE\" provided: ", 
```
use commas instead of concatenation

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -647,12 +660,287 @@ void stream_generated_tokens(std::shared_ptr<ov::genai::StreamerBase> streamer_p
...
+) : LLMPipelineImplBase(tokenizer, generation_config) {
+
+    bool use_blobs = false;
+    auto anyopt = get_option<bool>(properties, "USE_BLOBS");
+    if (anyopt.has_value()) {
+        use_blobs = *anyopt;
+    }
+    // Using model_str and weights_tensor with blobs is meaningless.
+    if (use_blobs) {
+        OPENVINO_THROW("Blobs cannot be used with model string and weights tensor");
```
ilya-lavrenov, line 716: condition + throw = `OPENVINO_ASSERT(!use_blobs, "Blobs cannot be used with model string and weights tensor")`

the same in other places.

AsyaPronina: Did this, Anatolii suggested `THROW` instead, @TolyaTalamanov?

AsyaPronina: Sorry, I was wrong, I've used `OPENVINO_ASSERT(!use_blobs && "Blobs cannot be used with model string and weights tensor")` instead of `OPENVINO_ASSERT(!use_blobs, "Blobs cannot be used with model string and weights tensor")`. Fixed now

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Fixed

src/cpp/src/llm_pipeline_static.cpp
AsyaPronina: Simple function won't work per my opinion, as it should call method of the class inside (`generate()` for encoder) and lambda-s won't look good I think. I suggest to do base class in a separate task

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -647,12 +660,269 @@ void stream_generated_tokens(std::shared_ptr<ov::genai::StreamerBase> streamer_p
...
+        raw_perf_counters.m_new_token_times.emplace_back(std::chrono::steady_clock::now());
+        raw_perf_counters.m_batch_sizes.emplace_back(batch_size);
+        if (streamer_ptr && streamer_ptr->put(last_token)) {
+            break;
+        }
+
+        if (last_token == config.eos_token_id && !config.ignore_eos) {
+            break;
+        }
+    }
```
ilya-lavrenov, line 897: looks like Sampler should be used here similar to https://github.com/openvinotoolkit/openvino.genai/pull/1431

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: udoli

src/cpp/src/llm_pipeline_static.hpp
TolyaTalamanov: device NPU, config NPUW

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: out `StaticLLMPipeline` parameters are documented stings, not exposed through headers

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: I'd prefer having `Exception` rather than `Assert` in cases where the error is important for user...

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Yes, I believe next iteration

src/cpp/src/llm_pipeline_static.cpp
ilya-lavrenov: assert is the same as exception, but on C++ syntax level you write a single line instead of condition + throw

it's not C's assert which is working only in debug mode

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Let's discuss it first

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Ah ok, then it makes sense

PR #1246, opened 2024-11-21
Add parametrization for the detokenization/decoding

pavel-esir:
image

Tokenizers IRs should be converted after https://github.com/openvinotoolkit/openvino_tokenizers/pull/325 is merged

Ticket CVS-154151


src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -87,23 +87,59 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
-    std::vector<std::string> decode(ov::Tensor tokens);
+    template <typename... Properties>
+    util::EnableIfAllStringAny<std::vector<std::string>, Properties...> decode(ov::Tensor tokens, Properties&&... properties) {
+        return decode(tokens, AnyMap{std::forward<Properties>(properties)...});
+    }
 
     /**
     * @brief batched decoding of tokens. 
     * @param tokens vector of vectors with tokens, tokens.size() is equal to batch_size
+    * @param tokenization_params AnyMap with detokenization parameters, e.g. {'skip_special_tokens', false}
```
ilya-lavrenov, line 128: tokenization_params => detokenization_params 
and replace python style of example to C++ one

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -87,23 +87,59 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
+    std::string decode(std::vector<int64_t> tokens, const ov::AnyMap& detokenization_params = {});
+
+    /**
+    * @brief decode sequence of tokens
+    * @param tokens vector storing tokens
+    * @param tokenization_params detokenization parameters,  e.g. ov::genai::skip_special_tokens(true)
+    * @return sequence string
+    */
+    template <typename... Properties>
+    util::EnableIfAllStringAny<std::string, Properties...> decode(std::vector<int64_t>& tokens, Properties&&... properties) {
```
ilya-lavrenov, line 102: detokenization_params here and in docs

src/cpp/src/tokenizer.cpp
```diff
@@ -86,11 +87,16 @@ class Tokenizer::TokenizerImpl {
...
+        ov::genai::utils::read_anymap_param(params, skip_special_tokens.name(), skip_special_tokens_flag);
+
         // If user requested add_special_tokens mode different from the current one,
         // need to set state variable.
         // If requested mode matches the stored state set, then don't touch states.
-        if (add_special_tokens == m_add_special_tokens) {
+        if (add_special_tokens_flag == m_add_special_tokens && skip_special_tokens_flag == m_skip_special_tokens) {
             return;
         }
         if (m_older_than_24_5) {
```
ilya-lavrenov, line 102: does m_skip_special_tokens work with IRs produced by 2024.4 and older?

src/cpp/src/tokenizer.cpp
```diff
@@ -86,11 +87,16 @@ class Tokenizer::TokenizerImpl {
 
     std::string m_chat_template = {};
 
-    void set_state_if_necessary(CircularBufferQueueElementGuard<ov::InferRequest>& infer_request_guard, bool add_special_tokens) {
+    void set_state_if_necessary(CircularBufferQueueElementGuard<ov::InferRequest>& infer_request_guard, const ov::AnyMap& params) {
+        bool add_special_tokens_flag = true;
+        bool skip_special_tokens_flag = false;
+        ov::genai::utils::read_anymap_param(params, add_special_tokens.name(), add_special_tokens_flag);
+        ov::genai::utils::read_anymap_param(params, skip_special_tokens.name(), skip_special_tokens_flag);
```
ilya-lavrenov, line 94: does it mean that if I run:
```c++
tok.encode("ccc", ov::genai::skip_special_tokens(true))
```
it will also allowed? 
there is distinguishing between tokenization and de-tokenization parameters.

src/python/py_tokenizer.cpp
```diff
@@ -63,27 +63,33 @@ void init_tokenizer(py::module_& m) {
...
             "decode",
-            [](Tokenizer& tok, ov::Tensor& tokens) -> py::typing::List<py::str> {
-                return pyutils::handle_utf8(tok.decode(tokens));
+            [](Tokenizer& tok, ov::Tensor& tokens, bool skip_special_tokens) -> py::typing::List<py::str> {
+                ov::AnyMap detokenization_params;
+                detokenization_params[ov::genai::skip_special_tokens.name()] = skip_special_tokens;
+                return pyutils::handle_utf8(tok.decode(tokens, detokenization_params));
             },
-            py::arg("tokens"),
+            py::arg("tokens"), py::arg("skip_special_tokens") = true,
```
ilya-lavrenov, line 82: @pavel-esir I suppose default behavior should be aligned with how users converted this de-tokenizer model. E.g. user can specify skip_special_tokens=False, but via GenAI python it will be still overwritten to `skip_special_tokens=True` and behavior of C++ and Python will be different.

src/cpp/src/tokenizer.cpp
ilya-lavrenov: if this PR depends on https://github.com/openvinotoolkit/openvino_tokenizers/pull/325, than we need to add condition for 2025.0 for `skip_special_tokens`?

src/python/py_tokenizer.cpp
pavel-esir: Thanks for noticing that, sure, here should've been `false`. Will fix that

src/cpp/src/tokenizer.cpp
pavel-esir: Yes, tokenization and detokenization parameters differ. This is done to be aligned with HF

tests/python_tests/test_chat_generate_api.py
```diff
@@ -217,3 +217,25 @@ def test_add_special_tokens(add_special_tokens, prompt):
     res_genai = genai_tokenzier.encode(prompt, add_special_tokens).input_ids.data
     res_hf = hf_tokenizer(prompt, return_tensors="np", add_special_tokens=add_special_tokens)["input_ids"]
     assert np.all(res_genai == res_hf)
+
+@pytest.mark.precommit
+@pytest.mark.xfail(reason="Need to turn them back on when openvino_tokenizers will be updated.")
```
pavel-esir, line 222: This tests will be green on CI only after genai will be update to the latest openvino_tokenizers. Meantime checked them locally, will enable back as soon as openvino_tokenziers will be updated.
image

src/cpp/include/openvino/genai/tokenizer.hpp

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: fixed

src/python/py_tokenizer.cpp
pavel-esir: fixed

src/cpp/src/tokenizer.cpp
pavel-esir: about  2025.0 condition, need to discuss with @apaniukov

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -87,23 +87,59 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     /**
     * @brief decode sequence of tokens
     * @param tokens vector storing tokens
+    * @param tokenization_params AnyMap with detokenization parameters, e.g. {"skip_special_tokens", false}
```
ilya-lavrenov, line 90:
```suggestion
    * @param detokenization_params AnyMap with detokenization parameters, e.g. ov::genai::skip_special_tokens(false)
```

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -87,23 +87,59 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
+    */
+    template <typename... Properties>
+    util::EnableIfAllStringAny<std::string, Properties...> decode(std::vector<int64_t>& tokens, Properties&&... detokenization_params) {
+        return decode(tokens, AnyMap{std::forward<Properties>(detokenization_params)...});
+    }
+
     /**
     * @brief decode tokens. 
     * @param tokens ov::Tensor with tokens with shape [batch_size, seq_len]
+    * @param tokenization_params AnyMap with detokenization parameters, e.g. {"skip_special_tokens", false}
```
ilya-lavrenov, line 109: please, correct docs in all places.

src/python/py_tokenizer.cpp
```diff
@@ -63,27 +63,33 @@ void init_tokenizer(py::module_& m) {
...
             "decode",
-            [](Tokenizer& tok, std::vector<std::vector<int64_t>>& tokens) -> py::typing::List<py::str> {
-                return pyutils::handle_utf8(tok.decode(tokens));
+            [](Tokenizer& tok, std::vector<std::vector<int64_t>>& tokens, bool skip_special_tokens) -> py::typing::List<py::str> {
+                ov::AnyMap detokenization_params;
+                detokenization_params[ov::genai::skip_special_tokens.name()] = skip_special_tokens;
+                return pyutils::handle_utf8(tok.decode(tokens, detokenization_params));
             },
-            py::arg("tokens"),
+            py::arg("tokens"), py::arg("skip_special_tokens") = false,
```
ilya-lavrenov, line 92: it's also incorrect, because actual default depends on how de-tokenize is converted

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Applied the first part with `tokenization_params` -> `detokenization_params `. But the other half should remain, since it's method for AnyMap and it accepts map not `ov::Properties`, for `ov::Properties` we have a separate method.

src/cpp/include/openvino/genai/tokenizer.hpp

src/python/py_tokenizer.cpp
pavel-esir: We discussed with @apaniukov and came to conclusion, that decode from genai should not rely on IRs default behavior, when someone calls `pipe.decode([1, 2, 42])` then flag from default argument of `decode` is used instead of IRs defaults. Hereby we will have `decode` interface aligned with HF.

For users who use directly `openvino_tokenizers` these 2 PR don't change anything, they still will have default behavior which is embedded into IR.

src/cpp/src/tokenizer.cpp
pavel-esir: We don't need condition for 2025.0. If IR is older then mechanism from this genai PR simply will not be activated (`MakeVocabDecoderSatateful` pass simply will not be matched) and the results will be the same as with older genai.

src/python/py_tokenizer.cpp
ilya-lavrenov: ok, I see that now C++ and Python behavior are aligned

src/cpp/src/tokenizer.cpp
```diff
@@ -343,23 +354,25 @@ class Tokenizer::TokenizerImpl {
         return {input_ids_, attention_mask_};
     }
 
-    std::string decode(std::vector<int64_t> tokens) {
+    std::string decode(std::vector<int64_t> tokens, const ov::AnyMap& detokenization_params = {}) {
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/tokenizer.cpp
```diff
@@ -135,18 +145,25 @@ class Tokenizer::TokenizerImpl {
...
-        ov::pass::Manager manager;
-        manager.register_pass<MakeCombineSegmentsSatateful>();
-        manager.run_passes(ov_tokenizer);
+        ov::pass::Manager manager_tok;
+        manager_tok.register_pass<MakeCombineSegmentsSatateful>();
+        manager_tok.run_passes(ov_tokenizer);
+        
+        ov::pass::Manager manager_detok;
+        manager_detok.register_pass<MakeVocabDecoderSatateful>();
+        manager_detok.run_passes(ov_detokenizer);
```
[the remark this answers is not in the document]
pavel-esir: That was a mistake. Corrected, now pass is run only if `ov_detokenizer` is not `nullptr`

ilya-lavrenov (approved):

ilya-lavrenov:
@pavel-esir please, manually port this PR as cherry pick does not work:
```
$ git cherry-pick 5ee41ec
error: commit 5ee41ecc5689db8da70b1e31c3b1a726307209d6 is a merge but no -m option was given.
fatal: cherry-pick failed
```
CC @andrei-kochin

pavel-esir:
Discussed with @andrei-kochin, since OVMS will not use this feature in 2024.6, no need to port it there. Ilya already removed label

PR #1255, opened 2024-11-26
Text2Img models from buffer

likholat:


likholat:
https://github.com/openvinotoolkit/openvino.genai/pull/1265

PR #1267, opened 2024-11-27
StaticLLMPipeline: Handle CACHE_DIR for NPUW

smirnov-alexey:
E-149053

### DIFF
--- a/src/cpp/src/llm_pipeline_static.cpp
+++ b/src/cpp/src/llm_pipeline_static.cpp
@@ -571,8 +571,10 @@ ov::Tensor make_tensor_slice(ov::Tensor tensor, size_t dim, size_t start_pos, si
     return ov::Tensor(tensor, start_shape, end_shape);
 }
 
-void drop_cache_dir(ov::AnyMap& config) {
-    if (config.count("NPU_USE_NPUW") != 0u) {
+void set_npuw_cache_dir(ov::AnyMap& config) {
+    std::optional<std::string> cache_dir = get_option<std::string>(config, "CACHE_DIR");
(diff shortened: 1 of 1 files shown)


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -571,8 +571,10 @@ ov::Tensor make_tensor_slice(ov::Tensor tensor, size_t dim, size_t start_pos, si
     return ov::Tensor(tensor, start_shape, end_shape);
 }
 
-void drop_cache_dir(ov::AnyMap& config) {
-    if (config.count("NPU_USE_NPUW") != 0u) {
+void set_npuw_cache_dir(ov::AnyMap& config) {
+    std::optional<std::string> cache_dir = get_option<std::string>(config, "CACHE_DIR");
+    if (config.count("NPU_USE_NPUW") != 0u && cache_dir) {
```
dmatveev, line 576: if NPUW is not used, but "CACHE_DIR" is specified, it should be passed as "CACHE_DIR" to the model.

dmatveev (commented):
Need to think how this behavior changes once we introduce the blob compilation (weightless or full) and what'd happen in this case.

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: In this case `CACHE_DIR` is already in the config and this `if` won't trigger, what's the problem?

dmatveev: ah, ok - makes sense, thanks!

dmatveev (approved):

PR #1272, opened 2024-11-28
StaticLLMPipeline: Decide when to enable NPUW_DQ_FULL property

smirnov-alexey:
Mirror of https://github.com/openvinotoolkit/openvino.genai/pull/1258


dmatveev (approved):

dmatveev:
This one can be merged, I believe - functionally it depends on https://github.com/openvinotoolkit/openvino.genai/pull/1272 to be in the release branch, but as the config is text-based there's no ABI-level dependency on that PR - it is generally a look-ahead feature.

PR #1286, opened 2024-12-03
Set onnx version to <=1.16.1 - who_what_benchmark requirements

sswierze:


eaidova:
@sswierze that should not be proper solution and fixing onnx should be done on infra level on windows ci machines. I added you to thread where we discussed the same issue and suggestion how to resolve it for llm bench

eaidova:
will be fixed as part of https://github.com/openvinotoolkit/openvino.genai/pull/1301

PR #1293, opened 2024-12-04
StaticWhisperPipeline: fix encoder input_features reshape

eshiryae:


dmatveev (approved):
LTGM, hope it helps to handle the new transformers!

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -441,6 +441,7 @@ void reshape_to_static_encoder(std::shared_ptr<ov::Model> model) {
             const auto& partial_shape = input.get_partial_shape();
             new_shape = partial_shape;
             new_shape[0] = 1;  // batch_dim
+            new_shape[1] = feature_size;
```
dmatveev, line 444: Just in case, what's our expectation on the shape size here - how much dimensions should be there?

eshiryae: There's 3 dims for this tensor ("input_features") : [batch_size, feature_size, sequence_length]
From what I see from new converted model, only first 2 dims are dynamic.

dmatveev: so would it worth to put an assert on the partial_shape.size() ?

eshiryae: Done

dmatveev (approved):

PR #1302, opened 2024-12-04
fix TextCallbackStreamer when characters are eaten by regex

pavel-esir:
- In some cases adding the next token can shorten the text,  e.g. when apostrophe removing regex had worked after adding new tokens. Need to hold on before flushing if apostrophe is the last symbol.
- ticket CVS-157216

### DIFF
--- a/samples/python/multinomial_causal_lm/multinomial_causal_lm.py
+++ b/samples/python/multinomial_causal_lm/multinomial_causal_lm.py
@@ -85,6 +85,7 @@ class IterableStreamer(openvino_genai.StreamerBase):
         text = self.tokenizer.decode(self.tokens_cache)
 
         word = ''
+        delay_n_chars = 4
         if len(text) > self.print_len and '\n' == text[-1]:
             # Flush the cache after the new line symbol.
             word = text[self.print_len:]            
@@ -93,11 +94,14 @@ class IterableStreamer(openvino_genai.StreamerBase):
         elif len(text) >= 3 and text[-3:] == chr(65533):
             # Don't print incomplete text.
             pass
-        elif len(text) > self.print_len:
+        elif len(text) > self.print_len + delay_n_chars:
             # It is possible to have a shorter text after adding new token.
             # Print to output only if text length is increaesed.
-            word = text[self.print_len:]
-            self.print_len = len(text)
+            # Also, in some cases adding the next token can shorten the text, 
+            # e.g. when apostrophe removing regex had worked after adding new tokens.
+            # Several last characters are delayed before flushed to output.
+            word = text[self.print_len:-delay_n_chars]
+            self.print_len = len(text) - delay_n_chars
         self.put_word(word)        
         
         if self.get_stop_flag():
--- a/src/cpp/src/text_callback_streamer.cpp
+++ b/src/cpp/src/text_callback_streamer.cpp
@@ -15,6 +15,14 @@ bool TextCallbackStreamer::put(int64_t token) {
     std::stringstream res;
     m_tokens_cache.push_back(token);
     std::string text = m_tokenizer.decode(m_tokens_cache);
+    
+    // In some cases adding the next token can shorten the text, 
+    // e.g. when apostrophe removing regex had worked after adding new tokens.
+    // Need to hold on before flushing if apostrophe is the last symbol.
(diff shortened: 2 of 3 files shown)


apaniukov (approved):

src/cpp/src/text_callback_streamer.cpp
```diff
@@ -15,6 +15,14 @@ bool TextCallbackStreamer::put(int64_t token) {
     std::stringstream res;
     m_tokens_cache.push_back(token);
     std::string text = m_tokenizer.decode(m_tokens_cache);
+    
+    // In some cases adding the next token can shorten the text, 
+    // e.g. when apostrophe removing regex had worked after adding new tokens.
+    // Need to hold on before flushing if apostrophe is the last symbol.
+    if (text.size() > 1 && text.back() == '\'') {
+        return on_finalized_subword_callback(res.str());
+    }
```
Wovchena, line 24: There's one more implementation which should be fixed: https://github.com/openvinotoolkit/openvino.genai/blob/e2fa0d002117bbfe6d9a6a4cc413e583c4455131/samples/python/multinomial_causal_lm/multinomial_causal_lm.py#L11

Wovchena, line 24: Is it guaranteed that this may happen to `'` only? Maybe the streamer should always keep the last symbol unprinted until `\n` or `end()`

Wovchena (changes requested):

src/cpp/src/text_callback_streamer.cpp
pavel-esir: We also thought about that. That would've been better than the current solution. 

I am afraid i can make a mistake when i manipulate with `print_len` with the current approach risks are minimal. But let me check with printing only last n - 1 (or n - k) tokens

pavel-esir: applied fix for it as well

pavel-esir: Done. Now the last N characters are not printed just in case if anything will change after the tokens arrive. Set N = 4, even setting 1 already fixes the issue, but to ensure set it a bit bigger. Please take a look.

tests/python_tests/test_generate_api.py
```diff
@@ -687,7 +687,7 @@ def test_unicode_pybind_decoding_3():
     model_id, path = 'katuni4ka/tiny-random-phi3', Path('tiny-random-phi3')
     pipe = read_model((model_id, path))[4]
     res_str = []
-    pipe.generate(",", max_new_tokens=4, streamer=lambda x: res_str.append(x))
+    pipe.generate(",", max_new_tokens=4, streamer=lambda x: res_str.extend(list(x)))
```
pavel-esir, line 690: Since with the new streamer `�`  can be spitted out with other characters as well, collect all character into a list instead of collecting list of subwords.

src/cpp/src/text_callback_streamer.cpp
pavel-esir: BTW, HF streamer has the same issue with eating characters. @apaniukov can confirm. If we merge this PR we will be one step ahead of HF 🚀

samples/python/multinomial_causal_lm/multinomial_causal_lm.py
```diff
@@ -93,11 +94,14 @@ def put(self, token_id: int) -> bool:
...
             # Don't print incomplete text.
             pass
-        elif len(text) > self.print_len:
+        elif len(text) > self.print_len + delay_n_chars:
             # It is possible to have a shorter text after adding new token.
             # Print to output only if text length is increaesed.
-            word = text[self.print_len:]
-            self.print_len = len(text)
+            # Also, in some cases adding the next token can shorten the text, 
+            # e.g. when apostrophe removing regex had worked after adding new tokens.
```
Wovchena, line 101:
```suggestion
            # E.g. when apostrophe removing regex had worked after adding new tokens.
```
This statements are equivalent:
> It is possible to have a shorter text after adding new token.

and
> in some cases adding the next token can shorten the text

Wovchena (approved):
There's no much time left. Address https://github.com/openvinotoolkit/openvino.genai/pull/1302/files#r1870703054 when porting to master

PR #1309, opened 2024-12-05
Genai/optimum support streaming mode

zhaohb:
Support chunk streaming mode, mainly to reduce the number of decode calls, thereby improving performance.


andrei-kochin:
@zhaohb now for 24.6 =)

zhaohb:
> @zhaohb now for 24.6 =)

I had submit a new PR for 24.6.

PR #1320, opened 2024-12-06
Demo for chunk streaming

zhaohb:
Add chat example for chunk streaming


samples/python/chat_sample/chat_chunk_streaming_sample.py
Wovchena, line 1: There's already on reimplementation of streamer in sample: https://github.com/openvinotoolkit/openvino.genai/blob/ee91fcfa1b43c7cd42be58688832318fe6e8b0af/samples/python/multinomial_causal_lm/multinomial_causal_lm.py#L11 Would you patch it instead

zhaohb: Yes, it looks like the implementation is normal streaming. OK, I will change this file to chunk streaming mode as well.

samples/python/multinomial_causal_lm/README.md
```diff
@@ -19,7 +19,10 @@ optimum-cli export openvino --trust-remote-code --model TinyLlama/TinyLlama-1.1B
 
 Install [deployment-requirements.txt](../../deployment-requirements.txt) via `pip install -r ../../deployment-requirements.txt` and then, run a sample:
 
-`python multinomial_causal_lm.py TinyLlama-1.1B-Chat-v1.0 "Why is the Sun yellow?"`
+#### Simple streaming
+`python multinomial_causal_lm.py -m TinyLlama-1.1B-Chat-v1.0 -p "Why is the Sun yellow?" -tl 1`
+#### Chunk streaming
+`python3 multinomial_causal_lm.py -m TinyLlama-1.1B-Chat-v1.0 -p "Why is the Sun yellow?" -tl 10`
```
Wovchena, line 25: Since I requested to remove `t1` cmd arg, this section should be removed.

There's a section below about streaming. Please extend it to cover chunking as well.

Wovchena (changes requested):

samples/python/multinomial_causal_lm/multinomial_causal_lm.py
```diff
@@ -120,23 +120,91 @@ def end(self):
...
+        Processes a token and manages the decoding buffer. Adds decoded text to the queue.
+
+        Args:
+            token_id (int): The token_id to process.
+
+        Returns:
+            bool: True if generation should be stopped, False otherwise.
+        """
+        self.tokens_cache.append(token_id)
+        if len(self.tokens_cache) % self.tokens_len == 0:
```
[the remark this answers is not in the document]
zhaohb: as follows：
```python
    def put(self, token_id: int) -> bool:
        if (len(self.tokens_cache) + 1) % self.tokens_len != 0:
            self.tokens_cache.append(token_id)
            return False
        return super().put(token_id)
```

Wovchena (approved):

PR #1333, opened 2024-12-06
bump openvino_tokenizers version

pavel-esir:


slyalin (approved):

ilya-lavrenov (changes requested):
Please, address your comment in another PR https://github.com/openvinotoolkit/openvino.genai/pull/1246/files#r1853668649

pavel-esir:
> Please, address your comment from another PR https://github.com/openvinotoolkit/openvino.genai/pull/1246/files#r1853668649

done

pavel-esir:
Difference between HF chat_sample and our chat_sample arises from the fact that in tests we used LlamaTokenzier which is different from tokenzer used by default with AutoTokenzier. LlamaTokenzier gives slightly different results that AutoTokenzier, and openvino_tokenizers is alighned with the AutoTokenizers.

When openvino_tokenizers fixed and issue and aligned to AutoTokenzier/LlamaTokenizerFast then 29871 tokens appeared which is missing in LlamaTokenzier and precommit tests started to differ.

image

@apaniukov please confirm that analysis is correct

ilya-lavrenov (approved):

PR #1337, opened 2024-12-06
Add perf metrics support for WhisperStaticPipeline

eshiryae:


src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -136,17 +144,29 @@ void set_decoder_input_ids_attention_mask(ov::InferRequest& decoder,
     std::fill(attention_mask_data + init_ids.size(), attention_mask_data + attention_mask_tensor.get_size(), 0u);
 }
 
+void infer_with_perf_metrics(ov::InferRequest& request, ov::genai::RawPerfMetrics& raw_metrics) {
```
TolyaTalamanov, line 147: Duplication from: https://github.com/openvinotoolkit/openvino.genai/blob/d91123a375c62a0012e3239adfb1c6fc14b4bcbf/src/cpp/src/whisper/whisper.cpp#L81-L90

@as-suvorov  does it make sense to re-use this function in both pipelines?

TolyaTalamanov:
@as-suvorov could you have a look, please?

as-suvorov:
I think we should add `features_extraction_durations` as well to be aligned with stateful pipeline. [Example](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/whisper/whisper.cpp#L310)

src/cpp/src/whisper_pipeline_static.cpp
eshiryae: Added also functions for non segment tokens filtering, which are also duplicate functions in whisper.cpp.
Maybe we could move common functions to common space and reuse them? (e.g. create whisper_utils.hpp/cpp for this)

src/cpp/src/whisper/whisper_utils.hpp
```diff
@@ -0,0 +1,27 @@
...
+namespace ov {
+
+namespace genai {
+
+namespace utils {
+
+void infer_with_perf_metrics(ov::InferRequest& request, ov::genai::RawPerfMetrics& raw_metrics);
+
+template <typename T>
+void filter_by_ranges(std::vector<T>& value, size_t offset, std::vector<std::pair<size_t, size_t>>& ranges);
```
as-suvorov, line 19: It seems `filter_by_ranges` used only in `whisper_ustils.cpp`. If you don't want to use it somewhere else it might make sense do not expose it to `ov::genai::utils` namespace. You can use unnamed namespace in `whisper_utils.cpp` (example: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/whisper/whisper.cpp#L23).

as-suvorov (approved):

src/cpp/src/whisper/whisper_utils.cpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2023-2024 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "whisper_utils.hpp"
+
+namespace {
+
+template <typename T>
```
TolyaTalamanov, line 8: Well, in doesn't have sense to put templated function into anonym namespace, does it?

src/cpp/src/whisper/whisper_utils.cpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2023-2024 Intel Corporation
```
TolyaTalamanov, line 1: I believe just `2024`

src/cpp/src/whisper/whisper_utils.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+void filter_non_segment_metrics(ov::genai::RawPerfMetrics& raw_metrics,
+                                size_t offset,
+                                std::vector<std::pair<size_t, size_t>>& ranges) {
+    filter_by_ranges(raw_metrics.m_token_infer_durations, offset, ranges);
+    filter_by_ranges(raw_metrics.m_new_token_times, offset, ranges);
+    filter_by_ranges(raw_metrics.m_batch_sizes, offset, ranges);
+}
+
+} // namespace utils
```
TolyaTalamanov, line 44: identation is broken

TolyaTalamanov (approved):
LGTM 👍 

Only minor comments!

src/cpp/src/whisper/whisper_utils.cpp
as-suvorov: Why not?
Example from static pipeline: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/whisper_pipeline_static.cpp#L32

PR #1350, opened 2024-12-09
llm_pipeline_static: flush streamer after generation loop is complete

RyanMetcalfeInt8:
Without these changes, chat_sample with NPU device produces responses that are clipped by 4 characters:
image

Flushing the streamer (as *get_lm_encoded_results()* does in non-static LLM case) seems to resolve the issue.


dmatveev (approved):

dmatveev:
Thanks a lot for your analysis & fix @RyanMetcalfeInt8 !

PR #1368, opened 2024-12-12
[LLM Bench] Allow Image Generation Models to Run in BF16

anzr299:
This change allows setting image generation models to BF16 using config passed while running benchmark.


tools/llm_bench/llm_bench_utils/pt_utils.py
```diff
@@ -131,6 +131,7 @@ def create_image_gen_model(model_path, device, **kwargs):
             model_class = PT_MODEL_CLASSES_MAPPING[model_type]
             start = time.perf_counter()
             pipe = model_class.from_pretrained(model_path)
+            pipe = set_bf16(pipe, device, **kwargs)
```
wgzintel, line 134: Is set_bf16 required or can it be made optional?

anzr299: It is based on the conditions inside the `set_bf16` funciton, if it is passed in load_config flag, it sets the model to bf16 otherwise skips it.

eaidova (approved):

wgzintel (approved):

PR #1373, opened 2024-12-12
[WWB]: Add ImageText-to-Image pipeline validation

AlexKoff88:
CVS-159223


AlexKoff88:
> is image to image sensitive to random `echarlaix/tiny-random-stable-diffusion-xl-image-to-image` models?

the problem came from the difference in the resolutions of generated images by HF and GenAI libs.

tools/who_what_benchmark/whowhatbench/image2image.py
```diff
@@ -0,0 +1,143 @@
...
+import torch
+import openvino_genai
+
+from .registry import register_evaluator
+from .text2image_evaluator import Text2ImageEvaluator
+
+from .whowhat_metrics import ImageSimilarity
+
+
+class Generator(openvino_genai.Generator):
```
eaidova, line 17: is it still will be needed after https://github.com/openvinotoolkit/openvino.genai/pull/1379 merge?

tools/who_what_benchmark/whowhatbench/image2image.py
```diff
@@ -0,0 +1,143 @@
...
+@register_evaluator("image-to-image")
+class Image2ImageEvaluator(Text2ImageEvaluator):
+    def __init__(
+        self,
+        base_model: Any = None,
+        gt_data: str = None,
+        test_data: Union[str, list] = None,
+        metrics="similarity",
+        similarity_model_id: str = "openai/clip-vit-large-patch14",
+        resolution=(512, 512),
```
eaidova, line 56: not sure that providing resolution without resizing input image has effect for all pipelines (I know that only flux ignore input image resolution other models uses it as input resolution)

tools/who_what_benchmark/whowhatbench/image2image.py
```diff
@@ -0,0 +1,143 @@
...
+        num_inference_steps=4,
```
eaidova, line 57: is it always enough or there is way to override?
BTW, it also makes sense to support strength parameter (it used for  blending input image and text prompt and regulate number of steps in image2image, 1 - original image preserved text ignored, 0 - image ignored text prompt used)

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -0,0 +1,252 @@
...
+        try:
+            model = AutoModelForVision2Seq.from_pretrained(
+                model_id, trust_remote_code=True, device_map=device.lower()
+            )
+        except ValueError:
+            try:
+                model = AutoModel.from_pretrained(
+                    model_id, trust_remote_code=True, device_map=device.lower()
+                )
+            except ValueError:
```
eaidova, line 156: I think it may be more robust if you will use model_type from config for selection model loading approach (e.g. logic bellow required only for phi3-vision, but there is no guarantee that some other model will not go by this path)

tools/who_what_benchmark/whowhatbench/image2image.py
AlexKoff88: What is your suggestion here? Remove this parameter and let the pipeline decide based on the input image or the pipeline's intrinsic parameters?

tools/who_what_benchmark/whowhatbench/image2image.py
AlexKoff88: yes, it can be overridden in the CLI as well. Do we need this "strength" to understand results convergence across pipeline implementations?

tools/who_what_benchmark/whowhatbench/model_loaders.py
AlexKoff88: Right, and as you can see this is the last clause in the try-except chain. Anyway `load_visual_text_model` is called if "visual-text" model_type is passed explicitly.

AlexKoff88: To tell the truth, I like it more rather than hardcode `model_type == "phi3_v"`

tools/who_what_benchmark/whowhatbench/image2image.py
ilya-lavrenov: merged, we can drop custom generator

tools/who_what_benchmark/whowhatbench/image2image.py
eaidova: I think for now not, but for development experiments it may be interested

tools/who_what_benchmark/whowhatbench/image2image.py
eaidova: it will depend on is behavior of all backends (diffusers, optimum, genai) aligned or not. The basic problem here is if baseline and target model have different preferences between input image resolution and manually specified height/width we may get images of different size (not sure but may be problem for embeddings comparison and required some additional postprocessing steps)
in optimum it is aligned with diffusers
@ilya-lavrenov which logic uses genai for selection image size?

ilya-lavrenov: >which logic uses genai for selection image size?

based on input image size
https://github.com/openvinotoolkit/openvino.genai/blob/f177ffc9799ef34a57b257e1811a60c68c167eb2/src/cpp/src/image_generation/stable_diffusion_pipeline.hpp#L424-L431

AlexKoff88: Ideally, the behavior should be aligned across various APIs to provide a smooth transition from one API to another.

ilya-lavrenov: it's aligned with HF:
```python
                init_latents = [
                    retrieve_latents(self.vae.encode(image[i : i + 1]), generator=generator[i])
                    for i in range(batch_size)
                ]
```
https://github.com/huggingface/diffusers/blob/8ce37ab055372dedf4e9621ed63374a019d93f5d/src/diffusers/pipelines/stable_diffusion/pipeline_stable_diffusion_img2img.py#L750-L753

Here initial `image` is converted to latent space and used as initial latent.

Input `width` or `height` parameters are not used as in text to image.

VAE encode uses https://github.com/huggingface/diffusers/blob/8ce37ab055372dedf4e9621ed63374a019d93f5d/src/diffusers/image_processor.py#L540-L585

AlexKoff88: So, I believe I should remove the resolution as a parameter and let the pipeline decide on it, right?

ilya-lavrenov: for image2image - yes

eaidova (approved):

AlexKoff88:
I am getting good accuracy convergence for big models, e.g. SD-XL but cannot make tests working with GenAI for any dummy model (I tried a several).

AlexKoff88:
@ilya-lavrenov, CI is passed with non-dummy model.

ilya-lavrenov:
WWB tests on master:
{1B722F12-7A91-46C6-910F-C419862FD060}

WWB tests on current PR:
{74B06E4B-C0BD-484D-991A-F4D2ABA9AA72}

WWB tests with dummy random model (which still is huge):
{C71C9DDE-648B-440B-9A3A-FBC08384E0A8}

Let's switch to dummy model once the related issue is fixed by CPU team.

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -467,8 +269,21 @@ def genai_gen_image(model, prompt, num_inference_steps, generator=None):
     return image
 
 
+def genai_gen_image2image(model, prompt, image, num_inference_steps, generator=None):
+    image_data = ov.Tensor(np.array(image.getdata()).reshape(1, image.size[1], image.size[0], 3).astype(np.uint8))
+    image_tensor = model.generate(
+        prompt,
+        image=image_data,
+        num_inference_steps=num_inference_steps,
+        strength=0.8,
```
ilya-lavrenov, line 278: do you run optimum / diffusers with the same `strength` value?

tools/who_what_benchmark/tests/test_cli_image.py
```diff
@@ -65,6 +67,7 @@ def test_image_model_types(model_id, model_type, backend):
 @pytest.mark.parametrize(
     ("model_id", "model_type"),
     [
+        ("dreamlike-art/dreamlike-anime-1.0", "image-to-image"),
```
ilya-lavrenov, line 70: maybe `echarlaix/tiny-random-latent-consistency` can work here instead of non-working `echarlaix/tiny-random-stable-diffusion-xl` ?

AlexKoff88: I tried several of them including the one you mentioned and they do not work due the known bug.

eaidova: do you need torch model in these tests or openvino only enough?
We have several preconverted models here https://huggingface.co/collections/OpenVINO/image-generation-67697d9952fb1eee4a252aa8 (I requested dreamlike models too, but current status is waiting approval)

among trained but still small is https://huggingface.co/segmind/tiny-sd
but not sure that used in model scheduler currently supported

tools/who_what_benchmark/whowhatbench/wwb.py
AlexKoff88: Good point. It is 0.8 by default in Diffusers but it is better to set it explicitly.

AlexKoff88: Actually, it is already set to 0.8 as well in default_gen_image_fn of im2im pipeline.

tools/who_what_benchmark/tests/test_cli_image.py
eaidova: also for speedup testing, you can expose num_infeence_steps parameter in cli (I suppose even with streigh=0.8, it runs 40 steps

AlexKoff88: > do you need torch model in these tests or openvino only enough? We have several preconverted models here https://huggingface.co/collections/OpenVINO/image-generation-67697d9952fb1eee4a252aa8 (I requested dreamlike models too, but current status is waiting approval)
> 
> among trained but still small is https://huggingface.co/segmind/tiny-sd but not sure that used in model scheduler currently supported

Yes, I need a Torch model and I tried "[segmind/tiny-sd](https://huggingface.co/segmind/tiny-sd)" but it didn't work OOB.

And, yes `num_infeence_steps` is already exposed and I am planning to use it in the test.

PR #1405, opened 2024-12-18
add performance statistics for image generation

xufang-lisa:
tickets: CVS-157338


ilya-lavrenov:
Could you please provide example of such prints?
E.g. for SDXL and FLUX

likholat:
@xufang-lisa let's add a custom struct `ImageGenerationPerfMetrics` and collect the metrics there:
```cpp
struct OPENVINO_GENAI_EXPORTS ImageGenerationPerfMetrics {
    float load_time; // model load time 
    float generate_duration; // duration of method generate(...)

    MeanStdPair iteration_duration; // Mean-Std time of one generation iteration
    std::map<std::string, float> encoder_inference_duration; // inference duration for each encoder
    MeanStdPair unet_inference_duration; // inference duration for unet model, should be filled with zeros if we don't have unet
    MeanStdPair transformer_inference_duration; // inference duration for transformer model, should be filled with zeros if we don't have transformer
    float vae_encoder_inference_duration; // inference duration of vae_encoder model, should be filled with zeros if we don't use it
    float vae_decoder_inference_duration; // inference duration of  vae_decoder model
 
    bool m_evaluated = false;
 
    RawImageGenerationPerfMetrics raw_metrics;
};
 
struct OPENVINO_GENAI_EXPORTS RawImageGenerationPerfMetrics {
    std::vector<MicroSeconds> unet_inference_durations; // unet durations for each step
    std::vector<MicroSeconds> transformer_inference_durations; //  transformer durations for each step
    std::vector<MicroSeconds> iteration_durations;  //   iteration durations for each step
};
```

I'd also like to propose return `ov::Tensor` as a `generate` method output and add `get_perfomance_metrics()` method

xufang-lisa:
> @xufang-lisa let's add a custom struct `ImageGenerationPerfMetrics` and collect the metrics there:
> 

updated

ilya-lavrenov:
> > @xufang-lisa let's add a custom struct `ImageGenerationPerfMetrics` and collect the metrics there:
> 
> updated

could you please add benchmark for image generation similar to benchmark for GenAI LLM, which can print all these detailed statistic?
See https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/benchmark_genai/benchmark_genai.cpp#L54-L61

And provide here an output of such sample.

BTW, original samples should be kept as is and avoid printed timings.

xufang-lisa:
> > > @xufang-lisa let's add a custom struct `ImageGenerationPerfMetrics` and collect the metrics there:
> > 
> > 
> > updated
> 
> could you please add benchmark for image generation similar to benchmark for GenAI LLM, which can print all these detailed statistic? See https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/benchmark_genai/benchmark_genai.cpp#L54-L61
> 
> And provide here an output of such sample.
> 
> BTW, original samples should be kept as is and avoid printed timings.

@ilya-lavrenov Do you mean to add support for image generation in benchmark_genai?

ilya-lavrenov:
> > > > @xufang-lisa let's add a custom struct `ImageGenerationPerfMetrics` and collect the metrics there:
> > > 
> > > 
> > > updated
> > 
> > 
> > could you please add benchmark for image generation similar to benchmark for GenAI LLM, which can print all these detailed statistic? See https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/benchmark_genai/benchmark_genai.cpp#L54-L61
> > And provide here an output of such sample.
> > BTW, original samples should be kept as is and avoid printed timings.
> 
> @ilya-lavrenov Do you mean to add support for image generation in benchmark_genai?

no, create dedicated benchmark application for image generation, which is similar to VLM / LLM benchmarks

samples/cpp/image_generation/benchmark_text2image.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+#include "openvino/genai/image_generation/text2image_pipeline.hpp"
+#include <cxxopts.hpp>
+#include "imwrite.hpp"
+
+int main(int argc, char* argv[]) try {
+    cxxopts::Options options("benchmark_image_generation", "Help command");
+
+    options.add_options()
+    ("m,model", "Path to model and tokenizers base directory", cxxopts::value<std::string>()->default_value("."))
```
ilya-lavrenov, line 12: please, don't imply that model by default is located in current folder.
I would make this parameters as required as in other benchmark applications:
https://github.com/openvinotoolkit/openvino.genai/blob/4fb48deed27d23cbb517eb1247e8f1f48bdc8596/samples/cpp/text_generation/benchmark_genai.cpp#L10-L11

samples/cpp/image_generation/benchmark_text2image.cpp
```diff
@@ -0,0 +1,106 @@
...
+                                               return acc + duration;
+                                           });
+    if (!total_inference_durations.empty()) {
+        inference_mean /= total_inference_durations.size();
+    }
+
+    std::cout << std::fixed << std::setprecision(2);
+    std::cout << "Load time: " << load_time << " ms" << std::endl;
+    std::cout << "One generate avg time: " << generate_mean << " ms" << std::endl;
+    std::cout << "Total inference for one generate avg time: " << inference_mean << " ms" << std::endl;
```
ilya-lavrenov, line 93: can we print more information? E.g. how much time is taken by text encoders, VAE encode / decode, first / other iterations for main denoising loop

Because currently printed information is non informative (about what happens in pipeline) and can be achieved by external benchmarking around generate method.

samples/cpp/image_generation/README.md
```diff
@@ -148,3 +149,35 @@ The resuling image is:
    ![](./inpainting.bmp)
 
 Note, that LoRA, heterogeneous execution and other features of `Text2ImagePipeline` are applicable for `InpaintingPipeline`.
+
+## benchmarking sample for text to image pipeline
+
+This `benchmark_text2image.cpp` sample script demonstrates how to benchmark the text to image pipeline. The script includes functionality for warm-up iterations, generating image, and calculating various performance metrics.
```
ilya-lavrenov, line 155: can we generalize this benchmark to support inpainting / image to image as well?
We can add an argument to specify pipeline type

src/cpp/src/image_generation/models/autoencoder_kl.cpp
```diff
@@ -223,11 +223,14 @@ AutoencoderKL& AutoencoderKL::compile(const std::string& device, const ov::AnyMa
...
 }
 
-ov::Tensor AutoencoderKL::decode(ov::Tensor latent) {
+ov::Tensor AutoencoderKL::decode(ov::Tensor latent, float& infer_duration) {
     OPENVINO_ASSERT(m_decoder_request, "VAE decoder model must be compiled first. Cannot infer non-compiled model");
 
     m_decoder_request.set_input_tensor(latent);
+    const auto infer_start = std::chrono::steady_clock::now();
     m_decoder_request.infer();
+    infer_duration = ov::genai::PerfMetrics::get_microsec(std::chrono::steady_clock::now() - infer_start);
```
ilya-lavrenov, line 232: can external code (within Image generation pipeline) measure time of `decode` function? W/ such approach we don't need extra output arguments for this method.

The same approach for all other models within this `image_generation/models` folder.

src/cpp/include/openvino/genai/image_generation/image2image_pipeline.hpp
```diff
@@ -87,8 +87,11 @@ class OPENVINO_GENAI_EXPORTS Image2ImagePipeline {
 
     ov::Tensor decode(const ov::Tensor latent);
 
+    ImageGenerationPerfMetrics get_performance_metrics();
+
 private:
     std::shared_ptr<DiffusionPipeline> m_impl;
+    ImageGenerationPerfMetrics m_perf_metrics;
```
ilya-lavrenov, line 94: this field should be hidden inside `std::shared_ptr<DiffusionPipeline> m_impl;`

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -477,6 +500,7 @@ class FluxPipeline : public DiffusionPipeline {
     std::shared_ptr<T5EncoderModel> m_t5_text_encoder = nullptr;
     std::shared_ptr<AutoencoderKL> m_vae = nullptr;
     ImageGenerationConfig m_custom_generation_config;
+    ImageGenerationPerfMetrics m_perf_metrics;
```
ilya-lavrenov, line 503: should it be moved to base `DiffusionPipeline` class? All derived pipelines will inherit this field.

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -358,28 +364,45 @@ class FluxPipeline : public DiffusionPipeline {
         float* timestep_data = timestep.data<float>();
 
         for (size_t inference_step = 0; inference_step < timesteps.size(); ++inference_step) {
```
ilya-lavrenov, line 366: should not we measure time of whole iteration (inference of denoising model + scheduler overhead + some other steps like in inpainting) ?

samples/cpp/image_generation/README.md
```diff
@@ -148,3 +149,59 @@ The resuling image is:
...
+For example:
+
+`./benchmark_text2image -pt text2image -m dreamlike_anime_1_0_ov/FP16 -n 10 -d CPU`
+
+Performance output:
+
+```
+[warmup-0] generate time: 4978.00 ms, total infer time:4967.62 ms, infer step num:20
+[warmup-0] encoder infer time: (text_encoder-195.93 ms) (text_encoder_2-145.52 ms)
+[warmup-0] unet total infer time:3920.33 ms, infer number:20
```
ilya-lavrenov, line 186: could you please add `iteration duration` for transformer and Unet ?

It would be great to see difference between first and other iterations.

src/cpp/src/image_generation/image_generation_perf_metrics.cpp
```diff
@@ -0,0 +1,109 @@
...
+        float total = std::accumulate(raw_metrics.transformer_inference_durations.begin(),
+                                      raw_metrics.transformer_inference_durations.end(),
+                                      0.0f,
+                                      [](const float& acc, const ov::genai::MicroSeconds& duration) -> float {
+                                          return acc + duration.count() / 1000.0f;
+                                      });
+        total_duration += total;
+    }
+
+    total_duration += vae_decoder_inference_duration;
```
ilya-lavrenov, line 92: why `vae_encoder_inference_duration ` is not counted here? It's used in image2image and inpainting

xufang-lisa: Updated, `vae_encoder_inference_duration` is included in `get_encoder_infer_duration()`.

samples/cpp/image_generation/CMakeLists.txt
```diff
@@ -107,4 +107,30 @@ set_target_properties(inpainting PROPERTIES
 install(TARGETS inpainting
         RUNTIME DESTINATION samples_bin/
         COMPONENT samples_bin
-        EXCLUDE_FROM_ALL)
\ No newline at end of file
+        EXCLUDE_FROM_ALL)
+
+# benchmark_text2image
```
ilya-lavrenov, line 112:
```suggestion
# benchmark_image_gen
```

samples/cpp/image_generation/CMakeLists.txt
```diff
@@ -107,4 +107,30 @@ set_target_properties(inpainting PROPERTIES
...
+    cmake_policy(SET CMP0135 NEW)
+endif()
+
+FetchContent_Declare(cxxopts
+    URL https://github.com/jarro2783/cxxopts/archive/refs/tags/v3.1.1.tar.gz
+    URL_HASH SHA256=523175f792eb0ff04f9e653c90746c12655f10cb70f1d5e6d6d9491420298a08)
+FetchContent_MakeAvailable(cxxopts)
+
+add_executable(benchmark_image benchmark_image.cpp load_image.cpp imwrite.cpp)
+target_include_directories(benchmark_image PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}" "${CMAKE_BINARY_DIR}")
```
ilya-lavrenov, line 126:
```suggestion
target_include_directories(benchmark_image PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}")
```
am I right that binary dir is not required?

samples/cpp/image_generation/README.md
```diff
@@ -148,3 +149,63 @@ The resuling image is:
    ![](./inpainting.bmp)
 
 Note, that LoRA, heterogeneous execution and other features of `Text2ImagePipeline` are applicable for `InpaintingPipeline`.
+
+## benchmarking sample for text to image pipeline
```
ilya-lavrenov, line 153:
```suggestion
## benchmarking sample for image generation pipelines
```

samples/cpp/image_generation/README.md
```diff
@@ -148,3 +149,63 @@ The resuling image is:
...
+./benchmark_image [OPTIONS]
+```
+Options:
+- `-t, --pipeline_type` (default: `"text2image"`): Pipeline type(text2image, image2image, inpainting).
+- `-m, --model`: Path to the model and tokenizers base directory.
+- `-p, --prompt` (default: `"The Sky is blue because"`): The prompt to generate text.
+- `--nw, --num_warmup` (default: `1`): Number of warmup iterations.
+- `-n, --num_iter` (default: `3`): Number of iterations.
+- `-d, --device` (default: `"CPU"`): Device to run the model on.
+- `--wh, --width` (default: `512`): The width of the output image.
```
ilya-lavrenov, line 168:
```suggestion
- `--w, --width` (default: `512`): The width of the output image.
```

samples/cpp/image_generation/README.md
```diff
@@ -148,3 +149,63 @@ The resuling image is:
...
+- `--ht, --height` (default: `512`): The height of the output image.
```
ilya-lavrenov, line 169:
```suggestion
- `--h, --height` (default: `512`): The height of the output image.
```

samples/cpp/image_generation/benchmark_image.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include "load_image.hpp"
+#include "progress_bar.hpp"
+
+inline void print_one_generate(ov::genai::ImageGenerationPerfMetrics& metrics, std::string prefix, int idx) {
+    std::string prefix_idx = "[" + prefix + "-" + std::to_string(idx) + "]";
+    std::cout << "\n";
+    std::cout << prefix_idx << " generate time: " << metrics.get_generate_duration()
+              << " ms, total infer time:" << metrics.get_all_infer_duration()
+              << " ms" << std::endl;
+    std::cout << prefix_idx << " encoder infer time: " << metrics.get_encoder_infer_duration() << " ms"<< std::endl;
```
ilya-lavrenov, line 19: Is encoder infer time assumes both text encoder and image encoders?

for image to image and inpainting we also run VAE encoder to convert initial image to latent space. I think we need different values for text encoder (runs in all types of pipelines) and image encoder (only when initial image is passed)

src/cpp/include/openvino/genai/image_generation/image_generation_perf_metrics.hpp
```diff
@@ -0,0 +1,52 @@
...
+#include <chrono>
+#include <map>
+#include <string>
+
+#include "openvino/genai/visibility.hpp"
+#include "openvino/genai/perf_metrics.hpp"
+
+namespace ov::genai {
+
+struct OPENVINO_GENAI_EXPORTS RawImageGenerationPerfMetrics {
```
ilya-lavrenov, line 16:
```suggestion
struct RawImageGenerationPerfMetrics {
```

src/cpp/include/openvino/genai/image_generation/image_generation_perf_metrics.hpp
```diff
@@ -0,0 +1,52 @@
...
+    float generate_duration; // duration of method generate(...), ms
+
+    MeanStdPair iteration_duration; // Mean-Std time of one generation iteration, ms
+    std::map<std::string, float> encoder_inference_duration; // inference durations for each encoder, ms
+    MeanStdPair unet_inference_duration; // inference duration for unet model, should be filled with zeros if we don't have unet, ms
+    MeanStdPair transformer_inference_duration; // inference duration for transformer model, should be filled with zeros if we don't have transformer, ms
+    float vae_encoder_inference_duration; // inference duration of vae_encoder model, should be filled with zeros if we don't use it, ms
+    float vae_decoder_inference_duration; // inference duration of vae_decoder model, ms
+
+    bool m_evaluated = false;
```
ilya-lavrenov, line 33: maybe need to hide this field into a `private` section? Looks like it's not useful for users.

src/cpp/src/image_generation/diffusion_pipeline.hpp
```diff
@@ -92,6 +92,13 @@ class DiffusionPipeline {
 
     virtual ov::Tensor decode(const ov::Tensor latent) = 0;
 
+    virtual ImageGenerationPerfMetrics get_performance_metrics() = 0;
+
+    void save_load_time(std::chrono::steady_clock::time_point start_time) {
+        auto stop_time = std::chrono::steady_clock::now();
+        m_load_time_ms = std::chrono::duration_cast<std::chrono::milliseconds>(stop_time - start_time).count();
```
ilya-lavrenov, line 99: should we store it directly in `m_perf_metrics.load_time` instead of `m_load_time_ms` ?

It this case `clean_up` should not clean load time

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -335,12 +341,16 @@ class FluxPipeline : public DiffusionPipeline {
                                height,
                                width};
         ov::Tensor latent, noise, proccesed_image, image_latents;
+        float encoder_duration = 0.0f;
```
ilya-lavrenov, line 344:
```suggestion
        float vae_encoder_duration = 0.0f;
```

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -488,7 +504,9 @@ class FluxPipeline : public DiffusionPipeline {
         m_latent_timestep = timesteps[0];
 
         ov::Tensor latents, processed_image, image_latent, noise;
-        std::tie(latents, processed_image, image_latent, noise) = prepare_latents(initial_image, m_custom_generation_config);
+        float vae_encoder_duration;
+        std::tie(latents, processed_image, image_latent, noise, vae_encoder_duration) = prepare_latents(initial_image, m_custom_generation_config);
+        m_perf_metrics.vae_encoder_inference_duration = vae_encoder_duration;
```
ilya-lavrenov, line 509: can `prepare_latents` write directly to `m_perf_metrics.vae_encoder_inference_duration` ? instead of returning this value from a function and manual assigning later

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -501,24 +519,41 @@ class FluxPipeline : public DiffusionPipeline {
...
+            auto infer_start = std::chrono::steady_clock::now();
             ov::Tensor noise_pred_tensor = m_transformer->infer(latents, timestep);
+            auto infer_duration = ov::genai::PerfMetrics::get_microsec(std::chrono::steady_clock::now() - infer_start);
+            m_perf_metrics.raw_metrics.transformer_inference_durations.emplace_back(MicroSeconds(infer_duration));
 
             auto scheduler_step_result = m_scheduler->step(noise_pred_tensor, latents, inference_step, m_custom_generation_config.generator);
             latents = scheduler_step_result["latent"];
 
+            auto step_ms = ov::genai::PerfMetrics::get_microsec(std::chrono::steady_clock::now() - step_start);
+            m_perf_metrics.raw_metrics.iteration_durations.emplace_back(MicroSeconds(step_ms));
```
ilya-lavrenov, line 534: I think iteration duration should also cover `blend_latents` which contribute to iteration time for inpainting use case (BTW, I see that for SD pipeline you measure this time as `step_time`, so we need to be aligned)

Ans also, we can measure `callback`, which is supposed to be lightweight, but anyway

src/cpp/src/image_generation/stable_diffusion_3_pipeline.hpp
```diff
@@ -273,25 +273,36 @@ class StableDiffusion3Pipeline : public DiffusionPipeline {
...
         if (m_t5_text_encoder) {
+            infer_start = std::chrono::steady_clock::now();
             text_encoder_3_output = m_t5_text_encoder->infer(prompt_3_str,
                                                              negative_prompt_3_str,
                                                              do_classifier_free_guidance(generation_config.guidance_scale),
                                                              generation_config.max_sequence_length);
+            auto infer_duration =
+                std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::steady_clock::now() - infer_start)
+                    .count();
+            m_perf_metrics.encoder_inference_duration["text_encode_3"] = infer_duration;
```
ilya-lavrenov, line 305: maybe we need to explicitly write `0` as `text_encoder_3` time in case it's not used?

src/cpp/include/openvino/genai/image_generation/image_generation_perf_metrics.hpp
```diff
@@ -0,0 +1,52 @@
...
+    RawImageGenerationPerfMetrics raw_metrics;
+
+    void clean_up();
+    void evaluate_statistics();
+
+    MeanStdPair get_unet_infer_meanstd();
+    MeanStdPair get_transformer_infer_meanstd();
+    MeanStdPair get_iteration_meanstd();
+    float get_encoder_infer_duration();
+    float get_decoder_infer_duration();
```
ilya-lavrenov, line 44:
```suggestion
    float get_vae_decoder_infer_duration();
    float get_vae_encoder_infer_duration();
```

src/cpp/include/openvino/genai/image_generation/image_generation_perf_metrics.hpp
```diff
@@ -0,0 +1,52 @@
...
+    float get_all_infer_duration();
```
ilya-lavrenov, line 45:
```suggestion
    float get_total_infer_duration();
```

samples/cpp/image_generation/README.md
```diff
@@ -8,6 +8,7 @@ There are several sample files:
  - [`heterogeneous_stable_diffusion.cpp`](./heterogeneous_stable_diffusion.cpp) shows how to assemble a heterogeneous txt2image pipeline from individual subcomponents (scheduler, text encoder, unet, vae decoder)
  - [`image2image.cpp`](./image2image.cpp) demonstrates basic usage of the image to image pipeline
  - [`inpainting.cpp`](./inpainting.cpp) demonstrates basic usage of the inpainting pipeline
+ - [`benchmark_image.cpp](./benchmark_image.cpp) demonstrates how to benchmark the text to image / image to image / inpainting pipeline
```
likholat, line 11:
```suggestion
 - [`benchmark_image.cpp`](./benchmark_image.cpp) demonstrates how to benchmark the text to image / image to image / inpainting pipeline
```

src/cpp/include/openvino/genai/image_generation/image_generation_perf_metrics.hpp
xufang-lisa: That means we also need to provide functions for setting these parameters, right?

ilya-lavrenov: hm, forgot about it.

Ok, let's keep fields as is

src/cpp/src/image_generation/flux_pipeline.hpp
xufang-lisa: `virtual std::tuple<ov::Tensor, ov::Tensor, ov::Tensor, ov::Tensor> prepare_latents(ov::Tensor initial_image, const ImageGenerationConfig& generation_config) const = 0;`
This function is defined as a const function, so `m_perf_metrics.vae_encoder_inference_duration` can not be written in `prepare_latents` .

samples/cpp/image_generation/README.md
xufang-lisa: `-h` refers to `help`, so the abbreviation of height here needs to be `ht`

src/cpp/src/image_generation/flux_pipeline.hpp
ilya-lavrenov: good point. I think `const` can be removed

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -448,6 +462,8 @@ class FluxPipeline : public DiffusionPipeline {
                         ov::Tensor initial_image,
                         ov::Tensor mask_image,
                         const ov::AnyMap& properties) override {
+        const auto gen_start = std::chrono::steady_clock::now();
+        m_perf_metrics.clean_up();
```
ilya-lavrenov, line 466: will it clean load time ?

xufang-lisa: It does, so another variable is needed to save load time or `clean_up` does not clean load time. Which of these two methods is better?

src/cpp/src/image_generation/flux_pipeline.hpp
xufang-lisa: updated

ilya-lavrenov (approved):

ilya-lavrenov (commented):
Could you please add running of these apps to GHA CI?
I think you can use some tiny random model like `katuni4ka/lcm-tiny-sd` to minimize execution time.

You can extend this workflow with extra steps https://github.com/openvinotoolkit/openvino.genai/blob/master/.github/workflows/lcm_dreamshaper_cpp.yml

likholat (approved):

samples/python/image_generation/benchmark_image_gen.py
```diff
@@ -0,0 +1,227 @@
...
+    warmup_metrics = []
+    for i in range(num_warmup):
+        pipe.generate(prompt, image_input, strength)
+        metrics = pipe.get_performance_metrics()
+        warmup_metrics.append(metrics)
+        print_one_generate(metrics, "warmup", i)
+    
+    iter_metrics = []
+    for i in range(num_iter):
+        image_tensor = pipe.generate(prompt, image_input, strength)
```
ilya-lavrenov, line 147:
```suggestion
        image_tensor = pipe.generate(prompt, image_input, strength=strength)
```

PR #1414, opened 2024-12-19
[test] Ensure that the first token generation is not included into TPOT

pavel-esir:
CVS-155098

### DIFF
--- a/src/cpp/src/perf_metrics.cpp
+++ b/src/cpp/src/perf_metrics.cpp
@@ -101,7 +101,7 @@ void PerfMetrics::evaluate_statistics(std::optional<TimePoint> start_time) {
 
         auto ttft = tok_times[0] - start_time_val;
         raw_metrics.m_times_to_first_token = std::vector<MicroSeconds>();
-        raw_metrics.m_times_to_first_token.emplace_back(ttft / batch_sizes[0]);
+        raw_metrics.m_times_to_first_token.emplace_back(ttft);
         num_generated_tokens = batch_sizes[0];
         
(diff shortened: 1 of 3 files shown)


src/cpp/src/perf_metrics.cpp
```diff
@@ -101,7 +101,7 @@ void PerfMetrics::evaluate_statistics(std::optional<TimePoint> start_time) {
 
         auto ttft = tok_times[0] - start_time_val;
         raw_metrics.m_times_to_first_token = std::vector<MicroSeconds>();
-        raw_metrics.m_times_to_first_token.emplace_back(ttft / batch_sizes[0]);
+        raw_metrics.m_times_to_first_token.emplace_back(ttft);
```
pavel-esir, line 104: If we have batch 10 and it takes 1 sec until the first token is generates then time to the first token is still 1 sec not 100 ms! Therefore i removed `/ batch_sizes[0]`

Wovchena: I agree, but should we notify llm_bench about the change?

pavel-esir: I asked @eaidova what she thinks. Let's wait her answer

eaidova: @peterchen-intel  @wgzintel  do you have objections? What the reason to divide first token latency on batch?

Wovchena (approved):

eaidova (approved):

src/cpp/src/perf_metrics.cpp
pavel-esir: As i see from blame, it was added by @ialbrecht.

PR #1419, opened 2024-12-20
remove redundant `.tolist()`

pavel-esir:
image

- To fix remove redundant `.tolist()` since it was already done above.


tools/llm_bench/task/text_generation.py
```diff
@@ -324,7 +324,7 @@ def token_printer():
         num,
         iter_data,
         tm_list.tolist(),
```
pavel-esir, line 326:
```suggestion
        tm_list,
```

pavel-esir: done

eaidova (approved):

PR #1476, opened 2025-01-03
Add a choice of how to end streaming from callback: STOP or CANCEL

sbalandi:


sbalandi:
TODO: add CANCEL for ContinuousBatching

sbalandi:
> TODO: add CANCEL for ContinuousBatching

done

samples/python/prompt_lookup_decoding_lm/prompt_lookup_decoding_lm.py
```diff
@@ -5,11 +5,11 @@
...
-def streamer(subword): 
-        print(subword, end='', flush=True) 
-        # Return flag corresponds whether generation should be stopped. 
-        # False means continue generation. 
-        return False
+def streamer(subword):
+    print(subword, end='', flush=True) 
+    # Return flag corresponds whether generation should be stopped. 
+    # False means continue generation. 
+    return False
```
ilya-lavrenov, line 12: BTW, should we also support callback w/o return value?

E.g. when user don't care about any stop / cancellation

samples/cpp/chat_sample/chat_sample.cpp
```diff
@@ -15,11 +15,13 @@ int main(int argc, char* argv[]) try {
...
     config.max_new_tokens = 100;
-    std::function<bool(std::string)> streamer = [](std::string word) { 
+
+    std::function<ov::genai::StreamerRunningStatus(std::string)> streamer = [](std::string word) {
         std::cout << word << std::flush;
         // Return flag corresponds whether generation should be stopped.
         // false means continue generation.
-        return false; 
+
+        return ov::genai::StreamerRunningStatus::RUNNING;
```
ilya-lavrenov, line 24:
```suggestion
        return ov::genai::StreamingStatus::CONTINUE;
```

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -30,6 +31,9 @@ struct EncodedGenerationResult {
 
     // Status of generation
     GenerationStatus m_status = GenerationStatus::RUNNING;
+
+    // Status of streaming
+    StreamerRunningStatus m_streaming_status = ov::genai::StreamerRunningStatus::UNDEF;
```
ilya-lavrenov, line 36: maybe we can extend `GenerationStatus` ? E.g. DROPPED_BY_HANDLE means STOP in its current implementation, while for CANCEL we can add a new value.

BTW, looks like we can drop `DROPPED_BY_PIPELINE` as unused.

src/python/openvino_genai/__init__.py
```diff
@@ -15,7 +15,7 @@
     RawPerfMetrics,
     PerfMetrics,
     StreamerBase,
-    get_version,
```
ilya-lavrenov, line 18: can we keep it?

src/cpp/src/text_callback_streamer.hpp
```diff
@@ -11,12 +11,17 @@ namespace genai {
 
 class TextCallbackStreamer: public StreamerBase {
 public:
+    StreamerRunningStatus streaming_status = StreamerRunningStatus::UNDEF;
```
ilya-lavrenov, line 14: as I see `StreamerBase` already contains this field ?

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,16 +4,28 @@
...
 namespace genai {
 
+enum class StreamerRunningStatus {
+    UNDEF = 0, // Streaming is not run
+    RUNNING = 1, // Continue to run of inference
+    STOP = 2, // Stop generation, keep history as is, KV cache includes last request and generated tokens
+    CANCEL = 3 // Stop generate, drop last prompt and all generated tokens from history, KV cache include history but last step
+};
+
+using CallbackTypeVariant = std::variant<bool, StreamerRunningStatus>;
```
ilya-lavrenov, line 19:
```suggestion
using CallbackTypeVariant = std::variant<void, bool, StreamerRunningStatus>;
```
to support callback which just "prints"

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -22,6 +34,10 @@ class OPENVINO_GENAI_EXPORTS StreamerBase {
     /// @brief end is called at the end of generation. It can be used to flush cache if your own streamer has one
     virtual void end() = 0;
 
+    virtual StreamerRunningStatus get_finish_streaming_reason() {
```
ilya-lavrenov, line 37:
```suggestion
    StreamingStatus get_streaming_status() {
```

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -373,10 +373,10 @@ class ContinuousBatchingPipeline:
     def add_request(self, request_id: int, prompt: str, generation_config: GenerationConfig) -> GenerationHandle:
         ...
     @typing.overload
-    def generate(self, input_ids: list[openvino._pyopenvino.Tensor], generation_config: list[GenerationConfig], streamer: typing.Callable[[str], bool] | StreamerBase | None = None) -> list[EncodedGenerationResult]:
+    def generate(self, input_ids: list[openvino._pyopenvino.Tensor], generation_config: list[GenerationConfig], streamer: typing.Callable[[str], bool] | typing.Callable[[str], ...] | StreamerBase | None = None) -> list[EncodedGenerationResult]:
         ...
     @typing.overload
-    def generate(self, prompts: list[str], generation_config: list[GenerationConfig], streamer: typing.Callable[[str], bool] | StreamerBase | None = None) -> list[GenerationResult]:
+    def generate(self, prompts: list[str], generation_config: list[GenerationConfig], streamer: typing.Callable[[str], bool] | typing.Callable[[str], ...] | StreamerBase | None = None) -> list[GenerationResult]:
```
ilya-lavrenov, line 379: should we propagate StreamingStatus to Python API? to use enum instead of str

ilya-lavrenov (commented):
Please, add tests for new functionality.

src/cpp/include/openvino/genai/generation_handle.hpp
ilya-lavrenov: More thoughts: 
- maybe we should deprecated drop() method and introduce stop() instead
- similarly for GenerationStatus 
- and extend both GenerationHandle and GenerationStatus with cancel() functionality

In this case CB and LLM pipelines logic / naming will be aligned

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,16 +4,28 @@
...
+using CallbackTypeVariant = std::variant<bool, StreamerRunningStatus>;
+
 /**
  * @brief base class for streamers. In order to use inherit from from this class and implement put, and methods
  *
  * @param m_tokenizer tokenizer
  */
 class OPENVINO_GENAI_EXPORTS StreamerBase {
+protected:
+    StreamerRunningStatus streaming_finish_status = StreamerRunningStatus::UNDEF;
```
Wovchena, line 28:
```suggestion
    StreamerRunningStatus m_streaming_finish_status = StreamerRunningStatus::UNDEF;
```

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -187,6 +187,9 @@ class ov::genai::VLMPipeline::VLMPipelineImpl {
         SequenceGroup::Ptr sequence_group = std::make_shared<SequenceGroup>(request_id, prompt_ids, generation_config, block_size);
         requests.push_back(sequence_group);
 
+        OPENVINO_ASSERT((!m_is_chat_conversation || !std::get_if<std::function<StreamerRunningStatus(std::string)>>(&streamer)),
+                         "For chat mode, please, use Steamer as StreamerBase class or as callback with a bool return value.");
```
Wovchena, line 191: Why is `StreamerRunningStatus` return value is prohibited for chat? You wrap it with `TextCallbackStreamer` below

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,16 +4,28 @@
...
 
 #include "openvino/genai/tokenizer.hpp"
+#include <variant>
 
 namespace ov {
 namespace genai {
 
+enum class StreamerRunningStatus {
+    UNDEF = 0, // Streaming is not run
+    RUNNING = 1, // Continue to run of inference
```
Wovchena, line 14: RUNNING and UNDEF seem to be equivalent. In that case you should keep only one of them. Moreover callback should never return UNDEF, so merging them fixes the API.

Wovchena (changes requested):

src/cpp/include/openvino/genai/generation_handle.hpp
sbalandi: moved to GenerationStatus

src/cpp/include/openvino/genai/streamer_base.hpp
sbalandi: removed it , moved to GenerationStatus

src/cpp/include/openvino/genai/streamer_base.hpp
sbalandi: renamed

src/cpp/src/text_callback_streamer.hpp
sbalandi: removed

src/python/openvino_genai/__init__.py
sbalandi: added

src/cpp/include/openvino/genai/streamer_base.hpp
Wovchena: Merging with `GenerationStatus` allows a callback to return FINISHED and IGNORED which aren't related to this. I'd guess https://github.com/openvinotoolkit/openvino.genai/pull/1476#discussion_r1909969883 was about aligning API but not merging. @ilya-lavrenov, is that so?

ilya-lavrenov: yes

sbalandi: updated

samples/python/text_generation/chat_sample.py
```diff
@@ -10,8 +10,7 @@ def streamer(subword):
     print(subword, end='', flush=True)
     # Return flag corresponds whether generation should be stopped.
     # False means continue generation.
```
Wovchena, line 12:
```suggestion
    # Return flag corresponds whether generation should be stopped.
```
Same for samples/cpp/text_generation/chat_sample.cpp.

There are also
1. samples/cpp/text_generation/multinomial_causal_lm.cpp
2. samples/cpp/text_generation/prompt_lookup_decoding_lm.cpp
3. samples/cpp/text_generation/speculative_decoding_lm.cpp

samples/python/text_generation/prompt_lookup_decoding_lm.py
```diff
@@ -5,11 +5,11 @@
...
-def streamer(subword): 
-        print(subword, end='', flush=True) 
-        # Return flag corresponds whether generation should be stopped. 
-        # False means continue generation. 
-        return False
+def streamer(subword):
+    print(subword, end='', flush=True) 
+    # Return flag corresponds whether generation should be stopped. 
+    # False means continue generation. 
+    return False
```
Wovchena, line 12:
```suggestion
    return openvino_genai.StreamerRunningStatus.RUNNING
```

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -19,7 +19,7 @@ namespace ov {
 namespace genai {
 
 // Return flag corresponds whether generation should be stopped: false means continue generation, true means stop.
-using StreamerVariant = std::variant<std::function<bool(std::string)>, std::shared_ptr<StreamerBase>, std::monostate>;
+using StreamerVariant = std::variant<std::function<bool(std::string)>, std::function<StreamerRunningStatus(std::string)>, std::shared_ptr<StreamerBase>, std::monostate>;
```
Wovchena, line 22: Add a deprecation notice.
You can also print a warning from `create_streamer()`.

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,24 +4,39 @@
...
 public:
+    StreamerRunningStatus m_streaming_finish_status = StreamerRunningStatus::RUNNING;
     /// @brief put is called every time new token is decoded,
     /// @return bool flag to indicate whether generation should be stopped, if return true generation stops
     virtual bool put(int64_t token) = 0;
 
     /// @brief end is called at the end of generation. It can be used to flush cache if your own streamer has one
     virtual void end() = 0;
 
+    StreamerRunningStatus get_streaming_status() {
```
Wovchena, line 36: Explain in docstrings that `get_streaming_status()` is called by the pipelines and `m_streaming_finish_status` is supposed to be overwritten by `put()` to provide a more detailed streaming status compared to boolean return value.

Wovchena, line 36: `StreamerBase` is expected to be inherited. I think you should mark `get_streaming_status()` as virtual to provide the full control over the class. It would be annoying to debug if someone overrides this function but `LLMPipeline` still calls the function from the parent class using base pointer. Plus that gives you an excuse for introducing `get_streaming_status()` instead of just reading the member field.

Wovchena (changes requested):

samples/python/text_generation/chat_sample.py
sbalandi: updated

samples/python/text_generation/prompt_lookup_decoding_lm.py
sbalandi: updated

src/cpp/include/openvino/genai/streamer_base.hpp
sbalandi: added

sbalandi: added

src/cpp/include/openvino/genai/llm_pipeline.hpp
sbalandi: I have added comment here and OPENVINO_DEPRECATED to create_streamer function. Warning seems more logical to me, but it will trigger on each generate and it looks too much. What do you think is better way ?

Wovchena (changes requested):

src/cpp/src/utils.hpp
```diff
@@ -81,6 +93,7 @@ Config from_config_json_if_exists(const std::filesystem::path& models_path, cons
     return std::filesystem::exists(config_file_path) ? Config{config_file_path} : Config{};
 }
 
+OPENVINO_DEPRECATED("Callback streamer with `bool` return value is depricated. Please, use `ov::genai::StreamerRunningStatus` instead.")
```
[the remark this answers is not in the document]
sbalandi: removed

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,16 +4,29 @@
 #pragma once
 
 #include "openvino/genai/tokenizer.hpp"
+#include "openvino/genai/generation_handle.hpp"
+#include <variant>
 
 namespace ov {
 namespace genai {
 
+enum class StreamerRunningStatus {
```
ilya-lavrenov, line 13:
```suggestion
enum class StreamingStatus {
```
@Wovchena @sbalandi what do you think?

src/cpp/src/generation_handle.cpp
```diff
@@ -9,32 +9,44 @@
...
+    m_generation_stream->stop();
+}
+
+void GenerationHandleImpl::cancel() {
+    m_generation_stream->cancel();
 }
 
 std::unordered_map<uint64_t, GenerationOutput> GenerationHandleImpl::back() {
-    OPENVINO_ASSERT(!is_dropped(), "GenerationHandle cannot be used after it is dropped.");
+    OPENVINO_ASSERT(!is_stopped(), "GenerationHandle cannot be used after it is stopped.");
```
ilya-lavrenov, line 44:
```suggestion
    OPENVINO_ASSERT(!is_stopped() && !is_canceled(), "GenerationHandle cannot be used after it is stopped / canceled.");
```
the same in other places.

src/cpp/src/icontinuous_batching.cpp
```diff
@@ -110,6 +110,10 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::generate(
         });
     }
 
+    // if streaming was canceled, prompt/answer of current step shouldn't be presented in history, so let's remove prompt from history
+    if (m_is_chat_conversation && !encoded.empty() && encoded[0].m_status == ov::genai::GenerationStatus::CANCEL)
```
ilya-lavrenov, line 114: why do we need `!encoded.empty()` ? I suppose `encoded` always has the same size as number of inputs

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,16 +4,29 @@
...
  *
  * @param m_tokenizer tokenizer
  */
 class OPENVINO_GENAI_EXPORTS StreamerBase {
+protected:
+    StreamerRunningStatus m_streaming_finish_status = StreamerRunningStatus::RUNNING;
+
 public:
     /// @brief put is called every time new token is decoded,
     /// @return bool flag to indicate whether generation should be stopped, if return true generation stops
```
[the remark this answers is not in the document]
sbalandi: have added new method, called it `write` . `put_token` good for one token, but for chunk `put_chunk` is used now and I'm not sure chunk is common name or it's okay change it to `tokens` ? @as-suvorov what do you think ?
Also maybe `post` is possible ?

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -88,10 +91,19 @@ class OPENVINO_GENAI_EXPORTS GenerationHandleImpl {
...
+    bool is_cancelled();
+
+    OPENVINO_DEPRECATED("Please, use `stop()` instead of `drop()`.")
     void drop();
 
+    void stop();
+
+    void cancel();
+
+    GenerationOutputs back();
```
ilya-lavrenov, line 106: please, drop this method 

it was explicitly removed in another PR

src/python/py_utils.cpp
```diff
@@ -316,20 +316,33 @@ ov::genai::StreamerVariant pystreamer_to_streamer(const PyBindStreamerVariant& p
...
+                        result = StreamingStatus::STOP;
+                    }
+                }
+
+                return result;
+            };
+            streamer = callback_wrapped;
+        },
+        [&streamer](std::shared_ptr<StreamerBase> streamer_cls){
+            std::cout << "streamer_cls " << std::endl;
```
ilya-lavrenov, line 342: do we need this `cout`?

sbalandi: debug thing, sorry, removed

src/cpp/include/openvino/genai/generation_handle.hpp
sbalandi: removed

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -18,8 +18,10 @@
 namespace ov {
 namespace genai {
 
-// Return flag corresponds whether generation should be stopped: false means continue generation, true means stop.
-using StreamerVariant = std::variant<std::function<bool(std::string)>, std::shared_ptr<StreamerBase>, std::monostate>;
+// Return flag corresponds whether generation should be stopped. It could be:
+// ov::genai::StreamingStatus flag, RUNNING means continue generation, STOP means stop generation, CANCEL means stop generation and remove last propmt and answer from history
+// *DEPRECATED* bool flag, false means continue generation, true means stop. Please, use `ov::genai::StreamingStatus` instead.
+using StreamerVariant = std::variant<std::function<bool(std::string)>, std::function<StreamingStatus(std::string)>, std::shared_ptr<StreamerBase>, std::monostate>;
```
Wovchena, line 24: Is void returning function missing

src/cpp/src/text_callback_streamer.cpp
```diff
@@ -46,7 +46,21 @@ bool TextCallbackStreamer::put(int64_t token) {
...
+    StreamingStatus status = StreamingStatus::RUNNING;
+    if (auto res = std::get_if<StreamingStatus>(&callback_status)) {
+        status = *res;
+    } else if (auto res = std::get_if<bool>(&callback_status)) {
+        status = *res ? StreamingStatus::STOP : StreamingStatus::RUNNING;
+    } else if (auto res = std::get_if<std::monostate>(&callback_status)) {
+        status = StreamingStatus::RUNNING;
+    }
+
+    return status;
```
Wovchena, line 63:
```suggestion
    if (auto res = std::get_if<StreamingStatus>(&callback_status)) {
        return *res;
    if (auto res = std::get_if<bool>(&callback_status)) {
        return *res ? StreamingStatus::STOP : StreamingStatus::RUNNING;
    return StreamingStatus::RUNNING;
```
Same for utils and whisper bindings

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -11,14 +11,17 @@
...
+
 enum class GenerationStatus {
     RUNNING = 0, // Default status for ongoing generation
     FINISHED = 1, // Status set when generation has been finished
     IGNORED = 2, // Status set when generation run into out-of-memory condition and could not be continued
-    DROPPED_BY_PIPELINE = 3, // Currently not used, TODO: implement abort functionality
-    DROPPED_BY_HANDLE = 4 // Status set when generation handle is dropped
+    CANCEL = 3, // Status set when generation handle is cancelled
+    STOP = 4, // Status set when generation handle is stopped
+    DROPPED_BY_HANDLE OPENVINO_ENUM_DEPRECATED("Please, use `STOP` instead of `DROPPED_BY_HANDLE`.") = GenerationStatus::STOP // Status set when generation handle is dropped
```
as-suvorov, line 21: Should we add something like `Support will be removed in {version}(2026.0?)`

src/cpp/include/openvino/genai/streamer_base.hpp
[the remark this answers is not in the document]
as-suvorov: I think `write` is fine

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,20 +4,38 @@
...
  *
  * @param m_tokenizer tokenizer
  */
 class OPENVINO_GENAI_EXPORTS StreamerBase {
 public:
-    /// @brief put is called every time new token is decoded,
+    /// @brief put is called every time new token is decoded. Deprecated. Please, use write instead.
     /// @return bool flag to indicate whether generation should be stopped, if return true generation stops
-    virtual bool put(int64_t token) = 0;
+    OPENVINO_DEPRECATED("Please, use `write()` instead of `put()`.")
```
as-suvorov, line 29: Should deprecation messages have support removal version?

src/cpp/include/openvino/genai/streamer_base.hpp
```diff
@@ -4,20 +4,38 @@
...
 namespace ov {
 namespace genai {
 
+enum class StreamingStatus {
+    RUNNING = 0, // Continue to run of inference
+    STOP = 1, // Stop generation, keep history as is, KV cache includes last request and generated tokens
+    CANCEL = 2 // Stop generate, drop last prompt and all generated tokens from history, KV cache include history but last step
+};
+
+using CallbackTypeVariant = std::variant<bool, StreamingStatus, std::monostate>;
```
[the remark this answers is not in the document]
sbalandi: moved it to src/cpp/src/text_callback_streamer.hpp. Do you mean that or should i move it outside ov::genai namespace ?

src/cpp/include/openvino/genai/llm_pipeline.hpp
sbalandi: I am not sure. It is not possible to return void from c++ streamer now and I didn't add it. But it is possible to return None form python and it will stay that way. Should I add handling for void for c++ ?

sbalandi:
- I was wrong about tests for `StreamerBase` with exactly `bool` return value. There are tests with `put()` with None to return. I have added with bool and with status.
- It was added two tests: `test_callback_terminate_by_bool `and `test_callback_terminate_by_status`, which checks that output generated seq stops on exactly the token that the user requested via the streamer, but there is one extra token in generation sequence now. I would remove it in another PR, if it's required (need to be discussed)

src/cpp/include/openvino/genai/llm_pipeline.hpp
Wovchena:
```C++
StreamingStatus TextCallbackStreamer::set_streaming_status(CallbackTypeVariant callback_status) {
    if (auto res = std::get_if<StreamingStatus>(&callback_status))
        return *res;
    if (auto res = std::get_if<bool>(&callback_status))
        return *res ? StreamingStatus::STOP : StreamingStatus::RUNNING;
    return StreamingStatus::RUNNING;
}
```
(or the version that was before my suggestion) handles the case when `callback_status` is a monostate by returning `StreamingStatus::RUNNING`, so I interpreted this as `void` is also allowed. If `void` isn't actually supported,
```C++
    if (auto res = std::get_if<bool>(&callback_status))
        return *res ? StreamingStatus::STOP : StreamingStatus::RUNNING;
    return StreamingStatus::RUNNING;
```
should become
```C++
    return std::get<bool>(&callback_status) ? StreamingStatus::STOP : StreamingStatus::RUNNING;
```

As for the Python, I think `None` is interpreted as `0` which corresponds to `StreamingStatus::RUNNING`, so it's not possible at all to return analog of void from python.

sbalandi: removed std::monostate  from CallbackTypeVariant and update string with variants

src/python/py_whisper_pipeline.cpp
```diff
@@ -247,13 +262,23 @@ class ConstructableChunkStreamer : public ChunkStreamerBase {
...
+                                return static_cast<ChunkStreamerVariant>([py_callback](std::string subword) -> ov::genai::StreamingStatus {
                                     auto py_str = PyUnicode_DecodeUTF8(subword.data(), subword.length(), "replace");
-                                    return py_callback(py::reinterpret_borrow<py::str>(py_str));
+                                    std::optional<uint16_t> callback_output = py_callback(py::reinterpret_borrow<py::str>(py_str));
+                                    if (callback_output.has_value()) {
+                                        if (*callback_output == (uint16_t)StreamingStatus::RUNNING)
+                                            return StreamingStatus::RUNNING;
+                                        else if (*callback_output == (uint16_t)StreamingStatus::CANCEL)
+                                            return StreamingStatus::CANCEL;
+                                        std::cout << " AAA STOP " << std::endl;
```
Wovchena, line 277:
```suggestion
```

Wovchena (changes requested):

src/cpp/include/openvino/genai/llm_pipeline.hpp
Wovchena: Whisper and utils bindings have `std::optional<uint16_t> callback_output = py_callback(py::reinterpret_borrow<py::str>(py_str));` It should be aligned with this now, assuming that `None` is interpreted as 0 by pybind.

src/python/py_whisper_pipeline.cpp
sbalandi: sorry, debug print, remove

src/cpp/include/openvino/genai/llm_pipeline.hpp
sbalandi: when Python callback `return None` it is treated as `callback_output` has no value

Wovchena (approved):

ilya-lavrenov (approved):

ilya-lavrenov:
Please, resolve conflicts.

PR #1515, opened 2025-01-09
[LLM] [NPU] StaticLLMPipeline: Compiler DQ update

smirnov-alexey:
Depends on https://github.com/openvinotoolkit/openvino/pull/28316
Related to https://github.com/openvinotoolkit/openvino/pull/28343


src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -526,6 +520,7 @@ ov::AnyMap get_default_prefill_config(const std::shared_ptr<ov::Model>& model,
     }
     if (npudesc.has_value() && npudesc->compiler_dq) {
         config.emplace("NPUW_DQ_FULL", "NO");
+        config.emplace("NPU_COMPILATION_MODE_PARAMS", "enable-weights-dynamic-dequantization=true");
```
smirnov-alexey, line 523: @TolyaTalamanov Could this property be present in the config at this point?
Could it be overwritten later by someone?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -516,6 +515,10 @@ ov::AnyMap get_default_prefill_config(const std::shared_ptr<ov::Model>& model,
     auto config = get_default_common_config(model);
     if (is_cw_compressed(model)) {
         config.emplace("NPUW_DQ", "YES");
+        if (npudesc.has_value() && npudesc->compiler_dq) {
+            config.emplace("NPUW_DQ_FULL", "NO");
+            config.emplace("NPU_COMPILER_DYNAMIC_QUANTIZATION", true);
+        }
```
dmatveev, line 521: Why do you enable it for CW compressed models only? The whole point of this feature was to make it for prefill.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -539,15 +539,16 @@ ov::AnyMap get_default_generate_config(const std::shared_ptr<ov::Model>& model,
     }
     // NB: Unconditionally set for generation model
     config.emplace("NPUW_DQ", "YES");
+    if (npudesc.has_value() && npudesc->compiler_dq) {
+        config.emplace("NPUW_DQ_FULL", "NO");
+        config.emplace("NPU_COMPILER_DYNAMIC_QUANTIZATION", true);
+    }
```
dmatveev, line 545: You certainly don't need to make it twice, just do this in the shared section of the config (so it goes to both prefill and kvcache).

dmatveev: Also, when you switch to compiler DQ, you'll have to disable DCOFF since otherwise DCOFF will be applied to this model and it will run in FP16.

dmatveev (changes requested):

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -475,11 +475,10 @@ std::optional<NPUDesc> extract_npu_descriptor(ov::Core& core) {
     }
     const auto arch = core.get_property("NPU", ov::device::architecture);
     const auto max_tiles = core.get_property("NPU", ov::intel_npu::max_tiles);
-
     bool compiler_dq = false;
-    const auto device_caps = core.get_property("NPU", ov::device::capabilities);
-    if (std::find(device_caps.begin(), device_caps.end(),
-                  "COMPILER_DYNAMIC_QUANTIZATION") != device_caps.end()) {
+    const auto supported_properties = core.get_property("NPU", ov::supported_properties);
+    if (std::find(supported_properties.begin(), supported_properties.end(),
```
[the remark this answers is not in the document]
TolyaTalamanov: Seems already simple enough

TolyaTalamanov:
> Propose to encapsulate it like this:
> 
> ```
> void enable_dq() {
>     if (npudesc.has_value() && npudesc->compiler_dq) {
>         "NPUW_DQ_FULL": "NO"
>         "NPU_COMPILER_DYNAMIC_QUANTIZATION": true
>     } else {
>         "NPUW_DQ": "YES"
>     }
> }
> ```
> 
> We need to call `enable_dq` in all places where previously was just `NPUW_DQ: YES`
> 
> As for the logic to enable dq or not, it was the following:
> 
> 1. For prefill model DQ is enabled only when model is channel-wise compressed
> 2. For generation model DQ is always enabled.
>    @dmatveev Could you confirm this, please?
> 
> If it should be enabled for both `CW` and `GQ` models we can call `enable_dq()` function uncoditionaly for both configs.

Should be done here: https://github.com/openvinotoolkit/openvino.genai/blob/fa76cf716eaf80f361c0a468898068b649530fdf/src/cpp/src/llm_pipeline_static.cpp#L502-L511

dmatveev:
@TolyaTalamanov I updated our internal guide (the one you've used before)

src/cpp/src/llm_pipeline_static.cpp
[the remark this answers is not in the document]
dmatveev: How it is a sub-string search, looks more like a container. At least std::find works on it (this `auto` sometimes makes things harder to understand)

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -475,17 +475,16 @@ std::optional<NPUDesc> extract_npu_descriptor(ov::Core& core) {
...
+                  "NPU_COMPILER_DYNAMIC_QUANTIZATION") != supported_properties.end()) {
         compiler_dq = true;
     }
     return std::make_optional(NPUDesc{arch, max_tiles, compiler_dq});
 }
 
-ov::AnyMap get_baseline_common_config() {
+ov::AnyMap get_baseline_common_config(bool enable_compiler_dq) {
```
dmatveev, line 487: Please pass the `NPUDesc` here instead - there's a request to get rid of "high-precision" options in the future as well.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -497,11 +496,23 @@ ov::AnyMap get_baseline_common_config() {
         { "NPUW_SLICE_OUT", "YES" },
         { "NPUW_FUNCALL_ASYNC", "YES" }
     };
+    if (enable_compiler_dq) {
+        config.emplace("NPUW_DQ", "YES");
+        config.emplace("NPUW_DQ_FULL", "NO");
+        config.emplace("NPU_COMPILER_DYNAMIC_QUANTIZATION", true);
```
dmatveev, line 502: "YES", otherwise all our past "YES" or "NO" should be `true` or `false`

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -497,11 +496,23 @@ ov::AnyMap get_baseline_common_config() {
...
+        config.erase("NPUW_DCOFF_TYPE");
+        config.erase("NPUW_DCOFF_SCALE");
+    }
     return config;
 }
 
-ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model) {
-    auto config = get_baseline_common_config();
+bool enable_compiler_dq(const std::optional<NPUDesc>& npudesc) {
```
dmatveev, line 509:
```suggestion
bool have_compiler_dq(const std::optional<NPUDesc>& npudesc) {
```
`enable_` may look ambiguous depending on the context (as it already does in the code below)

dmatveev (commented):
It gets more and more complex.

If it works, let's keep it like this, then propagate to the LLMCompiledModel (down to NPUW), remove it here, and refactor/rethink there.

@TolyaTalamanov please be sure to provide your review here as well.

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -497,11 +496,23 @@ ov::AnyMap get_baseline_common_config() {
...
+    return npudesc.has_value() && npudesc->compiler_dq;
```
TolyaTalamanov, line 510: What's the purpose of this function? Why don't just use `npudesc.has_value() && npudesc->compiler_dq` straightaway

TolyaTalamanov (commented):
The whole point was to encapsulate logic of compiler and non-compiler DQ into function and where the DQ is needed just call that function

TolyaTalamanov (approved):
`enable_compiler_dq` seems to be redundant

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: Done

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: Done

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: Done

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: Done

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: Removed this function

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -513,40 +521,39 @@ ov::AnyMap get_default_common_config(const std::shared_ptr<ov::Model>& model) {
...
+    auto config = get_default_common_config(model, npudesc);
     if (npudesc.has_value() &&
         npudesc->arch == "4000" &&
         npudesc->max_tiles != -1) {
         config.emplace("NPU_DPU_GROUPS", npudesc->max_tiles);
     }
-    if (npudesc.has_value() && npudesc->compiler_dq) {
-        config.emplace("NPUW_DQ_FULL", "NO");
+    // Specify NPUW DQ if Compiler DQ is not enabled
+    if (!npudesc.has_value() || !npudesc->compiler_dq) {
```
dmatveev, line 531: Not gonna lie @TolyaTalamanov, it was much better WITH that tiny one-liner than without that.

dmatveev (approved):

dmatveev:
Let's wait for the testing results

PR #1518, opened 2025-01-09
[Tokenizers] add max_lengh parametrisation to encode

pavel-esir:
Works in collaboration with tokenizers changes https://github.com/openvinotoolkit/openvino_tokenizers/pull/362
Ticket: CVS-157356, CVS-159924


src/cpp/src/tokenizer.cpp
```diff
@@ -109,17 +111,47 @@ class Tokenizer::TokenizerImpl {
...
     std::string m_chat_template = {};
 
+    std::pair<int, int> get_padding_values(std::string padding_mode, size_t max_length) {
+        if (padding_mode == "truncate") {
+            return {max_length, std::numeric_limits<int32_t>::max()};
+        } else if (padding_mode == "longest") {
+            return {std::numeric_limits<int32_t>::max(), std::numeric_limits<int32_t>::max()};
+        } else if (padding_mode == "max_length") {
+            return {std::numeric_limits<int32_t>::max(), max_length};
+        } else if (padding_mode == "do_not_pad") {
```
ilya-lavrenov, line 121: why is string, but not enum?

src/python/py_tokenizer.cpp
```diff
@@ -44,21 +44,35 @@ void init_tokenizer(py::module_& m) {
             return std::make_unique<ov::genai::Tokenizer>(tokenizer_path, kwargs_properties);
         }), py::arg("tokenizer_path"), py::arg("properties") = ov::AnyMap({}))
 
-        .def("encode", [](Tokenizer& tok, std::vector<std::string>& prompts, bool add_special_tokens) {
+        .def("encode", [](Tokenizer& tok, std::vector<std::string>& prompts, 
+                          bool add_special_tokens, 
+                          size_t max_length,
+                          std::string padding_mode) {
```
ilya-lavrenov, line 50: why are not these parameters are `kwargs` ?

src/cpp/src/tokenizer.cpp
```diff
@@ -109,17 +111,47 @@ class Tokenizer::TokenizerImpl {
...
+            // bahves exactly as longest
```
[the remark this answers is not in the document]

src/cpp/src/tokenizer.cpp
pavel-esir: done with enum

tests/python_tests/test_tokenizer.py
```diff
@@ -236,6 +236,26 @@ def test_encode_decode_with_special_tokens_option(add_special_tokens, skip_speci
...
+def test_padding(add_special_tokens, max_length, pad_mode, prompt):
+    import numpy as np
+    model_descr = get_chat_models_list()[0]
+    model_id, path, hf_tokenizer, model_opt, ov_pipe = read_model((model_descr[0], model_descr[1] / '_test_chat'))
+    genai_tokenzier = ov_pipe.get_tokenizer()
+    
+    # Calling encode with 'add_special_tokens' will set state flag.
+    ov_res = genai_tokenzier.encode(prompt, add_special_tokens=add_special_tokens, max_length=max_length, padding_mode=pad_mode).input_ids.data
+    hf_res = hf_tokenizer(prompt, return_tensors="np", add_special_tokens=add_special_tokens, max_length=max_length, padding=pad_mode)["input_ids"]
+    assert np.all(ov_res == hf_res)
```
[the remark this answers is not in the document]
pavel-esir: Thanks for noticing. Will add comparison with `attention_mask` as well.

pavel-esir:
Currently it hangs because of incorrect default truncation. Trying to fix it, but mainly code is ready for review

src/README.md
```diff
@@ -396,6 +396,34 @@ print(f'Median from token to token duration: {np.median(durations):.2f} ms')
...
+tokenizer = ov_genai.Tokenizer(models_path)
+
+// Or from LLMPipeline.
+pipe = ov_genai.LLMPipeline(models_path, "CPU")
+tokenizer = pipe.get_tokenizer()
+```
+
+`Tokenizer` has `encode` and `decode` methods which support the following arguments: `add_special_tokens`, `skip_special_tokens`, `padding_mode`, `max_length` arguments.
+
+For padding use
```
pavel-esir, line 426: TODO: finish this readme

src/cpp/src/make_tokenizer_stateful.hpp
```diff
@@ -76,6 +96,8 @@ class MakeVocabDecoderSatateful : public ov::pass::ModelPass {
 
 const std::string ADD_SPECIAL_TOKENS_VAR_ID = "add_special_tokens";
 const std::string SKIP_SPECIAL_TOKENS_VAR_ID = "skip_special_tokens";
+const std::string MAX_PAD_LENGTH_VAR_ID = "max_pad_length";
+const std::string MAX_TRUNCATION_LENGTH_VAR_ID = "max_pad_length";
```
ilya-lavrenov, line 100: values should not defined in .hpp files, but only declared via `extern`. Definition should be moved to .cpp

Otherwise, if I include such header in other .cpp files, I will have multiple definitions when link final OpenVINO GenAI library.

src/cpp/src/tokenizer.cpp
```diff
@@ -109,17 +112,46 @@ class Tokenizer::TokenizerImpl {
...
+                return {max_length, 0};
+            case PaddingMode::LONGEST:
+                return {std::numeric_limits<int32_t>::max(), 0};
+            case PaddingMode::MAX_LENGTH:
+                return {std::numeric_limits<int32_t>::max(), max_length};
+            case PaddingMode::DO_NOT_PAD:
+                // behaves exactly as longest
+                // TODO: need to find a way to disable padding automatically so that it will match to HF.
+                return {std::numeric_limits<int32_t>::max(), 0};
+            default:
```
ilya-lavrenov, line 127: NONE is missed here. Do we need NONE at all?

src/cpp/src/tokenizer.cpp
```diff
@@ -206,6 +249,9 @@ class Tokenizer::TokenizerImpl {
         if (ov_tokenizer) {
             ov::pass::Manager manager;
             manager.register_pass<MakeCombineSegmentsSatateful>();
+            manager.register_pass<MakePaddingSatateful>();
+            manager.register_pass<MakeTruncationSatateful>();
+            manager.register_pass<ov::pass::VisualizeTree>("after.svg");
```
ilya-lavrenov, line 254: it's used only for debug, please, remove when code is ready

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1666,12 +1666,12 @@ class Tokenizer:
         Decode a batch of tokens into a list of string prompt.
         """
     @typing.overload
-    def encode(self, prompts: list[str], add_special_tokens: bool = True) -> TokenizedInputs:
+    def encode(self, prompts: list[str], add_special_tokens: bool = True, max_length: int = 2147483647, padding_mode: str = 'truncate') -> TokenizedInputs:
```
ilya-lavrenov, line 1669: how to read it? 
E.g. `max_length: int = 2147483647` can not be used for ClipTokenizer 

`padding_mode: str` -> `padding_mode: PaddingMode`. 
BTW, I see that `src/python/py_tokenizer.cpp` defines Variant for types, while here we see only `str`

src/python/py_tokenizer.cpp
```diff
@@ -19,8 +19,41 @@ namespace pyutils = ov::genai::pybind::utils;
...
+namespace {
+
+PaddingMode get_pad_mode(const PadModesVariant& pad_mode) {
+    PaddingMode result;
+    std::visit(pyutils::overloaded {
+    [&](PaddingMode padding_mode) {
+        result = padding_mode;
+    },
+    [&](std::basic_string<char> pad_mode_str) {
+        if (pad_mode_str == "truncate") {
```
ilya-lavrenov, line 34: why do we need to allow `str` for this value?

src/README.md
```diff
@@ -396,6 +396,34 @@ print(f'Median from token to token duration: {np.median(durations):.2f} ms')
 
 For more examples of how metrics are used, please refer to the Python [benchmark_genai.py](../samples/python/text_generation/README.md) and C++ [benchmark_genai](../samples/cpp/text_generation/README.md) samples.
 
+### Tokenization
+
+OpenVINO™ GenAI provides a simple way to tokenize and detokenize text using the `ov::genai::Tokenizer` class. The `Tokenizer` is a high level abstraction over the OpenVINO Tokenizers library, which is used to tokenize and detokenize text. The `Tokenizer` class provides a simple interface to tokenize and detokenize text.
```
Wovchena, line 401:
```suggestion
OpenVINO™ GenAI provides a way to tokenize and detokenize text using the `ov::genai::Tokenizer` class. The `Tokenizer` is a high level abstraction over the OpenVINO Tokenizers library.
```
You stated the same thing twice.

src/README.md
```diff
@@ -396,6 +396,34 @@ print(f'Median from token to token duration: {np.median(durations):.2f} ms')
...
+
+It can be initialized from the path as a string or from the `ov::genai::LLMPipeline` object.
```
Wovchena, line 403:
```suggestion
It can be initialized from the path, in-memory IR representation or obtained from the `ov::genai::LLMPipeline` object.
```

src/cpp/src/tokenizer.cpp
```diff
@@ -109,17 +112,46 @@ class Tokenizer::TokenizerImpl {
...
+        std::tie(max_trunc_length_val, max_pad_length_val) = get_padding_values(padding_mode_val, max_length_val);
 
         // If user requested add_special_tokens mode different from the current one,
         // need to set state variable.
         // If requested mode matches the stored state set, then don't touch states.
-        if (add_special_tokens_flag == m_add_special_tokens && skip_special_tokens_flag == m_skip_special_tokens) {
+        if (add_special_tokens_flag == m_add_special_tokens 
+            && skip_special_tokens_flag == m_skip_special_tokens
+            && max_trunc_length_val == m_max_trunc_length
+            && max_pad_length_val == m_max_pad_length) {
```
Wovchena, line 154: Extend the comment

src/cpp/src/tokenizer.cpp
pavel-esir: sorry🤦will remove that

tests/python_tests/test_tokenizer.py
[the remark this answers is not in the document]
pavel-esir: Done. Now masks are compared as well

src/cpp/src/make_tokenizer_stateful.hpp
pavel-esir: Done. Even better to use `inline`

src/cpp/src/tokenizer.cpp
pavel-esir: Simplified regimes a little bit. There no NONE anymore. Now there is only 3 regimes. Please take a look to the latest changes.

src/cpp/src/tokenizer.cpp
pavel-esir: removed

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -125,7 +125,9 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     /**
     * @brief encode a single prompt
     * @param prompt std::string with input prompt
-    * @param properties tokenization properties, e.g. ov::genai::add_special_tokens(false)
+    * @param add_special_tokens whether to add special tokens
+    * @param max_length maximum length to which output will be padded or truncated
+    * @param padding_mode whether to pad result, allowed values are ["truncate", "longest", "max_length", "do_not_pad"]
```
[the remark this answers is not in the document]
pavel-esir: Simplified regimes. None is not present anymore. Updated docstring.

src/python/py_tokenizer.cpp
pavel-esir: removed string. Left only enums.

tests/python_tests/test_tokenizer.py
```diff
@@ -254,6 +254,26 @@ def test_encode_decode_with_special_tokens_option(prompt):
     assert decoded_genai_skip_spec != decoded_genai_no_skip
     assert decoded_hf_skip_spec != decoded_hf_no_skip
 
+prompts = [
+    ['1+1=', 'What is the previous answer?']
```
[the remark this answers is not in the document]
pavel-esir: - Yes, these arguments can be used with single batch as well. Added such test cases as well.
- Added longer prompts that exceed `max_length` to ensure truncation is done.

src/README.md
pavel-esir: Thx! Applied

src/README.md
pavel-esir: Done

src/python/openvino_genai/py_openvino_genai.pyi
pavel-esir: I changed that. The current approach is way way simpler and clearer. Only enum is accepted. Please take a look to the latest version

src/cpp/src/tokenizer.cpp
```diff
@@ -11,6 +11,7 @@
 #include <jinja2cpp/generic_list_iterator.h>
 
 #include "openvino/pass/manager.hpp"
+#include "openvino/pass/visualize_tree.hpp"
```
pavel-esir, line 14: remove this

pavel-esir: removed

src/cpp/src/tokenizer.cpp
```diff
@@ -22,39 +22,6 @@
 
 namespace {
 
-// todo: remove when openvino-tokenizers will support left padding
```
pavel-esir, line 25: We already have left padding in tokenizers

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -174,7 +178,7 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     /**
     * @brief decode sequence of tokens
     * @param tokens ov::Tensor with tokens with shape [batch_size, seq_len]
-    * @param detokenization_params detokenization parameters,  e.g. ov::genai::skip_special_tokens(true)
+    * @param detokenization_params detokenization parameters,  e.g. ov::genai::skip_special_tokens(true)s
```
apaniukov, line 181:
```suggestion
    * @param detokenization_params detokenization parameters,  e.g. ov::genai::skip_special_tokens(true)
```

apaniukov (approved):

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -241,8 +245,31 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
+ * IMPORTANT NOTICE: even in TRUNCATE mode the padding is applied to the longest sequence in the batch
+ * since resulting tokenization is stored as a signe ov::Tensor which cannot store irregular/ragged array.
+ * For the same reason also in MAX_LENGTH mode truncation is also applied for sequences longer that max_length,
+ * so that we have the same lengths for each sequence in the batch.
+ *
+ * @var PaddingMode::TRUNCATE
+ * Truncate the sequence to the maximum length specified. (But also LONGEST mode is implicitly applied.)
+ *
+ * @var PaddingMode::LONGEST
+ * Pad the sequence to the length of the longest sequence in the batch. In this mode truncation is switched off.
```
Wovchena, line 262: > But also LONGEST mode is implicitly applied.

> In this mode truncation is switched off.

So TRUNCATE activates LONGEST which disables TRUNCATE. TRUNCATE switches itself off

src/cpp/src/tokenizer.cpp
```diff
@@ -137,15 +130,31 @@ class Tokenizer::TokenizerImpl {
         ov::Tensor skip_special_tensor = ov::Tensor(ov::element::i32, {1});
         *skip_special_tensor.data<int>() = skip_special_tokens_flag;
 
+        ov::Tensor max_trunc_length_tensor = ov::Tensor(ov::element::i32, {});
+        *max_trunc_length_tensor.data<int>() = max_trunc_length_val;
+        ov::Tensor max_pad_length_tensor = ov::Tensor(ov::element::i32, {1});
+        *max_pad_length_tensor.data<int>() = max_pad_length_val;
```
Wovchena, line 136: ints aren't always 32 bit. Use int32_t. Although if someone tries to compile genai on such system it would probably fail earlier. You use int32_t in `get_padding_values()` after all

src/README.md
```diff
@@ -398,6 +398,93 @@ print(f'Median from token to token duration: {np.median(durations):.2f} ms')
...
+print(tokens.input_ids.shape)
+# out_shape: [2, 6]
+
+# String are short and resulting tokens sequence will be padded to 1024.
+tokens = tokenizer.encode(prompts, padding_mode=ov_genai.PaddingMode.MAX_LENGTH, max_length=1024)
+print(tokens.input_ids.shape)
+# out_shape: [2, 1024]
+
+# Very long string will be truncated to the same 1024 length for the longer sequences 
+# and padded to the same 1024 length for the shorter sequences.
```
Wovchena, line 479: > We should add that truncation is the default behaviour and the max_length value comes from the original tokenizer if not explicitly defined.

Same for docstrings

src/cpp/src/tokenizer.cpp
```diff
@@ -109,17 +78,41 @@ class Tokenizer::TokenizerImpl {
...
-        // need to set state variable.
-        // If requested mode matches the stored state set, then don't touch states.
-        if (add_special_tokens_flag == m_add_special_tokens && skip_special_tokens_flag == m_skip_special_tokens) {
+        ov::genai::utils::read_anymap_param(params, padding_mode.name(), padding_mode_val);
+        ov::genai::utils::read_anymap_param(params, max_length.name(), max_length_val);
+        
+        int max_trunc_length_val = m_max_trunc_length;
+        int max_pad_length_val = m_max_pad_length;
+
+        std::tie(max_trunc_length_val, max_pad_length_val) = get_padding_values(padding_mode_val, *max_length_val);
```
Wovchena, line 108: MAX_LENGTH and missing max_length_val is  undefined behavior.

tests/python_tests/test_continuous_batching.py
```diff
@@ -74,11 +74,11 @@ def test_e2e_real_models(tmp_path, model_id):
 @pytest.mark.parametrize("generation_config", test_configs)
 @pytest.mark.parametrize("prompt", batched_prompts[1:])  # num_beams=15 diverges on the first prompt.
 @pytest.mark.precommit
-def test_continuous_batching_vs_stateful(prompt, generation_config):
+def test_continuous_batching_vs_stateful(prompt, generation_config):  # TODO check this test
```
Wovchena, line 77: Describe TODO more clearly or apply

Wovchena (changes requested):

src/cpp/src/make_tokenizer_stateful.hpp
```diff
@@ -32,9 +33,29 @@ namespace genai {
...
+class MakeAddSpecialTokensSatateful : public ov::pass::ModelPass {
 public:
-    OPENVINO_MODEL_PASS_RTTI("MakeCombineSegmentsSatateful");
+    OPENVINO_MODEL_PASS_RTTI("MakeAddSpecialTokensSatateful");
+    bool run_on_model(const std::shared_ptr<ov::Model>& model) override;
+};
+
+/** 
+ * @brief This pass modifies tokenizer ov::Model so that max_pad_length input to
+ * RaggedToDense is made modifiable during runtime so that padding can be controlled.
```
pavel-esir, line 44:
```suggestion
 * Make ends main input for CombineSegments stateful so that padding can be controlled at runtime.
```

tests/python_tests/test_continuous_batching.py
pavel-esir: I already checked that test, but forget to delete comment. Will remove it

src/cpp/src/tokenizer.cpp
```diff
@@ -137,15 +130,31 @@ class Tokenizer::TokenizerImpl {
...
+        bool set_truncation = padding_mode_val == PaddingMode::LONGEST || max_length_val.has_value();
+
         for (auto& state: infer_request_guard.get().query_state()) {
-            if (state.get_name().find(ov::genai::ADD_SPECIAL_TOKENS_VAR_ID) != std::string::npos) {
+            if (state.get_name().find(add_special_tokens.name()) != std::string::npos) {
                 state.set_state(add_special_tensor);
-            } else if (state.get_name().find(ov::genai::SKIP_SPECIAL_TOKENS_VAR_ID) != std::string::npos) {
+            } else if (state.get_name().find(skip_special_tokens.name()) != std::string::npos) {
                 state.set_state(skip_special_tensor);
+            } else if (state.get_name().find(MAX_TRUNCATION_LENGTH_VAR_ID) != std::string::npos && set_truncation) {
```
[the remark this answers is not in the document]
pavel-esir: It's not possible because we have a single ov::Parameter max_length and 2 different states for padding and truncation.

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Looks a bit complicated. I realized that 2 regimes is enough. Please take a look to the latest version.

src/cpp/src/tokenizer.cpp

src/README.md
pavel-esir: Simplified regimes. Now we have 

- `max_length` which by default defines truncation value and padding to the longest sequence in the batch is also performed. (If `max_length` is not set, default value is taken from IR).

- and `pad_to_max_length` flag, which enables padding to the `max_length` instead of the longest sequence in the batch. 

If resulting sequence exceeds `max_length` it will be still truncated. So now we always have truncation, if someone need to completely disable it they can set it to int32_max.

@apaniukov please take a look once more to the latest version, feed free to withdraw your approve if you don't agree with the latest version :)

src/cpp/src/tokenizer.cpp
pavel-esir: I had a solution for this. But since approach changed a bit, this code is not present anymore.

src/cpp/src/make_tokenizer_stateful.hpp
pavel-esir: applied

src/README.md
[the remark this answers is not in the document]
pavel-esir: Answered below. Please take a look

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -136,7 +138,9 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     /**
     * @brief encode batch of prompts. Left padding will be applied by default
     * @param prompts vector storing batch of prompts
-    * @param properties tokenization properties, e.g. ov::genai::add_special_tokens(false)
+    * @param add_special_tokens whether to add special tokens
+    * @param max_length maximum length to which output will be padded or truncated
```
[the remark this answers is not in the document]
pavel-esir: applied separately and pushed

src/cpp/include/openvino/genai/tokenizer.hpp

Wovchena (changes requested):

src/cpp/src/tokenizer.cpp
ilya-lavrenov: should we remove this test then ? (which is skipped now, see CVS-159924) https://github.com/openvinotoolkit/openvino.genai/blob/523ea7c40667fd4c78ae6b0c817f32419a71db99/tests/python_tests/test_llm_pipeline.py#L454-L487 ?

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1678,12 +1678,12 @@ class Tokenizer:
         Decode a batch of tokens into a list of string prompt.
         """
     @typing.overload
-    def encode(self, prompts: list[str], add_special_tokens: bool = True) -> TokenizedInputs:
+    def encode(self, prompts: list[str], add_special_tokens: bool = True, pad_to_max_length: bool = False, max_length: int | None = None) -> TokenizedInputs:
```
ilya-lavrenov, line 1681: I suppose these default arguments can overrides values stored in XML files?

E.g. `CLIPtokenizer` pads to 77 (its max_length), while doing the same via GenAI will use `pad_to_max_length: bool = False` ?

tests/python_tests/ov_genai_test_utils.py
```diff
@@ -89,6 +88,7 @@ def read_model(params, **tokenizer_kwargs):
     from optimum.intel.openvino import OVModelForCausalLM
     from transformers import AutoTokenizer
     hf_tokenizer = AutoTokenizer.from_pretrained(model_id, trust_remote_code=True)
+    hf_tokenizer.padding_side = tokenizer_kwargs.pop("padding_side", "right")
```
ilya-lavrenov, line 91: why do we need to patch HF's Tokenizer padding side? 

OpenVINO Tokenizers always use `paddind_size=left`, so why is it not the same for HF?

src/python/py_tokenizer.cpp
```diff
@@ -44,21 +44,40 @@ void init_tokenizer(py::module_& m) {
             return std::make_unique<ov::genai::Tokenizer>(tokenizer_path, kwargs_properties);
         }), py::arg("tokenizer_path"), py::arg("properties") = ov::AnyMap({}))
 
-        .def("encode", [](Tokenizer& tok, std::vector<std::string>& prompts, bool add_special_tokens) {
+        .def("encode", [](Tokenizer& tok, std::vector<std::string>& prompts, 
+                          bool add_special_tokens, 
+                          bool pad_to_max_length,
+                          std::optional<size_t> max_length) {
```
ilya-lavrenov, line 50: should not we wrap all parameters with std::optional? to be aligned with C++ where we don't have default arguments for `encode()`

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -238,8 +242,12 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     std::shared_ptr<TokenizerImpl> m_pimpl;
 };
 
+enum class PaddingMode { TRUNCATE = 0, LONGEST = 1, MAX_LENGTH = 2, DO_NOT_PAD = 3, NONE = 4};
```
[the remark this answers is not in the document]
ilya-lavrenov: > About clip, they it will truncate to 77 and padded to longest

But since longest is still 77, it makes no difference ?

src/python/openvino_genai/py_openvino_genai.pyi
apaniukov: This union notation for typing was added in Python 3.10: https://docs.python.org/3/whatsnew/3.10.html#pep-604-new-type-union-operator
We should use `Optional` until we drop support for 3.8 and 3.9

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -243,6 +247,7 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
 
 static constexpr ov::Property<bool> add_special_tokens{"add_special_tokens"};
 static constexpr ov::Property<bool> skip_special_tokens{"skip_special_tokens"};
+static constexpr ov::Property<bool> pad_to_max_length{"pad_to_max_length"};
```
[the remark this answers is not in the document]
pavel-esir: It's not missing. It's already defined in `generation_config.hpp`, and we safely reuse it. Moreover redeclaration causes compilation error

src/python/py_tokenizer.cpp
pavel-esir: In our approach with AnyMap/ov::Parameters indeed we have optional arguments. When user calls "encode" without arguments we come to `set_state_if_necessary` with empty anymap and optional max_length remains without value, while pad_to_max_length_val will be false.

https://github.com/openvinotoolkit/openvino.genai/blob/2f6c08c9615f212dac91f31d76880daa8d5371fa/src/cpp/src/tokenizer.cpp#L84-L90

src/cpp/src/tokenizer.cpp
pavel-esir: Yes, we can remove that test. Current tests cover more cases.

tests/python_tests/ov_genai_test_utils.py
pavel-esir: >OpenVINO Tokenizers always use paddind_size=left, so why is it not the same for HF?

By default we do not enforce any side during conversion. OV Tokenizers uses the padding side from hf_tokenizer object which can both right/left.

src/cpp/include/openvino/genai/tokenizer.hpp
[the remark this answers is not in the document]
pavel-esir: It will turn into 77 only if user explicitly specified to pad to `max_length` both in HF and in OV Tokenizers. If user do not specify explicitly max_length it will be taken from the IR.

I simplified this regimes even more: now we have only 2 regimes `pad_to_max_length=true/false` let's continue this discussion bellow.

src/python/openvino_genai/py_openvino_genai.pyi
Wovchena: > This default value is aligned with the default value that is written into IR.

Someone could change rt_info manually.

tests/python_tests/ov_genai_test_utils.py
Wovchena: > In CB we don't have padding, let's assume we have batch 2 with, in CB it will be
```
[1, 42, 98]
[1, 67, 89, 0, 56, 8]
```
> While in the vanilla LLMPipeline, since most of the models are right padded, in most cases we would have
```
[1, 42, 98, pad, pad, pad]
[1, 67, 89,     0,    56,    8]
```
> And results would not match for right padded models.

Shouldn't vanilla LLMPipeline suppress pad tokens with attention_mask?

src/cpp/src/tokenizer.cpp
```diff
@@ -112,14 +81,20 @@ class Tokenizer::TokenizerImpl {
     void set_state_if_necessary(CircularBufferQueueElementGuard<ov::InferRequest>& infer_request_guard, const ov::AnyMap& params) {
         bool add_special_tokens_flag = m_add_special_tokens;
         bool skip_special_tokens_flag = m_skip_special_tokens;
-
+        std::optional<int32_t> max_length_val;
+        bool pad_to_max_length_val = false;
```
Wovchena, line 85: `add_special_tokens_flag` and `skip_special_tokens_flag` are initialized from the previous values but `max_length_val` and `pad_to_max_length_val` have always the same initial values. So if `add_special_tokens` was used earlier, it would be preserved if not specified again, but `pad_to_max_length` must always be specified if true i needed. Is that really how it should work?

src/python/openvino_genai/py_openvino_genai.pyi
ilya-lavrenov: @pavel-esir ok, thank you. Looks like I have mixed Clip encoder (which always expect 77 as input len) and tokenizer (which does not perform padding)

> Someone could change rt_info manually.

I suppose it's not in rt_info, but in `.bin` file

tests/python_tests/ov_genai_test_utils.py
ilya-lavrenov: > Shouldn't vanilla LLMPipeline suppress pad tokens with attention_mask?

It can, but position IDs between tokens will different on an amount of # pad tokens.

ilya-lavrenov (approved):

src/cpp/src/tokenizer.cpp
pavel-esir: Thanks for noticing this!

The approach for `max_length_val` `pad_to_max_length_val` was correct for add/skip special tokens was wrong, we should've not initialized them with values from `m_*`. 

After your question i checked behavior in c++, and it's strange. If we do call `encode` without arguments they will hold last passed values, e.g. here after we called encode with add_special_tokens=False and after that without arguments it does not add special tokens although we intended to add them by default.

image
image

This issue happened only in c++, if we call without arguments then anymap is empty. In python if we didn't specify arguments anymap was always is filled with default values because they are defined in src/python/py_tokenizer.cpp.

Indeed this is the place where default arguments are defined for c++, i changed them to default values instead of initializing with `m_*`.

Wovchena (approved):

PR #1540, opened 2025-01-14
Fixed streamer without corrupting hieroglyphs

pavel-esir:
- Cherry-pick of #1492 and #1302
- Fix for streamer when adding new tokens can shorten result. This is the fix CVS-157216 which was merged to the previous release branch and did not make it into master. And this is a correct fix this does not break hieroglyphs.

Ticket: CVS-157216

### DIFF
--- a/tests/python_tests/common.py
+++ b/tests/python_tests/common.py
@@ -474,3 +474,22 @@ def get_image_by_link(link):
         image = image.convert('RGB')
     image_data = np.array((np.array(image.getdata()) - 128).astype(np.byte)).reshape(1, 3, image.size[1], image.size[0])
     return Tensor(image_data)
+
+def get_streamer_with_results():
+    # Return a streamer which accumulates results in order to compare with results returned from generate.
+    class StreamerWithResults:
+        results: List[str] = []
+        def __init__(self):
+            self.results = []
+
+        def accumulate(self, subword) -> bool:
+            self.results.append(subword)
+            return False
+        
+        def get_result_str(self) -> str:
+            return ''.join(self.results)
+        
+        def reset(self):
+            self.results = []
+
+    return StreamerWithResults()
--- a/tests/python_tests/common.py
+++ b/tests/python_tests/common.py
@@ -335,8 +335,10 @@ class StreamerWithResults:
         self.results.append(subword)
         return False
     
-    def get_result_str(self) -> str:
-        return ''.join(self.results)
+    def get_results(self) -> List[GenerationResult]:
+        streaming_result = GenerationResult()
+        streaming_result.m_generation_ids = [''.join(self.results)]
+        return [streaming_result]
     
     def reset(self):
         self.results = []
--- a/tests/python_tests/common.py
+++ b/tests/python_tests/common.py
@@ -350,22 +350,14 @@ def run_llm_pipeline(
     prompts: List[str],
     generation_config : GenerationConfig,
     use_cb : bool = False,
-    streamer: StreamerWithResults | Callable | StreamerBase = None
+    streamer: Callable | StreamerBase = None
 ) -> List[GenerationResult]:
     properties = get_default_properties()
     if use_cb:
         properties['scheduler_config'] = SchedulerConfig()
     ov_pipe = LLMPipeline(models_path, device='CPU', **properties)
 
-    if (isinstance(streamer, StreamerWithResults)):
-        # Clear the accumulated strings to avoid side effects
-        streamer.reset()
-
-    generate_outputs : DecodedResults = ov_pipe.generate(
-        inputs=prompts, 
-        generation_config=generation_config, 
-        streamer=streamer.accumulate if isinstance(streamer, StreamerWithResults) else streamer
-    )
+    generate_outputs : DecodedResults = ov_pipe.generate(inputs=prompts, generation_config=generation_config, streamer=streamer)
 
     index = 0
     generation_results = []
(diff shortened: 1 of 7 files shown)


tests/python_tests/common.py
```diff
@@ -474,3 +474,22 @@ def get_image_by_link(link):
         image = image.convert('RGB')
     image_data = np.array((np.array(image.getdata()) - 128).astype(np.byte)).reshape(1, 3, image.size[1], image.size[0])
     return Tensor(image_data)
+
+def get_streamer_with_results():
```
ilya-lavrenov, line 478: as we agreed - we need to add to `run_llm_pipeline_with_ref` and `run_cb_pipeline_with_ref` and as they cover more cases (like streaming when stop strings are enabled)

pavel-esir: Extented `run_llm_pipeline` to use with streamer and reused `compare_generation_results`

pavel-esir:
Current changes are final. @ilya-lavrenov please review

tests/python_tests/common.py
```diff
@@ -412,6 +440,27 @@ def convert_models(opt_model : OVModelForCausalLM, hf_tokenizer : AutoTokenizer,
     serialize(detokenizer, models_path / "openvino_detokenizer.xml")
 
 
+def run_llm_pipeline_with_streamer(model_id: str,
```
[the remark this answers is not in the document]
pavel-esir: I thought that comparing inside `run_llm_pipeline` isn't quite right because function was introduced to get results. I run comparison together with other comparators inside `run_llm_pipeline_with_ref`.

Removed `run_llm_pipeline_with_streamer', it's not necessary anymore

tests/python_tests/test_llm_pipeline.py
```diff
@@ -194,6 +194,27 @@ def test_callback_kwargs_one_string(callback):
...
+    'The Sun is yellow because', 
+    '你好！ 你好嗎？'
+    'I have an interview about product speccing with the company Weekend Health. Give me an example of a question they might ask with regards about a new feature'
+])
+@pytest.mark.parametrize("generation_config", [dict(max_new_tokens=10), dict(max_new_tokens=300)])
+@pytest.mark.parametrize("model_descr", get_models_list())
+@pytest.mark.parametrize("use_cb", [True, False])
+@pytest.mark.precommit
+@pytest.mark.nightly
+def test_streamer_compare_texts(model_descr, generation_config, prompt, streamer, use_cb):
```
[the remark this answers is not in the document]
pavel-esir: Added comparison in `run_llm_pipeline_with_ref`. Please take a look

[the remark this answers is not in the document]
pavel-esir: But we still need one more test. Because there are some specific long prompts and large max_new_tokens. We can't rely on only on the currently existing ones

tests/python_tests/common.py
```diff
@@ -410,9 +430,14 @@ def convert_models(opt_model : OVModelForCausalLM, hf_tokenizer : AutoTokenizer,
...
+ 
 
-
-def run_llm_pipeline_with_ref(model_id: str, prompts: List[str], generation_config: GenerationConfig | dict, tmp_path: Path, use_cb : bool = False):
+def run_llm_pipeline_with_ref(model_id: str, 
+                              prompts: List[str], 
+                              generation_config: GenerationConfig | dict, 
+                              tmp_path: Path, 
+                              use_cb : bool = False,
+                              streamer: StreamerWithResults | Callable | StreamerBase = None):
```
ilya-lavrenov, line 440: why streamer is None by default? 
The idea is to enable comparison streamer output vs generate output for all tests which use `run_llm_pipeline`

tests/python_tests/common.py
```diff
@@ -325,19 +325,39 @@ def get_default_properties():
...
+
+
+
 def run_llm_pipeline(
     models_path : Path,
     prompts: List[str],
     generation_config : GenerationConfig,
-    use_cb : bool = False
+    use_cb : bool = False,
+    streamer: Callable | StreamerBase = None
```
ilya-lavrenov, line 353: why streamer is None by default? 
The idea is to enable comparison streamer output vs generate output for all tests which use `run_llm_pipeline`

tests/python_tests/test_llm_pipeline.py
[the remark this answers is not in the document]
ilya-lavrenov: here you run greedy generation, can you extend this test 
https://github.com/openvinotoolkit/openvino.genai/blob/8ea14145509c52baa6c37cb1425f0d8d7529ad9d/tests/python_tests/test_sampling.py#L58-L67
instead of new one? the do the same things

tests/python_tests/common.py
ilya-lavrenov: > to use with streamer

they don't use streamer now as default argument is None

tests/python_tests/common.py
pavel-esir: I was afraid that it will fail in case if there is a beam search, multibatch prompt. AFAIK these are the only 2 states when we don't allow streaming. But i'm not sure maybe there are some more new cases introduced.

Enabled streaming results check for every test which is not multibatch and not beam_search. Fingers crossed, let's hope. Nothing will fail.

But `run_llm_pipeline` is not used in tests, it requires many steps to call it, load HF model, convert and save them. When i used it in `test_llm_pipeline` it looked ugly and with many duplicates with `common.py`.  `run_llm_pipeline_with_ref` is better place to add streamer for each test case.

tests/python_tests/test_llm_pipeline.py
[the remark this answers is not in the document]
pavel-esir: done

tests/python_tests/common.py
```diff
@@ -410,19 +430,34 @@ def convert_models(opt_model : OVModelForCausalLM, hf_tokenizer : AutoTokenizer,
...
+                              tmp_path: Path, 
+                              use_cb : bool = False,
+                              streamer: StreamerWithResults | Callable | StreamerBase = None):
     models_path : Path = tmp_path / model_id
     opt_model, hf_tokenizer = get_hugging_face_models(model_id)
 
     if type(generation_config) is dict:
         generation_config = GenerationConfig(**generation_config)
 
+    if streamer is None and (not generation_config.is_beam_search()) and len(prompts) == 1:
```
ilya-lavrenov, line 447:
```suggestion
    if streamer is None and not (generation_config.is_beam_search() or generation_config.num_return_sequences > 1) and len(prompts) == 1:
```

ilya-lavrenov (approved):

PR #1425, opened 2024-12-23
Move lm model to async infer

sbalandi:

PR #1554, opened 2025-01-15
Fix finish generation from streamer

sbalandi:


sbalandi:
Bug was appeared here https://github.com/openvinotoolkit/openvino.genai/pull/1425

PR #1555, opened 2025-01-15
2025 Copyright year update

jacekpawlak:


andrei-kochin (approved):

Wovchena:
Can you update other places as well? For example https://github.com/openvinotoolkit/openvino.genai/blob/ef1f0f60c58cf0452cfbc6305f5cf30e9137ceb7/src/cpp/src/visual_language/clip.hpp#L1

PR #1583, opened 2025-01-17
Change openvino.runtime imports to openvino for 2025.0

helena-intel:
Prevent DeprecationWarnings in 2025.0. 

PR to master branch: https://github.com/openvinotoolkit/openvino.genai/pull/1579

### DIFF
--- a/tools/llm_bench/llm_bench_utils/ov_utils.py
+++ b/tools/llm_bench/llm_bench_utils/ov_utils.py
@@ -17,7 +17,10 @@ from llm_bench_utils.config_class import (
     DEFAULT_MODEL_CLASSES,
     IMAGE_GEN_CLS
 )
-import openvino.opset13 as opset
+try:
+    import openvino.opset13 as opset
+except ImportError:
+    import openvino.runtime.opset13 as opset
 from transformers import pipeline
 import openvino_genai as ov_genai
 import queue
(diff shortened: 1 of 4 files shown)


tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -17,7 +17,7 @@
     DEFAULT_MODEL_CLASSES,
     IMAGE_GEN_CLS
 )
-import openvino.runtime.opset13 as opset
+import openvino.opset13 as opset
```
eaidova, line 20: can we temporary keep old namespace with try-except here:
```suggestion
try:
    import openvino.opset13 as opset
except ImportError:
    import openvino.runtime.opset13 as opset
```

openvino namespace doesn't contain opsets in 2024.6 and more likely some december nightly builds for 2025/0/0 runtime benchmarking may become broken due to this

or you can move this import closer to its usage, I;m planning to remove part of legacy transformation from llm bench soon anyway, we need just make sure that this import does not affect main usage scenario

eaidova (approved):

tools/llm_bench/llm_bench_utils/ov_utils.py
helena-intel: Thank you! Agreed with the try-except approach for now, also because it's not great to have to checkout the release branch to run llm_banch every time you want to compare 2025.0 with 2024.6.

PR #1588, opened 2025-01-17
[RTTI] Fix passes rtti definitions

t-jankowski:
Follow up of #1577


ilya-lavrenov (approved):
Please, port to releases/2025/0 as well

PR #1601, opened 2025-01-20
[LLM] [NPU] StaticLLMPipeline: Export blob

smirnov-alexey:
Release: https://github.com/openvinotoolkit/openvino.genai/pull/1603

### DIFF
--- a/src/cpp/src/llm_pipeline_static.cpp
+++ b/src/cpp/src/llm_pipeline_static.cpp
@@ -698,11 +698,15 @@ StatefulLLMPipeline::StatefulLLMPipeline(
                         utils::from_config_json_if_exists(models_path)),
     m_sampler(m_tokenizer) {
     ov::AnyMap properties = config;
-    const auto use_blob = pop_or_default(properties, "USE_BLOB", false);
-    if (use_blob) {
-        auto blob_path = pop_or_default(properties, "BLOB_PATH", std::string{});
+
+    auto blob_path = pop_or_default(properties, "BLOB_PATH", std::string{});
+    const auto export_blob = pop_or_default(properties, "EXPORT_BLOB", false);
+
+    bool do_import = (!blob_path.empty() && !export_blob);
+
+    if (do_import) {
         if (blob_path.empty()) {
-            blob_path = (models_path / "openvino_model.blob").string();
+            OPENVINO_THROW("Please provide a path to blob in BLOB_PATH: " + blob_path);
         }
         if (!std::filesystem::exists(blob_path)) {
             OPENVINO_THROW("Blob file is not found at: " + blob_path);
@@ -722,9 +726,7 @@ StatefulLLMPipeline::StatefulLLMPipeline(
         ov::AnyMap properties = config;
         auto compiled = setupAndCompileModel(model, model_desc, properties);
         // Also export compiled model if required
-        const auto export_blob = pop_or_default(properties, "EXPORT_BLOB", false);
         if (export_blob) {
-            auto blob_path = pop_or_default(properties, "BLOB_PATH", std::string{});
             if (blob_path.empty()) {
                 blob_path = (models_path / "openvino_model.blob").string();
             }
--- a/src/cpp/src/llm_pipeline_static.cpp
+++ b/src/cpp/src/llm_pipeline_static.cpp
@@ -721,6 +721,19 @@ StatefulLLMPipeline::StatefulLLMPipeline(
         ModelConfigDesc model_desc = get_modeldesc_from_json(models_path / "config.json");
         ov::AnyMap properties = config;
         auto compiled = setupAndCompileModel(model, model_desc, properties);
+        // Also export compiled model if required
+        const auto export_blob = pop_or_default(properties, "EXPORT_BLOB", false);
+        if (export_blob) {
+            auto blob_path = pop_or_default(properties, "BLOB_PATH", std::string{});
+            if (blob_path.empty()) {
+                blob_path = (models_path / "openvino_model.blob").string();
+            }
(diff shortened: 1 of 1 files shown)


dmatveev (approved):
Not too bad, @TolyaTalamanov thoughts?

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -721,6 +721,19 @@ StatefulLLMPipeline::StatefulLLMPipeline(
         ModelConfigDesc model_desc = get_modeldesc_from_json(models_path / "config.json");
         ov::AnyMap properties = config;
         auto compiled = setupAndCompileModel(model, model_desc, properties);
+        // Also export compiled model if required
+        const auto export_blob = pop_or_default(properties, "EXPORT_BLOB", false);
```
TolyaTalamanov, line 725: What kind of path can be provided there?

1) If it's absolute path e.g `C:\workspace\openvino_model.blob` - then, I believe we shouldn't combine it with `model_path`.
2) If it's relative path, I believe the handling is right.

BTW, should the user provide path to the file or folder? How do you check this?

TolyaTalamanov: Why not `DUMP_BLOB`?

smirnov-alexey: Only full path is accepted.
I only combine `models_path / "openvino_model.blob"` if the provided `blob_path` is empty.
Added a check that `.blob` file is provided

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -721,6 +721,26 @@ StatefulLLMPipeline::StatefulLLMPipeline(
...
+        if (export_blob) {
+            auto blob_path = pop_or_default(properties, "BLOB_PATH", std::string{});
+            if (blob_path.empty()) {
+                blob_path = (models_path / "openvino_model.blob").string();
+            }
+            // Check the path is full
+            if (blob_path.size() < 5) {
+                OPENVINO_THROW("Please provide a full path to blob file in BLOB_PATH: " + blob_path);
+            }
+            if (strncmp(".blob", &blob_path[blob_path.size() - 5], 5) != 0) {
```
TolyaTalamanov, line 735: `const int EXT_SIZE=5`

TolyaTalamanov, line 735: Why not `std::string::compare` btw?

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: > Only full path is accepted.

Perhaps it can be checked: https://en.cppreference.com/w/cpp/filesystem/path/is_absrel

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Maybe this can help: https://en.cppreference.com/w/cpp/filesystem/path/extension

TolyaTalamanov (approved):

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: I took it from OV util. Couldn't include it here

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -698,11 +698,15 @@ StatefulLLMPipeline::StatefulLLMPipeline(
...
+    auto blob_path = pop_or_default(properties, "BLOB_PATH", std::string{});
+    const auto export_blob = pop_or_default(properties, "EXPORT_BLOB", false);
+
+    bool do_import = (!blob_path.empty() && !export_blob);
+
+    if (do_import) {
         if (blob_path.empty()) {
-            blob_path = (models_path / "openvino_model.blob").string();
+            OPENVINO_THROW("Please provide a path to blob in BLOB_PATH: " + blob_path);
         }
```
dmatveev, line 710: By definition this is dead code, `do_import` can't be `true` if `blob_path.empty()` is true

src/cpp/src/llm_pipeline_static.cpp
```diff
@@ -721,6 +725,25 @@ StatefulLLMPipeline::StatefulLLMPipeline(
...
+                blob_path = (models_path / "openvino_model.blob").string();
+            }
+            // Check the path is full
+            const int EXT_SIZE = 5; // ".blob"
+            if (blob_path.size() < EXT_SIZE) {
+                OPENVINO_THROW("Please provide a full path to blob file in BLOB_PATH: " + blob_path);
+            }
+            if (strncmp(".blob", &blob_path[blob_path.size() - EXT_SIZE], EXT_SIZE) != 0) {
+                OPENVINO_THROW("Please provide a full path to blob file in BLOB_PATH: " + blob_path);
+            }
```
dmatveev, line 740: Out-of-bounds access if `BLOB_PATH` is specified as `a`?

smirnov-alexey: I check the size 1 line above

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: I'd like to check it before I try to create a file

src/cpp/src/llm_pipeline_static.cpp
smirnov-alexey: Fixed

dmatveev (approved):

src/cpp/src/llm_pipeline_static.cpp
TolyaTalamanov: Makes sense

PR #1607, opened 2025-01-20
[llm_bench] enable prompt permutations for prevent prefix caching and fix vlm image load

eaidova:
CVS-160892

### DIFF
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -161,6 +161,8 @@ def get_argprser():
     parser.add_argument("--num_steps", type=int, required=False, help="Number of inference steps for image generation")
     parser.add_argument("--height", type=int, required=False, help="Generated image height. Applicable only for Image Generation.")
     parser.add_argument("--width", type=int, required=False, help="Generated image width. Applicable only for Image Generation.")
+    parser.add_argument("--disable_prompt_permutation", action="store_true", help="Disable modification prompt from run to run for avoid prefix caching")
+    
     return parser.parse_args()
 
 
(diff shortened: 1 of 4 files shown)


tools/llm_bench/benchmark.py
```diff
@@ -161,6 +161,8 @@ def get_argprser():
     parser.add_argument("--num_steps", type=int, required=False, help="Number of inference steps for image generation")
     parser.add_argument("--height", type=int, required=False, help="Generated image height. Applicable only for Image Generation.")
     parser.add_argument("--width", type=int, required=False, help="Generated image width. Applicable only for Image Generation.")
+    parser.add_argument("--disable_prompt_permutation", action="store_true", help="Disable modification prompt from run to run for avoid prefix caching")
+
```
eaidova, line 165:
```suggestion
```

sbalandi (approved):

yeonbok:
Hi @eaidova  From this commit, we see wrong result for gemma 7b it model on MTL. 
Used cmd: python benchmark.py -d GPU -m /c/dev/models/WW03_llm-optimum_2025.0.0-17891/gemma-7b-it/pytorch/ov/OV_FP16-4BIT_DEFAULT/ -n 2 -ic 128 -pf /c/dev/taylor/repo-prompts/32_1024/gemma-7b-it.jsonl 
With this commit, the output length of the 32 input's result becomes 0. 
Seems that disable_prompt_permutation is None by default and enable_prompt_permutation is turned on. 
If I set false to enable_prompt_permutation, the result is restored. 
Could you please look into this?

eaidova:
@yeonbok 
this looks like an issue with generation config parameters, flag that prevent early stop generation was not added to pipeline.
I believe for some models output can be 0 even without this behavior changes, should be fixed here https://github.com/openvinotoolkit/openvino.genai/pull/1625

PR #1624, opened 2025-01-24
Correct model name

wgzintel:


peterchen-intel (approved):

eaidova:
@wgzintel could you please explain why these changes needed?

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -203,18 +203,24 @@ def get_use_case(model_name_or_path):
                 if config.get("model_type").lower().replace('_', '-').startswith(model_id):
                     log.info(f'==SUCCESS FOUND==: use_case: {case}, model_type: {model_id}')
                     return case, model_ids[idx]
+
+    case, model_name = get_model_name(model_name_or_path)
+    if case == None:
```
eaidova, line 208:
```suggestion
    if case is None:
```

wgzintel:
> @wgzintel could you please explain why these changes needed? (just for my understanding and to simplify maintenance efforts in future)

We need to output the correct model name to the csv file and json file. Currently, when the config.json file exists, according to the get_use_case(), the model name output to the csv file and json file is model_id instead of the real model name. For example, model qwen2-7b output to the csv and json file becomes qwen.

eaidova (approved):

PR #1638, opened 2025-01-29
[WWB]: Added config to preprocessor call in VLMs

AlexKoff88:


eaidova (approved):

AlexKoff88:
Fixes intern-vl model in Optimum and Transformers

PR #1641, opened 2025-01-29
[JS] Add download binaries script for npm package

vishniakov-nikolai:
When user installs `genai-node` package, first js sources downloads as the package content.
Then, as the part of post-install process `download_binaries` script runs. It downloads compiled binaries from openvino storage, based on detected user OS and architecture.

## Task
- 161447


Retribution98:
This PR is outdated, I took these changes and updated them in the PR #1775

PR #1645, opened 2025-01-29
Fix Qwen2VL generation without images

yatarkan:


src/cpp/src/visual_language/inputs_embedder.cpp
Wovchena, line 1: Is there small version of the model? Add it to tests

tests/python_tests/common.py
```diff
@@ -523,7 +523,7 @@ def get_image_by_link(link):
     image = Image.open(requests.get(link, stream=True).raw)
     if image.mode != 'RGB':
         image = image.convert('RGB')
-    image_data = np.array((np.array(image.getdata()) - 128).astype(np.byte)).reshape(1, 3, image.size[1], image.size[0])
+    image_data = np.array((np.array(image.getdata()) - 128).astype(np.byte)).reshape(1, image.size[1], image.size[0], 3)
```
Wovchena, line 526: That's a strange change. The sample uses NHWC https://github.com/openvinotoolkit/openvino.genai/blob/ec50b5b68baf2736da7f36e2543ec68667b4e064/samples/python/visual_language_chat/visual_language_chat.py#L39

Wovchena (approved):

PR #1660, opened 2025-01-31
Fix tiny-random-llava-next in VLM Pipeline

yatarkan:
Ticket: CVS-160437

llava models store patch size parameter in `config.json`, not in `preprocessor_config.json` like other vlm models. In GenAI we have default value for patch size (14) set in processor config.
However, tiny versions of llava models have different patch size value (2).
In this PR the patch size value from vlm config is propagated to vision encoder.


src/cpp/src/visual_language/vlm_config.cpp
```diff
@@ -19,6 +19,9 @@ ov::genai::VLMConfig::VLMConfig(const std::filesystem::path& json_path) {
 
     // Setting llava_next specific config params
     read_json_param(parsed, "image_newline", image_newline);
+    if (parsed.contains("vision_config")) {
+        read_json_param(parsed.at("vision_config"), "patch_size", vision_config_patch_size);
+    }
```
Wovchena, line 24: `ProcessorConfig` already has `patch_size` field. Why doesn't `ProcessorConfig` check for `vision_config` in json?

Wovchena (approved):

PR #1664, opened 2025-02-03
Fix Falcon validation

AlexKoff88:
The issue is not reproduced if using optimum-cli for export but only when cloning the entire model repo from the HF Hub.


AlexKoff88:
cc'ed @sstrehlk

eaidova (approved):

PR #1667, opened 2025-02-03
[JS] Add a dependency on the openvino-node package

Retribution98:
CVS-161725


Retribution98:
Now all files are saved in the archive, but only a few of them are used:
- genai_node_addon.node
- libcore_tokenizers.so
- libopenvino_genai.so*
- libopenvino_tokenizers.so

We can remove other files from the JS bindings archive, can you tell me how I can do this?

**I tested these changes only for ubuntu.**

src/js/CMakeLists.txt
```diff
@@ -78,6 +78,7 @@ set_target_properties(${PROJECT_NAME} PROPERTIES
 if(LINUX)
     # to find libopenvino_genai.so in the same folder
     set(rpaths "$ORIGIN")
+    list(APPEND rpaths "$ORIGIN/../../openvino-node/bin")
```
Wovchena, line 81: Does that change how the project should be built described in https://github.com/openvinotoolkit/openvino.genai/blob/3ea96c05d89c2c6aaa354df48fabe48eb0168166/src/js/README.md? As far as I understand, it's no longer required to use OPENVINO_EXTRA_MODULES. If that's true, you can probably merge the instructions with https://github.com/openvinotoolkit/openvino.genai/blob/3ea96c05d89c2c6aaa354df48fabe48eb0168166/src/docs/BUILD.md

It's true that there' a disclaimer
> Tested on Ubuntu, another OS didn't tested yet

But I was hoping that it'll go away after some time. The problem is that rpath is Linux specific

Wovchena (changes requested):

src/js/README.md
```diff
@@ -13,44 +13,26 @@ This is preview version, do not use it in production!
...
+```
+"genai-node": "file:*path-to-current-directory*"
+```
+
+### Installation verification
+
+Run a simple NodeJS script:
+```nodejs
+const { Pipeline } = require('genai-node');
+```
```
Wovchena, line 38: We didn't think about it when we were writing `BUILD.md`, but maybe you can provide a one-liner like `python -c "import openvino_genai; print(openvino_genai.__version__)"` for JS. That would simplify verification

.github/workflows/linux.yml
```diff
@@ -467,16 +460,24 @@ jobs:
...
+      - name: Install npm package tests dependencies
         working-directory: ${{ env.SRC_DIR }}/src/js
         run: |
-          cp -R ${{ env.OV_INSTALL_DIR }}/nodejs bin
-          npm install
-          npm test
+          cp -R ${{ env.INSTALL_DIR }} bin
+          npm install --verbose
+          rm -rf node_modules/openvino-node/bin
+          cp -R ${{ env.OV_INSTALL_DIR }}/openvino_js_package node_modules/openvino-node/bin
```
Wovchena, line 469: That's the first mention of `node_modules` folder in the file. Can you add an explanation to the yaml comment what it is for? This step isn't described in the instruction

src/js/BUILD.md
```diff
@@ -0,0 +1,70 @@
+### Build OpenVINO™ GenAI Node.js bindings (preview)
+
+1. Build and install OpenVINO from sources following the [instructions](https://github.com/openvinotoolkit/openvino/wiki#how-to-build).
```
Wovchena, line 3: Is OpenVINO required to be built with JS bindings? I wasn't able to find JS instruction on the wiki. If JS bindings aren't required, it's worth specifying it here anyway to be more verbose.

Wovchena (changes requested):

src/js/README.md
Wovchena: https://github.com/openvinotoolkit/openvino.genai/pull/1739

Retribution98: Sure, we can

.github/workflows/linux.yml
```diff
@@ -51,7 +51,7 @@ jobs:
       id: openvino_download
       with:
         platform: ubuntu22
-        commit_packages_to_provide: wheels
+        commit_packages_to_provide: all
```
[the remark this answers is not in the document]
akladiev: Please try with package extension
```suggestion
        commit_packages_to_provide: wheels,openvino_js_package.tar.gz
```
Since we're supporting both folder & file download ("wheels" is a folder); we need archives specified with extension

.github/workflows/linux.yml
Retribution98: I added comments here.

src/js/BUILD.md
Retribution98: Here we need to build OpenVINO without JS bindings because it is used to build GenAI.
But you are right that the user should use the GenAI JS package from source with the correct version of the OpenVINO JS package.
I've added information about this in the README that says how it should be used.

Retribution98: Yes, I have done.

.github/workflows/linux.yml
Retribution98: Oh, I see. Sure, let's do that.

.github/workflows/linux.yml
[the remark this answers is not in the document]
Retribution98: > I think wheels could be skipped for JS

Wheels is used for the other steps.

> Please try with package extension
> 
> Since we're supporting both folder & file download ("wheels" is a folder); we need archives specified with extension

Thank you, it works!

Wovchena (approved):

PR #1677, opened 2025-02-05
Switch NPU LLM execution to ov::genai::StatefulLLMPipeline

TolyaTalamanov:


src/cpp/src/llm_pipeline_stateful.hpp
```diff
@@ -48,7 +48,8 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
         const ov::genai::Tokenizer& tokenizer,
         const std::string& device,
         const ov::AnyMap& config,
-        const ov::genai::GenerationConfig& generation_config
+        const ov::genai::GenerationConfig& generation_config,
+        const std::filesystem::path& models_path = {}
```
TolyaTalamanov, line 52: Tmp solution...

src/cpp/src/utils.cpp
```diff
@@ -408,6 +474,75 @@ void print_compiled_model_properties(ov::CompiledModel& compiled_Model, const ch
...
+        kv_desc.min_response_len = compiled.get_property("NPUW_MIN_RESPONSE_LEN").as<uint32_t>();
+    } else {
+        kv_desc.max_prompt_len = pop_int_and_cast(properties, "MAX_PROMPT_LEN").value_or(1024u);
+        kv_desc.min_response_len = pop_int_and_cast(properties, "MIN_RESPONSE_LEN").value_or(128u);
+        update_npu_config(properties, model, kv_pos, kv_desc);
+        auto cache_mode = get_option<CacheMode>(config, "CACHE_MODE");
+        // NB: OPTIMIZE_SIZE is only possible when model_path is defined!
+        // Otherwise, switch to OPTIMIZE_SPEED
+        if (cache_mode.has_value() && *cache_mode == CacheMode::OPTIMIZE_SIZE && !model_path.empty()) {
+            compiled = ov::genai::utils::singleton_core().compile_model(model_path, "NPU", properties);
```
AsyaPronina, line 510: Thus way we will read the model by `model_path` twice, won't we?

src/cpp/src/utils.hpp
```diff
@@ -196,6 +205,24 @@ template<class... Ts> struct overloaded : Ts... {using Ts::operator()...;};
...
 std::shared_ptr<StreamerBase> create_streamer(StreamerVariant streamer, Tokenizer tokenizer);
 
+std::optional<ov::Any> pop_option(ov::AnyMap& config, const std::string& option_name);
+
+template <typename T>
+T pop_or_default(ov::AnyMap& config, const std::string& key, const T& default_value) {
+    auto anyopt = pop_option(config, key);
+    if (anyopt.has_value()) {
+        if (anyopt.value().empty()) {
+            if (ov::genai::utils::is_container<T>)
```
AsyaPronina, line 215: I would vote against such lifehacks..

AsyaPronina (approved):
LGTM, thanks! Minor comments left.

src/cpp/src/utils.cpp
```diff
@@ -408,6 +474,75 @@ void print_compiled_model_properties(ov::CompiledModel& compiled_Model, const ch
...
+        }
+        compiled = ov::genai::utils::singleton_core().import_model(fin, "NPU", {});
+        kv_desc.max_prompt_len = compiled.get_property("NPUW_MAX_PROMPT_LEN").as<uint32_t>();
+        kv_desc.min_response_len = compiled.get_property("NPUW_MIN_RESPONSE_LEN").as<uint32_t>();
+    } else {
+        kv_desc.max_prompt_len = pop_int_and_cast(properties, "MAX_PROMPT_LEN").value_or(1024u);
+        kv_desc.min_response_len = pop_int_and_cast(properties, "MIN_RESPONSE_LEN").value_or(128u);
+        update_npu_config(properties, model, kv_pos, kv_desc);
+        auto cache_mode = get_option<CacheMode>(config, "CACHE_MODE");
+        // NB: OPTIMIZE_SIZE is only possible when model_path is defined!
```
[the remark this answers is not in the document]
TolyaTalamanov: @ilya-lavrenov As I understood this is how it work on `openvino` side:

https://docs.openvino.ai/2025/api/c_cpp_api/classov_1_1_core.html
>This can be more efficient than using the [Core::read_model](https://docs.openvino.ai/2025/api/c_cpp_api/group__ov__transformation__common__api.html#classov_1_1_core_1af3143c652debb91411a186281593bb1a) + Core::compile_model(model_in_memory_object) flow, especially for cases when caching is enabled and a cached model is available.

src/cpp/src/llm_pipeline_stateful.cpp
```diff
@@ -284,12 +296,23 @@ EncodedResults StatefulLLMPipeline::generate(
         config.set_eos_token_id(m_generation_config.eos_token_id);
     config.validate();
 
+    auto batch_size = input_ids.get_shape().at(0);
+
+    auto execution_devices = m_model_runner.get_compiled_model().get_property(ov::execution_devices);
```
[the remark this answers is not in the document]
TolyaTalamanov: I don't mind, let's to this

src/cpp/src/utils.cpp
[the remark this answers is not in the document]
ilya-lavrenov: I suppose those are different things.

Definitely compile_model with path is preferable, but CACHE_MODE optimize should also work with ov::Model as well, IMO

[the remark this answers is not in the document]
ilya-lavrenov: BTW, model has been already read via `read_model` using a "common" code and some transformations are applied. 
So, looks strange that you ignore this model and read your own one inside the plugin.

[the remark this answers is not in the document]
TolyaTalamanov: @ilya-lavrenov You're right, it's an artifact of copy-paste, will fix, thanks!

src/cpp/src/utils.hpp
TolyaTalamanov: Oh yes, you're right, we've already discussed this. In fact, such handling isn't really needed any longer as it was used only for `STATELESS`. I'll drop it!

src/cpp/src/utils.cpp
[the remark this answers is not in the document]
TolyaTalamanov: > BTW, model has been already read via `read_model` using a "common" code and some transformations are applied. So, looks strange that you ignore this model and read your own one inside the plugin.

@ilya-lavrenov The idea is that when it's `CacheMode::OPTIMIZE_SIZE` the model should be compiled by using `models_path`.

src/cpp/src/utils.cpp
TolyaTalamanov: I believe your point is the same as @ilya-lavrenov. @smirnov-alexey please explain :)

src/cpp/src/utils.hpp
TolyaTalamanov: Done

src/cpp/src/llm_pipeline_stateful.cpp
```diff
@@ -17,7 +17,17 @@ StatefulLLMPipeline::StatefulLLMPipeline(
...
     OptionalGenerationConfig generation_config)
     : LLMPipelineImplBase(tokenizer, generation_config.has_value() ? *generation_config : GenerationConfig()),
-    m_model_runner(request) {}
+    m_model_runner(request) {
+    auto compiled_model = m_model_runner.get_compiled_model();
+    auto execution_devices = compiled_model.get_property(ov::execution_devices);
+    if (execution_devices[0].find("NPU") != std::string::npos) {
+        OPENVINO_ASSERT(execution_devices.size() == 1u);
+        m_is_npu = true;
+        const auto max_prompt_len = compiled_model.get_property("NPUW_MAX_PROMPT_LEN").as<uint32_t>();
```
TolyaTalamanov, line 26: @smirnov-alexey are these two properties will be exposed once model is imported from blob?

tests/python_tests/test_llm_pipeline_static.py
```diff
@@ -66,13 +70,40 @@ def test_generation_compare_with_stateful(generation_config):
     assert ref_out == actual_out
 
 
+@pytest.mark.precommit
+@pytest.mark.nightly
+@pytest.mark.parametrize("config", pipeline_configs)
+def test_pipeline_from_blob(config):
+    prompt = 'What is OpenVINO?'
+    model_path = read_model(get_models_list()[0])[1]
+    blob_path = "compiled_model.blob"
```
ilya-lavrenov, line 79: could you please clean up this blob upon test finalize?

tests/python_tests/test_llm_pipeline_static.py
```diff
@@ -26,13 +27,15 @@
...
-                      'NPUW_ONLINE_PIPELINE': 'NONE',
-                      'PREFILL_CONFIG': { },
-                      'GENERATE_CONFIG': { }
-                } | get_default_llm_properties()
+default_config = {
+                     'NPUW_DEVICES': 'CPU',
+                     'NPUW_ONLINE_PIPELINE': 'NONE'
+                 } | get_default_llm_properties()
+
+static_config = { **default_config, 'STATIC_PIPELINE': 'STATEFUL' }
```
smirnov-alexey, line 35: Do we still need `'STATIC_PIPELINE': 'STATEFUL'`?

src/cpp/src/utils.cpp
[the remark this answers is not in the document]
smirnov-alexey: @ilya-lavrenov I believe read_model() + compile_model(model) doesn't work. I previously discussed it with GPU and I think the only way to import model via CACHE_DIR is using compile_model(**model_path**). It has something to do with https://github.com/openvinotoolkit/openvino/pull/27162/files
Please correct me if I'm wrong, but I could achieve weightless caching with CACHE_DIR and model. If I recall correctly, WEIGHTS_PATH is not set in this case.

src/cpp/src/utils.cpp
```diff
@@ -408,6 +474,75 @@ void print_compiled_model_properties(ov::CompiledModel& compiled_Model, const ch
...
+        kv_desc.max_prompt_len = compiled.get_property("NPUW_LLM_MAX_PROMPT_LEN").as<uint32_t>();
+        kv_desc.min_response_len = compiled.get_property("NPUW_LLM_MIN_RESPONSE_LEN").as<uint32_t>();
+    } else {
+        kv_desc.max_prompt_len = pop_int_and_cast(properties, "MAX_PROMPT_LEN").value_or(1024u);
+        kv_desc.min_response_len = pop_int_and_cast(properties, "MIN_RESPONSE_LEN").value_or(128u);
+        update_npu_config(properties, model, kv_pos, kv_desc);
+        auto cache_mode = get_option<CacheMode>(config, ov::cache_mode.name());
+        // NB: OPTIMIZE_SIZE is only possible when model_path is defined!
+        // Otherwise, switch to OPTIMIZE_SPEED
+        if (cache_mode.has_value() && *cache_mode == CacheMode::OPTIMIZE_SIZE && !model_path.empty()) {
```
smirnov-alexey, line 509: @TolyaTalamanov previous behavior was in default case we just use `model_path` (weightless). And user would need to explicitly specify `OPTIMIZE**_SPEED**`. Please, return to this behavior - no need for user to specify `OPTIMIZE_SIZE` here, it's our default option

src/cpp/src/utils.cpp
smirnov-alexey: @AsyaPronina  Actually, we read model once in case of `OPTIMIZE_SPEED` and don't explicitly read by default

src/cpp/src/llm_pipeline.cpp
```diff
@@ -139,8 +139,8 @@ ov::genai::LLMPipeline::LLMPipeline(
         m_pimpl = std::make_unique<ContinuousBatchingAdapter>(models_path, tokenizer, scheduler_config, device, device_properties);
     }
 
-    if (m_pimpl == nullptr && device == "NPU") {
-        m_pimpl = static_llm::LLMPipelineFactory::create(models_path, tokenizer, device, properties);
+    if (m_pimpl == nullptr && device == "NPU" && properties.count("STATIC_PIPELINE")) {
```
smirnov-alexey, line 142: Following my previous review - can we just get rid of `STATIC_PIPELINE` since our only option is `STATEFULL`?

smirnov-alexey (changes requested):
My only concern here is `OPTIMIZE_SPEED` and `OPTIMIZE_SIZE` behavior

src/cpp/src/llm_pipeline.cpp
TolyaTalamanov: Most likely it's not the case, but what if we have another static pipeline implementation? I plan to remove it anyway next week preferably when switching to generic pipeline doesn't bring any perf regressions

src/cpp/src/utils.cpp
TolyaTalamanov: Sorry, changed it now, could you check?

tests/python_tests/test_llm_pipeline_static.py
TolyaTalamanov: sure, done

src/cpp/src/utils.cpp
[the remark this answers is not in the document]
smirnov-alexey: I'll try

smirnov-alexey (approved):

smirnov-alexey:
@TolyaTalamanov please wait for merge until I test Ilya's changes. Likely we could remove `cache_mode` logic

src/cpp/src/utils.cpp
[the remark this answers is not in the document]
smirnov-alexey: Unfortunately we get  increased memory consumption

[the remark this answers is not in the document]
ilya-lavrenov: Yet another attempt https://github.com/openvinotoolkit/openvino/pull/29107 😄

smirnov-alexey:
> @TolyaTalamanov please wait for merge until I test Ilya's changes. Likely we could remove `cache_mode` logic

Tested - doesn't work that simple. I suggest we implement it separately

ilya-lavrenov (approved):

PR #1700, opened 2025-02-10
Make `TextStreamer` public & add unit-tests

pavel-esir:
- Check whether text is incomplete should be done as early as possible in order to write to `m_decoded_lengths` relevant information that text in the position of this token is incomplete (is expressed with `-1`).
- Made `TextStreamer` public.
- Also added unit-tests for `TextStreamer` ~which use pybind to make available this object from Python for test purposes. But `TextCallbackStreamer` is still private and visible only for developers.~

Ticket: CVS-148635 CVS-160780


tests/python_tests/test_text_streamer.py
```diff
@@ -0,0 +1,65 @@
...
+    streamer = TextCallbackStreamer(ov_tokenizer, lambda x: accumulated.append(x))
+    accumulated = []
+    for token in tokens:
+        res = streamer.write(token)
+    streamer.end()
+
+    assert ''.join(accumulated) == ov_tokenizer.decode(tokens)
+
+encoded_prompts = [
+    [    2,  3479,   990,   122,   254,     9,    70,   498,   655]  # This tokens caused error in Meta-Llama-3-8B-Instruct
```
pavel-esir, line 47: TODO: add more test cases

CMakeLists.txt
```diff
@@ -97,6 +97,9 @@ endif()
 if(EXISTS "${OpenVINOGenAI_SOURCE_DIR}/tests/cpp")
```
ilya-lavrenov, line 97:
```suggestion
if(EXISTS "${OpenVINOGenAI_SOURCE_DIR}/tests")
```
and add `CMakeLists.txt` to `tests`

.github/workflows/windows.yml
```diff
@@ -243,15 +243,16 @@ jobs:
           . "${{ env.OV_INSTALL_DIR }}/setupvars.ps1"
           python -m pip install ./thirdparty/openvino_tokenizers/[transformers] -r ./tests/python_tests/requirements.txt --find-links ${env:OV_INSTALL_DIR}/wheels
           python -m pytest -v ./tests/python_tests/test_tokenizer.py::test_set_chat_template
+          python -m pytest -v tests/python_tests/test_text_streamer.py
         env:
-          PYTHONPATH: "./build/" # cmd evaluates variables in a different way. Setting PYTHONPATH before setupvars.bat instead of doing that after solves that.
+          PYTHONPATH: "./build/;build/tests/python_tests/pybind_wrapper" # cmd evaluates variables in a different way. Setting PYTHONPATH before setupvars.bat instead of doing that after solves that.
```
ilya-lavrenov, line 248: maybe output binary should also be copied to `./build` to have a single folder with all binaries?

.github/workflows/linux.yml
```diff
@@ -259,6 +259,15 @@ jobs:
           path: ${{ env.INSTALL_DIR }}/*.tar.gz
           if-no-files-found: 'error'
 
+  genai_test_text_streamer:
```
ilya-lavrenov, line 262: why do we need a dedicated job for it?
I would run these tests as part of `LLM / VLM`

.github/workflows/windows.yml
pavel-esir: Agreed.

Just one question. Won't pybind_wrapper be accidentally copied into the product?

ilya-lavrenov: no, `/build` is used during development only

.github/workflows/linux.yml
ilya-lavrenov: in any case, in current job you don't have access to all this files build in `genai_build_cmake` job.

You need to archive those files in build job, then extract in `genai_tests_wheel` and setup extra path to `pywrapper_for_tests` binary

src/cpp/src/text_callback_streamer.cpp
```diff
@@ -26,19 +26,19 @@ StreamingStatus TextCallbackStreamer::write(int64_t token) {
         return set_streaming_status(on_finalized_subword_callback(res.str()));
     }
 
+    constexpr char replacement[] = "\xef\xbf\xbd";  // MSVC with /utf-8 fails to compile � directly with newline in string literal error.
+    if (text.size() >= 3 && text.compare(text.size() - 3, 3, replacement) == 0) {
+        m_decoded_lengths[m_decoded_lengths.size() - 1] = -1;
+        // Don't print incomplete text
+        return set_streaming_status(on_finalized_subword_callback(res.str()));
```
Wovchena, line 33: `IterableStreamer` needs a similar change. Please add `ChunkStreamer` to tests as well.

CMakeLists.txt
Wovchena: Why is this check is needed in the first place? These folders are always here because we put add them

pavel-esir: done

.github/workflows/linux.yml
pavel-esir: We discussed with @mryzhov about sharing utility. And come up the solution to pack this utilities into a separate wheel. In that case we will not need to modify environment variables.

Please take a look and tell whether it looks good now

.github/workflows/windows.yml
pavel-esir: Since introduced a wheel package with `genai_test_utils`. Answered below.

src/cpp/src/text_callback_streamer.cpp
pavel-esir: 1. Turned out `IterableStreamer` already handles this situation correctly.

2. We talked with @as-suvorov, looks like he will add such unit-tests together with implementation of `TextStreamer` with vectorized write.

src/python/py_openvino_genai.cpp
```diff
@@ -131,6 +143,11 @@ PYBIND11_MODULE(py_openvino_genai, m) {
         .def(py::init<>())
         .def("write", &StreamerBase::write, "Write is called every time new token is decoded. Returns a StreamingStatus flag to indicate whether generation should be stopped or cancelled", py::arg("token"))
         .def("end", &StreamerBase::end, "End is called at the end of generation. It can be used to flush cache if your own streamer has one");
+    
+    py::class_<TextStreamer, std::shared_ptr<TextStreamer>>(m, "TextStreamer", text_streamer_docstring)
```
ilya-lavrenov, line 147: I think we need to highlight that `TextStreamerBase` is a base class

I would also create a dedicated `py_text_streamer.cpp` file for all streamers on python side

tests/python_tests/test_text_streamer.py
```diff
@@ -0,0 +1,92 @@
...
+
+tokenizer_model_ids = [
+    "microsoft/phi-1_5",
+    "openbmb/MiniCPM-o-2_6",
+    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+    "NousResearch/Meta-Llama-3-8B-Instruct", # Open analog for gated "meta-llama/Meta-Llama-3-8B-Instruct",
+    # ("black-forest-labs/FLUX.1-dev", dict(subfolder="tokenizer")),  # FLUX.1-dev has tokenizer in subfolder
+]
+
+# Chekc that fix for CVS-157216 works.
```
ilya-lavrenov, line 16:
```suggestion
# Check that fix for CVS-157216 works.
```

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1706,6 +1706,21 @@ class Text2ImagePipeline:
...
+class TextStreamer:
+    """
+    
+    TextStreamer is used to decode tokens into text and call a user-defined callback function.
+    
+    tokenizer: Tokenizer object to decode tokens into text.
+    callback: User-defined callback function to process the decoded text, callback should return either boolean flag or StreamingStatus.
+    
+    """
+    def __init__(self, tokenizer: ..., callback: typing.Callable[[str], bool | StreamingStatus]) -> None:
```
ilya-lavrenov, line 1718: type `...` is unexpected

`Tokenizer` class should be defined before `TextStreamer` in pybind11 `PYBIND11_MODULE(py_openvino_genai, ..)`

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1706,6 +1706,21 @@ class Text2ImagePipeline:
...
+        ...
+    def end(self) -> None:
+        ...
+    def write(self, arg0: int) -> StreamingStatus:
```
ilya-lavrenov, line 1722: please, use explicit argument name via `py::arg` to avoid automatic `arg0` names

src/python/openvino_genai/__init__.pyi
```diff
@@ -46,5 +47,5 @@ from openvino_genai.py_openvino_genai import draft_model
...
-__version__: str
+__all__ = ['Adapter', 'AdapterConfig', 'AggregationMode', 'AutoencoderKL', 'CLIPTextModel', 'CLIPTextModelWithProjection', 'CacheEvictionConfig', 'ChunkStreamerBase', 'ContinuousBatchingPipeline', 'CppStdGenerator', 'DecodedResults', 'EncodedResults', 'FluxTransformer2DModel', 'GenerationConfig', 'GenerationResult', 'Generator', 'Image2ImagePipeline', 'ImageGenerationConfig', 'InpaintingPipeline', 'LLMPipeline', 'PerfMetrics', 'RawPerfMetrics', 'SD3Transformer2DModel', 'Scheduler', 'SchedulerConfig', 'StopCriteria', 'StreamerBase', 'StreamingStatus', 'T5EncoderModel', 'Text2ImagePipeline', 'TextStreamer', 'TokenizedInputs', 'Tokenizer', 'TorchGenerator', 'UNet2DConditionModel', 'VLMPipeline', 'WhisperGenerationConfig', 'WhisperPerfMetrics', 'WhisperPipeline', 'WhisperRawPerfMetrics', 'draft_model', 'get_version', 'openvino', 'os', 'py_openvino_genai']
+__version__: str = '2025.1.0.0-1760-6e22001f986-add_text_streamer_tests'
```
ilya-lavrenov, line 51: having a version is unexpected 

it's explicitly removed to avoid regeneration of this file for each commit because of different hashes

src/python/openvino_genai/py_openvino_genai.pyi
pavel-esir: done

src/python/openvino_genai/py_openvino_genai.pyi
pavel-esir: done

src/python/py_openvino_genai.cpp
pavel-esir: Moved to `py_streamers.cpp` and highlighted that `StreamerBase` is a base class (we don't have TextStreamerBase).

ilya-lavrenov (approved):

as-suvorov (approved):

Wovchena (changes requested):

src/cpp/include/openvino/genai/text_streamer.hpp
```diff
@@ -0,0 +1,42 @@
...
+    StreamingStatus run_callback_if_needed(const std::string& text);
+
+public:
+    StreamingStatus write(int64_t token) override;
+
+    void end() override;
+
+    TextStreamer(const Tokenizer& tokenizer, std::function<CallbackTypeVariant(std::string)> callback);
+
+protected:
```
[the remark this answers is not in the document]
pavel-esir: `ChunkTextCallbackStreamer` uses some of these values. Should we specify different access specifiers?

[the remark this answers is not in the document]
Wovchena: Ok. Keep it `protected`. I think `ChunkTextCallbackStreamer` will be able to call write(vector) when it's added, so accessing fields should go away

Wovchena (approved):

PR #1713, opened 2025-02-11
Use Sampler for StaticWhisperPipeline

eshiryae:
CVS-161218

### DIFF
--- a/src/cpp/src/whisper_pipeline_static.cpp
+++ b/src/cpp/src/whisper_pipeline_static.cpp
@@ -326,7 +326,7 @@ std::pair<ov::genai::EncodedResults, bool> full_decode(ov::Tensor& encoder_hidde
         const auto running_sequences = sequence_group->get_running_sequences();
         OPENVINO_ASSERT(running_sequences.size() == 1u);
         auto last_token = running_sequences.front()->get_generated_ids().back();
-        auto last_idx = running_sequences.front()->get_generated_ids().size();
+        auto last_idx = running_sequences.front()->get_generated_ids().size() - 1;
 
         auto logits = decode_with_past(models.decoder_with_past,
                                        last_token,
--- a/src/cpp/src/whisper_pipeline_static.cpp
+++ b/src/cpp/src/whisper_pipeline_static.cpp
@@ -289,11 +289,10 @@ void stream_generated_tokens(const std::shared_ptr<ov::genai::StreamerBase> stre
     }
 
     std::unordered_map<uint64_t, ov::genai::GenerationOutput> token = handle->read();
-    for (const auto& gen_token : token.begin()->second.generated_ids) {
-        if (streamer_ptr->put(gen_token)) {
-            handle->drop();
-            break;
-        }
+
+    auto streaming_status = streamer_ptr->write(token.begin()->second.generated_ids);
+    if (streaming_status != ov::genai::StreamingStatus::RUNNING) {
+        streaming_status == ov::genai::StreamingStatus::CANCEL ? handle->cancel() : handle->stop();
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -640,14 +665,22 @@ WhisperDecodedResults WhisperPipeline::StaticWhisperPipeline::generate(
...
-                                                            m_models,
-                                                            init_ids,
-                                                            max_new_tokens - output_tokens.size(),
-                                                            return_timestamps,
-                                                            raw_metrics,
-                                                            streamer_ptr);
+        std::vector<int64_t> chunk_init_tokens;
+        chunk_init_tokens.insert(chunk_init_tokens.end(), init_ids.begin(), init_ids.end());
+
+        SequenceGroup::Ptr sequence_group = std::make_shared<SequenceGroup>(0, chunk_init_tokens, config, 1);
```
as-suvorov, line 671: `chunk_init_tokens` looks redundant. Can't we pass `init_ids` direclty?

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -288,49 +278,81 @@ std::vector<int32_t> prepare_init_ids(ov::Tensor& encoder_hidden_state,
...
+                             const bool return_timestamps) {
+    if (return_timestamps || !streamer_ptr || !handle->can_read()) {
+        return;
     }
 
-    if (max_new_tokens == 1) {
-        return {false, output_tokens};
+    std::unordered_map<uint64_t, ov::genai::GenerationOutput> token = handle->read();
+    for (const auto& gen_token : token.begin()->second.generated_ids) {
+        if (streamer_ptr->put(gen_token)) {
```
as-suvorov, line 290: fyi `streamer->put` is deprecated. Stateful pipeline reference: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/whisper/whisper.cpp#L70

src/cpp/src/whisper_pipeline_static.cpp
```diff
@@ -288,49 +278,81 @@ std::vector<int32_t> prepare_init_ids(ov::Tensor& encoder_hidden_state,
...
+    // NB: Only batch=1 is supported now
+    results.scores.resize(1u);
+    results.scores[0] = 0u;
+    results.tokens.resize(1u);
 
-        if (!return_timestamps && streamer && streamer->put(output_token)) {
-            return {true, output_tokens};
-        }
-    }
+    OPENVINO_ASSERT(sequence_group->get_finished_sequences().size() == 1u);
```
sbalandi, line 346: What situation can that there will be more than one sequence here?

TolyaTalamanov: Such cases not supported for NPU yet

src/cpp/src/whisper_pipeline_static.cpp
eshiryae: fixed, thanks!

src/cpp/src/whisper_pipeline_static.cpp
eshiryae: They have different types, std::vector<int64_t> and std::vector<int32_t> accordingly.

as-suvorov: Is it possible to change `init_ids` to `std::vector<int64_t>`? I see `static_cast`s in `prepare_init_ids` are they required?

TolyaTalamanov (approved):

TolyaTalamanov:
@eshiryae Looks good! Should we extend whisper tests?

src/cpp/src/whisper_pipeline_static.cpp
eshiryae: Looks like there's no reason against it. Changed to std::vector<int64_t>.

PR #1715, opened 2025-02-11
Threaded streamer: add tests

as-suvorov:


tests/cpp/CMakeLists.txt
```diff
@@ -13,6 +12,11 @@ if(NOT TARGET gtest)
   FetchContent_MakeAvailable(googletest)
 endif()
 
+FetchContent_Declare(safetensors.h
+    URL https://github.com/hsnyder/safetensors.h/archive/974a85d7dfd6e010558353226638bb26d6b9d756.tar.gz
+    URL_HASH SHA256=9aaf5961609601cf9aaa96582a207bce7c6e5fbf57ed2cc669bb7bde6a937d4b)
+FetchContent_MakeAvailable(safetensors.h)
```
ilya-lavrenov, line 18: let's move `safetensors.h` downloading to `thirdparty/CMakeLists.txt` as well

ilya-lavrenov (approved):

tests/cpp/CMakeLists.txt
as-suvorov: It was my initial intent, but lora_adapter can't find `safetensors.h`:
image

ilya-lavrenov:
For failed builds of samples - please, use larger runners.

ilya-lavrenov (approved):

tests/cpp/threaded_streamer.cpp
```diff
@@ -0,0 +1,139 @@
...
+    threaded_streamer.write(value);
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(200));
+
+    threaded_streamer.end();
+
+    EXPECT_THAT(streamer->tokens, ElementsAreArray(generated_tokens));
+}
+
+TEST_F(MockStreamerBaseFixture, put_end_test) {
```
Wovchena, line 100: Is this test really needed? `general_test` also calls `end()` right after `write()`.

Wovchena (changes requested):

Wovchena:
There's timeout error https://github.com/openvinotoolkit/openvino.genai/actions/runs/13325713617/job/37219499907?pr=1715. `whisper_speech_recognition.py` doesn't use streamer, so the problem doesn't relate to this. But you may want to run an endless loop on your Windows to see if it hangs again on the very exit.

tests/cpp/threaded_streamer.cpp
as-suvorov: I think we can have this test. `general_test` calls `write()` several times. It also calls `threaded_streamer.get_status()` between `write()` and `end()`.

as-suvorov:
> There's timeout error https://github.com/openvinotoolkit/openvino.genai/actions/runs/13325713617/job/37219499907?pr=1715. `whisper_speech_recognition.py` doesn't use streamer, so the problem doesn't relate to this. But you may want to run an endless loop on your Windows to see if it hangs again on the very exit.

Thanks, will check

.github/workflows/causal_lm_cpp.yml
```diff
@@ -674,7 +674,7 @@ jobs:
           PYTHONPATH: "./build"
 
   cpp-chat_sample-ubuntu:
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-24.04-8-cores
```
ilya-lavrenov, line 677:
```suggestion
    runs-on: ubuntu-22.04-8-cores
```

PR #1735, opened 2025-02-14
Parsing model names of deepseek and flan-t5-xxl

wgzintel:
model_utils.py/get_model_name() cannot parse the model name of deepseek and flan-t5-xxl, since text_gen in USE_CASES does not contain the keywords deepseek and flan-t5. Using the model id obtained from USE_CASES to parse the model name.


eaidova:
@wgzintel why it is required to use model path for parcing name? it may does not contains something related at all. for me, it is better to use just parent directory as model name (skipping precision and framework part is subdirs available)

eaidova:
@wgzintel I mean something like that:

```
 from pathlib import Path
model_path = Path("/mnt/llm_irs/llm/models_f499bb00_ww07.4_optimum/deepseek-r1-distill-llama-8b/pytorch/dldt/FP16")
model_name = None
for path_part in [model_path.name] + list(model_path.parents:
    if not (path_part.name in KNOWN_PRECISIONS or path_part.name in KNOWN_FRAMEWORKS): 
          model_name = path_part.name
          break
```

wgzintel:
> @wgzintel I mean something like that:
> 
> ```
> from pathlib import Path
> model_path = Path("/mnt/llm_irs/llm/models_f499bb00_ww07.4_optimum/deepseek-r1-distill-llama-8b/pytorch/dldt/FP16")
> model_name = None
> for path_part in reversed(model_path.parts):
>     if not (path_part in KNOWN_PRECISIONS or path_part in KNOWN_FRAMEWORKS): 
>           model_name = path_part
>           break
> ```
> 
> so idea is filter out pytorch/dldt as KNOWN_FRAMEWORKS and precision as KNOWN_PRECISIONS
> 
> the first directory that not in these lists (deepseek-r1-distill-llama-8b) should be treated as name

Updated the code according to your suggestion

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -10,6 +10,20 @@
 import librosa
 
 
+KNOWN_PRECISIONS = [
```
eaidova, line 13: not sure if this function still in use, but can we reuse precisions mapping here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/llm_bench_utils/model_utils.py#L272

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -10,6 +10,20 @@
...
+        'FP16-INT8', 'INT8', 'INT8_compressed_weights', 'INT8_quantized', 'PT_compressed_weights',
+        'OV_FP32-INT8', 'OV_FP16-INT8',
+        'OV_FP32-INT8_ASYM', 'OV_FP32-INT8_SYM', 'OV_FP16-INT8_ASYM', 'OV_FP16-INT8_SYM', 'OV_FP16-INT8_ASYM_HYBRID',
+        'PT_FP32-INT8', 'PT_FP16-INT8', 'PT_FP32-INT8_ASYM', 'PT_FP32-INT8_SYM', 'PT_FP16-INT8_ASYM', 'PT_FP16-INT8_SYM',
+        'GPTQ_INT4-FP32', 'GPTQ_INT4-FP16', 'INT4',
+        'OV_FP16-INT4_SYM', 'OV_FP16-INT4_ASYM', 'OV_FP32-INT4_SYM', 'OV_FP32-INT4_ASYM',
+        'OV_FP32-4BIT_DEFAULT', 'OV_FP16-4BIT_DEFAULT', 'OV_FP32-4BIT_MAXIMUM', 'OV_FP16-4BIT_MAXIMUM']
+
+
+KNOWN_FRAMEWORKS = ['pytorch', 'ov', 'dldt', 'compressed_weights']
```
eaidova, line 24: compressed_weighs is not framework.
Maybe it is better to have common list of skip model directory:

IGNORE_MODEL_PATH_PARTS = KNOWN_FRAMEWORKS + KNOWN_PRECISIONS + ["com]pressed_weights"]

eaidova: or move it to precisions (weights compression is more related to precision than framework)

eaidova (approved):

wgzintel:
@eaidova  Compare IGNORE_MODEL_PATH_PARTS with lowercase, is it OK to do that?

eaidova:
> @eaidova Compare IGNORE_MODEL_PATH_PARTS with lowercase, is it OK to do that?

@wgzintel not sure that I understand your comment. do you mean if these parts will be found in path, but in different case?

wgzintel:
> > @eaidova Compare IGNORE_MODEL_PATH_PARTS with lowercase, is it OK to do that?
> 
> @wgzintel not sure that I understand your comment. do you mean if these parts will be found in path, but in different case?

I mean convert IGNORE_MODEL_PATH_PARTS to lowercase first, then the path_part are also converted to lowercase, and then they are compared.

def get_model_name_with_path_part(model_name_or_path):
    **IGNORE_MODEL_PATH_PARTS = [x.lower() for x in (KNOWN_FRAMEWORKS + KNOWN_PRECISIONS + OTHER_IGNORE_MODEL_PATH_PARTS)]**
    model_path = Path(model_name_or_path)
    model_name = None
    for path_part in reversed(model_path.parts):
        if not **path_part.lower()** in IGNORE_MODEL_PATH_PARTS:
            model_name = path_part
            break
    return model_name

eaidova:
> > > @eaidova Compare IGNORE_MODEL_PATH_PARTS with lowercase, is it OK to do that?
> > 
> > 
> > @wgzintel not sure that I understand your comment. do you mean if these parts will be found in path, but in different case?
> 
> I mean convert IGNORE_MODEL_PATH_PARTS to lowercase first, then the path_part are also converted to lowercase, and then they are compared.
> 
> def get_model_name_with_path_part(model_name_or_path): **IGNORE_MODEL_PATH_PARTS = [x.lower() for x in (KNOWN_FRAMEWORKS + KNOWN_PRECISIONS + OTHER_IGNORE_MODEL_PATH_PARTS)]** model_path = Path(model_name_or_path) model_name = None for path_part in reversed(model_path.parts): if not **path_part.lower()** in IGNORE_MODEL_PATH_PARTS: model_name = path_part break return model_name

Yes, it may make sense. on some platforms paths are non-case sensitive as I know

wgzintel:
@eaidova  Can this PR be merged?

PR #1742, opened 2025-02-14
[DO NOT MERGE] Whisper pipeline: debug py sample

as-suvorov:


as-suvorov:
It looks like `librosa` hangs sample. Even without genai API calls sample takes about 1m to run:  https://github.com/openvinotoolkit/openvino.genai/actions/runs/13334950348/job/37249247068
Switching to scipy solves the problem.
`librosa` troubleshooting section: https://librosa.org/doc/latest/troubleshooting.html#import-librosa-hangs-indefinitely

as-suvorov:
GH Action timeout increase solves the failing sample. Closing in a favor of sample tests refactoring: https://github.com/openvinotoolkit/openvino.genai/pull/1780.

PR #1768, opened 2025-02-19
Text2ImagePipeline heterogenous compile 

RyanMetcalfeInt8:
To simplify creation of a heterogenous stable diffusion txt2image pipeline, this adds a new API to `Text2ImagePipeline` class:

```
/**
 * Compiles image generation pipeline for given devices for text encoding, denoising, and vae decoding.
 * @param text_encode_device A device to compile text encoder(s) with
 * @param denoise_device A device to compile denoiser (e.g. UNet, SD3 Transformer, etc.) with
 * @param vae_decode_device A device to compile VAE decoder(s) with
 * @param properties A map of properties which affect models compilation
 * @note If pipeline was compiled before, an exception is thrown.
 */
void compile(const std::string& text_encode_device,
             const std::string& denoise_device,
             const std::string& vae_decode_device,
             const ov::AnyMap& properties = {});
``` 

(Need some feedback here.. especially on if we technically need 3 sets of properties.. one per device?)

This API greatly simplifies heterogenous pipeline setup to this:
```
ov::genai::Text2ImagePipeline pipe(models_path);
pipe.reshape(1, width, height, pipe.get_generation_config().guidance_scale);
pipe.compile(text_encoder_device, unet_device, vae_decoder_device);
```

And so now with these changes, heterogenous stable diffusion sample can support all variants of stable diffusion (SD1.5, LCM, XL, SD3, etc.) with the same code. With the old method (creating sub-components and assembling pipeline object), it would have been difficult to achieve this. 

With that said, this PR is tested and working with the following pipelines (with NPU running denoise):
* SD1.5 / LCM
* SDXL

TODO:
* ~~Add python bindings for the new API~~ 
* ~~Update python heterogenous sample~~
**FUTURE WORK** (outside the scope of this PR):
  * Add support for SD3 (this will be separate PR)
      * In general, this requires fixes to this issue: https://github.com/openvinotoolkit/openvino/issues/29113
      * Also some weirdness in current reshape() path I need to figure out.
      * For NPU, this requires a 'batch 1' implementation for Transformer2D -- similar as we did for UNet.
  * Add support for FLUX (this will be separate PR)
  * Add equivalent API for IMAGE2IMAGE / INPAINTING (separate PR's)


src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp
```diff
@@ -12,96 +12,64 @@ namespace genai {
...
+            if (unet == "UNet2DConditionModel") {
+                m_unet = std::make_shared<UNet2DConditionModel>(root_dir / "unet");
+            } else {
+                OPENVINO_THROW("Unsupported '", unet, "' UNet type");
+            }
 
-        const std::string vae = data["vae"][1].get<std::string>();
-        if (vae == "AutoencoderKL") {
-            if (m_pipeline_type == PipelineType::TEXT_2_IMAGE)
-                m_vae = std::make_shared<AutoencoderKL>(root_dir / "vae_decoder", device, *updated_properties);
```
ilya-lavrenov, line 88: when you pass `device` and properties to class ctor, it internally goes with `compile_model(filepath)` call which is faster if model caching is enabled.
See https://docs.openvino.ai/2025/openvino-workflow/running-inference/optimize-inference/optimizing-latency/model-caching-overview.html

RyanMetcalfeInt8: Okay, I could see how passing the model path + device could enable that kind of optimization... but I didn't see any of the SD components (text encoder, unet, vae) actually taking advantage of this, unless I missed it...

For example, these are the constructors:
```
CLIPTextModel::CLIPTextModel(const std::filesystem::path& root_dir,
                             const std::string& device,
                             const ov::AnyMap& properties) :
    CLIPTextModel(root_dir) {
    compile(device, properties);
}

...

UNet2DConditionModel::UNet2DConditionModel(const std::filesystem::path& root_dir,
                                           const std::string& device,
                                           const ov::AnyMap& properties) :
    UNet2DConditionModel(root_dir) {
    compile(device, properties);
}

...

AutoencoderKL::AutoencoderKL(const std::filesystem::path& vae_decoder_path,
                             const std::string& device,
                             const ov::AnyMap& properties)
    : AutoencoderKL(vae_decoder_path) {
    compile(device, *extract_adapters_from_properties(properties));
}
```

So I don't see how this uses `compile_model(filepath)`, but anyway, I'll revert the commit anyway as I can see how my changes would now *prevent* that kind of optimization to be added.

src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp
```diff
@@ -131,21 +99,35 @@ class StableDiffusionXLPipeline : public StableDiffusionPipeline {
         check_image_size(height, width);
 
         const size_t batch_size_multiplier = m_unet->do_classifier_free_guidance(guidance_scale) ? 2 : 1;  // Unet accepts 2x batch in case of CFG
-        m_clip_text_encoder->reshape(batch_size_multiplier);
-        m_clip_text_encoder_with_projection->reshape(batch_size_multiplier);
+        m_clip_text_encoder->reshape(1);
+        m_clip_text_encoder_with_projection->reshape(1);
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: Okay, good point. This isn't correct. For some reason I was thinking that text encoder output was combined in a way to produce the batch-2 input to unet but this isn't the case.

When I run the text2image sample with SDXL, the batch-size of input_id's passed to these text encoders is 1. What I didn't realize is that this is because the default sample only passes positive prompt, and so there's some optimization being enabled there. 

I need to figure out what the correct logic is here (within SDXL's reshape method), as at this point it doesn't know whether user will pass negative prompt or not... 

Perhaps the only reasonable thing to do is to force using empty negative prompt when SDXL reshape is invoked...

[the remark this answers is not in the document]
RyanMetcalfeInt8: Okay, I've changed reshape function to do this:
```
const size_t batch_size_multiplier = m_unet->do_classifier_free_guidance(guidance_scale) ? 2 : 1;  // Unet accepts 2x batch in case of CFG
// We use the batch_size_multiplier as the batch-size used reshape our text encoders here.
// At generate() time though, if user doesn't pass a negative prompt and m_force_zeros_for_empty_prompt is true,
// then the text encoders will get passed batch-1 sizes tensors... which will cause a tensor shape mismatch.
// So, here we force m_force_zeros_for_empty_prompt to false so that text encoders will be invoked with
// 'batch_size_multiplier' batch size at infer() time, regardless of whether user passes negative prompt or not.
m_force_zeros_for_empty_prompt = false;
m_clip_text_encoder->reshape(batch_size_multiplier);
m_clip_text_encoder_with_projection->reshape(batch_size_multiplier);
```

Which seems to work properly...

RyanMetcalfeInt8:
@ilya-lavrenov -- Thanks for the review! I pushed changes to address your comments.

src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp
```diff
@@ -131,21 +165,30 @@ class StableDiffusionXLPipeline : public StableDiffusionPipeline {
         check_image_size(height, width);
 
         const size_t batch_size_multiplier = m_unet->do_classifier_free_guidance(guidance_scale) ? 2 : 1;  // Unet accepts 2x batch in case of CFG
+        // We use the batch_size_multiplier as the batch-size used reshape our text encoders here.
+        // At generate() time though, if user doesn't pass a negative prompt and m_force_zeros_for_empty_prompt is true,
+        // then the text encoders will get passed batch-1 sizes tensors... which will cause a tensor shape mismatch.
+        // So, here we force m_force_zeros_for_empty_prompt to false so that text encoders will be invoked with
+        // 'batch_size_multiplier' batch size at infer() time, regardless of whether user passes negative prompt or not.
+        m_force_zeros_for_empty_prompt = false;
```
ilya-lavrenov, line 173: in this case generated image will be slightly different. 

Text embeddings computed on empty prompt are not the same as filled with zeros. 

I suppose in case of reshaped model, we either need to fix this call 
https://github.com/openvinotoolkit/openvino.genai/blob/c239a536ae48026a03dd6738586dc32df1590856/src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp#L200
and pass `true` instead of `false` (but only when model is reshaped to batch size 2) and perform some post-processing.

Or we can fix `CLIPTextModelWithProjection` itself to check if it was reshaped, then during inference the second batch is filled with dummy input which is ignored after inference and `infer()` returns expected result with respect to `do_classifier_free_guidance` argument:
https://github.com/openvinotoolkit/openvino.genai/blob/c239a536ae48026a03dd6738586dc32df1590856/src/cpp/src/image_generation/models/clip_text_model_with_projection.cpp#L101
I prefer this solution since it incapsulates all "complexity" within `CLIPTextModelWithProjection` block and remaining pipeline does not know about it.

ilya-lavrenov:
Please, also ensure that .pyi files are up to date:
{7E2496C8-45C4-4EF4-A084-19BB01BE6393}


Once you built Python API, pyi files in source tree are automatically updated and you can commit those changes.

samples/python/image_generation/heterogeneous_stable_diffusion.py
```diff
@@ -36,72 +35,41 @@ def main():
...
+    #    "DEVICE_PROPERTIES":
+    #    {
+    #        "CPU": {"CACHE_DIR": "cpu_cache"},
+    #        "GPU": {"CACHE_DIR": "gpu_cache"},
+    #        "NPU": {"CACHE_DIR": "npu_cache"}
+    #    }
+    #}
 
-    pipe = openvino_genai.Text2ImagePipeline.stable_diffusion(scheduler, text_encoder, unet, vae)
+    pipe.compile(args.text_encoder_device, args.unet_device, args.vae_decoder_device, config=properties );
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: This suggestion would work fine for the default properties map:
```
properties = {"CACHE_DIR": "cache"}
```

But if the user were to comment out the device-specific example, then `**properties` wouldn't work here. I might've missed some way to define these such that it's compatible with kwargs expansion style.. but the only way I could get it to work was using the special `config=properties` way (as there is some special handing if key=="config"...
```
# Note that if there are device-specific properties that are needed, they can
    # be added using a "DEVICE_PROPERTIES" entry, like this:
    #properties = {
    #    "DEVICE_PROPERTIES":
    #    {
    #        "CPU": {"CACHE_DIR": "cpu_cache"},
    #        "GPU": {"CACHE_DIR": "gpu_cache"},
    #        "NPU": {"CACHE_DIR": "npu_cache"}
    #    }
    #}
```

src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp
RyanMetcalfeInt8: I agree -- seems like I'd need to apply this logic to both `CLIPTextModelWithProjection` and `CLIPTextModel`. Comparing these classes, they seem just about identical...

RyanMetcalfeInt8: Okay, take a look at set of changes in https://github.com/openvinotoolkit/openvino.genai/pull/1768/commits/97d986ccc1ed96816860c473c9638957d8d121af

Not super thrilled that I needed to split logic between infer() and get_output_tensor() -- but let me know if this is what you had in mind.

Same changes for both CLIP classes. I was tempted to do some refactoring to cut down on the duplicate code (pretty much entire implementation), but didn't want to complicate this PR any more :)

RyanMetcalfeInt8:
> Once you built Python API, pyi files in source tree are automatically updated and you can commit those changes.

I'm trying to figure out how this pyi generation works... In my case, I don't see the .pyi getting generated. Looking at [the CMakeLists.txt](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/python/CMakeLists.txt#L99), it seems like it only enables this if `OpenVINODeveloperPackage_FOUND`. 

Seems like `OpenVINODeveloperPackage` is not found in my case (I am setting up OV env. by sourcing nightly package setupvars.bat). Do I need to build OpenVINO from source to get this to work?

src/cpp/src/image_generation/models/clip_text_model.cpp
```diff
@@ -124,12 +124,25 @@ ov::Tensor CLIPTextModel::infer(const std::string& pos_prompt, const std::string
...
+
+    if (compiled_input_partial_shape.is_dynamic()) {
+        input_ids.set_shape({text_embedding_batch_size, m_config.max_position_embeddings});
+    } else {
+        auto compiled_input_shape = input_ids.get_shape();
+        OPENVINO_ASSERT(compiled_input_shape.size() == 2, "CLIP text encoder model input must have rank of 2");
+        OPENVINO_ASSERT(text_embedding_batch_size <= compiled_input_shape[0],
+                        "text_embedding_batch_size > CLIP text encoder model batch size");
+        OPENVINO_ASSERT(m_config.max_position_embeddings == compiled_input_shape[1],
+                        "max_position_embeddings != what CLIP text encoder model was compiled for.");
```
ilya-lavrenov, line 140: can we print in these asserts that are actual numbers?
E.g. `max_position_embeddings (77) != what CLIP text encoder model was compiled for (55).` is more informative than `max_position_embeddings != what CLIP text encoder model was compiled for.`

src/cpp/src/image_generation/flux_pipeline.hpp
```diff
@@ -256,6 +256,13 @@ class FluxPipeline : public DiffusionPipeline {
         m_transformer->compile(device, *updated_properties);
     }
 
+    void compile(const std::string& text_encode_device,
+                 const std::string& denoise_device,
+                 const std::string& vae_device,
+                 const ov::AnyMap& properties) override {
+        OPENVINO_THROW("not supported yet.");
```
ilya-lavrenov, line 263: can we implement it ? Implementation is trivial 
the same for SD3

src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp
```diff
@@ -12,25 +12,59 @@ namespace genai {
 class StableDiffusionXLPipeline : public StableDiffusionPipeline {
 public:
     StableDiffusionXLPipeline(PipelineType pipeline_type, const std::filesystem::path& root_dir) :
-        StableDiffusionPipeline(pipeline_type, root_dir) {
```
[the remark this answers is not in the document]
ilya-lavrenov: good catch! thank you

src/cpp/src/image_generation/stable_diffusion_xl_pipeline.hpp
ilya-lavrenov: > the duplicate code

PR https://github.com/openvinotoolkit/openvino.genai/pull/1809

samples/python/image_generation/heterogeneous_stable_diffusion.py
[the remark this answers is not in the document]
ilya-lavrenov: ok, at least we need to drop `;` at the end 😄

ilya-lavrenov (approved):

ilya-lavrenov:
build_jenkins

PR #1791, opened 2025-02-24
Replace ov::element::undefined with dynamic

olpipi:
ov::element::undefined is deprecated https://github.com/openvinotoolkit/openvino/pull/28766


as-suvorov:
There is already PR with a fix https://github.com/openvinotoolkit/openvino.genai/pull/1785

PR #1794, opened 2025-02-24
[GHA] Replaced cpp-speculative_decoding_lm-ubuntu

mryzhov:
Replaced cpp-speculative_decoding_lm-ubuntu pipeline


Wovchena (approved):

Wovchena:
Greedy sample doesn't use streaming but speculative decoding does. It's either a streamer who refused to print the last incomplete token or speculative decoding algo which didn't put the last token to streamer. @iefode, can you check?

iefode (approved):

PR #1796, opened 2025-02-24
Switch workflows to aks-linux-mediumworker

ababushk:


.github/workflows/coverity.yml
```diff
@@ -30,10 +30,10 @@ jobs:
     defaults:
       run:
         shell: bash
-    runs-on: aks-linux-2-cores-8gb
+    runs-on: aks-linux-mediumworker
```
mryzhov, line 33: Could we use a more understandable name for the runners?

mryzhov (approved):

PR #1810, opened 2025-02-26
Fixed VLM metrics test.

popovaan:
Ticket: CVS-160580

### DIFF
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -273,27 +273,31 @@ def test_sampling(config, cache):
 @pytest.mark.nightly
 def test_perf_metrics(cache):
     import numpy as np
-    import datetime
+    from time import perf_counter_ns
     models_path = get_ov_model("katuni4ka/tiny-random-minicpmv-2_6", cache)
 
     images = [get_image_by_link(image_links[0])]
+    max_new_tokens = 30
 
-    start_time = datetime.datetime.now()
+    start_time = perf_counter_ns()
     pipe = VLMPipeline(models_path, "CPU")
-    result = pipe.generate(prompts[0], images=images, generation_config=GenerationConfig(max_new_tokens=30))
-    generate_time = (datetime.datetime.now() - start_time).total_seconds() * 1000
+    start_generate = perf_counter_ns()
+    load_time = (start_generate - start_time) / 1000000
+
+    result = pipe.generate(prompts[0], images=images, generation_config=GenerationConfig(max_new_tokens=max_new_tokens))
+    generate_time = (perf_counter_ns() - start_generate) / 1000000
 
     perf_metrics = result.perf_metrics
 
     assert perf_metrics is not None
 
-    assert 0 < perf_metrics.get_load_time() < generate_time
+    assert 0 < perf_metrics.get_load_time() < load_time
     assert 0 < perf_metrics.get_num_generated_tokens() < generate_time
     assert 0 < perf_metrics.get_num_input_tokens() < generate_time
     assert 0 < perf_metrics.get_ttft().mean < generate_time
     assert 0 < perf_metrics.get_tpot().mean < generate_time
     assert 0 < perf_metrics.get_ipot().mean < generate_time
-    assert 0 < perf_metrics.get_throughput().mean < generate_time
+    assert 0 < perf_metrics.get_throughput().mean < max_new_tokens / ((generate_time - perf_metrics.get_ttft().mean) / 1000)
     assert 0 < perf_metrics.get_inference_duration().mean < generate_time
     assert 0 < perf_metrics.get_generate_duration().mean < generate_time
     assert 0 < perf_metrics.get_tokenization_duration().mean < generate_time
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -107,6 +107,7 @@ def test_vlm_continuous_batching_generate_vs_add_request(config, cache):
     ov_pipe = VLMPipeline(models_path, "CPU", scheduler_config=scheduler_config, **get_default_llm_properties())
     generation_config = config
     generation_config.max_new_tokens = 30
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_vlm_pipeline.py
```diff
@@ -269,34 +269,36 @@ def test_sampling(config, cache):
...
     import numpy as np
+    import datetime
     models_path = get_ov_model("katuni4ka/tiny-random-minicpmv-2_6", cache)
 
     images = [get_image_by_link(image_links[0])]
 
+    start_time = datetime.datetime.now()
     pipe = VLMPipeline(models_path, "CPU")
     result = pipe.generate(prompts[0], images=images, generation_config=GenerationConfig(max_new_tokens=30))
+    generate_time = (datetime.datetime.now() - start_time).total_seconds() * 1000
```
Wovchena, line 284: Use https://docs.python.org/3/library/time.html#time.perf_counter_ns instead. I don't think `datetime` is monotonic

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -269,34 +269,36 @@ def test_sampling(config, cache):
...
-    assert 0 < perf_metrics.get_tokenization_duration().mean < 100
-    assert 0 < perf_metrics.get_detokenization_duration().mean < 10
-    assert 0 < perf_metrics.get_prepare_embeddings_duration().mean < 100
+    assert 0 < perf_metrics.get_load_time() < generate_time
+    assert 0 < perf_metrics.get_num_generated_tokens() < generate_time
+    assert 0 < perf_metrics.get_num_input_tokens() < generate_time
+    assert 0 < perf_metrics.get_ttft().mean < generate_time
+    assert 0 < perf_metrics.get_tpot().mean < generate_time
+    assert 0 < perf_metrics.get_ipot().mean < generate_time
+    assert 0 < perf_metrics.get_throughput().mean < generate_time
```
Wovchena, line 296: Throughput is measured in tokens according to benchmark_genai.cpp. It's not a duration. You should probably compare that with `max_new_tokens/generate_time`

tests/python_tests/test_vlm_pipeline.py
Wovchena, line 284: I remember that it was me who proposed that upper bound, but there's a way to refine the idea. Can you add another timepoint between `VLMPipeline()` and `generate()` so the test for `get_load_time()` becomes stricter.

Wovchena (changes requested):

tests/python_tests/test_vlm_pipeline.py
popovaan: Done.

popovaan: Done.

tests/python_tests/test_vlm_pipeline.py
popovaan: I used formula `max_new_tokens/(generate_time - TTFT)`, because we subtract TTFT when calculate throughput, so I did the same here. Otherwise estimated throughput is too low. 
https://github.com/openvinotoolkit/openvino.genai/blob/894e2aadff505cfd03b74560f6c90a8106bdad70/src/cpp/src/perf_metrics.cpp#L107-L110

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -269,34 +269,40 @@ def test_sampling(config, cache):
...
+    max_new_tokens = 30
 
+    start_time = perf_counter_ns()
     pipe = VLMPipeline(models_path, "CPU")
-    result = pipe.generate(prompts[0], images=images, generation_config=GenerationConfig(max_new_tokens=30))
+    start_generate = perf_counter_ns()
+    load_time = (start_generate - start_time) / 1000000
+
+    result = pipe.generate(prompts[0], images=images, generation_config=GenerationConfig(max_new_tokens=max_new_tokens))
+    generate_time = (perf_counter_ns() - start_generate) / 1000000
```
Wovchena, line 288:
```suggestion
    result = pipe.generate(prompts[0], images=images, generation_config=GenerationConfig(max_new_tokens=max_new_tokens))
    generate_time = (perf_counter_ns() - start_generate) / 1_000_000.0
    load_time = (start_generate - start_time) / 1_000_000.0
```
Strengthen the test :) (and make more readable)

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -269,34 +269,40 @@ def test_sampling(config, cache):
...
+    assert 0 < perf_metrics.get_num_input_tokens() < generate_time
+    assert 0 < perf_metrics.get_ttft().mean < generate_time
+    assert 0 < perf_metrics.get_tpot().mean < generate_time
+    assert 0 < perf_metrics.get_ipot().mean < generate_time
+    assert 0 < perf_metrics.get_throughput().mean < max_new_tokens / ((generate_time - perf_metrics.get_ttft().mean) / 1000)
+    assert 0 < perf_metrics.get_inference_duration().mean < generate_time
+    assert 0 < perf_metrics.get_generate_duration().mean < generate_time
+    assert 0 < perf_metrics.get_tokenization_duration().mean < generate_time
+    assert 0 < perf_metrics.get_detokenization_duration().mean < generate_time
+    assert 0 < perf_metrics.get_prepare_embeddings_duration().mean < generate_time
```
Wovchena, line 305: Can you do the same asserts with `std` members? `generate_time*generate_time` would be the upper bound. Not sure if it can be `(generate_time/2)^2`.

Wovchena (changes requested):

tests/python_tests/test_vlm_pipeline.py
popovaan: Done.

tests/python_tests/test_vlm_pipeline.py
popovaan: Fixed.

Wovchena (approved):

PR #1834, opened 2025-03-03
update supported model_ids

eaidova:
allow search for model_type from config for phi, opt and gpt models

CVS-163344


sbalandi (approved):

tools/llm_bench/llm_bench_utils/config_class.py
```diff
@@ -62,6 +62,7 @@
         't5',
         'falcon',
         "glm",
+        "gpt",
         'gpt-',
```
sbalandi, line 66: if we have `gpt`, do we need also `gpt-` and `gpt2` ?

eaidova: this is needed for backward compatibility. I have additional task for cleanup configuration types list, it will be done separately (this onle only for unblocking validation teams)

PR #1846, opened 2025-03-05
LLM: release plugin once pipeline is removed and WA for GPU

p-durandin:
Based on https://github.com/openvinotoolkit/openvino.genai/pull/1627 and WA for GPU oneDNN cache clean added


src/cpp/src/icontinuous_batching.cpp
```diff
@@ -16,8 +16,9 @@ std::string add_image_tags_to_prompt(const std::string& prompt, const std::vecto
 
 namespace ov::genai {
 
-template<class... Ts> struct overloaded : Ts... {using Ts::operator()...;};
-template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
+ContinuousBatchingPipeline::IContinuousBatchingPipeline::~IContinuousBatchingPipeline() {
+    m_tokenizer = {};
+}
```
Wovchena, line 21: `IContinuousBatchingPipeline` is `ContinuousBatchingImpl`'s parent. The order of destructor calls is reverse:
1. ~ContinuousBatchingImpl()
2. ~IContinuousBatchingPipeline()

By the time `~IContinuousBatchingPipeline()` is called, `utils::release_core_plugin(m_device)` from `ContinuousBatchingImpl()` had already been executed. Given that the solution is satisfactory, you could have defined `~IContinuousBatchingPipeline() = default;` and get the same result. But maybe a better solution would be to move `utils::release_core_plugin(m_device)` to `IContinuousBatchingPipeline()` (`m_device` is already there). This would fix the call order, remove the need to manually clear `ContinuousBatchingImpl`'s members and enable clearing for other children: `PromptLookupImpl` and `SpeculativeDecodingImpl`.

Wovchena (changes requested):

src/cpp/src/icontinuous_batching.cpp
sbalandi: @Wovchena I don't have right to modify that PR, so I have applied comment here https://github.com/openvinotoolkit/openvino.genai/pull/2102

ilya-lavrenov:
replaced by https://github.com/openvinotoolkit/openvino.genai/pull/2102

PR #1852, opened 2025-03-06
Add remove adapters for LLMpipeline

wenyi5608:
tickets: CVS-160984
Intermediate solution because adapters aren't actually unloaded.


src/cpp/src/continuous_batching_impl.cpp
```diff
@@ -150,10 +150,11 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::initialize_pipeline(
     const ov::AnyMap& properties,
     const std::vector<KVHeadConfig>& kv_cache_config) {
     // apply LoRA
-    auto filtered_properties = extract_adapters_from_properties(properties, &m_generation_config.adapters);
-    if (m_generation_config.adapters) {
-        m_generation_config.adapters->set_tensor_name_prefix("base_model.model.model.");
```
ilya-lavrenov, line 155: why have you removed storing adapters in `m_generation_config` ?

src/cpp/src/lora_adapter.cpp
```diff
@@ -1067,6 +1067,27 @@ struct AdapterControllerImpl {
...
+                        if (adapter1 == adapter2) {
+                            current_config.remove(adapter1);
+                            need_full_apply = true;
+                        }
+                    }
+                }
+            }
+
+            for (const auto& adapter2 : adapters2) {
+                adapter2.~Adapter();
```
ilya-lavrenov, line 1087: we do we need to call explicit dtor?

src/cpp/src/continuous_batching_impl.cpp
wenyi5608: If we used m_generation_config.adapters to record the adapters. When the program returns pipe.generate, the adapter will be recorded in m_generation_config. adapter. Using ~Adater() to release the buffer will not work.  In addition, m_generation_config.adapters is not used.

src/cpp/src/lora_adapter.cpp
wenyi5608: Do you have any suggestions?

ilya-lavrenov: why does not automatic dtor calling work here?

src/cpp/src/continuous_batching_impl.cpp
ilya-lavrenov: > m_generation_config.adapters is not used

Users can get adapters set to ctor and change them later:
```c++
Pipeline pipe(path, adapter_config);
// a lot of code, etc
...
// later on
pipe.generate(prompt1)
auto new_adapter_config = *pipe.get_generation_config().adapters;
new_adapter_config.remove_adapter(..) // any modifications here
pipe.generate(prompt2, adapters(new_adapter_config)) // pass updated config
```

slyalin (changes requested):
What is your use case that demands `remove_adapters` functionality?

Why not call `apply` with adapter removed from the config? If it works for you, you can propagate `apply` to the upper level API like we have already done in the form of `set_adapters` method implemented for part of the pipelines.

Can we just wait for the next `generate` call when unused adapters can be deallocated automatically if you don't use them in generation config and the adapter objects are not kept anywhere else?

Will using mmap for adapter content be helpful in your use case and eliminate the need for `remove_adapters`?

src/cpp/src/continuous_batching_impl.cpp
wenyi5608: Yes,  user can use "new_adapter_config.remove_adapter(..)" to remove adapters in m_generation_config. So this part of code will not be modified.

src/cpp/src/lora_adapter.cpp
```diff
@@ -1067,6 +1067,27 @@ struct AdapterControllerImpl {
...
+        const auto &adapters1 = current_config.get_adapters(), adapters2 = config.get_adapters();
+
+        if (adapters2.size() > 0) {
+            if (adapters1.size() > 0) {
+                // if current adpater need to remove, remove from current_config.adapters
+                for (const auto& adapter1 : adapters1) {
+                    for (const auto& adapter2 : adapters2) {
+                        if (adapter1 == adapter2) {
+                            current_config.remove(adapter1);
+                            need_full_apply = true;
```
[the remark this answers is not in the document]
wenyi5608: > What is your use case that demands remove_adapters functionality?
For example, the customer has four LoRA adapters,  namely LoRA_A, LoRA_B, LoRA_C, LoRA_D.  They don't want to load four LoRA adapters (LoRA_A, LoRA_B, LoRA_C, LoRA_D) at the same time. When they use LoRA_A pipe.generate, they only load the LoRA_A adapter; when they switch to LoRA_B pipe.generate, they will load LoRA_B and unload LoRA_A to save memory.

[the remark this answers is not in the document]
wenyi5608: > Why not call apply with adapter removed from the config? If it works for you, you can propagate apply to the upper level API like we have already done in the form of set_adapters method implemented for part of the pipelines.
Now the form of remove_adapters is the same  as the form of set_adapters method implemented for part of the pipelines.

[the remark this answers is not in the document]
wenyi5608: >Can we just wait for the next generate call when unused adapters can be deallocated automatically if you don't use them in generation config and the adapter objects are not kept anywhere else?
The adapter objects are allocated by user. generation_config and m_adapter_controller just reference the adapter objects. Should the adapter objects be deallocated by user himself?

src/cpp/src/lora_adapter.cpp
wenyi5608: The adapter objects are allocated by user. generation_config and m_adapter_controller just reference the adapter objects. Should the adapter objects be deallocated by user himself?

wenyi5608: I have deleted this dtor.

src/cpp/src/lora_adapter.cpp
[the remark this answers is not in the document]
wenyi5608: > Tensors from the removed adapter may still be used in the model states. This is applicable for cases when only one adapter is attached to the model and no precision change is applied before setting the state. In this case, a tensor set into `set_state` is created on top of the shared memory loaded in Adapter object, and, depending on a plugin implementation, it may or may not be used from the original destination. The dependency will be removed from the model state when you call `apply` method. This is a theory, I haven't checked that. CPU guys said they are copying each time. And if they are copying, this is another issue: even if you deallocate memory for the adapter object it is still allocated inside the plugin as a part of the state.

When the user calls the `apply` method, the adapter object is copied to the model state.  If the user continues using this LoRA adapter(LoRA_A) to generate, the adapter object(LoRA_A) will not be used and the model state will not be changed.  Case one: when the user changes to another LoRA adapter(LoRA_B), the adapter object(LoRA_B) will be copied to the model state.  Case two: when the user changes to an empty LoRA,  the dynamic dimension of the model state will be set to 0,  and then changes to LoRA adapter(LoRA_A), the adapter object(LoRA_A) will be copied to the model state.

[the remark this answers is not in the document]
slyalin: > > What is your use case that demands remove_adapters functionality?
> 
> For example, the customer has four LoRA adapters, namely LoRA_A, LoRA_B, LoRA_C, LoRA_D. They don't want to load four LoRA adapters (LoRA_A, LoRA_B, LoRA_C, LoRA_D) at the same time. When they use LoRA_A pipe.generate, they only load the LoRA_A adapter; when they switch to LoRA_B pipe.generate, they will load LoRA_B and unload LoRA_A to save memory.

This is exactly what mmap'ed LoRA adapters will give you: adapter file content will be loaded and unloaded automatically depending on the system memory availability, like OS filesystem cache. In the same way as OpenVINO IR bin file is loaded. The first PR that introduces this functionality is here: https://github.com/openvinotoolkit/openvino/pull/29175. There will be another one in GenAI that will allocate safetensors with that mmaped functionality.

The question is about the period between two generates, one with LORA_A and the next one with LORA_B: do you want to minimize memory consumed by LORA_A in this period, or it doesn't matter? If yes, you need an API like `set_adapters` that will allow you to (_theoretically_) deallocate plugin memory for Adapter states.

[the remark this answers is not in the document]
slyalin: > > Can we just wait for the next generate call when unused adapters can be deallocated automatically if you don't use them in generation config and the adapter objects are not kept anywhere else?
> 
> The adapter objects are allocated by user. generation_config and m_adapter_controller just reference the adapter objects. Should the adapter objects be deallocated by user himself?

User side should remove any references to the adapter, I believe it is the same for your PR. If user bothers about memory allocated by the adapters, yes, it is required to explicitly get rid of all adapter references, including the adapter object on the user side. Again, with mmap functionality, when it is merged, this amount of memory doesn't affect the amount of memory that can be allocated for other needs in the system and can be counted as free memory.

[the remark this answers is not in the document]
wenyi5608: > > > What is your use case that demands remove_adapters functionality?
> > 
> > 
> > For example, the customer has four LoRA adapters, namely LoRA_A, LoRA_B, LoRA_C, LoRA_D. They don't want to load four LoRA adapters (LoRA_A, LoRA_B, LoRA_C, LoRA_D) at the same time. When they use LoRA_A pipe.generate, they only load the LoRA_A adapter; when they switch to LoRA_B pipe.generate, they will load LoRA_B and unload LoRA_A to save memory.
> 
> This is exactly what mmap'ed LoRA adapters will give you: adapter file content will be loaded and unloaded automatically depending on the system memory availability, like OS filesystem cache. In the same way as OpenVINO IR bin file is loaded. The first PR that introduces this functionality is here: [openvinotoolkit/openvino#29175](https://github.com/openvinotoolkit/openvino/pull/29175). There will be another one in GenAI that will allocate safetensors with that mmaped functionality.

I used the mmap of this PR(openvinotoolkit/openvino#29175) to read lora file and found that the memory size of PrivateWorkingSetSize was reduced, but the size of WorkingSetSize increased.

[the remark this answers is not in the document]
slyalin: The difference in PrivateWorkingSetSize shows that the approach works. The saved amount contributes to WorkingSetSize, because (in my rough understanding) mmaped regions are counted there. I don't know why it is such a big increase, probably something else starts contributing there. Please share the diff in the code which is associated with the logs. Is this picture reproduced each time for both variants?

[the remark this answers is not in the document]
wenyi5608: > The difference in PrivateWorkingSetSize shows that the approach works. The saved amount contributes to WorkingSetSize, because (in my rough understanding) mmaped regions are counted there. I don't know why it is such a big increase, probably something else starts contributing there. Please share the diff in the code which is associated with the logs. Is this picture reproduced each time for both variants?

Using mmap to read lora adapter file can reduce memory size. The increase in  WorkingSetsize is due to different OpenVINO versions. When I used the OpenVINO version after this PR(https://github.com/openvinotoolkit/openvino/pull/29175), WorkingSetSize increases regardless of whether mmap is used to read Lora Adapter file. 
Without Lora, the memory data of different OpenVINO versions are as follows:
minicpm_without_lora

With Lora, the memory data of different OpenVINO versions are as follows:
minicpm_with_lora

wenyi5608:
> Will using mmap for adapter content be helpful in your use case and eliminate the need for `remove_adapters`?

I had used mmap for adapter content and eliminate 'romove_adapters'.   The user need to encrypt the lora adapter, so I added new API to read the lora adapter from the buffer.

src/cpp/src/lora_adapter.cpp
```diff
@@ -139,18 +134,30 @@ ConstantMap read_safetensors(const std::filesystem::path& filename) {
 
         OPENVINO_ASSERT(
             ov::shape_size(shape) <= tensor.end_offset_bytes - tensor.begin_offset_bytes,
-            "Tensor shape ", ov::shape_size(shape), " for tensor \"", name, "\" from Safetensors file \"", filename, "\" doesn't match the expected tensor size ",
+            "Tensor shape ", ov::shape_size(shape), " for tensor \"", name, "\" from Safetensors file \"", "safetensor", "\" doesn't match the expected tensor size ",
```
slyalin, line 137: Please don't lose a better diagnostics for a file due to the code unification. Most users will still use adapters from files, so this message becomes less useful for them.

slyalin (changes requested):

src/cpp/src/lora_adapter.cpp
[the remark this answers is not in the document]
slyalin: @wenyi5608, are you sure that you are using exactly the same build flow and the same environment for all the variants in the comparison table? Stable ~1GB increase is not something that can be ignored and requires investigation. @olpipi, have you noticed something like that after your PR?

src/cpp/src/icontinuous_batching.cpp
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "icontinuous_batching.hpp"
+#include "lora_helper.hpp"
```
iefode, line 5: Please remove unused headers

src/cpp/src/lora_adapter.cpp
```diff
@@ -115,19 +115,14 @@ struct AutoSafetensor: public safetensors_File {
...
 
-
-// Reads a file with a given filename expecting Safetensors file format.
-// The data is read to a solid memory block and the function returns a map of OV Constants allocated on top of that block.
 // The key in the map is a tensor name and the Constant uses a region of memory from the memory block.
 // Each Constant holds a shared pointer to the block in the runtime info.
 // The memory block will be deallocated when the last Constant is destroyed.
-ConstantMap read_safetensors(const std::filesystem::path& filename) {
-    auto buffer = read_file_helper(filename);
+ConstantMap safetensors_init(const ov::Tensor & safetensor) {
```
iefode, line 121: Minor, the name is not so clear for understanding.... Is it possible to name smt like:
* `ov_tensor_to_constant_map`
* `safetensor_to_constant_map`..

src/cpp/src/lora_adapter.cpp
```diff
@@ -139,18 +134,30 @@ ConstantMap read_safetensors(const std::filesystem::path& filename) {
...
+// Reads a file with a given filename expecting Safetensors file format.
+// The file data is mmaped to tensor.
+ConstantMap read_safetensors(const std::filesystem::path& filename) {
+    auto safetensor = ov::read_tensor_data(filename);
+
+    return safetensors_init(safetensor);
+}
+
+// Reads a safetensor and creates a constantmap from the memory.
+ConstantMap read_safetensors(const ov::Tensor& safetensor) {
```
iefode, line 158: Why do we need this function? It it whole duplication of `safetensors_init(safetensor)`. We can move second one to public section in this case

src/cpp/include/openvino/genai/lora_adapter.hpp
```diff
@@ -37,6 +37,7 @@ class OPENVINO_GENAI_EXPORTS Adapter {
     Adapter(const std::shared_ptr<AdapterImpl>& pimpl);
 public:
     explicit Adapter(const std::filesystem::path& path);
+    explicit Adapter(const ov::Tensor& safetensor);
```
ilya-lavrenov, line 40: please, add such ctor to Python API as well

wenyi5608: Python API has added.

src/cpp/src/icontinuous_batching.cpp
wenyi5608: Has removed unused headers

src/cpp/src/lora_adapter.cpp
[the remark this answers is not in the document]
wenyi5608: @olpipi @slyalin @ilya-lavrenov OpenVINO 2025.1.0RC3 fixed the bug of Workingsetsize increasing significantly.  I tested the memory footprint of adapter reading from file or using mmap to read file. When creating the adapter and llm pipe using mmap to read the adapter file, both PrivateWorkingSetSize and Workingsetsize are reduced. When pipeline uses lora to generate results, using mmap to read adapter file will reduce PrivateWorkingSetSize but increase Workingsetsize.
 
mtl_minicpm_mmap_compare

ilya-lavrenov:
build_jenkins

Wovchena (approved):

Wovchena:
build_jenkins

src/cpp/src/lora_adapter.cpp
```diff
@@ -136,20 +108,22 @@ ConstantMap read_safetensors(const std::filesystem::path& filename) {
         ov::Shape shape(tensor.shape, tensor.shape + tensor.n_dimensions);
         void* ptr = tensor.ptr;     // FIXME: needs a non-constant pointer because Tensor doesn't accept a constant pointer
 
-        OPENVINO_ASSERT(
```
ilya-lavrenov, line 139: why is this assert removed?

wenyi5608: We use ov::tensor to parse lora tensor, which is inconsistent with the original method of judging whether the file size is out of bounds.

ilya-lavrenov:
build_jenkins

PR #1860, opened 2025-03-07
[tokenizer] Fix setting max_legth and special tokens flags

pavel-esir:
- `MakePaddingSatateful` was skipping left padding graphs during match because it was checking for `Subtract` on the wrong `1st input`, while it should've do this on `0th input`. We did miss that, because models list in test by default was testing only right padding - now added explicitly check left padding as well.
- There was an issue with `add_special_tokens`, Even when we called `decode` it was rewriting `m_add_special_tokens` values as if we called `set_state`, therefore if we subsequently requested `encode` even with different params Tokenizer wrongly assumed that InferReuest already has that param in state and did not modify it and we end up having wrong output, e.g. below on the second call there should been 1 at the begining. Fixed this issue as well.

image

CVS-163154

### DIFF
--- a/src/cpp/src/tokenizer.cpp
+++ b/src/cpp/src/tokenizer.cpp
@@ -137,10 +137,8 @@ public:
 
     // To change the adding special tokens mode we use a statefull subgraph,
     // this flag holds the current state value of the CompiledModel.
-    bool m_add_special_tokens = true;
-    bool m_skip_special_tokens = true;
-    bool m_pad_to_max_length = false;
-    std::optional<int32_t> m_max_length;
+    ov::AnyMap m_state_flags;
+
     bool m_older_than_24_5 = false;
 
     int64_t m_pad_token_id = -1;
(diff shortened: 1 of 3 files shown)


src/cpp/src/tokenizer.cpp
```diff
@@ -137,10 +137,8 @@ class Tokenizer::TokenizerImpl {
 
     // To change the adding special tokens mode we use a statefull subgraph,
     // this flag holds the current state value of the CompiledModel.
-    bool m_add_special_tokens = true;
-    bool m_skip_special_tokens = true;
-    bool m_pad_to_max_length = false;
-    std::optional<int32_t> m_max_length;
+    ov::AnyMap m_state_flags;
```
pavel-esir, line 140: This solution is more concise and general. We will not need to add new field for every new tokenization parameter!

sbalandi (approved):
LGTM

ilya-lavrenov (approved):

PR #1890, opened 2025-03-11
Harcode long prompts

sstrehlk:


ilya-lavrenov:
Please, re-open if it's still valid.

PR #1897, opened 2025-03-11
Add perf metrics for CB VLM

pavel-esir:
- Added passing CB perf metric for VLM cases as well. 
- Numbers are similar to vanilla pipeline
image

Ticket CVS-163346


ilya-lavrenov:
inference duration varies significantly, do we know a reason?

pavel-esir:
> inference duration varies significantly, do we know a reason?

don't know the reason yet

pavel-esir:
> inference duration varies significantly, do we know a reason?

It turned out that here https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching_impl.cpp#L404-L410 CB metrics measure not pure inference time, but also preparation/setting of tensors,

If i measure more selectively here https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching_impl.cpp#L280-L283 i get values comparable to vanilla SDPA.

ilya-lavrenov:
> preparation/setting of tensors

Does it mean that this step takes so much time and requires optimization?

pavel-esir:
@ilya-lavrenov corrected inference duration measurements. Now it's more fine grained, and measures only forward time instead of full time of `step()`. After that proportions are similar to vanilla SDPA.

image

src/cpp/src/continuous_batching_impl.cpp
```diff
@@ -403,11 +406,13 @@ ContinuousBatchingPipeline::ContinuousBatchingImpl::generate(const std::vector<o
         try {
             const auto infer_start = std::chrono::steady_clock::now();
             step();
+            // The very first prefill step we don't generate new tokens, but still inference took place,
```
ilya-lavrenov, line 409: whether we generate new tokens or not, depends on max_batch_size.
If max_batch_size < prompt_len, then prompt is split into chunks and processed one by one. 

So, comment can be adjust to reflect actual logic, because says that "first inference does not generate tokens" is not always correct.

pavel-esir: adjusted comment accordingly

ilya-lavrenov (approved):

PR #1928, opened 2025-03-17
[DRAFT] Support of `lm_head` and `embed_tokens`

iefode:
Ticket:
* https://jira.devtools.intel.com/browse/CVS-159931

### DIFF
--- a/src/cpp/src/lora_common.hpp
+++ b/src/cpp/src/lora_common.hpp
@@ -18,13 +18,14 @@ namespace utils {
 
 template <typename T>
 struct LoRAParts {
-    T alpha, A, B;
+    T alpha, A, B, weight;
 
     LoRAParts() = default;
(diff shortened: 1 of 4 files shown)


src/cpp/src/lora_common.hpp
```diff
@@ -18,13 +18,14 @@ namespace utils {
 
 template <typename T>
 struct LoRAParts {
-    T alpha, A, B;
+    T alpha, A, B, weight;
```
slyalin, line 21: Technically, `weight` is not a LoRA, so I would suggest to consider representing separately a normal LoRA triplet (`A`, `B`, and `alpha`) and `weights` at a higher level: where we obtain grouped tensors from `safetensors` file by extending the `value` type in that map. Otherwise, adding `weight` here, in the basic template, will leak it everywhere.

as-suvorov:
Addressed in: https://github.com/openvinotoolkit/openvino.genai/pull/2395

PR #1933, opened 2025-03-18
Added GPTQ/AWQ support with HF Transformers

AlexKoff88:


tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -82,18 +85,39 @@ def load_text_llamacpp_pipeline(model_dir):
...
 
+def load_text_hf_pipeline(model_id, device):
+    model_kwargs = {}
+
+    if not torch.cuda.is_available or device.lower() == "cpu":
+        config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
+        is_gptq = False
+        is_awq = False
+        if getattr(config, "quantization_config", None):
+            is_gptq = config.quantization_config["quant_method"] == "gptq"
```
eaidova, line 96: not sure if it is common case, but I sow some models where quant_method is not specified in quatization_config (not sure if we need to handle it, it anyway can not be loaded by transformers, only by mlx), e.g. https://huggingface.co/mlx-community/QwQ-32B-8bit/blob/main/config.json#L22

eaidova (approved):

tools/who_what_benchmark/whowhatbench/model_loaders.py
AlexKoff88: thanks. I also noticed this but it is for custom models, not for Transformers.

PR #1924, opened 2025-03-14
[GHA] Introduce SDL pipeline

mryzhov:
- Added pipeline to consolidate different lint/dependencies checks 
- Added Trivy vulnerability scanning
- Added Dependency Review scanning
- Moved wwb tests

PR #1954, opened 2025-03-21
[GHA] Replaced genai-tools pipeline

mryzhov:
Moved genai-tools to the Linux Tools workflow


eaidova:
@mryzhov I see only llm bench tests here, what about wwb that was also defined in genai_tools?

mryzhov:
> @mryzhov I see only llm bench tests here, what about wwb that was also defined in genai_tools?

They have already been replaced in https://github.com/openvinotoolkit/openvino.genai/pull/1924

ilya-lavrenov (approved):

eaidova (approved):

PR #1959, opened 2025-03-24
Check provider fix

akladiev:
https://github.com/openvinotoolkit/openvino/pull/29650/files

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -44,7 +44,7 @@ jobs:
     continue-on-error: true
 
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/windows.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     continue-on-error: true
 
     steps:
-    - uses: openvinotoolkit/openvino/.github/actions/openvino_provider@master
+    - uses: openvinotoolkit/openvino/.github/actions/openvino_provider@@refs/pull/29650/head
```
mryzhov, line 47:
```suggestion
    - uses: openvinotoolkit/openvino/.github/actions/openvino_provider@refs/pull/29650/head
```

PR #1962, opened 2025-03-24
[GHA] Setup ov_cache

mryzhov:
- Set HF_HOME to cache HF models
- Set OV_CACHE to cache converted models an test data
- Enabled OV_CACHE cleanup on daily base


.github/workflows/cleanup_caches.yml
```diff
@@ -0,0 +1,36 @@
...
+
+    steps:
+      - name: Pre-Collecting Cache Info
+        run: |
+          echo "Cache info: "
+          du -h -d2 ${{ env.OV_CACHE }}
+
+      - name: Cleanup cache
+        run: |
+          echo "Delete cache files if they have not been used in over 30 days"
```
akladiev, line 30: Is it really supposed to cleanup cache files not used over 30 days? I see that it just wipes out the whole directory once a day (probably not very safe to do in runtime).

.github/workflows/windows.yml
```diff
@@ -483,9 +484,6 @@ jobs:
       INSTALL_DIR: ${{ github.workspace }}/install
       SRC_DIR: ${{ github.workspace }}/src
       BUILD_DIR: ${{ github.workspace }}/build
-      HF_HOME: C:/mount/caches/huggingface
-      OV_CACHE: ${{ github.workspace }}/ov_cache
-      CLEANUP_CACHE: 1
```
akladiev, line 488: Just in case - what this variable was supposed to do?

akladiev (changes requested):

.github/workflows/cleanup_caches.yml
mryzhov: yes, because we need to convert the models any time when the OV runtime is updated, so the easiest way to remove the once per day

.github/workflows/windows.yml
mryzhov: HF_HOME - cache for downloaded HF models and datasets, OV_CACHE - internal cache for OV converted models, CLEANUP_CACHE - this option enables automatic cleanup of the test content(models, media, etc) when the tests already used it.

.github/workflows/cleanup_caches.yml
mryzhov, line 30:
```suggestion
          echo "Delete the cached test data"
```

.github/workflows/windows.yml
akladiev: Looks like you removed CLEANUP_CACHE: 1 here, is it expected?

mryzhov: that is correct, because we will remove the test data by schedule (once per day)

.github/workflows/linux.yml
```diff
@@ -24,7 +24,8 @@ env:
   SCCACHE_SERVER_PORT: 35555
   SCCACHE_CACHE_SIZE: 30G
   SCCACHE_AZURE_KEY_PREFIX: genai/ubuntu/22_04/x64
-  HF_HOME: /mount/caches/huggingface
+  # HF_HOME: /mount/caches/huggingface
```
mryzhov, line 27:
```suggestion
  HF_HOME: /mount/caches/huggingface 
```

.github/workflows/windows.yml
```diff
@@ -22,6 +22,8 @@ env:
   CMAKE_CXX_COMPILER_LAUNCHER: ccache
   CMAKE_C_COMPILER_LAUNCHER: ccache
   CCACHE_MAXSIZE: 500Mi
+  # HF_HOME: C:/mount/caches/huggingface
```
mryzhov, line 25:
```suggestion
  HF_HOME: C:/mount/caches/huggingface
```

as-suvorov (approved):

PR #2000, opened 2025-04-03
[GHA] Use developer openvino packages

mryzhov:
Build genai packages using openvino developer package to pickup compilation flags


.github/workflows/linux.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - name: CMake Build
         run: |
           source ${{ env.OV_INSTALL_DIR }}/setupvars.sh
-          cmake -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -S ${{ env.SRC_DIR}} -B ${{ env.BUILD_DIR }}
+          cmake -DOpenVINODeveloperPackage_DIR=${{ env.OV_INSTALL_DIR }}/developer_package/cmake -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -S ${{ env.SRC_DIR}} -B ${{ env.BUILD_DIR }}
```
Wovchena, line 119: The comment in `CMakeLists.txt` says
> \# Find OpenVINODeveloperPackage first to compile with SDL flags

Is that the reason for the chanage?

mryzhov: Yes, that is a prerequisite for starting to use GHA artifacts for distribution.

ilya-lavrenov (approved):

.github/workflows/windows.yml
```diff
@@ -131,7 +131,7 @@ jobs:
         shell: pwsh
         run: |
           ${{ env.OV_INSTALL_DIR }}/setupvars.ps1
-          cmake -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -S ${{ env.SRC_DIR }} -B ${{ env.BUILD_DIR }}
+          cmake -DOpenVINODeveloperPackage_DIR=${{ env.OV_INSTALL_DIR }}/developer_package/cmake -DCMAKE_BUILD_TYPE=${{ matrix.build-type }} -S ${{ env.SRC_DIR }} -B ${{ env.BUILD_DIR }}
```
ilya-lavrenov, line 134: looks like gtest become a dynamic library 
{9D5D8850-768B-4676-9F48-E4AC09CA46A3}

I think we need to set it explicitly as static here:
https://github.com/openvinotoolkit/openvino.genai/blob/55ce655af15195f58a02d495650ec780af0a32ca/tests/cpp/CMakeLists.txt#L6-L13

tests/cpp/CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ file(GLOB tests_src "*.cpp")
 
 set(TEST_TARGET_NAME "tests_continuous_batching")
 
+set(BUILD_SHARED_LIBS OFF)
```
ilya-lavrenov, line 26: I suppose it should be right after `set(INSTALL_GTEST OFF CACHE BOOL "")` otherwise, `FetchContent_MakeAvailable` will not respect this value

tests/cpp/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ file(GLOB tests_src "*.cpp")
 
 set(TEST_TARGET_NAME "tests_continuous_batching")
 
-set(BUILD_SHARED_LIBS OFF)
+set(BUILD_SHARED_LIBS OFF CACHE BOOL "")
```
ilya-lavrenov, line 26: don't change global state

tests/cpp/CMakeLists.txt
mryzhov: reverted

ilya-lavrenov:
Please, fix macOS wheel build.

PR #2002, opened 2025-04-03
Update tokenizers

mryzhov:
Update tokenizers  to pick up changes from https://github.com/openvinotoolkit/openvino_tokenizers/pull/461


mryzhov:
Depends on https://github.com/openvinotoolkit/openvino.genai/pull/2002

PR #2006, opened 2025-04-04
Set xfail to failing tests.

popovaan:


Wovchena (approved):

tests/python_tests/samples/test_greedy_causal_lm.py
```diff
@@ -11,6 +11,7 @@
 class TestGreedyCausalLM:
     @pytest.mark.llm
     @pytest.mark.samples
+    @pytest.mark.xfail(reason="CVS-165581")
```
mryzhov, line 14: it is better to skip only the tests with TinyLlama-1.1B-Chat-v1.0 model

mryzhov:
Maybe it would be easier to remove this model or temporary replace it with another one?

tests/python_tests/samples/test_greedy_causal_lm.py
popovaan: There are currently two types of errors, related to TinyLlama-1.1B-Chat-v1.0 (CVS-165581) and related to json.decoder.JSONDecodeError (CVS-165582) on windows. In this test both errors happen, so I skipped it entirely but I agree that it is better to skip tests more precisely. 
Corrected in this commit: https://github.com/openvinotoolkit/openvino.genai/pull/2006/commits/4773e6bc5d89ed409962a5db0009e38143d10bf0

mryzhov (approved):

PR #2019, opened 2025-04-08
[do not merge] [do not review] output speculative decoding metrics

xufang-lisa:


ilya-lavrenov:
Please, re-open if it's still valid

PR #2020, opened 2025-04-08
Use OV_CACHE for python tests

as-suvorov:


tests/python_tests/conftest.py
```diff
@@ -1,28 +1,58 @@
...
+import logging
+from utils.constants import get_ov_cache_models_dir
+
+# Configure logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def setup_and_teardown():
```
Wovchena, line 13: Can you merge two `conftest.py` instead?

as-suvorov: In order to reuse `conftest.py` download and convert approaches should be aligned between API and sample tests. I propose to address it in another PR.

mryzhov (approved):

mryzhov:
samples tests will be also fixed by https://github.com/openvinotoolkit/openvino.genai/pull/2021

Wovchena (approved):

PR #2028, opened 2025-04-09
samples: Adds optional device selection to some samples

apram0d:
Updates CLI argument parsing in C++ and Python samples to allow optional specification of the device (e.g., CPU, GPU, NPU).


Wovchena:
Why do you need this change?

The absence of device flag is intentional and invites readers to modify the implementation.

apram0d:
@Wovchena
1. All the samples in openvino toolkit https://github.com/openvinotoolkit/openvino/tree/master/samples have device as parameter.
2. While I was testing three use cases as in VLM, LLM and stable-diffusion on different devices CPU,GPU and NPU it felt much easier and intuitive to pass as an argument (defaulting to CPU) rather than making code change every time.

And that's the reason I did not change for every sample, only one sample from each to try with.

Wovchena:
1. A newer sample doesn't have the device: https://github.com/openvinotoolkit/openvino/tree/master/samples/python/benchmark/bert_benchmark
2. The hope is that you tested other things as well, not just device modification.

apram0d:
@Wovchena 

For whisper_speech_recognition I tried with whisper small model for both python and C++ sample on CPU, GPU and NPU devices.
For chat_sample I tried with llama 3.1 3B model for both python and C++ samples on CPU, GPU and NPU devices.
For vlm I tried with phi-3.5-vision model for both python and C++ sample on CPU device. I have issues with GPU device since its a pre-prod platform but its able to compile and offload to GPU I can see in GPU utilization tool. NPU device is not supported for VLM.

Wovchena:
Should I do anything with this report?

apram0d:
@Wovchena 

The samples that I've changed have only take model directory or prompt. Passing device is the only change that I added, so I tested thoroughly the changes along with existing parameters which are working good as before. Do you want me to test any other things? Would happy to know what else you are expecting :)

Wovchena:
build_jenkins

popovaan:
There are also other samples with hardcoded device:

- text2image samples
- LoRA, multinomial, prompt lookup LLM samples and so on.

Do we need to change them accordingly?

apram0d:
@popovaan , yes they can be. But, for now as part of functional validation of three use cases as mentioned,

> 2. While I was testing three use cases as in VLM, LLM and stable-diffusion on different devices CPU,GPU and NPU it felt much easier and intuitive to pass as an argument (defaulting to CPU) rather than making code change every time.

the above samples will be sufficient for validating the models and include as part of the release. If further needed, I can raise PR again for those models.

PR #2046, opened 2025-04-14
[WWB]: move diffusers imports closer to usage

eaidova:
CVS-165875

in some cases diffusers import may breaks transformers models loading, however that should not be used at all.
Moved diffusers closer to their usage for avoid possible issues

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/utils.py
+++ b/tools/who_what_benchmark/whowhatbench/utils.py
@@ -26,6 +26,11 @@ def patch_diffusers():
 
 @contextmanager
 def mock_torch_cuda_is_available(to_patch):
+    try:
+        # import bnb before patching for avoid attempt to load cuda extension during first import
+        import bitsandbytes as bnb # noqa: F401 
+    except ImportError:
+        pass
     original_is_available = torch.cuda.is_available
     if to_patch:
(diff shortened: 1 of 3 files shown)


andreyanufr (approved):

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -22,11 +20,17 @@ def new_randn_tensor(
...
 def patch_diffusers():
+    from diffusers.utils import torch_utils
     torch_utils.randn_tensor = new_randn_tensor
 
 
 @contextmanager
 def mock_torch_cuda_is_available(to_patch):
+    try:
+        # import bnb before patching for avoid attempt to load cuda extension during first import
+        import bitsandbytes as bnb # noqa: F401
```
eaidova, line 31:
```suggestion
        import bitsandbytes as bnb # noqa: F401
```

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -22,11 +20,17 @@ def new_randn_tensor(
...
```
eaidova, line 31:
```suggestion
        import bitsandbytes as bnb  # noqa: F401
```

PR #2050, opened 2025-04-14
[llm bench] allow loading onnx models using optimum-intel

eaidova:


tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -627,121 +627,125 @@ def is_genai_available(log_msg=False):
             return False
     return True
 
+def get_genai_chunk_streamer():
```
eaidova, line 630:
```suggestion

def get_genai_chunk_streamer():
```

sbalandi (approved):

PR #2057, opened 2025-04-15
[RAG] Add text embedding pipeline

as-suvorov:
todo:
* add samples


src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,304 @@
...
+    };
+
+    void start_embed_documents_async(std::vector<std::string>& texts) {
+        if (!m_config.embed_instruction) {
+            return start_embed_async(texts);
+        }
+
+        auto formatted_texts = format_texts(texts);
+        start_embed_async(formatted_texts);
+    };
```
Wovchena, line 90: `format_texts()` already handles `!m_config.embed_instruction`. I think you can remove `!m_config.embed_instruction` part from `start_embed_documents_async()`

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+    void start_embed_documents_async(std::vector<std::string>& texts);
+    std::vector<EmbeddingResult> wait_embed_documents();
+
+    /**
+     * @brief Computes embedding for a query
+     */
+    EmbeddingResult embed_query(std::string& text);
+
+    void start_embed_query_async(std::string& text);
```
Wovchena, line 71: Add const to all references in the interface. The current interface won't compile for `start_embed_query_async("");`

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,304 @@
...
+
+    std::string format_query(const std::string& text) {
+        if (!m_config.query_instruction) {
+            return text;
+        }
+
+        return *m_config.query_instruction + text;
+    }
+
+    std::vector<EmbeddingResult> to_embedding_result(const Tensor last_hidden_state) {
```
Wovchena, line 167:
```suggestion
    std::vector<EmbeddingResult> to_embedding_result(const Tensor& last_hidden_state) {
```

src/cpp/src/tokenizer.cpp
```diff
@@ -455,7 +455,7 @@ class Tokenizer::TokenizerImpl {
 
     TokenizedInputs encode(std::vector<std::string>& prompts, const ov::AnyMap& tokenization_params = {}) {
         OPENVINO_ASSERT(m_ireq_queue_tokenizer, "Either openvino_tokenizer.xml was not provided or it was not loaded correctly. "
-                                                "Tokenizer::encode is not available");
+                        "Tokenizer::encode is not available");
```
Wovchena, line 458: There are other asserts written with such indentation in this file. Revert it.

src/python/py_rag.cpp
```diff
@@ -0,0 +1,139 @@
...
+            if (config.has_value()) {
+                return std::make_unique<TextEmbeddingPipeline>(models_path,
+                                                               device,
+                                                               *config,
+                                                               pyutils::kwargs_to_any_map(kwargs));
+            }
+            return std::make_unique<TextEmbeddingPipeline>(models_path, device, pyutils::kwargs_to_any_map(kwargs));
+        }),
+        py::arg("models_path"),
+        "folder with openvino_model.xml and openvino_tokenizer[detokenizer].xml files",
```
Wovchena, line 126:
```suggestion
        "folder with openvino_model.xml and openvino_tokenizer.xml files",
```
Do you care about detokenizer?
Also add this to C++ docs

.github/workflows/windows.yml
```diff
@@ -422,6 +422,9 @@ jobs:
           - name: 'API tests'
             cmd: 'tests/python_tests/test_continuous_batching.py tests/python_tests/test_generation_config.py tests/python_tests/test_sampling.py tests/python_tests/test_text_streamer.py'
             timeout: 60
+          - name: 'Rag tests'
+            cmd: 'tests/python_tests/test_rag.py'
+            timeout: 15
```
Wovchena, line 427:
```suggestion
            timeout: 30
```
Failed in https://github.com/openvinotoolkit/openvino.genai/actions/runs/14644342760/job/41097717431?pr=2057

Wovchena (changes requested):

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @brief Computes embeddings for a vector of texts
+     */
+    std::vector<EmbeddingResult> embed_documents(std::vector<std::string>& texts);
+
+    void start_embed_documents_async(std::vector<std::string>& texts);
+    std::vector<EmbeddingResult> wait_embed_documents();
+
+    /**
+     * @brief Computes embedding for a query
```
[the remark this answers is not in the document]
as-suvorov: `query_instruction` and `text_instruction` can be applied for `query` and `text`. This naming used in [llamaindex](https://github.com/run-llama/llama_index/blob/main/llama-index-core/llama_index/core/base/embeddings/base.py#L116) and [langchain](https://python.langchain.com/api_reference/core/embeddings/langchain_core.embeddings.embeddings.Embeddings.html).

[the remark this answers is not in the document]
Wovchena: I can't find `query_instruction` or `text_instruction` in the given links.

I cant find definitions of query, text or document in this links neither.

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: Good point. There is no const signature for [tokenizer.encode](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/tokenizer.hpp#L136). 
```c++
    TokenizedInputs encode(std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
    TokenizedInputs encode(std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
    TokenizedInputs encode(std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
```

So as this API is intended to be used with large amount of text I removed const from the interfaces to get rid of copying before tokenizer.

I thought it was the same reason why llm_pipeline has no const in interface
```c++
using StringInputs = std::variant<std::string, std::vector<std::string>>;
DecodedResults generate(
    StringInputs inputs,
    OptionalGenerationConfig generation_config = std::nullopt,
    StreamerVariant streamer=std::monostate()
);
```
But I just realized that there is no reference as well.

I'll add const back to interfaces but it would require copying before `tokenizer.encode`.

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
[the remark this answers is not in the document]
as-suvorov: Above links shows the naming conventions.
Here is the link for instructions: https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/embeddings/llama-index-embeddings-huggingface-openvino/llama_index/embeddings/huggingface_openvino/base.py#L28

tests/python_tests/test_rag.py
```diff
@@ -0,0 +1,197 @@
...
+        normalize=True,
+        pooling_type=TextEmbeddingPipeline.PoolingType.MEAN,
+    )
+    TextEmbeddingPipeline(
+        models_path,
+        "CPU",
+        normalize=True,
+        pooling_type=TextEmbeddingPipeline.PoolingType.MEAN,
+        PERFORMANCE_HINT_NUM_REQUESTS=2,
+    )
```
apaniukov, line 139: Please parametrize the test with different model configs and model_ids: `TextEmbeddingPipeline(models_path, "CPU", **config)`

tests/python_tests/test_rag.py
```diff
@@ -0,0 +1,197 @@
...
+    print(f"Max error: {max_error}")
+    assert (
+        np.abs(np_genai_result - np_langchain_result).max() < 1e-6
+    ), f"Max error: {max_error}"
+
+
+@pytest.mark.precommit
+def test_constructors():
+    model_id = "BAAI/bge-small-en-v1.5"
+    _, _, models_path = download_and_convert_embeddings_models(model_id)
```
apaniukov, line 117: This function call should be in a fixture.

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1890,6 +1890,94 @@ class Text2ImagePipeline:
...
+        @property
+        def name(self) -> str:
+            ...
+        @property
+        def value(self) -> int:
+            ...
+    def __init__(self, models_path: os.PathLike, device: str, config: TextEmbeddingPipeline.Config | None = None, **kwargs) -> None:
+        """
+        TextEmbeddingPipeline class constructor.
+        models_path (os.PathLike): Path to the model file.
```
apaniukov, line 1952: We should settle on one docstring format in genai.

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,304 @@
...
+#include "openvino/op/divide.hpp"
+#include "openvino/op/maximum.hpp"
+#include "openvino/op/multiply.hpp"
+#include "openvino/op/normalize_l2.hpp"
+#include "openvino/op/reduce_mean.hpp"
+#include "openvino/op/reduce_sum.hpp"
+#include "openvino/op/shape_of.hpp"
+#include "openvino/op/slice.hpp"
+#include "openvino/op/squeeze.hpp"
+#include "openvino/op/unsqueeze.hpp"
```
apaniukov, line 21: You can use opset to get the operations, like here: https://github.com/openvinotoolkit/openvino_tokenizers/blob/979cba29ffc1f99d3c53e28a2c29b750390e59d9/src/tensorflow_translators.cpp#L9

tests/python_tests/test_rag.py
as-suvorov: Test cannot be fully parametrized with `TextEmbeddingPipeline(models_path, "CPU", **config)` as positional args checked as well.

tests/python_tests/test_rag.py
as-suvorov: I agree, but this function is not a fixture at the moment.

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
Wovchena: If you care about copy, you can add the overload with `&&`. Alternatively you can try casting const away because we know that it's `ov::Tensor`'s API weirdness rather than actual string modification.

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace ov {
+namespace genai {
+
+using EmbeddingResult = std::variant<std::vector<float>, std::vector<int8_t>, std::vector<uint8_t>>;
+
+class OPENVINO_GENAI_EXPORTS TextEmbeddingPipeline {
+public:
+    enum class OPENVINO_GENAI_EXPORTS PoolingType {
+        CLS = 0,
```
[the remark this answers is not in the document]
as-suvorov: Docstings added. `CLS` stands for `classification`

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,87 @@
...
+        bool normalize = false;
+        std::optional<std::string> query_instruction;
+        std::optional<std::string> embed_instruction;
+
+        Config() = default;
+        explicit Config(const ov::AnyMap& properties);
+    };
+
+    /**
+     * @brief Constructs an pipeline from xml/bin files, tokenizers and configuration in the same dir.
```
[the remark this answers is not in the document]
as-suvorov: Applied

src/cpp/src/rag/text_embedding_pipeline.cpp
as-suvorov: It was a copy optimization attempt. I will apply it in further prs after testing if applicable.

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: I returned const to interfaces. I would rather address it on tokenizer interface, will sync with @apaniukov to discuss.

src/cpp/src/rag/text_embedding_pipeline.cpp
as-suvorov: Is reference required for `Tensor`? I guess there is no data copy when passing by value.

src/cpp/src/tokenizer.cpp
as-suvorov: Reverted

src/python/py_rag.cpp
as-suvorov: Fixed

.github/workflows/windows.yml
as-suvorov: Timeout increased

src/cpp/src/rag/text_embedding_pipeline.cpp
as-suvorov: Opset headers used instead of exact ops.

Wovchena (approved):

tests/python_tests/test_rag.py
```diff
@@ -0,0 +1,211 @@
...
+    assert (
+        np.abs(np_genai_result - np_langchain_result).max() < 1e-6
+    ), f"Max error: {max_error}"
+
+
+@pytest.mark.parametrize(
+    "download_and_convert_embeddings_models",
+    [{"model_id": "BAAI/bge-small-en-v1.5"}],
+    indirect=True,
+)
```
apaniukov, line 116: You can pass just a list of `model_id`, so in a fixture it will be:
```python
@pytest.fixture()
def download_and_convert_embeddings_models(request):
    model_id = request.param
    return _download_and_convert_model(model_id, OVModelForFeatureExtraction)
```

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include "openvino/genai/tokenizer.hpp"
+
+namespace ov {
+namespace genai {
+
+using EmbeddingResult = std::variant<std::vector<float>, std::vector<int8_t>, std::vector<uint8_t>>;
+
+class OPENVINO_GENAI_EXPORTS TextEmbeddingPipeline {
+public:
+    enum class OPENVINO_GENAI_EXPORTS PoolingType {
```
ilya-lavrenov, line 19:
```suggestion
    enum class PoolingType {
```
`export` is not required for classes / enums, etc which don't have any methods in .cpp files

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,93 @@
...
+    /**
+     * @brief Constructs a pipeline from xml/bin files, tokenizer and configuration in the same dir.
+     *
+     * @param models_path Path to the dir model xml/bin files, tokenizer
+     * @param device optional device
+     * @param plugin_config optional plugin_config properties
+     */
+    TextEmbeddingPipeline(const std::filesystem::path& models_path,
+                          const std::string& device,
+                          const std::optional<Config>& config = std::nullopt,
```
ilya-lavrenov, line 50: do we need `std::optional` here? I suppose instead of `TextEmbeddingPipeline(path, device, std::nullopt, properties)` users can use the ctor below `TextEmbeddingPipeline(path, device, properties)` which don't accept `Config`

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    /**
+     * @brief Constructs a pipeline from xml/bin files, tokenizer and configuration in the same dir.
+     *
+     * @param models_path Path to the dir model xml/bin files, tokenizer
+     * @param device optional device
+     * @param plugin_config optional plugin_config properties
+     */
+    TextEmbeddingPipeline(const std::filesystem::path& models_path,
```
ilya-lavrenov, line 48: do we need `template <typename... Properties` version for this ctor as well?

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+    template <typename... Properties,
+              typename std::enable_if<ov::util::StringAny<Properties...>::value, bool>::type = true>
+    TextEmbeddingPipeline(const std::filesystem::path& models_path,
+                          const std::string& device,
+                          Properties&&... properties)
+        : TextEmbeddingPipeline(models_path, device, ov::AnyMap{std::forward<Properties>(properties)...}) {}
+
+    /**
+     * @brief Computes embeddings for a vector of texts
```
ilya-lavrenov, line 65: please, add documentation for parameters 

and other methods / API entities within this file.

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,290 @@
...
+    return properties_copy;
+}
+}  // namespace
+
+namespace ov {
+namespace genai {
+using utils::read_anymap_param;
+
+TextEmbeddingPipeline::Config::Config(const ov::AnyMap& properties) {
+    read_anymap_param(properties, ov::genai::max_length.name(), max_length);
```
ilya-lavrenov, line 35:
```suggestion
    read_anymap_param(properties, max_length.name(), max_length);
```
as we are already within `ov::genai` namespace

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,290 @@
...
+    std::string format_query(const std::string& text) {
+        if (!m_config.query_instruction) {
+            return text;
+        }
+
+        return *m_config.query_instruction + text;
+    }
+
+    std::vector<EmbeddingResult> to_embedding_result(const Tensor& last_hidden_state) {
+        const auto last_hidden_state_data = last_hidden_state.data<float>();
```
ilya-lavrenov, line 154: as I see from API standpoint, `EmbeddingResult` can also store 8 bit types. I think you need to handle such case here, then

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+        return processor.build();
+    }
+
+    /**
+     * CLS pooling slices first element from seq_length dimension
+     * [batch_size, seq_length, hidden_size] -> [batch_size, seq_length[0], hidden_size]
+     * [10, 5, 768] -> [10, 768]
+     */
+    std::shared_ptr<op::Op> get_cls_pooling_op(const ov::Output<ov::Node>& last_hidden_state_node) {
```
ilya-lavrenov, line 200: as I see, these post-processing functions don't use class state. Should they be separate free functions, defined in unnamed namespace?

src/python/py_rag.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include "tokenizers_path.hpp"
+
+namespace py = pybind11;
+using ov::genai::EmbeddingResult;
+using ov::genai::TextEmbeddingPipeline;
+
+namespace pyutils = ov::genai::pybind::utils;
+
+void init_rag_pipelines(py::module_& m) {
+    m.doc() = "Pybind11 binding for RAG Pipelines";
```
ilya-lavrenov, line 21: looks like here we override global documentation title for `src/python/openvino_genai/py_openvino_genai.pyi`

I think we need to drop `m.doc()` from all `init_XXX` functions and keep only this description:
https://github.com/openvinotoolkit/openvino.genai/blob/782109739f8a92fbad82819a6584912ace6f92ea/src/python/py_openvino_genai.cpp#L73

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: We need `std::optional` here to support creation with no config or properties. Current ctors supports next calls:
```c++
    ov::genai::TextEmbeddingPipeline pipeline(models_path, device);
    ov::genai::TextEmbeddingPipeline pipeline(models_path, device, config);
    ov::genai::TextEmbeddingPipeline pipeline(models_path,
                                               device,
                                               ov::genai::normalize(true),
                                               ov::cache_dir(".ov_cache"));
```

src/cpp/src/rag/text_embedding_pipeline.cpp
as-suvorov: We are in `Config` struct method where we have `max_length` struct member. So we need to differentiate between `max_length` property and struct member.

tests/python_tests/test_rag.py
```diff
@@ -0,0 +1,205 @@
...
+def run_gc_after_test():
+    """
+    Fixture to run garbage collection after each test class.
+    This is a workaround to minimize memory consumption during tests and allow the use of less powerful CI runners.
+    """
+    yield
+    gc.collect()
+
+
+@pytest.fixture(scope="module")
```
apaniukov, line 34:
```suggestion
@pytest.fixture
```

as-suvorov: What's the intention here?

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: export removed

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: I think we have it covered with existing templated ctor:
```c++
    ov::genai::TextEmbeddingPipeline pipeline(models_path,
                                               device,
                                               ov::genai::normalize(true),
                                               ov::cache_dir(".ov_cache"));
```

src/cpp/src/rag/text_embedding_pipeline.cpp
as-suvorov: Yeah, makes sense, moved to unnamed namespace.

src/python/py_rag.cpp
as-suvorov: Thanks, fixed

as-suvorov:
@ilya-lavrenov @Wovchena I have a question regarding return value. Would it makes sense to add and return `using EmbeddingResults = std::variant<std::vector<std::vector<float>>, std::vector<std::vector<int8_t>>, std::vector<std::vector<int8_t>>>;` instead of `EmbeddingResult`. Seems to provide better DX with a single `std::get`. What do you think?
```c++
    // using EmbeddingResult = std::variant<std::vector<float>, std::vector<int8_t>, std::vector<uint8_t>>;
    std::vector<ov::genai::EmbeddingResult> embeddings = pipeline.embed_documents(documents);

    for (const ov::genai::EmbeddingResult& embedding : embeddings) {
        const std::vector<float>& float_embedding = std::get<std::vector<float>>(embedding);
    }

    // using EmbeddingResults = std::variant<std::vector<std::vector<float>>, std::vector<std::vector<int8_t>>, std::vector<std::vector<int8_t>>>;
    ov::genai::EmbeddingResults embeddings_v2 = pipeline.embed_documents_v2(documents);
    const std::vector<std::vector<float>>& float_embeddings_v2 = std::get<std::vector<std::vector<float>>>(embeddings_v2);
```

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
ilya-lavrenov: discussed offline

ilya-lavrenov (approved):

Wovchena:
File /runner/_work/openvino.genai/openvino.genai/src/src/python/openvino_genai/py_openvino_genai.pyi is outdated and need to be regenerated with pybind11-stubgen

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: Docstrings added

as-suvorov:
> maybe we can keep both? and use them dependent on input arguments to a function
Sure, that was my proposal. Implemted as:
```c++
EmbeddingResults embed_documents(const std::vector<std::string>& texts);
EmbeddingResult embed_query(const std::string& text);
```

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,172 @@
...
+         */
+        std::optional<size_t> max_length;
+
+        /**
+         * @brief Pooling strategy applied to model output tensor
+         */
+        PoolingType pooling_type = PoolingType::CLS;
+
+        /**
+         * @brief If 'true' normalization is applied to embeddings
```
Wovchena, line 44:
```suggestion
         * @brief If 'true' L2 normalization along axis 1 is applied to embeddings
```
Maybe the member and the Property should be renamed to `normize_l2` as well to make it more verbose.

.github/workflows/linux.yml
```diff
@@ -369,6 +369,9 @@ jobs:
           - name: 'API tests'
             cmd: 'tests/python_tests/test_continuous_batching.py tests/python_tests/test_generation_config.py tests/python_tests/test_sampling.py tests/python_tests/test_text_streamer.py'
             timeout: 60
+          - name: 'Rag tests'
+            cmd: 'tests/python_tests/test_rag.py'
+            timeout: 15
```
Wovchena, line 374:
```suggestion
            timeout: 30
```
Align with windows. Imagine someone not familiar with the tests. If they happened to read Windows first, they would expect it to be the same for the other two.

.github/workflows/mac.yml
```diff
@@ -373,6 +373,9 @@ jobs:
           - name: 'API tests'
             cmd: 'tests/python_tests/test_continuous_batching.py tests/python_tests/test_generation_config.py tests/python_tests/test_sampling.py tests/python_tests/test_text_streamer.py'
             timeout: 60
+          - name: 'Rag tests'
+            cmd: 'tests/python_tests/test_rag.py'
+            timeout: 15
```
Wovchena, line 378:
```suggestion
            timeout: 30
```

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
```diff
@@ -0,0 +1,172 @@
...
+                          Properties&&... properties)
+        : TextEmbeddingPipeline(models_path, device, ov::AnyMap{std::forward<Properties>(properties)...}) {}
+
+    /**
+     * @brief Computes embeddings for a vector of texts
+     */
+    EmbeddingResults embed_documents(const std::vector<std::string>& texts);
+
+    /**
+     * @brief Asynchronously computes embeddings for a vector of texts
```
Wovchena, line 120:
```suggestion
     * @brief Asynchronously computes embeddings for a vector of texts. Only one method of async family can be active.
```
Same for start_embed_query_async()

Wovchena (changes requested):

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
as-suvorov: The normalization algorithm is implementation detail. [llama_index](https://github.com/run-llama/llama_index/blob/main/llama-index-integrations/embeddings/llama-index-embeddings-huggingface-optimum-intel/llama_index/embeddings/huggingface_optimum_intel/base.py#L18) and [langchain](https://github.com/langchain-ai/langchain-community/blob/main/libs/community/langchain_community/embeddings/openvino.py#L150) do not specify explicitly what type of normalization would be applied.

.github/workflows/linux.yml
as-suvorov: Aligned

src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
Wovchena: This implementation detail affects the result so it can't be fully hidden. An example of the detail that can be hidden is sorting algorithm in sort(). And even then it's specified if the order of equal elements is guaranteed to be preserved.

As for llama_index and langchain, you can add a note to the docstring that L2 is aligned with them. You can also submit them PRs adding this to see their opinion.

Wovchena: Mean comes before L2, so the last point is fine

as-suvorov: I added L2 norm note to docstrings. Don't think `along axis 1` is needed as normalization is done after pooling.

PR #2058, opened 2025-04-15
[JS] Add LLMPipeline samples

Retribution98:
## Details:
- Implement `beam_search_causal_lm` sample
- Implement `multinomial_causal_lm` sample

## Tickets:
- CVS-165108


samples/js/text_generation/beam_search_causal_lm.js
```diff
@@ -0,0 +1,34 @@
...
+async function main() {
+    const modelPath = process.argv[2];
+    const prompts = process.argv.slice(3);
+    
+    if (!modelPath) {
+        console.error('Please specify path to model directory\n'
+                    + 'Run command must be: `node chat_sample.js *path_to_model_dir* *prompt*`');
+        process.exit(1);
+    }
+    if (!modelPath) {
```
Wovchena, line 14:
```suggestion
    if (!prompts) {
```
same for samples/js/text_generation/multinomial_causal_lm.js and greedy

src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -57,16 +57,7 @@ export class LLMPipeline {
...
-    const castedOptions: Options = {};
-
-    for (const key in options)
-      castedOptions[key] = String(options[key]);
-
-    return castedOptions;
-  }
-
-  getAsyncGenerator(prompt: string, generationOptions: Options = {}) {
+  stream(prompt: string, generationOptions: Options = {}) {
```
Wovchena, line 60: Why is it added? It should have been a class similar to multinomial_causal_lm.py

src/js/src/helper.cpp
```diff
@@ -42,12 +42,37 @@ ov::Any js_to_cpp<ov::Any>(const Napi::Env& env, const Napi::Value& value) {
...
-        return ov::Any(value.ToBoolean());
+        return ov::Any(static_cast<bool>(value.ToBoolean()));
     } else {
         OPENVINO_THROW("Cannot convert to ov::Any");
     }
 }
 
+template <>
+ov::genai::StringInputs js_to_cpp<ov::genai::StringInputs>(const Napi::Env& env, const Napi::Value& value) {
+    const auto elem = value;
```
Wovchena, line 53: If the intention is to rename the var, you can rename the function arg. If the intention is to copy, you can remove reference from `const Napi::Value&` so it copies for you.

Wovchena (changes requested):

samples/js/text_generation/beam_search_causal_lm.js
Retribution98: Thanks!

src/js/src/helper.cpp
Retribution98: It's just an artifact of debugging. Thank you for your attentiveness.

Wovchena (approved):

PR #2065, opened 2025-04-16
Print speculative decoding perf metrics in Debug mode

sbalandi:
Ticket 165469


src/cpp/src/prompt_lookup/prompt_lookup_impl.cpp
```diff
@@ -84,10 +84,12 @@ void ContinuousBatchingPipeline::PromptLookupImpl::step() {
         raw_perf_counters.m_new_token_times.emplace_back(main_timer.get_end_time());
     }
 
-    if (generated_len_after.empty() && 0) {
+#ifdef LOG_LEVEL_DEBUG
```
Wovchena, line 87: Check `NDEBUG` instead. Example: https://github.com/openvinotoolkit/openvino.genai/blob/e312b49e6a874a4401632734554998f25b1c5265/src/cpp/src/lora_adapter.cpp#L48

Wovchena, line 87: Can you mention somewhere this feature? I guess `SpeculativeDecodingMetrics`'s docstrings is a good place, but maybe you can come up with a better one

Wovchena (changes requested):

src/cpp/src/prompt_lookup/prompt_lookup_impl.cpp
Wovchena: I realized the following problem only now. It doesn't make much sense to collect metrics in debug, because they are going to be different for release. `OPENVINO_LOG_LEVEL` should have been used similar to `print_compiled_model_properties()`.

sbalandi: thank you, moved to use env variable as trigger

Wovchena (approved):

Wovchena (changes requested):
https://github.com/openvinotoolkit/openvino.genai/pull/2065#discussion_r2048603712 is left

src/cpp/src/prompt_lookup/prompt_lookup_impl.cpp
Wovchena: There's DEBUG_LOG.md with `OPENVINO_LOG_LEVEL` so it's worth adding the description there as well

sbalandi: updated, I thought that it was connected to `LOG_LEVEL_DEBUG` only

Wovchena (approved):

iefode (approved):

PR #2071, opened 2025-04-17
[C] Add ov::Property as arguments to the ov_genai_llm_pipeline_create function

apinge:
In order to enhance the functionality of the `ov_genai_llm_pipeline_create` function, I've included `ov::Property` as arguments. This will allow user to pass additional properties when compiling the model.

### DIFF
--- a/src/c/include/openvino/genai/c/llm_pipeline.h
+++ b/src/c/include/openvino/genai/c/llm_pipeline.h
@@ -70,19 +70,30 @@ typedef struct ov_genai_llm_pipeline_opaque ov_genai_llm_pipeline;
 
 /**
  * @brief Construct ov_genai_llm_pipeline.
+ *
+ * Initializes a ov_genai_llm_pipeline instance from the specified model directory and device. Optional property
+ * parameters can be passed as key-value pairs.
+ *
  * @param models_path Path to the directory containing the model files.
  * @param device Name of a device to load a model to.
  * @param property_args_size How many properties args will be passed, each property contains 2 args: key and value.
  * @param ov_genai_llm_pipeline A pointer to the newly created ov_genai_llm_pipeline.
  * @param ... property paramater: Optional pack of pairs: <char* property_key, char* property_value> relevant only
  * @return ov_status_e A status code, return OK(0) if successful.
+ *
+ * @example
+ * Example with no properties:
+ * ov_genai_llm_pipeline_create(model_path, "CPU", 0, &pipe);
+ *
+ * Example with properties:
+ * ov_genai_llm_pipeline_create(model_path, "GPU", 2, &pipe,
+ *                             "CACHE_DIR", "cache_dir");
  */
 OPENVINO_GENAI_C_EXPORTS ov_status_e ov_genai_llm_pipeline_create(const char* models_path,
                                                                   const char* device,
                                                                   const size_t property_args_size,
                                                                   ov_genai_llm_pipeline** pipe,
                                                                   ...);
-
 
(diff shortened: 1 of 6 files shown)


src/c/src/types_c.h
```diff
@@ -6,6 +6,48 @@
...
+        auto decrypt_func = _value->decrypt_func;                                                              \
+        std::function<std::string(const std::string&)> encrypt_value = [encrypt_func](const std::string& in) { \
+            size_t out_size = 0;                                                                               \
+            std::string out_str;                                                                               \
+            encrypt_func(in.c_str(), in.length(), nullptr, &out_size);                                         \
+            if (out_size > 0) {                                                                                \
+                std::unique_ptr<char[]> output_ptr(new char[out_size]);                                        \
+                if (output_ptr) {                                                                              \
+                    char* output = output_ptr.get();                                                           \
+                    encrypt_func(in.c_str(), in.length(), output, &out_size);                                  \
```
Wovchena, line 23: `encrypt_func()` is called twice with `nullptr` in the first call. Will it be better if `encrypt_func()` is expected to allocate `output` and `ov_genai_llm_pipeline_create()` is expected to free the memory?

src/c/include/openvino/genai/c/llm_pipeline.h
```diff
@@ -72,12 +72,16 @@ typedef struct ov_genai_llm_pipeline_opaque ov_genai_llm_pipeline;
  * @brief Construct ov_genai_llm_pipeline.
  * @param models_path Path to the directory containing the model files.
  * @param device Name of a device to load a model to.
+ * @param property_args_size How many properties args will be passed, each property contains 2 args: key and value.
  * @param ov_genai_llm_pipeline A pointer to the newly created ov_genai_llm_pipeline.
+ * @param ... property paramater: Optional pack of pairs: <char* property_key, char* property_value> relevant only
```
Wovchena, line 77: I don't think it would be possible to figure out how to use it without a sample and cleaner docstrings. Please add the sample with its test similar to C++. https://github.com/openvinotoolkit/openvino.genai/blob/e312b49e6a874a4401632734554998f25b1c5265/tests/python_tests/samples/test_encrypted_model_causal_lm.py#L24 If you aren't be able to come up with docstring formulation, I guess you can simply refer to the sample here.

Wovchena (changes requested):

src/c/src/types_c.h
apinge: https://github.com/openvinotoolkit/openvino/blob/d9c2aee59bc059797f20c26f32e3bd70b2e1ce2f/src/bindings/c/include/openvino/c/ov_common.h#L201-L218 `ov_encryption_callbacks` is aligned with the ov's function. The usage involves first calling to get the size, and the second call is to allocate memory. Is there any method of calling once to allocate memory？

src/c/include/openvino/genai/c/llm_pipeline.h
apinge: I've updated the docstring of this function. 
The example `encrypted_model_causal_lm` exceeds the current capabilities of the C interface. It involves elements such as `ov::genai::Tokenizer` and overloaded versions of the `LLMPipeline` constructor
https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/llm_pipeline.hpp#L118-L125

Wovchena (approved):

ilya-lavrenov (approved):

PR #2075, opened 2025-04-17
[Docs] Add installation, guides & concepts pages

yatarkan:
1. Added separate [Installation](https://yatarkan.github.io/openvino.genai/docs/getting-started/installation) page with links to Selector Tool in OpenVINO docs, Build instruction in GitHub, added dependencies section, moved system requirements.
2. Added new pages to Guides:
    - [Performance Metrics](https://yatarkan.github.io/openvino.genai/docs/guides/performance-metrics)
    - [Tokenization](https://yatarkan.github.io/openvino.genai/docs/guides/tokenization)
    - [Debug Logging](https://yatarkan.github.io/openvino.genai/docs/guides/debug-logging)
3. Added new pages to Concepts:
    - [How It Works](https://yatarkan.github.io/openvino.genai/docs/concepts/how-it-works)
    - [LoRA](https://yatarkan.github.io/openvino.genai/docs/concepts/lora) (with cross-reference to LoRA Adapters page in Guides section)


site/docs/guides/performance-metrics.mdx
```diff
@@ -0,0 +1,182 @@
...
+In addition to mean and standard deviation values, the `perf_metrics` object has a `raw_metrics` field.
+This field stores raw data, including:
+
+- Timestamps for each batch of generated tokens
+- Batch sizes for each timestamp
+- Tokenization durations
+- Detokenization durations
+- Other relevant metrics
+
+These metrics can be use for more fine grained analysis, such as getting exact calculating median values, percentiles, etc.
```
Wovchena, line 129:
```suggestion
These metrics can be use for more fine grained analysis, such as calculating exact median values, percentiles, etc.
```

site/static/img/lora.png
Wovchena, line 1: 1. `B` is upside down. I guess the intention was to show that B increases the matrix size, but it's already represented with it's shape.
2. `W` is depicted as trapezoid, but it's dimensions are `d x d`. The paper uses `d x k` in the text and `d x d` on the image, but with square matrix. I think `d x k` is better because it's more general, although I don't know if it's common for LLMs to keep dimensionality between layers.
3. Would be nice if the length of top edge of `W` matched the length of top edge of `B` because the results are going to be summed, so the dimensions must match.
4. Why is `.` used in `W.x`? It's a result of multiplication of `W` and `x`, so it should be just `Wx` or `W*x`.
5. Dimensions for `W.x`. `x` dimensions aren't specified so the result of `Wx` also can't be fully specified.
6. There is an arrow from `W` to `W.x` but arrows from lora matrixes to `dW.x` are missing.
7. I think `dW.x` should be renamed to `BAx`: it's the name used by the Output tensor and fixed the order of multiplication.

If you have this image sources, fix this please.

site/static/img/stateful.jpg
Wovchena, line 1: I assume old files and texts aren't removed because the docs are still work in progress. Is that so? In that case you'll need to double check if the content matches before the removal because, for example, https://github.com/openvinotoolkit/openvino.genai/pull/2065 is going to alter DEBUG_LOG.md

Wovchena (changes requested):

site/static/img/lora.png
yatarkan: I took this image from our [LLM LoRA notebook](https://github.com/openvinotoolkit/openvino_notebooks/blob/latest/notebooks/llm-lora/llm-lora.ipynb).
Unfortunately, I don't have its sources. We can use another image (e.g. from [HF guide](https://huggingface.co/docs/peft/conceptual_guides/adapter) that we are mentioning in the end of our page). Our we can redraw it from scratch.

site/static/img/stateful.jpg
yatarkan: You are right, old markdown files are not removed yet as docs are still WIP. My plan was to remove old files as a final step and I will double check the content is up to date.
When the docs are ready, I think we should keep only the following files:
- root `README.md` - should be slightly modified, add link to docs, remove some sections
- `src/README.md` - used as a description in PyPi, should be also modified
- `BUILD.md` - we are referring build instruction in docs and navigating user to GitHub (this is a common practice)
- all README files in samples directory
Other documents (from `src/docs` directory) can be removed as we will have them in docs.

yatarkan: As for DEBUG_LOG.md change in #2065 - I have updated the docs.

site/static/img/lora.png
Wovchena: @sbalandi, do you have the image sources to fix it?

yatarkan: I have redrew and updated the LoRA diagram. Does it look better now?

Wovchena (approved):

PR #2080, opened 2025-04-18
Add paired input into genai::Tokenizer

pavel-esir:
CVS-164556

### DIFF
--- a/src/cpp/include/openvino/genai/tokenizer.hpp
+++ b/src/cpp/include/openvino/genai/tokenizer.hpp
@@ -124,7 +124,7 @@ public:
     * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
     * @return pair of [input_ids, attention_mask]
     */
-    TokenizedInputs encode(const std::string prompt, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::string& prompt, const ov::AnyMap& tokenization_params = {});
 
     /**
     * @brief encode batch of prompts.
@@ -132,9 +132,9 @@ public:
     * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
     * @return pair of [input_ids, attention_mask]
     */
-    TokenizedInputs encode(std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
-    TokenizedInputs encode(std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
-    TokenizedInputs encode(std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
    
     /**
     * @brief encode paired prompts.
@@ -143,7 +143,7 @@ public:
     * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
     * @return pair of [input_ids, attention_mask]
     */
-   TokenizedInputs encode(std::vector<std::pair<std::string, std::string>>& prompt, const ov::AnyMap& tokenization_params = {});
+   TokenizedInputs encode(const std::vector<std::pair<std::string, std::string>>& prompt, const ov::AnyMap& tokenization_params = {});
 
    /**
    * @brief encode paired prompts.
@@ -154,7 +154,7 @@ public:
    * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
    * @return pair of [input_ids, attention_mask]
    */
-    TokenizedInputs encode(std::vector<std::string>& prompts_1, std::vector<std::string>& prompts_2, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::vector<std::string>& prompts_1, const std::vector<std::string>& prompts_2, const ov::AnyMap& tokenization_params = {});
 
     /**
     * @brief encode a single prompt
@@ -165,7 +165,7 @@ public:
     * @return pair of [input_ids, attention_mask]
     */
     template <typename... Properties>
-    util::EnableIfAllStringAny<TokenizedInputs, Properties...> encode(std::string& prompt, Properties&&... properties) {
+    util::EnableIfAllStringAny<TokenizedInputs, Properties...> encode(const std::string& prompt, Properties&&... properties) {
         return encode(prompt, AnyMap{std::forward<Properties>(properties)...});
     }
 
@@ -178,7 +178,7 @@ public:
     * @return pair of [input_ids, attention_mask]
     */
     template <typename... Properties>
-    util::EnableIfAllStringAny<TokenizedInputs, Properties...> encode(std::vector<std::string>& prompts, Properties&&... properties) {
+    util::EnableIfAllStringAny<TokenizedInputs, Properties...> encode(const std::vector<std::string>& prompts, Properties&&... properties) {
         return encode(prompts, AnyMap{std::forward<Properties>(properties)...});
     }
 
@@ -188,7 +188,7 @@ public:
     * @param detokenization_params AnyMap with detokenization parameters, e.g. {"skip_special_tokens", false}
     * @return sequence string
     */
-    std::string decode(std::vector<int64_t> tokens, const ov::AnyMap& detokenization_params = {});
+    std::string decode(const std::vector<int64_t>& tokens, const ov::AnyMap& detokenization_params = {});
 
     /**
     * @brief decode sequence of tokens
@@ -197,7 +197,7 @@ public:
     * @return sequence string
     */
     template <typename... Properties>
-    util::EnableIfAllStringAny<std::string, Properties...> decode(std::vector<int64_t>& tokens, Properties&&... detokenization_params) {
+    util::EnableIfAllStringAny<std::string, Properties...> decode(const std::vector<int64_t>& tokens, Properties&&... detokenization_params) {
         return decode(tokens, AnyMap{std::forward<Properties>(detokenization_params)...});
     }
 
@@ -207,7 +207,7 @@ public:
     * @param detokenization_params AnyMap with detokenization parameters, e.g. {"skip_special_tokens", false}
     * @return vector of std::string, with size = batch_size
     */
-    std::vector<std::string> decode(ov::Tensor tokens, const ov::AnyMap& detokenization_params = {});
+    std::vector<std::string> decode(const ov::Tensor& tokens, const ov::AnyMap& detokenization_params = {});
 
     /**
     * @brief decode sequence of tokens
@@ -216,7 +216,7 @@ public:
     * @return vector of std::string, with size = batch_size
     */
     template <typename... Properties>
-    util::EnableIfAllStringAny<std::vector<std::string>, Properties...> decode(ov::Tensor tokens, Properties&&... detokenization_params) {
+    util::EnableIfAllStringAny<std::vector<std::string>, Properties...> decode(const ov::Tensor& tokens, Properties&&... detokenization_params) {
         return decode(tokens, AnyMap{std::forward<Properties>(detokenization_params)...});
     }
 
@@ -226,7 +226,7 @@ public:
     * @param detokenization_params AnyMap with detokenization parameters, e.g. {"skip_special_tokens", false}
     * @return vector of std::string, with size equal to batch_size
     */
-    std::vector<std::string> decode(std::vector<std::vector<int64_t>> tokens, const ov::AnyMap& detokenization_params = {});
+    std::vector<std::string> decode(const std::vector<std::vector<int64_t>>& tokens, const ov::AnyMap& detokenization_params = {});
 
     /**
     * @brief decode sequence of tokens
@@ -235,7 +235,7 @@ public:
     * @return vector of std::string, with size = batch_size
     */
     template <typename... Properties>
-    util::EnableIfAllStringAny<std::vector<std::string>, Properties...> decode(std::vector<std::vector<int64_t>> tokens, Properties&&... detokenization_params) {
+    util::EnableIfAllStringAny<std::vector<std::string>, Properties...> decode(const std::vector<std::vector<int64_t>>& tokens, Properties&&... detokenization_params) {
         return decode(tokens, AnyMap{std::forward<Properties>(detokenization_params)...});
     }
 
--- a/src/cpp/include/openvino/genai/tokenizer.hpp
+++ b/src/cpp/include/openvino/genai/tokenizer.hpp
@@ -135,6 +135,15 @@ public:
     TokenizedInputs encode(std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
     TokenizedInputs encode(std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
     TokenizedInputs encode(std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
+   
+    /**
+    * @brief encode paired prompts. Left padding will be applied by default
+    * @param prompts vector storing batch of prompts
+    * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
+    * @return pair of [input_ids, attention_mask]
+    */
+    TokenizedInputs encode(std::vector<std::pair<std::string, std::string>>& prompt, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(std::vector<std::string>& prompts_1, std::vector<std::string>& prompts_2, const ov::AnyMap& tokenization_params = {});
 
     /**
     * @brief encode a single prompt
--- a/src/cpp/include/openvino/genai/tokenizer.hpp
+++ b/src/cpp/include/openvino/genai/tokenizer.hpp
@@ -127,7 +127,7 @@ public:
     TokenizedInputs encode(const std::string prompt, const ov::AnyMap& tokenization_params = {});
 
     /**
-    * @brief encode batch of prompts. Left padding will be applied by default
+    * @brief encode batch of prompts.
     * @param prompts vector storing batch of prompts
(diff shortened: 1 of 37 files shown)


tests/python_tests/test_tokenizer.py
```diff
@@ -356,7 +356,7 @@ def hf_ov_genai_models(request, tmp_path_factory):
         ("katuni4ka/tiny-random-phi3", {"padding_side": None}),
         ("TinyLlama/TinyLlama-1.1B-Chat-v1.0", {"padding_side": None}),
         ("katuni4ka/tiny-random-llava-next", {"padding_side": "right"}),
-        ("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}),
+        ("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}),  # This fails
```
pavel-esir, line 359: fails with segfault

pavel-esir, line 359:
```suggestion
        ("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}),
```

src/cpp/src/tokenizer/make_tokenizer_stateful.cpp
```diff
@@ -165,18 +185,31 @@ bool ov::genai::MakePaddingSatateful::run_on_model(const std::shared_ptr<ov::Mod
...
+
+    var_info = {ov::Shape{}, ov::element::boolean, ov::genai::IS_MAX_LENGTH_SET};
+    auto is_max_len_set_var = std::make_shared<op::util::Variable>(var_info);
+    auto defaul_false_const = std::make_shared<v0::Constant>(ov::element::boolean, ov::Shape{}, std::vector{false});
+    auto is_max_len_set_rv = std::make_shared<v6::ReadValue>(defaul_false_const, is_max_len_set_var);
+    auto is_max_len_set_assign = std::make_shared<v6::Assign>(is_max_len_set_rv, is_max_len_set_var);
+    model->add_sinks({is_max_len_set_assign});
+
+    auto int32_max_constant = std::make_shared<v0::Constant>(ov::element::i32, ov::Shape{}, std::vector{std::numeric_limits<int32_t>::max() - 1000});
+    // auto int32_max_constant = std::make_shared<v0::Constant>(ov::element::i32, ov::Shape{}, std::vector{10000});
```
apaniukov, line 200:
```suggestion
```

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -135,6 +135,15 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     TokenizedInputs encode(std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
     TokenizedInputs encode(std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
     TokenizedInputs encode(std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
+   
+    /**
+    * @brief encode paired prompts. Left padding will be applied by default
```
pavel-esir, line 140: Left padding is not applied by default. Adjust this docstring

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -11,6 +11,7 @@
 #include <jinja2cpp/generic_list_iterator.h>
 
 #include "openvino/pass/manager.hpp"
+#include "openvino/pass/visualize_tree.hpp"
```
pavel-esir, line 14: remove

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -280,6 +284,7 @@ class Tokenizer::TokenizerImpl {
             ov::pass::Manager manager;
             manager.register_pass<MakeAddSpecialTokensSatateful>();
             manager.register_pass<MakePaddingSatateful>();
+            manager.register_pass<ov::pass::VisualizeTree>("after.svg");
```
pavel-esir, line 287: remove this

tests/python_tests/test_tokenizer.py
apaniukov: Delete the comment

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: done

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: done

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Corrected this docstring. Also added information about broadcasting

src/cpp/src/tokenizer/make_tokenizer_stateful.cpp
pavel-esir: removed. There is a strange behavior regarding int32_max left a TODO.

tests/python_tests/test_tokenizer.py
pavel-esir: removed

pavel-esir: deleted this comment

src/python/py_tokenizer.cpp
```diff
@@ -107,8 +107,58 @@ void init_tokenizer(py::module_& m) {
...
+                }
+                return tok.encode(prompts_1, prompts_2, tokenization_params);
+            },
+            py::arg("prompts_1"),
+            py::arg("prompts_2"),
+            py::arg("add_special_tokens") = true,
+            py::arg("pad_to_max_length") = false,
+            py::arg("max_length") = std::nullopt,
+            R"(Encodes a list of prompts into tokenized inputs. Prompts should be of the same length, or one of them should be of length 1. 
+            In the latest case, the prompt will be broadcasted to the length of the other prompt)")
```
apaniukov, line 131:
```suggestion
            R"(Encodes a list of prompts into tokenized inputs. The number of strings must be the same, or one of the inputs can contain one string.
            In the latter case, the single-string input will be broadcast into the shape of the other input, which is more efficient than repeating the string in pairs.)")
```

tests/python_tests/test_tokenizer.py
```diff
@@ -356,7 +357,7 @@ def hf_ov_genai_models(request, tmp_path_factory):
...
```
Wovchena, line 360:
```suggestion
        ("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}),
```

Wovchena (changes requested):

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -471,6 +474,56 @@ class Tokenizer::TokenizerImpl {
...
+        for(size_t i = 0; i < batch_size; ++i) {
+            prompts_0[i] = prompts_pairs[i].first;
+            prompts_1[i] = prompts_pairs[i].second;
+        }
+        inputs[0] = ov::Tensor{ov::element::string, {batch_size}, prompts_0.data()};
+        inputs[1] = ov::Tensor{ov::element::string, {batch_size}, prompts_1.data()};
+        return encode(inputs, tokenization_params);
+    }
+
+    TokenizedInputs encode(std::vector<std::string> prompts_1, std::vector<std::string> prompts_2, const ov::AnyMap& tokenization_params = {}) {
```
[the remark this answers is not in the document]
pavel-esir: Agreed and corrected. But regarding const added only for `std::vector<ov::Tensor>`. 

For string, vector<string>, etc. arguments where i create `ov::Tensor` from them there is a error on openvino side, it requires that  `prompts` is not a const object here `ov::Tensor{ov::element::string, {batch_size}, prompts.data()}`, otherwise we end up having the following error

```
 /home/epavel/opt/openvino_master/runtime/include/openvino/runtime/allocator.hpp:65:22: error: request for member ‘allocate’ in ‘((ov::Allocator::Impl<const std::__cxx11::basic_string<char>*>*)this)->ov::Allocator::Impl<const std::__cxx11::basic_string<char>*>::a’, which is of pointer type ‘const std::__cxx11::basic_string<char>*’ (maybe you meant to use ‘->’ ?)
[build]    65 |             return a.allocate(bytes, alignment);
```

tests/python_tests/test_tokenizer.py
pavel-esir: removed comment

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -471,6 +474,56 @@ class Tokenizer::TokenizerImpl {
...
+        std::vector<std::string> prompts_0(batch_size);
+        std::vector<std::string> prompts_1(batch_size);
+
+        for(size_t i = 0; i < batch_size; ++i) {
+            prompts_0[i] = prompts_pairs[i].first;
+            prompts_1[i] = prompts_pairs[i].second;
+        }
+        inputs[0] = ov::Tensor{ov::element::string, {batch_size}, prompts_0.data()};
+        inputs[1] = ov::Tensor{ov::element::string, {batch_size}, prompts_1.data()};
+        return encode(inputs, tokenization_params);
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -462,7 +465,7 @@ class Tokenizer::TokenizerImpl {
         CircularBufferQueueElementGuard<ov::InferRequest> infer_request_guard(m_ireq_queue_tokenizer.get());
         set_state_if_necessary(infer_request_guard, tokenization_params);
         size_t batch_size = 1;
-        infer_request_guard.get().set_input_tensor(ov::Tensor{ov::element::string, {batch_size}, &prompt});
+        infer_request_guard.get().set_input_tensor(0, ov::Tensor{ov::element::string, {batch_size}, &prompt});
```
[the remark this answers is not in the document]
pavel-esir: If we have 2 input IR we should be able to tokenize even if input is a single prompt. In case if model has 2 inputs, if we call set_input_tensor error will be thrown. 

Therefore we set explicitly input index 0. When we do so, the other input/Parameter remains an empty tensor which is correctly handles by subgraph and we return a valid tokens as if we had an old/simple IR with single input.

Added comment about that

src/cpp/src/tokenizer/tokenizer.cpp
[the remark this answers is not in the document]
Wovchena: Const is needed. See https://github.com/openvinotoolkit/openvino.genai/pull/2057#discussion_r2059860891 for the explanation why. The possible solution is in ticket 167027

pavel-esir:
@mryzhov pushed with updated caches. Please take a look

src/cpp/src/tokenizer/tokenizer.cpp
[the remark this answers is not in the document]
pavel-esir: Added const modifier to all encode/decode methods in genai::Tokenizer. Also tokens/prompts now are always passed by reference. This should remove unnecessary memcopies

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -471,16 +477,66 @@ class Tokenizer::TokenizerImpl {
         );
     }
 
-    TokenizedInputs encode(std::vector<std::string>& prompts, const ov::AnyMap& tokenization_params = {}) {
+    TokenizedInputs encode(const std::vector<std::pair<std::string, std::string>>& prompts_pairs, const ov::AnyMap& tokenization_params = {}) {
+        OPENVINO_ASSERT(m_ireq_queue_tokenizer, "Either openvino_tokenizer.xml was not provided or it was not loaded correctly. "
+                                                "Tokenizer::encode is not available");
+        std::vector<ov::Tensor> inputs(2);
```
Wovchena, line 483:
```suggestion
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -471,16 +477,66 @@ class Tokenizer::TokenizerImpl {
...
+    TokenizedInputs encode(const std::vector<std::string>& prompts_1, const std::vector<std::string>& prompts_2, const ov::AnyMap& tokenization_params = {}) {
+        OPENVINO_ASSERT(m_ireq_queue_tokenizer, "Either openvino_tokenizer.xml was not provided or it was not loaded correctly. "
+                                                "Tokenizer::encode is not available");
+        std::vector<ov::Tensor> prompts(2);
+        size_t batch_size = prompts_1.size();
+        OPENVINO_ASSERT(prompts_1.size() == prompts_2.size() || prompts_1.size() == 1 | prompts_2.size() == 1, 
+                        "prompts_1 and prompts_2 should be of the same size or one of them should be of size 1");
+        prompts[0] = ov::Tensor{ov::element::string, {batch_size}, const_cast<std::string*>(prompts_1.data())};
+        prompts[1] = ov::Tensor{ov::element::string, {batch_size}, const_cast<std::string*>(prompts_2.data())};
+        std::vector<std::string> prompts_1_, prompts_2_;
```
Wovchena, line 504:
```suggestion
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -471,16 +477,66 @@ class Tokenizer::TokenizerImpl {
...
+        return encode(prompts, tokenization_params);
+    }
+
+    TokenizedInputs encode(const std::vector<ov::Tensor>& prompts, const ov::AnyMap& tokenization_params = {}) {
```
Wovchena, line 508:
```suggestion
    TokenizedInputs encode(const std::array<ov::Tensor, 2>& prompts, const ov::AnyMap& tokenization_params = {}) {
```
Or `pair`, That way the assert isn't needed. The second Tensor can be default constructed if it's not required

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -124,17 +124,37 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
+    TokenizedInputs encode(const std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
+   
+    /**
+    * @brief encode paired prompts.
+    * 
+    * @param prompts vector storing batch of prompts
+    * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
+    * @return pair of [input_ids, attention_mask]
+    */
+   TokenizedInputs encode(const std::vector<std::pair<std::string, std::string>>& prompt, const ov::AnyMap& tokenization_params = {});
```
Wovchena, line 146: Add a comment that this overload is less efficient because it copies the strings to match the signature with two vectors.

An overload with non const && could be added in addition to that one to fix this, but I don't see how to avoid code duplication. This function could be templated to implement the perfect forwarding, but this would replace specific `prompt` type with a template type which is harder to read

Wovchena (changes requested):

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -455,14 +458,17 @@ class Tokenizer::TokenizerImpl {
...
                                                 "Tokenizer::encode is not available");
 
         CircularBufferQueueElementGuard<ov::InferRequest> infer_request_guard(m_ireq_queue_tokenizer.get());
         set_state_if_necessary(infer_request_guard, tokenization_params);
         size_t batch_size = 1;
-        infer_request_guard.get().set_input_tensor(ov::Tensor{ov::element::string, {batch_size}, &prompt});
+        // When the model has more than one input, calling set_input_tensor without specifying an index may fail.
+        // If the model has two inputs, explicitly set the first input while leaving the second input tensor empty.
+        // The subgraph within the ov::Model will handle this scenario, ensuring the output remains correct.
+        infer_request_guard.get().set_input_tensor(0, ov::Tensor{ov::element::string, {batch_size}, const_cast<std::string*>(&prompt)});
```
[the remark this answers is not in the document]
pavel-esir: Thanks for noticing this!

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: removed

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -124,17 +124,37 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
+    * @brief encode batch of prompts.
     * @param prompts vector storing batch of prompts
     * @param tokenization_params AnyMap with tokenization parameters, e.g. {{"add_special_tokens", false}, {"max_length", 128}}
     * @return pair of [input_ids, attention_mask]
     */
-    TokenizedInputs encode(std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
-    TokenizedInputs encode(std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
-    TokenizedInputs encode(std::initializer_list<std::string>& prompts, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::vector<std::string>& prompt, const ov::AnyMap& tokenization_params = {});
+    TokenizedInputs encode(const std::vector<std::string>&& prompts, const ov::AnyMap& tokenization_params = {});
```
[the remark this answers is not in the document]
pavel-esir: removed

src/cpp/src/image_generation/models/clip_text_model.cpp
```diff
@@ -117,10 +117,10 @@ ov::Tensor CLIPTextModel::infer(const std::string& pos_prompt, const std::string
 
         if (input_ids.get_element_type() == ov::element::i32) {
             std::fill_n(input_ids.data<int32_t>(), input_ids.get_size(), pad_token_id);
-            std::copy_n(input_ids_token.data<int64_t>(), input_ids_token.get_size(), input_ids.data<int32_t>());
+            std::copy_n(input_ids_token.data<int64_t>(), input_ids.get_size(), input_ids.data<int32_t>());
         } else {
             std::fill_n(input_ids.data<int64_t>(), input_ids.get_size(), pad_token_id);
-            std::copy_n(input_ids_token.data<int64_t>(), input_ids_token.get_size(), input_ids.data<int64_t>());
+            std::copy_n(input_ids_token.data<int64_t>(), input_ids.get_size(), input_ids.data<int64_t>());
```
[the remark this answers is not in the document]
pavel-esir: Thanks! I think your corrections are right. But let's wait Annas opinion

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: I removed this overload. Since it's used only in one place i moved code inside other function. Not passing between functions is needed now

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Added comment about efficiency.

Wovchena (approved):

PR #2100, opened 2025-04-22
Disable /sdl for gguf-tools

Wovchena:
It fails with _deps\gguflib-src\sds.c(347): error C4146: unary minus operator applied to unsigned type, result still unsigned


src/cpp/CMakeLists.txt
```diff
@@ -64,6 +64,9 @@ if(ENABLE_GGUF)
     add_library(gguflib STATIC ${gguflib_SOURCE_DIR}/fp16.c ${gguflib_SOURCE_DIR}/gguflib.c)
     set_target_properties(gguflib PROPERTIES POSITION_INDEPENDENT_CODE ON)
     target_include_directories(gguflib PUBLIC "${gguflib_SOURCE_DIR}")
+    if(MSVC)
+        target_compile_options(gguf-tools PRIVATE "/sdl-")
```
ilya-lavrenov, line 68: probably, it's better to add `EXCLUDE_FROM_ALL` for GGUF folder

in this case, only `gguflib` will be built

Wovchena: Fixed

ilya-lavrenov (approved):

PR #2101, opened 2025-04-23
docs: fix path to kv-cache-areas-diagram.svg

Wovchena:
https://wovchena.github.io/openvino.genai-public/docs/concepts/optimization-techniques/kvcache-eviction-algorithm


site/docs/concepts/optimization-techniques/kvcache-eviction-algorithm.md
```diff
@@ -11,7 +11,7 @@ The cache eviction algorithm is designed to manage KV (Key-Value) cache memory f
 ## Conceptual Model
 The KV cache for each sequence is divided into three logical areas:
 
-![KV cache layout with cache eviction](/img/kv-cache-areas-diagram.svg)
+![KV cache layout with cache eviction](./../../../static/img/kv-cache-areas-diagram.svg)
```
yatarkan, line 14: As I can see on the deployed version from master, `kv-cache-areas-diagram.svg` image is displayed correctly:
https://openvinotoolkit.github.io/openvino.genai/docs/concepts/optimization-techniques/kvcache-eviction-algorithm 
Am I missing something?

Wovchena: True. This fixes repo's render: https://github.com/openvinotoolkit/openvino.genai/blob/master/site/docs/concepts/optimization-techniques/kvcache-eviction-algorithm.md

site/docusaurus.config.ts
```diff
@@ -6,7 +6,8 @@ import GenAISamplesDocsPlugin from './src/plugins/genai-samples-docs-plugin';
 // This runs in Node.js - Don't use client-side code here (browser APIs, JSX...)
 
 const organizationName = 'openvinotoolkit';
```
yatarkan, line 8: I would suggest taking both `organizationName` and `projectName` from env var:
```suggestion
// GITHUB_REPOSITORY env var is set by GitHub Actions
const GITHUB_REPOSITORY = process.env.GITHUB_REPOSITORY || 'openvinotoolkit/openvino.genai';
const [organizationName, projectName] = GITHUB_REPOSITORY.split('/');
```

site/docs/concepts/optimization-techniques/kvcache-eviction-algorithm.md
yatarkan: It should work without changing image path from absolute to relative.
Docusaurus should prepend `baseUrl` (which is set in docusaurus.config.ts) to all paths under the hood. Otherwise all other image URLs would be affected.

Wovchena: GitHub pages are fine. The problem is with GitHub's/VS Code markdown view

site/docusaurus.config.ts
```diff
@@ -5,8 +5,11 @@ import GenAISamplesDocsPlugin from './src/plugins/genai-samples-docs-plugin';
 
 // This runs in Node.js - Don't use client-side code here (browser APIs, JSX...)
 
-const organizationName = 'openvinotoolkit';
-const projectName = 'openvino.genai';
+// GITHUB_REPOSITORY env var is set by GitHub Actions
+const GITHUB_REPOSITORY = process.env.GITHUB_REPOSITORY || 'openvinotoolkit/openvino.genai';
+const [organizationName, projectName] = GITHUB_REPOSITORY.split('/');
+// GITHUB_REPOSITORY is set by GitHub Actions
+const projectName = `${process.env.GITHUB_REPOSITORY?.split('/')[1] || 'openvino.genai'}`
```
yatarkan, line 12: Duplicated const will cause an error
```suggestion
```
GitHub does not support code suggestion for both deleted and added lines, so I was not able to change it correctly in my previous comment

yatarkan (approved):

PR #2102, opened 2025-04-23
LLM: release plugin once pipeline is removed and WA for GPU

sbalandi:


sbalandi:
checked the memory on Linux/Windows and got the same results as in the task, most of the memory is released after removing the pipeline, but 40-60 MB remains(but that tail is not the goal of that pr)

@Wovchena please, take a look

ilya-lavrenov (approved):

src/cpp/src/continuous_batching_impl.cpp
```diff
@@ -66,8 +66,17 @@ ContinuousBatchingPipeline::ContinuousBatchingImpl::ContinuousBatchingImpl(
 }
 
 ContinuousBatchingPipeline::ContinuousBatchingImpl::~ContinuousBatchingImpl() {
+    // manually release all blocks, which can re-initialize OpenVINO plugins during destruction
+    m_model_runner->get_infer_request().get_compiled_model().release_memory();
+
+    // manually release all blocks, which can re-initialize OpenVINO plugins during destruction
```
Wovchena, line 72: Destructor resets `ContinuousBatchingImpl`'s members before `~IContinuousBatchingPipeline()`. Are resets still needed here?

sbalandi:
> checked the memory on Linux/Windows and got the same results as in the task, most of the memory is released after removing the pipeline, but 40-60 MB remains(but that tail is not the goal of that pr)
> 
> @Wovchena please, take a look

Only for Windows. For Linux it is about 300MB, task: https://jira.devtools.intel.com/browse/CVS-166803

src/cpp/src/icontinuous_batching.hpp
```diff
@@ -33,6 +33,8 @@ class ContinuousBatchingPipeline::IContinuousBatchingPipeline {
 
     PipelineMetrics m_pipeline_metrics;
 
+    std::string m_device;
```
Wovchena, line 36: gtest fails. I'd guess the reason is that `CBForSDTest` inherits from `ContinuousBatchingPipeline` leaving `m_device` default initialized. Can you check if this inheritance really required?

Or maybe it's `PipelineTestInstance` inheriting from `ContinuousBatchingForSpeculativeDecodingImpl`

sammysun0711:
Please resolve the merge conflict and fixed test error, thanks.

sbalandi:
> Please resolve the merge conflict and fixed test error, thanks.

@sammysun0711 resolved

Wovchena (approved):

PR #2107, opened 2025-04-23
Bring Back get_vocab

apaniukov:
- Revert get_vocab and tokenizer tests optimizations revert
- Optimize `TestImageTags` VLM tests (1 min vs 2 min locally)
- Further optimizations of tokenizers tests
- Fix issue with multiple infer requests with different states

CVS-166547, CVS-164141


Wovchena (changes requested):

pyproject.toml
```diff
@@ -65,4 +65,14 @@ build-backend = "py_build_cmake.build"
...
+
+[tool.ruff]
+line-length = 119
+
+[tool.ruff.lint]
+ignore = ["C901", "E501", "E741", "W605"]
+select = ["C", "E", "F", "I", "W"]
+
+[tool.ruff.lint.isort]
+lines-after-imports = 2
```
Wovchena, line 78: The change is nice, but these rules are violated because of other python files you didn't change and the next contributor may break it for the files you touched. It's already hard to distinguish functional and style changes. If you applied it to every python file that would become a nightmare. Move to a separate PR with precommit enforcing this.

tests/python_tests/conftest.py
```diff
@@ -3,6 +3,7 @@
 import shutil
 import logging
 from utils.constants import get_ov_cache_models_dir
+from utils.tokenizers import model_tmp_path  # NOQA: F401
```
Wovchena, line 6: I missed this change in the original PR. I always struggled to follow the logic. That's pretty strange to me to enable style checks and then work around tool's bugs with comments. Not enabling the check at all would be easier at the cost of random unused imports. Style checkers are supposed to help to write readable the code, not make it harder.

Can you explain?

tests/python_tests/test_tokenizer.py
```diff
@@ -280,86 +283,132 @@ def test_special_tokens(tmp_path, prompt, model_id):
...
-        genai_tokenzier = Tokenizer(model_dir)
-        return hf_tokenizer, genai_tokenzier
-    return _get_model_path
+
+@pytest.mark.precommit
+@pytest.mark.nightly
+def test_miltiple_infer_request_state():
+    hf_tokenizer = retry_request(lambda: AutoTokenizer.from_pretrained("llamafactory/tiny-random-Llama-3"))
+    ov_tokenizer = convert_tokenizer(hf_tokenizer)
+    with TemporaryDirectory() as temp_dir:
```
Wovchena, line 294: Use pytest's `tmp_path`. It's not remove right after the run so you can use it for reproduction it if needed

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -54,13 +83,13 @@ def get_ov_model(model_id):
 image_links = [
     "https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11",
     "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/tasks/car.jpg",
-    "https://github.com/user-attachments/assets/8c9ae017-7837-4abc-ae92-c1054c9ec350"
+    "https://github.com/user-attachments/assets/8c9ae017-7837-4abc-ae92-c1054c9ec350",
```
Wovchena, line 86: This is a tiny thing not worth mentioning. The problem is that lots of people do similar changes, so it becomes big. I got an idea that I'm going to write this once and paste whenever I see this.

The point of adding a trailing comma is to avoid changing this line in future if another element is added. When you add `,` as a separate change, you increase the size of diff ruining the whole purpose of minimizing diffs including the future ones. It's hard to prove that something is going to be added to the list in future. Even if you prove that, the overall diff size of the current change and future one isn't going to change so you should never add `,` as a separate line change. You only break blame with this change. Revert this

tests/python_tests/test_vlm_pipeline.py
Wovchena, line 1: So the failing test which was a reason for the revert, was a random failure, right?

Wovchena (changes requested):

tests/python_tests/test_vlm_pipeline.py
apaniukov: Yes, I didn't change the test logic at all.

Wovchena: Increased attempts https://github.com/openvinotoolkit/openvino.genai/pull/2115

tests/python_tests/test_tokenizer.py
apaniukov: Done

tests/python_tests/conftest.py
apaniukov: This import is not unused, it is just the way pytest handles fixtures. Moved the `model_tmp_path` import back to `test_tokenizers`

tests/python_tests/test_vlm_pipeline.py
apaniukov: This is the formatter's decision, not mine. I don't see the issue with it, this is how we can gradually introduce formatting into the repository.

src/cpp/src/tokenizer.cpp
```diff
@@ -137,7 +169,7 @@ class Tokenizer::TokenizerImpl {
 
     // To change the adding special tokens mode we use a statefull subgraph,
     // this flag holds the current state value of the CompiledModel.
-    ov::AnyMap m_state_flags;
+    std::unordered_map<ov::InferRequest*, ov::AnyMap> m_request_to_state_flags;
```
Wovchena, line 172: Can you try std::reference_wrapper instead of ptr? They can't be null

Wovchena (approved):

src/cpp/src/tokenizer.cpp
apaniukov: `std::reference_wrapper` is not hashable, which adds complexity. If you still want to go this route, we can merge it in the next PR.

ilya-lavrenov:
@apaniukov please, resolve conflicts

PR #2111, opened 2025-04-24
Fix of filling of pixel_values tensor in llava_image_embed_make_with_bytes_slice()

popovaan:
Currently `pixel_values` tensor is not always completely filled with values and some values remain uninitialized and sporadically these values can be `nan`. These `nan` lead to appearing of `nan` in embeddings vectors which corrupt output of language model. 
In this PR I set these uninitialized values to zero, which prevents appearing of `nan` values.

Ticket: CVS-165499

### DIFF
--- a/src/cpp/src/visual_language/minicpm/classes.cpp
+++ b/src/cpp/src/visual_language/minicpm/classes.cpp
@@ -320,9 +320,8 @@ EncodedImage llava_image_embed_make_with_bytes_slice(clip_ctx& ctx_clip, const o
     for (size_t c_idx = 0; c_idx < channels; ++c_idx) {
         for (size_t k_idx = 0; k_idx < patch_size; k_idx++) {
             std::copy(clip_value_data, clip_value_data + d3_clip_pixel, pixel_value_data);
-            if (d3_all_pixel > d3_clip_pixel) {
-                memset(pixel_value_data + d3_clip_pixel, 0, (d3_all_pixel - d3_clip_pixel) * sizeof(float));
-            }
+            // pixel_values and patch_attention_mask are multiplied instead of ignoring the corresponding pixel_values resulting in NaN instead of 0.0f if pixel_values has NaN.
+            memset(pixel_value_data + d3_clip_pixel, 0, (d3_all_pixel - d3_clip_pixel) * sizeof(float));
             clip_value_data += d3_clip_pixel;
             pixel_value_data += d3_all_pixel;
         }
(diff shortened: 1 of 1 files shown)


src/cpp/src/visual_language/minicpm/classes.cpp
```diff
@@ -320,6 +320,9 @@ EncodedImage llava_image_embed_make_with_bytes_slice(clip_ctx& ctx_clip, const o
     for (size_t c_idx = 0; c_idx < channels; ++c_idx) {
         for (size_t k_idx = 0; k_idx < patch_size; k_idx++) {
             std::copy(clip_value_data, clip_value_data + d3_clip_pixel, pixel_value_data);
+            if (d3_all_pixel > d3_clip_pixel) {
+                memset(pixel_value_data + d3_clip_pixel, 0, (d3_all_pixel - d3_clip_pixel) * sizeof(float));
+            }
```
Wovchena, line 325:
```suggestion
            // pixel_values and patch_attention_mask are multiplied instead of ignoring the corresponding pixel_values resulting in NaN instead of 0.0f if pixel_values has NaN.
            memset(pixel_value_data + d3_clip_pixel, 0, (d3_all_pixel - d3_clip_pixel) * sizeof(float));
```
I guess we are fine if `d3_all_pixel == d3_clip_pixel` that would result is noop.

src/cpp/src/visual_language/minicpm/classes.cpp
```diff
@@ -341,6 +344,9 @@ EncodedImage llava_image_embed_make_with_bytes_slice(clip_ctx& ctx_clip, const o
                 for (size_t c_idx = 0; c_idx < channels; ++c_idx) {
                     for (size_t k_idx = 0; k_idx < patch_size; k_idx++) {
                         std::copy(clip_value_data, clip_value_data + d3_clip_pixel, pixel_value_data);
+                        if (d3_all_pixel > d3_clip_pixel) {
+                            memset(pixel_value_data + d3_clip_pixel, 0, (d3_all_pixel - d3_clip_pixel) * sizeof(float));
+                        }
```
Wovchena, line 349:
```suggestion
                        // pixel_values and patch_attention_mask are multiplied instead of ignoring the corresponding pixel_values resulting in NaN instead of 0.0f if pixel_values has NaN.
                        memset(pixel_value_data + d3_clip_pixel, 0, (d3_all_pixel - d3_clip_pixel) * sizeof(float));
```

src/cpp/src/visual_language/minicpm/classes.cpp
Wovchena, line 1: Given that FP16 is fine, it's worth asking the plugins to align the implementation so INT8 followed the same robust approach. We can merge this workaround meanwhile.

Wovchena (changes requested):

Wovchena (approved):

src/cpp/src/visual_language/minicpm/classes.cpp
popovaan: Potentially it can be reproduced on fp16 and fp32. I tried to set `pixel_values` to all NAN before filling `pixel_values` with actual values and the bug reproduced on both fp16 and fp32. So it was was accidental that it reproduced only on int4 and int8. So I don't think it's plugin specific.

Wovchena (approved):

PR #2115, opened 2025-04-25
Increase max_retries

Wovchena:


tests/python_tests/test_vlm_pipeline.py
```diff
@@ -480,7 +480,7 @@ def streamer(subword):
 
 
 def retry(func):
```
apaniukov, line 482: Maybe it is better to move `max_retries` to kwargs?
```python
def retry(func, max_retries=20):
    ...
```

Wovchena: No one uses other value - YAGNI

apaniukov (approved):

PR #2121, opened 2025-04-28
Update llm_bench requirements.txt to contain sentencepiece

skuros:


eaidova (approved):

eaidova:
@skuros  for me it will be better if you will use transformers[sentencepiece] instead of direct adding sentencepiece in requirements or add it into conversion extras (python3.13 is not supported by sentencepiece yet as I know, itwill block usage llm_bench for this python even for models where sentencepiece does not required

sstrehlk (approved):

PR #2122, opened 2025-04-28
Coverity: exclude C++ and Python Tokenizers

Wovchena:


Wovchena:
Moved to https://github.com/openvinotoolkit/openvino.genai/pull/2124 because my fork doesn't have the token added

PR #2125, opened 2025-04-28
[llm_bench] Fix batch size processing while image gen benchmark

apram0d:
currently, when passing `bs` parameter, it is not generating images of batch size and throwing below error while processing:

```
  File "/home/intel/apramod/openvino.genai/tools/llm_bench/task/image_generation.py", line 188, in run_image_generation_genai
    image = Image.fromarray(res[bs_idx])
```
Fixing it by passing correct parameter duing image pipeline:

Tested with model and command `python3 benchmark.py --model ~/lcm-dreamshaper-v7/pytorch/ov/OV_FP16-INT8_ASYM/ -p "Generate an image of a cat" --num_steps 50 -bs 4` and was able to see 4 images generated.


ilya-lavrenov:
build_jenkins

ilya-lavrenov:
build_jenkins

PR #2129, opened 2025-04-28
Switch VLM to ContinuousBatching by default.

popovaan:
CVS-166784


src/cpp/src/llm_pipeline.cpp
```diff
@@ -22,80 +22,6 @@ bool with_cpu_sve();
```
ilya-lavrenov, line 22: I suppose we need to move forward declaration for `with_cpu_sve();` to utils.cpp as well

src/cpp/src/utils.hpp
```diff
@@ -17,6 +17,15 @@
 #include "openvino/genai/generation_handle.hpp"
 #include "openvino/genai/streamer_base.hpp"
 
+namespace ov {
+namespace genai {
+
+const std::string PA_BACKEND = "PA";
+const std::string SDPA_BACKEND = "SDPA";
```
ilya-lavrenov, line 24:
```suggestion
extern const std::string SDPA_BACKEND;
```
and move definition to `utils.cpp`

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -237,7 +237,7 @@ def test_vlm_continuous_batching_vs_stateful(config):
 
     models_path = get_ov_model(model_ids[0])
     for idx, links in enumerate(image_links_list):
-        stateful_pipe = VLMPipeline(models_path, "CPU", **get_default_llm_properties())
+        stateful_pipe = VLMPipeline(models_path, "CPU", ATTENTION_BACKEND="SDPA", **get_default_llm_properties())
```
ilya-lavrenov, line 240: maybe we should also parametrize `test_vlm_pipeline` to have better per-model coverage?

And `test_vlm_with_scheduler_vs_default` seems redundant, as it compares PA vs PA

src/cpp/src/utils.hpp
popovaan: Done.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -324,19 +324,39 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+        m_pimpl = std::make_unique<VLMPipelineImpl>(models_dir, device, properties);
+    }
+
+    // try to call CB adapter one more time, but with safe guard to silent exception
+    if (m_pimpl == nullptr && attention_backend == PA_BACKEND) {
+        try {
+            auto [plugin_properties, scheduler_config] = utils::extract_scheduler_config(properties, utils::get_latency_oriented_scheduler_config());
+            // we need use CB only for x86, as for other architectures like arm64 or risc-v we can create Paged Attention based model
+            // but cannot perform its inference later
+#ifdef OPENVINO_ARCH_X86_64
```
ilya-lavrenov, line 348: looks like in LLM part the condition is also obsolete and we need to allow CB backend for arm64 as well (PA was implemented for arm64 using fp16 inference)

Also, we need to update the comment above.

ilya-lavrenov (approved):
Ok from CB perspective except a single comment about arm64.

Leaving review for Image embedder part for @Wovchena

src/cpp/src/visual_language/pipeline.cpp
popovaan: Done.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -227,9 +231,13 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::generate(
     if (m_is_chat_conversation) {
         if (encoded_results[0].m_status == ov::genai::GenerationStatus::CANCEL) {
             m_history.pop_back();
+            for (size_t i = 0; i < encode_images_num; i++) {
+                m_history_images.pop_back();
+            }
         }
         else {
             m_history.push_back({{"role", "assistant"}, {"content", results[0].texts[0]}});
+            m_image_id += rgbs_vector[0].size();
```
Wovchena, line 240:
```suggestion
            m_image_id += encoded_images.size();
```
So `encode_images_num` isn't needed

src/cpp/src/visual_language/internvl_chat/classes.cpp
```diff
@@ -228,12 +228,26 @@ InputsEmbedderInternVLChat::InputsEmbedderInternVLChat(
...
+    std::string unified_prompt;
+    std::vector<size_t> images_sequence;
 
+    if (image_sequence.size()) {
+        images_sequence = image_sequence;
+        unified_prompt = prompt;
+    } else {
+        auto normalized = normalize_prompt(prompt, m_image_id, images.size());
+        unified_prompt = normalized.first;
+        images_sequence = normalized.second;
```
Wovchena, line 249: Move to VLMPipelineIml, so the new contract is that normalize_prompt() always called outside
Maybee apply_chat_template() can also be called always outside. If this is the case, we could get rid of `m_inputs_embedder->set_apply_chat_template_status(false);`

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -324,19 +324,39 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+    // try to call CB adapter one more time, but with safe guard to silent exception
+    if (m_pimpl == nullptr && attention_backend == PA_BACKEND) {
+        try {
+            auto [plugin_properties, scheduler_config] = utils::extract_scheduler_config(properties, utils::get_latency_oriented_scheduler_config());
+            // we need use CB only for x86 and arm64, as for other architectures like risc-v we can create Paged Attention based model
+            // but cannot perform its inference later
+#if defined(OPENVINO_ARCH_X86_64) || defined(OPENVINO_ARCH_ARM64)
+            m_pimpl = std::make_unique<VLMContinuousBatchingAdapter>(models_dir, scheduler_config, device, plugin_properties);
+#endif
+        } catch (ov::Exception&) {
```
Wovchena, line 351: We'd probably need perf and accuracy validation runs. Please ensure llm_bench and wwb are able to enforce CB with using ATTENTION_BACKEND via config if they aren't. (They are probably fine)

Wovchena (changes requested):

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -176,15 +178,17 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::generate(
         if (!m_inputs_embedder->prompt_has_image_tag(prompt_with_tags)) {
             prompt_with_tags = add_image_tags_to_prompt(prompt_with_tags, rgbs, m_history_images.size());
         }
-        m_history.push_back({{"role", "user"}, {"content", prompt_with_tags}});
+        auto [unified_prompt, images_sequence] = m_inputs_embedder->normalize_prompt(prompt_with_tags, m_image_id, rgbs.size());
```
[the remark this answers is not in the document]
popovaan: Removed `add_image_tags_to_prompt()`.

src/cpp/src/visual_language/llava/classes.cpp
```diff
@@ -114,9 +114,24 @@ std::vector<ov::genai::EncodedImage> InputsEmbedderLLaVA::encode_images(const st
...
+    std::vector<size_t> images_sequence;
+    std::string image_token = m_vlm_config.im_start;
+    
+    if (image_sequence.size()) {
+        images_sequence = image_sequence;
+        unified_prompt = prompt;
+    } else {
+        auto normalized = normalize_prompt(prompt, m_image_id, images.size());
+        unified_prompt = normalized.first;
+        images_sequence = normalized.second;
```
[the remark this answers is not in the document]
popovaan: This code is not needed anymore and was removed.

src/cpp/src/continuous_batching/pipeline_base.cpp
popovaan: Yes, replaced with `encoded_images.size()`.

src/cpp/src/visual_language/internvl_chat/classes.cpp
popovaan: Now `normalize_prompt()` is always called outside and I removed `m_inputs_embedder->set_apply_chat_template_status(false)` from CB chat part.

src/cpp/src/visual_language/pipeline.cpp
popovaan: Yes, both llm_bench and wwb are able to enforce CB with using ATTENTION_BACKEND via config. 

Andrei made validation runs for performance and accuracy. According his report, there's no degradations in accuracy, regarding performance there's run-to-run deviations of Qwen2VL performance +-20%, 
the rest works approximately as stateful or have performance improvements comparing to stateful.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -168,34 +161,29 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::generate(
...
 
     if (m_is_chat_conversation) {
         OPENVINO_ASSERT(1 == prompts.size(), "Can't chat with multiple prompts");
         const auto& rgbs = rgbs_vector[0];
-        auto prompt_with_tags = prompts[0];
-        if (!m_inputs_embedder->prompt_has_image_tag(prompt_with_tags)) {
-            prompt_with_tags = add_image_tags_to_prompt(prompt_with_tags, rgbs, m_history_images.size());
-        }
-        m_history.push_back({{"role", "user"}, {"content", prompt_with_tags}});
+        const auto prompt = prompts[0];
```
Wovchena, line 169:
```suggestion
        const auto& prompt = prompts[0];
```

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -168,34 +161,29 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::generate(
...
+        const auto [unified_prompt, image_sequence] = m_inputs_embedder->normalize_prompt(prompt, m_image_id, rgbs.size());
```
Wovchena, line 170: `rgbs` maybe batched. `encoded_images` should be used

Wovchena (changes requested):

src/cpp/src/visual_language/internvl_chat/classes.cpp
Wovchena: Both lines are removed:
```C++
        std::string templated_history = m_tokenizer.apply_chat_template(m_history, true);
        m_inputs_embedder->set_apply_chat_template_status(false);
```
I suspect one of them is needed.

set_apply_chat_template_status() calls should be removed from everywhere:
> Maybee apply_chat_template() can also be called always outside. If this is the case, we could get rid of `m_inputs_embedder->set_apply_chat_template_status(false);`

src/cpp/src/continuous_batching/pipeline_base.cpp
popovaan: Done.

src/cpp/src/continuous_batching/pipeline_base.cpp
popovaan: Done.

src/cpp/src/visual_language/minicpm/classes.cpp
```diff
@@ -7,6 +7,7 @@
 #include "visual_language/clip.hpp"
 
 #include "utils.hpp"
+#include "debug_utils.hpp"
```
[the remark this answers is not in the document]
popovaan: Removed.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -246,7 +232,8 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::add_request(uint64_t re
     {
         std::lock_guard<std::mutex> lock(m_embeddings_mutex);
         m_inputs_embedder->set_apply_chat_template_status(sampling_params.apply_chat_template);
-        inputs = m_inputs_embedder->get_inputs_embeds(prompt, rgbs, metrics);
+        const auto [unified_prompt, image_sequrnce] = m_inputs_embedder->normalize_prompt(prompt, 0, rgbs.size());
+        inputs = m_inputs_embedder->get_inputs_embeds(unified_prompt, rgbs, metrics, image_sequrnce);
```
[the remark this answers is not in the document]
popovaan: Fixed.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -324,19 +333,39 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+        m_pimpl = std::make_unique<VLMContinuousBatchingAdapter>(models_dir, scheduler_config, device, plugin_properties);
+    }
+
+    if (m_pimpl == nullptr && device == "NPU") {
+        m_pimpl = std::make_unique<VLMPipelineImpl>(models_dir, device, properties);
     }
-    else {
+
+    // try to call CB adapter one more time, but with safe guard to silent exception
+    if (m_pimpl == nullptr && attention_backend == PA_BACKEND) {
```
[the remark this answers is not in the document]
popovaan: Done.

src/cpp/src/utils.cpp
```diff
@@ -21,6 +21,14 @@
 
 #include "sampling/sampler.hpp"
 
+namespace ov {
+
+namespace genai {
+const std::string PA_BACKEND = "PA";
+const std::string SDPA_BACKEND = "SDPA";
```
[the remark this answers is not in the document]
popovaan: Done.

src/cpp/src/visual_language/internvl_chat/classes.cpp
Wovchena: Maybe the passing prompt without history to `get_inputs_embeds()` similar to stateful was used to illustrate the similarity of CB and SDPA...

I definitely remember saying that `InputsEmbedder` should become almost stateless. That contradicts with
> History is now stored in inputs_embedder

CB has to maintain image ids to cache image embeddings. Because of that it's tempting to let CB to manage the whole state including the chat history. To avoid maintaining the history in two different places that are active at the same time, history management should be completely removed from `InputsEmbedder`. That requires Stateful implementation to manage the history similar to CB.

As a result the history must be maintained in only two classes: stateful and CB instead of every `InputsEmbedder` subclass.

src/cpp/src/utils.cpp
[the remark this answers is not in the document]
popovaan: Reverted this change, as we can't use `string_view` for changeable field in the line below and there's no implicit conversion from `string` to `string_view`. 
https://github.com/openvinotoolkit/openvino.genai/blob/34019312e65770ff4431e854f5407c582e2b31c4/src/cpp/src/llm/pipeline.cpp#L71

src/cpp/src/visual_language/internvl_chat/classes.cpp
popovaan: Moved history outside of `InputsEmbedder`, now CB and stateful pipeline store their chat history.

Stateful still uses `m_kv_cache_state` in InputsEmbedder, I couldn't move it outside, as there's a logic of cropping history tokens before merging them with embeddings, that use this state.
https://github.com/openvinotoolkit/openvino.genai/blob/34019312e65770ff4431e854f5407c582e2b31c4/src/cpp/src/visual_language/inputs_embedder.cpp#L140

https://github.com/openvinotoolkit/openvino.genai/blob/34019312e65770ff4431e854f5407c582e2b31c4/src/cpp/src/lm_encoding.cpp#L329-L350

`set_apply_chat_template_status` wasn't moved outside, as it is used not only to apply chat template, but also to set additional `add_special_tokens=false` flag in `encode()`.

https://github.com/openvinotoolkit/openvino.genai/blob/34019312e65770ff4431e854f5407c582e2b31c4/src/cpp/src/visual_language/inputs_embedder.cpp#L121-L127

Wovchena (approved):

src/cpp/src/visual_language/qwen2_5_vl/classes.hpp
```diff
@@ -37,8 +37,7 @@ class InputsEmbedderQwen2_5_VL : public InputsEmbedderQwen2VL {
 protected:
     ov::Tensor run_image_embeddings_merger(
         const std::vector<EncodedImage>& images, 
-        const std::vector<size_t>& images_sequence, 
-        const size_t image_id) override;
+        const std::vector<size_t>& images_sequence) override;
```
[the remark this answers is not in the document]
popovaan: Base class `InputsEmbedderQwen2V`L was changed as well and `run_image_embeddings_merger()` there accepts `const std::vector<EncodedImage>& images` and const` std::vector<size_t>& images_sequence`, so this suggested change is false.

src/cpp/src/visual_language/phi3_vision/classes.cpp
```diff
@@ -662,21 +662,21 @@ InputsEmbedderPhi3V::InputsEmbedderPhi3V(
     const ov::AnyMap device_config) :
     IInputsEmbedder(vlm_config, models_map, tokenizer, config_dir_path, device, device_config) {}
 
-ov::Tensor InputsEmbedderPhi3V::get_inputs_embeds(const std::string& prompt, const std::vector<ov::genai::EncodedImage>& images, ov::genai::VLMPerfMetrics& metrics, bool recalculate_merged_embeddings) {
+std::pair<std::string, std::vector<size_t>> InputsEmbedderPhi3V::normalize_prompt(const std::string& prompt, size_t base_id, const std::vector<EncodedImage>& images) const {
+    return {normalize_prompt_phi3(prompt, base_id, images.size()), {}};
```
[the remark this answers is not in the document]
popovaan: `get_inputs_embeds()` for Phi3 model doesn't use image sequence, so there's no need to propagate it there.

PR #2120, opened 2025-04-28
GGUF support load split files for Qwen2.5

TianmengChen:
Base on https://github.com/openvinotoolkit/openvino.genai/pull/2110

Mainly features:
1. support splited gguf files loading such as [qwen2.5-7b-instruct-q4_0](https://huggingface.co/Qwen/Qwen2.5-7B-Instruct-GGUF/tree/main?show_file_info=qwen2.5-7b-instruct-q4_0-00001-of-00002.gguf) by first file path.
2. change shared_embedding default value to false, because for some gguf model, it cause accuracy issue when GGUF token_embed.weights contains scales/biases with continuous 0 value. 

CVS-166026

PR #2135, opened 2025-04-29
GGUF Q6K WA for GPU

TianmengChen:
Base on https://github.com/openvinotoolkit/openvino.genai/pull/2120

Mainly features:

1. Q6_K is not supported by some platforms GPU because of group size 16, so we use F16 as a workaround
2. Format file gguf.cpp

https://jira.devtools.intel.com/browse/CVS-166026
https://jira.devtools.intel.com/browse/CVS-166027


ilya-lavrenov:
build_jenkins

FionaZZ92:
@TianmengChen Could you please consider to use ov::parallel_for threading, and replace std C++ matrix multiply operation to intrinsic.

ilya-lavrenov:
build_jenkins

ilya-lavrenov (approved):

PR #2146, opened 2025-05-01
Group source files for smart CI

ilya-lavrenov:
Preparation for CVS-165370

If any files needs to renamed - please, suggest in PR.


src/python/py_whisper_pipeline.cpp
```diff
@@ -11,7 +11,7 @@
 #include "openvino/genai/whisper_generation_config.hpp"
 #include "openvino/genai/whisper_pipeline.hpp"
 #include "py_utils.hpp"
-#include "tokenizers_path.hpp"
+#include "tokenizer/tokenizers_path.hpp"
```
Wovchena, line 14: How about removing duplicate words from paths? tokenizer/tokenizers_path.hpp -> tokenizer/path.hpp?
It also seems that continuous_batching_adapter.hpp shouldn't be part of continuous_batching because continuous_batching itself is fine without the adapter.

I'm fine submitting a PR to you if we agree on that

ilya-lavrenov: done. 

Please, check and provide suggestion about file naming / renaming if any

.github/labeler.yml
```diff
@@ -44,9 +44,9 @@
 - 'src/cpp/include/openvino/genai/lora_adapter.hpp'
 - 'src/cpp/src/lora_adapter.cpp'
 - 'src/cpp/src/lora_helper.cpp'
-- 'src/cpp/src/lora_helper.hpp'
+- 'src/cpp/src/helper.hpp'
```
Wovchena, line 47:
```suggestion
- 'src/cpp/src/lora/helper.hpp'
```
Other paths also need fixing here

Wovchena (changes requested):

.github/labeler.yml
ilya-lavrenov: [.github/labeler.yml](https://github.com/openvinotoolkit/openvino.genai/pull/2146/files/82cff0583b1897a03c8e05e83c5039e3ea2f2348#diff-a22c263686553013feaeb0677d26eeb0b8778a756c4311c1fce13384258026aa) is planned to be updated as a second step towards Smart CI

Wovchena (approved):

PR #2158, opened 2025-05-05
Added smart CI for Linux workflow

ilya-lavrenov:


.github/workflows/linux.yml
```diff
@@ -30,7 +30,43 @@ env:
...
+        timeout-minutes: 15
+        with:
+          repository: 'openvinotoolkit/openvino'
+          path: 'openvino'
+          ref: ${{ env.OV_BRANCH }}
+          sparse-checkout: .github/actions/smart-ci
+
+      - name: Get affected components
+        id: smart_ci
+        uses: ./.github/actions/smart-ci
```
akladiev, line 51: Since you're cloning to "openvino" path I think it will be something like
```suggestion
        uses: ./openvino/.github/actions/smart-ci
```

.github/labeler.yml
Wovchena, line 1: Just realized. Can you exclude `.md` from other labels so simple text updates don't trigger tests

ilya-lavrenov: we have:
```
skip_when_only_listed_files_changed: '*.md,*.rst,*.png,*.jpg,*.svg'
```
for it

Wovchena (approved):

PR #2160, opened 2025-05-05
[GHA] Unique ov_cache

mryzhov:
Add date and optimum-intel version to the cache name, it allows to automatically create the new cache when the optimum-intel  version is changed


.github/workflows/cleanup_caches.yml
```diff
@@ -24,11 +24,11 @@ jobs:
         run: |
           echo "Cache info: "
           du -h -d2 ${{ env.OV_CACHE }}
-
+
```
mryzhov, line 27:
```suggestion
```

tests/python_tests/samples/conftest.py
```diff
@@ -278,6 +292,7 @@ def generate_test_content(request):
 
         from PIL import Image
         import numpy as np
+
```
mryzhov, line 295:
```suggestion
```

Wovchena (approved):
Thank you!

Issue #2069, opened 2025-04-17
[WhisperPipeline] Batch Size Mismatch Error When Using Beam Search


HighBridgeDragon:
## Problem Description
When running speech recognition with OpenVINO WhisperPipeline, setting `num_beams=2` or higher causes a batch size mismatch error.

## Error Message
```bash
RuntimeError: Exception from src/inference/src/cpp/infer_request.cpp:245:
Exception from src/plugins/intel_cpu/src/graph.cpp:1613:
Node __module.model.model.decoder.layers.0.self_attn/aten::scaled_dot_product_attention/ScaledDotProductAttention of type ScaledDotProductAttentionWithKVCache
Check 'B == B_state' failed at src/plugins/intel_cpu/src/nodes/scaled_attn.cpp:1693:
ScaledDotProductAttentionWithKVCache node with name '__module.model.model.decoder.layers.0.self_attn/aten::scaled_dot_product_attention/ScaledDotProductAttention' beam idx batch: 1 is not equal to batch of state: 2
```

## Steps to Reproduce
1. Convert the model (first time only):
```bash
optimum-cli export openvino --model openai/whisper-large-v2 --trust-remote-code models/openai/whisper-large-v2/
```

2. Download the sample audio file:
```bash
curl -o samples/how_are_you_doing_today.wav https://storage.openvinotoolkit.org/models_contrib/speech/2021.2/librispeech_s5/how_are_you_doing_today.wav
```

3. Run the following code:

```python
from librosa import load
from openvino_genai import WhisperPipeline

audio, _ = load("samples/how_are_you_doing_today.wav", sr=16000)

pipe = WhisperPipeline("models/openai/whisper-large-v2/", device="CPU")

results = pipe.generate(
    audio,
    language="<|en|>",
    num_beams=2,  # Setting this value to 2 or higher causes the error
)
```
## Expected Behavior
Speech recognition should work normally with beam search enabled (`num_beams=2` or higher).

### OS Information
```
PRETTY_NAME="Ubuntu 24.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.1 LTS (Noble Numbat)"
```

### CPU Information
```
Architecture: x86_64
Model name: Intel(R) Xeon(R) w7-3445
CPU family: 6
Threads per core: 2
Cores per socket: 20
```

## Python Version
```bash
[...]

PR #2166, opened 2025-05-06
Fix whisper pipeline beam search decoding

as-suvorov:
Closes: https://github.com/openvinotoolkit/openvino.genai/issues/2069

### DIFF
--- a/src/cpp/include/openvino/genai/generation_config.hpp
+++ b/src/cpp/include/openvino/genai/generation_config.hpp
@@ -31,9 +31,10 @@ enum class StopCriteria { EARLY, HEURISTIC, NEVER };
  * be used while greedy and beam search parameters will not affect decoding at all.
  *
  * Generic parameters:
- * @param max_length the maximum length the generated tokens can have. Corresponds to the length of the input prompt +
+ * @param max_length the maximum length of generated tokens for each processed audio chunk (30 seconds). Corresponds to the length of the input prompt +
  *        `max_new_tokens`. Its effect is overridden by `max_new_tokens`, if also set.
- * @param max_new_tokens the maximum numbers of tokens to generate, excluding the number of tokens in the prompt. max_new_tokens has priority over max_length.
+ * @param max_new_tokens the maximum numbers of tokens to generate for each processed audio chunk (30 seconds), excluding the number of tokens in the prompt.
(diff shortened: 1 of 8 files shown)


tests/python_tests/test_whisper_pipeline.py
```diff
@@ -232,6 +233,8 @@ def run_pipeline_with_ref(
     for _sample in sample:
         genai_result = run_genai(genai_pipe, _sample, generation_config, streamer)
         hf_result = run_huggingface(hf_pipe, _sample, generation_config)
+        print(f"HF result: {hf_result['text']}")
+        print(f"GenAI result: {genai_result.texts[0]}")
```
Wovchena, line 237: Is it a leftover? I guess pytest's -vv should print the whole diff

as-suvorov: yes, debug print. Thanks for catch, removed

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -31,9 +31,10 @@ enum class StopCriteria { EARLY, HEURISTIC, NEVER };
  * be used while greedy and beam search parameters will not affect decoding at all.
  *
  * Generic parameters:
- * @param max_length the maximum length the generated tokens can have. Corresponds to the length of the input prompt +
+ * @param max_length the maximum length of generated tokens for each processed audio chunk (30 seconds). Corresponds to the length of the input prompt +
  *        `max_new_tokens`. Its effect is overridden by `max_new_tokens`, if also set.
- * @param max_new_tokens the maximum numbers of tokens to generate, excluding the number of tokens in the prompt. max_new_tokens has priority over max_length.
+ * @param max_new_tokens the maximum numbers of tokens to generate for each processed audio chunk (30 seconds), excluding the number of tokens in the prompt.
```
ilya-lavrenov, line 36: it's generic doc. Why do we mention chunks here? maybe we can add a note instead?

as-suvorov: My bad, I mixed it with Whisper config. Not sure where to add a whisper specific note. Maybe we can skip it

ilya-lavrenov (approved):

Wovchena (approved):

PR #2168, opened 2025-05-07
Adjusted smart CI / labeler configs

ilya-lavrenov:


.github/components.yml
```diff
@@ -120,6 +120,12 @@ JS_API:
   revalidate:
     - LLM_samples
 
+CPP_API: # does not participate in smart CI
+  revalidate: []
+
+Python_API: # does not participate in smart CI
+  revalidate: []
```
Wovchena, line 127: My understanding is that `components.yml` is for smart CI only. What's the point listing things that aren't in smart CI?

ilya-lavrenov: it's turned out that smart CI uses all labels, I've excluded CPP and Python API in Smart CI action run.

Wovchena (approved):

PR #2172, opened 2025-05-07
WA for cmake versions >= 4.0

olpipi:
WA for cmake versions >= 4.0
There is only 1 3rd party dependency which is not aligned with cmake 4.0

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -41,6 +41,7 @@ function(ov_genai_build_jinja2cpp)
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-suggest-override")
         endif()
 
+        set(CMAKE_POLICY_VERSION_MINIMUM 3.5)
         add_subdirectory("${jinja2cpp_SOURCE_DIR}" "${jinja2cpp_BINARY_DIR}" EXCLUDE_FROM_ALL)
 
         target_compile_definitions(jinja2cpp PUBLIC JINJA2CPP_LINK_AS_SHARED=0)


src/cpp/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ function(ov_genai_build_jinja2cpp)
             set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-suggest-override")
         endif()
 
+        set(CMAKE_POLICY_VERSION_MINIMUM 3.5)
         add_subdirectory("${jinja2cpp_SOURCE_DIR}" "${jinja2cpp_BINARY_DIR}" EXCLUDE_FROM_ALL)
```
Wovchena, line 45: Can you submit a PR to Jinja then?

olpipi: The problem is not in jinja repo itself but is in its sub-dependencies.
So I cannot solve the problem on jinja level - only move this WA to them. And that doesn't seem reasonable.

ilya-lavrenov: agree with @Wovchena 

I think Jinja users will benefit as well, because cmake 4.x will be working for Jinja then

olpipi: Jinja build works even now with additional cmake cli key. 
We will not solve the issue by moving this line to their repo. Just move WA from our level to their.

Wovchena: I haven't tried cmake 4 yet. What is the problem?
> The problem is not in jinja repo itself but is in its sub-dependencies.

How many sub-dependencies have this problem? Maybe they already fixed them and Jinja only needs to update its dependencies

olpipi: lemme check the latest Jinja revision

PR #2173, opened 2025-05-07
Update Tokenizers

apaniukov:


thirdparty/openvino_tokenizers
Wovchena, line 1: Whisper and VLM failures are probably random, but Tokenizer tests seem to detect the real problems

apaniukov:
Is part of https://github.com/openvinotoolkit/openvino.genai/pull/2080

PR #2113, opened 2025-04-25
[POC] Image generation multi-concurrency idea

dkalinowski:
Increased number of infer requests for each model
Added a way to select infer request by `request_idx`
Added GenerationRequest class to manage assignment of `request_idx`

PR #2190, opened 2025-05-09
Image generation multiconcurrency

dkalinowski:
As requested in https://github.com/openvinotoolkit/openvino.genai/pull/2113#issuecomment-2859299077


samples/cpp/image_generation/text2image.cpp
```diff
@@ -1,27 +1,55 @@
...
-    imwrite("image_%d.bmp", image, true);
+    
+    std::vector<std::thread> threads;
+
+    for (size_t i = 0; i < prompts.size(); ++i) {
+        const std::string p = prompts[i];
+        threads.emplace_back([i, &pipe, models_path, p] () {
+            std::cout << "Generating... " << i << std::endl;
+            ov::genai::Text2ImagePipeline::GenerationRequest request = pipe.create_generation_request();
+            request.set_scheduler(ov::genai::Scheduler::from_config(std::filesystem::path(models_path) / "scheduler/scheduler_config.json"));
```
likholat, line 36: Please move this step to `clone()` methods

src/cpp/src/image_generation/models/autoencoder_kl.cpp
```diff
@@ -182,7 +182,16 @@ AutoencoderKL::AutoencoderKL(const std::string& vae_encoder_model,
     compile(device, *extract_adapters_from_properties(properties));
 }
 
-AutoencoderKL::AutoencoderKL(const AutoencoderKL&) = default;
+AutoencoderKL::AutoencoderKL(const AutoencoderKL& rhs) = default;
+
+AutoencoderKL AutoencoderKL::clone() {
+    OPENVINO_ASSERT(!m_decoder_model, "AutoencoderKL must be compiled first. Cannot clone non-compiled model");
```
ilya-lavrenov, line 188: why not? we can clone even when model is not yet compiled?

src/cpp/src/image_generation/models/clip_text_model.cpp
```diff
@@ -67,6 +68,15 @@ CLIPTextModel::CLIPTextModel(const std::string& model,
 
 CLIPTextModel::CLIPTextModel(const CLIPTextModel&) = default;
 
+CLIPTextModel CLIPTextModel::clone() {
+    OPENVINO_ASSERT(!m_model, "CLIP text encoder model must be compiled first. Cannot clone non-compiled model");
+    CLIPTextModel cloned = *this;
+    cloned.m_request = m_request.get_compiled_model().create_infer_request();
+    // TODO: tokenizer?
```
ilya-lavrenov, line 75: we need to clone tokenizer as well and remove vector of requests from ov::Tokenizer class itself

src/cpp/src/image_generation/stable_diffusion_pipeline.hpp
```diff
@@ -163,6 +165,19 @@ class StableDiffusionPipeline : public DiffusionPipeline {
...
+        // m_scheduler is not cloned, but rather set_scheduler exposed for users via GenerationRequest
+        std::shared_ptr<AutoencoderKL> vae = std::make_shared<AutoencoderKL>(m_vae->clone());
+        std::shared_ptr<CLIPTextModel> clip_text_encoder = std::make_shared<CLIPTextModel>(m_clip_text_encoder->clone());
+        std::shared_ptr<UNet2DConditionModel> unet = std::make_shared<UNet2DConditionModel>(m_unet->clone());
+        std::shared_ptr<StableDiffusionPipeline> pipeline = std::make_shared<StableDiffusionPipeline>(
+            m_pipeline_type,
+            *clip_text_encoder,
+            *unet,
+            *vae);
+        return pipeline;
```
ilya-lavrenov, line 178: we probably need to copy generation config as well ?

dkalinowski: generation config is re-initialized in StableDiffusionPipeline constructor we use here
https://github.com/openvinotoolkit/openvino.genai/blob/f35313b6867d7fecdb904d4e0e1aa685925f3bc3/src/cpp/src/image_generation/stable_diffusion_pipeline.hpp#L130

src/cpp/src/image_generation/models/autoencoder_kl.cpp
dkalinowski: cloning produces new infer request. how can we allow for post-cloning model compilation? there is no ov::CompiledModel to produce infer request from @ilya-lavrenov

Wovchena: Extract the CompiledModel from infer request: https://github.com/openvinotoolkit/openvino/blob/994ac20d7fd7fc3ec415dbfa5f0bf6ad886a8162/src/inference/include/openvino/runtime/infer_request.hpp#L328

dkalinowski: This assertion checks if the model **is compiled**. Only when model is compiled we can get infer request @Wovchena @ilya-lavrenov

Wovchena: I see. It's still should be possible to clone by coping ov::Model. But we can postpone it until the time we agree on the rest of the things

src/cpp/src/image_generation/models/clip_text_model.cpp
dkalinowski: ov::Tokenizer is used in multiple places, including LLM text generation. we have already patched it once to use ireq queue in order to increase throughput in text generation multi concurrency scenarios (detokenization was a bottleneck with thousands of requests with streaming mode) @ilya-lavrenov @Wovchena 

For image generation the tokenizing part of the pipeline is performed only once, I don't think it will be a bottleneck in cloning scenario

samples/cpp/image_generation/text2image.cpp
dkalinowski: Done. Saved root_dir for subsequent usage. Using root_dir we can re-create the scheduler during cloning.

src/cpp/src/image_generation/models/autoencoder_kl.cpp
Wovchena: That's correct

src/cpp/src/image_generation/models/clip_text_model.cpp
Wovchena: This is not a performance problem solving part. This is functional enablement. To avoid a situation when vector<Text2ImagePipeline> results in multiple `CircularBufferQueue<ov::InferRequest>` underneath, Tokenizer should contain only one `InferRequest` and ContinuousBatching should have a mutex guarded `vector<Tokenizer>` in long run. For now it should be enough for `Tokenizer::clone()` to save a pointer to `CircularBufferQueue<ov::InferRequest>`, so we don't modify `ContinuousBatching` class at idea reviewing stage.

src/cpp/src/image_generation/stable_diffusion_pipeline.hpp
Wovchena: That's unfortunate `StableDiffusionPipeline` needs file system access even though everything is already in memory.
Would adding `StableDiffusionPipeline() = default` and cloning every member solve that?

dkalinowski: Even if it would solve that, we still need file system access at runtime due to fact that Scheduler is re-created using saved `root_dir`:

```
pipeline->m_root_dir = m_root_dir;
pipeline->set_scheduler(
  Scheduler::from_config(m_root_dir / "scheduler/scheduler_config.json"));
```

Do you see any way around that? Adding clone to all kinds of schedulers seems like overkill if we can create one from config file

Wovchena: @likholat, does `Scheduler` change it's state during pipeline usage? If not we, the pipeline could store a shared pointer

likholat: @Wovchena yes it changes the state, for example in `set_timesteps` function. I don't think we can reuse the scheduler

src/cpp/src/image_generation/models/clip_text_model.cpp
dkalinowski: There are no multiple circular buffers after cloning. The ClipTextModel has default copy constructor which performs shallow copy of all members, including Tokenizer. Tokenizer also has default copy constructor, which performs shallow copy of shared_ptr<TokenizerImpl> which contains CircularBufferQueue.

Please explain why is it wrong to share the tokenizer between multiple requests. I don't think we need tokenizer clone.

src/cpp/src/image_generation/stable_diffusion_pipeline.hpp
dkalinowski: Is using saved root_dir a no-go?

Wovchena: If the json changes between clone() calls the instances won't be identical. So the final solution can't read the file system.

At the same time this scenario is unlikely. Can you try one of the `Scheduler` with clone() just to ensure the problem can be resolved? After that you can come back to json reading as a temporary solution if @likholat is OK with this plan.

likholat: I think reading from json is fine at this point

samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    const std::string models_path = argv[1];
+    const std::string device = "CPU";  // GPU can be used as well
+
+    std::vector<std::string> prompts;
+    for (int i = 2; i < argc; ++i) {
+        prompts.push_back(argv[i]);
+    }
+
+    ov::genai::Text2ImagePipeline pipe(models_path, device);
```
likholat, line 22: Here in constructor we created one extra unused infer request

dkalinowski: @likholat  is it a big problem? if we dont create infer request, we need to hold ov::CompiledModel and code gets more complicated. I really like the idea of generating new infer request out of previous infer request (which holds ov::CompiledModel reference). What do you think?

samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+    std::vector<std::thread> threads;
+
+    for (size_t i = 0; i < prompts.size(); ++i) {
+        threads.emplace_back([i, &pipe, models_path, prompts] () {
+            std::cout << "Starting to generate with prompt: '" << prompts[i] << "'..." << std::endl;
+
+            ov::genai::Text2ImagePipeline request = pipe.clone();
```
likholat, line 30: As far as I understand we should clone pipeline outside of thread

samples/cpp/image_generation/text2image_concurrency.cpp
likholat: Let's just infer this original pipeline as well

samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -0,0 +1,61 @@
...
+        prompts.push_back(argv[i]);
+    }
+
+    ov::genai::Text2ImagePipeline pipe(models_path, device);
+
+    std::vector<std::thread> threads;
+
+    for (size_t i = 0; i < prompts.size(); ++i) {
+        threads.emplace_back([i, &pipe, models_path, prompts] () {
+            std::cout << "Starting to generate with prompt: '" << prompts[i] << "'..." << std::endl;
```
likholat, line 28: If multiple threads write to `std::cout` at the same time, in the worst case lines can mix. Let's get rid of `std::cout` inside of the thread

samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -0,0 +1,61 @@
...
+    for (int i = 2; i < argc; ++i) {
+        prompts.push_back(argv[i]);
+    }
+
+    ov::genai::Text2ImagePipeline pipe(models_path, device);
+
+    std::vector<std::thread> threads;
+
+    for (size_t i = 0; i < prompts.size(); ++i) {
+        threads.emplace_back([i, &pipe, models_path, prompts] () {
```
likholat, line 27: it's safer and clearer to pre-clone a pipeline (request) per thread and pass only the specific prompt and corresponding request to the thread:
```cpp
threads.emplace_back([i, prompt, &request]()
```

src/cpp/src/image_generation/models/autoencoder_kl.cpp
```diff
@@ -182,7 +182,18 @@ AutoencoderKL::AutoencoderKL(const std::string& vae_encoder_model,
     compile(device, *extract_adapters_from_properties(properties));
 }
 
-AutoencoderKL::AutoencoderKL(const AutoencoderKL&) = default;
+AutoencoderKL::AutoencoderKL(const AutoencoderKL& rhs) = default;
+
+AutoencoderKL AutoencoderKL::clone() {
+    // Support for cloning before compilation implemented below. TODO: Remove assertion when approved
+    //OPENVINO_ASSERT(!m_decoder_model, "AutoencoderKL must be compiled first. Cannot clone non-compiled model");
```
likholat, line 189: Please uncomment these asserts

- For future improvements: we should add `if(m_model)` and support both cases - when the model is compiled and when it's not. If the model isn’t compiled, we should create a copy of the original model

samples/cpp/image_generation/image2image_concurrency.cpp
```diff
@@ -0,0 +1,67 @@
+// Copyright (C) 2023-2025 Intel Corporation
```
likholat, line 1: Please add a couple of lines about these samples to readme

src/cpp/src/image_generation/models/autoencoder_kl.cpp
dkalinowski: Done

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: Why is that? This exactly mimics the use case inside OpenVINO Model Server.

In OVMS, we **do not** clone pipelines before hand - we **do** it in separate thread when HTTP generaton request appears. We keep the original compiled pipeline in memory, and do not use it for other tasks than being the clone source.

However, this is only a sample. Both scenarios should work, but I can change this sample to work differently. What do you think? @likholat

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: Please check https://github.com/openvinotoolkit/openvino.genai/pull/2190#discussion_r2102144487 comment

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: Good catch, will fix it

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: Sure, I can change it, however please check this comment: https://github.com/openvinotoolkit/openvino.genai/pull/2190#discussion_r2102144487

src/cpp/src/image_generation/models/autoencoder_kl.cpp
dkalinowski: I'm sorry, but I don't understand this request. In previous review you wanted me to add support for cloning-before-compilation. Reverting those asserts completely removes this feature...

samples/cpp/image_generation/image2image_concurrency.cpp
dkalinowski: Will do!

samples/cpp/image_generation/text2image_concurrency.cpp
Wovchena: I think `ireq.get_compiled_mode().create_infer_request()` isn't thread safe. Am I wrong?

dkalinowski: In OpenVINO there are `create_infer_request` stress tests which run creation in separate threads: https://github.com/openvinotoolkit/openvino/blob/master/tests/stress_tests/unittests/tests.cpp#L40-L42
So I'm assuming this is thread safe.

samples/cpp/image_generation/image2image_concurrency.cpp
dkalinowski: Pushed the changes.

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: Pushed the changes.

samples/cpp/image_generation/text2image_concurrency.cpp
likholat: It doesn't look like the same case, it creates new infer request each time from a new separate model object:

https://github.com/openvinotoolkit/openvino/blob/9315206a0eed9605fb96a82b8c3ba8686c0c616a/tests/stress_tests/unittests/tests_pipelines/tests_pipelines.cpp#L61

https://github.com/openvinotoolkit/openvino/blob/9315206a0eed9605fb96a82b8c3ba8686c0c616a/tests/stress_tests/common/ie_pipelines/pipelines.cpp#L66

src/cpp/src/image_generation/models/autoencoder_kl.cpp
dkalinowski: I understand now, we should also clone m_model in case of cloning before compilation
use clone method of ov::Model

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: TODO: Move cloning before thread creation, lock manually in OVMS scenario

src/cpp/src/image_generation/stable_diffusion_pipeline.hpp
dkalinowski: In separate task: find a way to clone scheduler without reading from disk

samples/cpp/image_generation/README.md
```diff
@@ -210,3 +212,36 @@ Test finish, load time: 9356.00 ms
...
+
+
+```cpp
+ov::genai::Text2ImagePipeline pipe(models_path, device);
+
+std::vector<std::thread> threads;
+
+for (int i = 0; i < 4; i++) {
+  threads.emplace_back([&pipe, i] {
+    auto request = pipe.clone();
```
likholat, line 238: I think, we need to align it with the sample code

samples/cpp/image_generation/image2image_concurrency.cpp
```diff
@@ -0,0 +1,68 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include <vector>
```
likholat, line 4: Tried to run this sample with 
```
./image2image_concurrency dreamlike_anime_1_0_ov/FP16 "A painting of a futuristic city" "A fantasy castle in the clouds" img.png
```
but got error (same for text2image_concurrency):
```
Starting to generate with prompt: 'A painting of a futuristic city'...
Starting to generate with prompt: 'A fantasy castle in the clouds'...
terminate called after throwing an instance of 'ov::Busy'
  what():  Infer Request is busy
Aborted (core dumped)
```

src/cpp/src/image_generation/models/autoencoder_kl.cpp
```diff
@@ -182,7 +182,19 @@ AutoencoderKL::AutoencoderKL(const std::string& vae_encoder_model,
...
+
+AutoencoderKL AutoencoderKL::clone() {
+    AutoencoderKL cloned = *this;
+    if (m_encoder_model) {
+        cloned.m_encoder_model = m_encoder_model->clone();
+    } else {
+        if (m_encoder_request) {
+            cloned.m_encoder_request = m_encoder_request.get_compiled_model().create_infer_request();
+        }
+    }
```
likholat, line 195: This implementation only handles the encoder, but `AutoencoderKL `also includes decoder (`m_decoder_model` / `m_decoder_request`)

src/cpp/src/image_generation/models/autoencoder_kl.cpp
```diff
@@ -182,7 +182,19 @@ AutoencoderKL::AutoencoderKL(const std::string& vae_encoder_model,
     compile(device, *extract_adapters_from_properties(properties));
 }
 
-AutoencoderKL::AutoencoderKL(const AutoencoderKL&) = default;
+AutoencoderKL::AutoencoderKL(const AutoencoderKL& rhs) = default;
+
+AutoencoderKL AutoencoderKL::clone() {
+    AutoencoderKL cloned = *this;
```
likholat, line 188: Let's add assert:
```
OPENVINO_ASSERT((m_encoder_model != nullptr) ^ (m_encoder_request!= nullptr), "AutoencoderKL must have exactly one of m_encoder_model or m_encoder_request initialized");
```
and then simplify to:
```cpp
 if (m_encoder_model) {
        cloned.m_encoder_model = m_encoder_model->clone();
 } else {
        cloned.m_encoder_request = m_encoder_request.get_compiled_model().create_infer_request();
 }
```
Same applies to other classes as well

src/cpp/src/image_generation/models/autoencoder_kl.cpp
dkalinowski: Good catch! Fixed it, cloning both now.

samples/cpp/image_generation/image2image_concurrency.cpp
likholat: No longer reproducible after the latest changes, it was related to decoder cloning

samples/cpp/image_generation/README.md
```diff
@@ -210,3 +212,43 @@ Test finish, load time: 9356.00 ms
 Warmup number:1, first generate warmup time:85008.00 ms, infer warmup time:84999.88 ms
 Generate iteration number:3, for one iteration, generate avg time: 84372.34 ms, infer avg time:84363.95 ms, all text encoders infer avg time:76.67 ms, vae encoder infer avg time:0.00 ms, vae decoder infer avg time:4470.33 ms
 ```
+
+### Run multiple generations with different prompt in parallel
```
likholat, line 216: We can do it later, but we need to align these instructions with those for other samples. They should show how to run the multiconcurrency samples

src/cpp/src/image_generation/models/sd3_transformer_2d_model.cpp
```diff
@@ -62,6 +62,20 @@ SD3Transformer2DModel::SD3Transformer2DModel(const std::string& model,
 
 SD3Transformer2DModel::SD3Transformer2DModel(const SD3Transformer2DModel&) = default;
 
+SD3Transformer2DModel SD3Transformer2DModel::clone() {
```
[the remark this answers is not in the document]
dkalinowski: Done

dtrawins (approved):

likholat (approved):

samples/cpp/image_generation/README.md
```diff
@@ -210,3 +212,43 @@ Test finish, load time: 9356.00 ms
...
+
+
+```cpp
+std::vector<ov::genai::Text2ImagePipeline> pipelines;
+
+// Prepare initial pipeline and compiled models into device
+pipelines.emplace_back(models_path, device);
+// Clone pipeline for concurrent usage
+for (int i = 1; i < 4; i++)
+   pipelines.emplace_back(pipelines.begin()->clone());
```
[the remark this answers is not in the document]
dkalinowski: Fixed

samples/cpp/image_generation/README.md
```diff
@@ -210,3 +212,43 @@ Test finish, load time: 9356.00 ms
...
+
+std::vector<std::thread> threads;
+
+for (int i = 0; i < 4; i++) {
+  auto& pipe = pipelines.at(i);
+  threads.emplace_back([&pipe, i] {
+    auto request = pipe.clone();
```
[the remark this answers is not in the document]
dkalinowski: No, this is not required, it is leftover. Removed it

src/cpp/include/openvino/genai/image_generation/clip_text_model.hpp
```diff
@@ -91,9 +94,12 @@ class OPENVINO_GENAI_EXPORTS CLIPTextModel {
 private:
     Config m_config;
     AdapterController m_adapter_controller;
+
+protected:
     ov::InferRequest m_request;
     std::shared_ptr<ov::Model> m_model;
 
+private:
     Tokenizer m_clip_tokenizer;
```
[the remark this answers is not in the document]
dkalinowski: Fixed!

PR #2201, opened 2025-05-12
Whisper static pipeline: fix for fp8 models

eshiryae:
Fixes transformation passes for fp8 model as it has additional FakeConvert layer for input key tensors.

Ticket: 166064


dmatveev:
Let's merge it!

PR #2211, opened 2025-05-14
Adding KVCrush technique for cache eviction

gopikrishnajha:


rkazants:
@gopikrishnajha, please provide description in PR: performance/accuracy benefits, code example

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
         return m_eviction_config.get_max_cache_size() + m_block_size - 1;
     }
 
+    std::vector<std::size_t> CacheEvictionAlgorithm::get_indices_of_blocks_to_retain_using_kvcrush(size_t decoder_layer_idx, size_t k, std::vector<std::size_t>& evicted_block_indices)
```
vshampor, line 28: This function is too large and has non-trivial, computational logic. I suggest setting up a new class `KVCrushAlgorithm`, extracting the steps 1-4 into its public methods (either each step into a separate method, or with another granularity that is still conducive to unit testing), and testing each method with a separate, parameterized unit test suite.

vshampor (changes requested):
In addition to resolving my inline comments, please run your new code through a formatter before pushing again. We have a `.clang-format` file in the repo root, but sadly, the rest of the repo does not conform to it, so you'll have to format your changes only. `git clang-formatter` works well, see the [info page](https://offlinemark.com/surgical-formatting-with-git-clang-format/) for instructions.

src/cpp/include/openvino/genai/cache_eviction.hpp
```diff
@@ -19,14 +19,23 @@ enum class AggregationMode {
...
+
 /**
 * @brief Configuration struct for the cache eviction algorithm.
 */
 class CacheEvictionConfig {
 public:
     CacheEvictionConfig() = default;
 
-    CacheEvictionConfig(size_t start_size, size_t recent_size, size_t max_cache_size, AggregationMode aggregation_mode_, bool apply_rotation_ = false) : aggregation_mode(aggregation_mode_), apply_rotation(apply_rotation_), m_start_size(start_size), m_recent_size(recent_size), m_max_cache_size(max_cache_size) {
+    CacheEvictionConfig(size_t start_size, size_t recent_size, size_t max_cache_size, size_t cl_size, AnchorPoints anchor_point_, AggregationMode aggregation_mode_, bool apply_rotation_ = false) : anchor_point(anchor_point_),aggregation_mode(aggregation_mode_), apply_rotation(apply_rotation_), m_start_size(start_size), m_recent_size(recent_size), m_max_cache_size(max_cache_size), m_cl_size(cl_size) {
```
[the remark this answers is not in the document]
gopikrishnajha: added new parameter before apply_rotation parameter, otherwise we will be forced to provide default values as apply_rotation_ = false will force all the subsequent parameters to have a value.

src/cpp/include/openvino/genai/cache_eviction.hpp
```diff
@@ -55,6 +64,12 @@ class CacheEvictionConfig {
         return m_max_cache_size;
     }
 
+    /** @return cache budget (number of tokens) used by KVCrush for representative tokens -
```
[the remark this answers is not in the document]
gopikrishnajha: Done

src/cpp/include/openvino/genai/cache_eviction.hpp
```diff
@@ -19,14 +19,23 @@ enum class AggregationMode {
                 * of a given token in cache */
 };
 
+//anchor points for kvcrush
+enum class AnchorPoints {
```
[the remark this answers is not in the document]
gopikrishnajha: Done

src/cpp/include/openvino/genai/cache_eviction.hpp
```diff
@@ -64,6 +79,8 @@ class CacheEvictionConfig {
     /** The mode used to compute the importance of tokens for eviction */
     AggregationMode aggregation_mode = AggregationMode::NORM_SUM;
 
+    AnchorPoints anchor_point = AnchorPoints::RANDOM;
```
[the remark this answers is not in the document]
gopikrishnajha: Done

src/cpp/include/openvino/genai/cache_eviction.hpp
```diff
@@ -89,6 +106,8 @@ class CacheEvictionConfig {
     std::size_t m_max_cache_size = 672;
     std::size_t m_evictable_size = 512;
 
+    /*KVCrush Cache budget - number of tokens*/
+    std::size_t m_cl_size = 128;
```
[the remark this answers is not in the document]
gopikrishnajha: took away the cluster abstraction.

regarding "_If not, there should probably be a runtime check that the budget size in tokens is divisible by the block size without remainder, as for the rest of the token-sized fields of this struct_." - we allow the user to specify any number of token as kvcrush budget, and we use the highest value divisible by the block size.

src/python/py_continuous_batching_pipeline.cpp
```diff
@@ -224,16 +225,27 @@ void init_continuous_batching_pipeline(py::module_& m) {
...
+            .value("ZEROS", AnchorPoints::ZEROS)
+            .value("ONES", AnchorPoints::ONES)
+            .value("MEAN", AnchorPoints::MEAN)
+            .value("ALTERNATE", AnchorPoints::ALTERNATE);
 
     py::class_<CacheEvictionConfig>(m, "CacheEvictionConfig", cache_eviction_config_docstring)
-            .def(py::init<>([](const size_t start_size, size_t recent_size, size_t max_cache_size, AggregationMode aggregation_mode, bool apply_rotation) {
-                return CacheEvictionConfig{start_size, recent_size, max_cache_size, aggregation_mode, apply_rotation}; }),
-                 py::arg("start_size"), py::arg("recent_size"), py::arg("max_cache_size"), py::arg("aggregation_mode"), py::arg("apply_rotation") = false)
+            .def(py::init<>([](const size_t start_size, size_t recent_size, size_t max_cache_size, size_t cl_size, AnchorPoints anchor_point, AggregationMode aggregation_mode, bool apply_rotation) {
```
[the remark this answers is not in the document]
gopikrishnajha: Done

tests/cpp/cache_eviction.cpp
```diff
@@ -227,7 +228,18 @@ TEST_P(CacheEvictionLowScoreBlocksParameterizedTest, EvictsLowestScoredBlocks) {
     auto test_evicted_blocks = algo.evict_logical_blocks();
     auto ref_evicted_blocks = test_struct.ref_evicted_blocks;
     for (size_t layer_idx = 0; layer_idx < num_decoder_layers; layer_idx++) {
-        EXPECT_EQ(test_evicted_blocks[layer_idx], ref_evicted_blocks[layer_idx]);
+        //if kvcrush is enabled
```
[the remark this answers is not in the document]
gopikrishnajha: I am not sure how extracting this as separate test would help. As KVCrush adds blocks on top of blocks selected by score based method. So, if kvcrush is not enabled, this will work like earlier, but if the kvcrush is enabled BOTH score based eviction and kvcrush should work, just kvcrush won't make any sense. Am I missing anything here?

tests/cpp/cache_eviction.cpp
```diff
@@ -384,23 +396,24 @@ struct CacheEvictionConfigInitParamsForTest {
...
 using CacheEvictionConfigInitializationTest = ::testing::TestWithParam<CacheEvictionConfigInitParamsForTest>;
 
 const std::vector<CacheEvictionConfigInitParamsForTest> INVALID_CONFIG_INIT_PARAMS_CASES = {
         // zero area sizes
-        {32, 32, 64},
-        {0, 13, 39},
-        {128, 0, 384},
+        {32, 32, 64, 10},
+        {0, 13, 39, 10},
+        {128, 0, 384, 10},
```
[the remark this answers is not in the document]
gopikrishnajha: I don't think any alignment on kvcrush_budget is required. We allow user to specify any number of token as the kvcrush budget, we use the largest number of blocks within that budget that can be accommodated within that budget.

tests/python_tests/test_kv_cache_eviction.py
```diff
@@ -193,6 +193,7 @@ class LongBenchTestData:
 def test_optimized_generation_longbench(device, test_struct):
     seqs_per_request = 32
     num_kv_blocks = 1000 if device == "CPU" else 500
+    print("device", device, "num_kv_blocks", num_kv_blocks)
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/src/continuous_batching/cache_eviction.cpp
gopikrishnajha: Broke this down to 4 functions within the existing class as I am not sure where the new class will exist in terms of hierarchy, within CacheEvictionAlgorithm class or independently. 

Can you make the necessary changes, and create test suite for these methods if required. I am not well versed with the google test benches, so it might require more time for me. Considering code freeze is tomorrow, your help will be appreciated.

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
...
+    {
+        const auto& keep_clus_eligible = evicted_block_indices;
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
...
+        std::iota(indices.begin(), indices.end(), 0);
+        std::partial_sort(
+            indices.begin(),
+            indices.begin() + k/2,
+            indices.end(),
+            [&](size_t i, size_t j) {
+                return m_scores[decoder_layer_idx][i] > m_scores[decoder_layer_idx][j];
+            }
+        );
+        std::vector<int> indicators(k, 0);
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
...
+        for (size_t i = 0; i < k/2; ++i) {
+            indicators[indices[i]] = 1;
+        }
+
+        // Step 2: Create a random binary vector of size m_block_size as anchor point
+        std::vector<int> anchor_point(m_block_size);
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
...
+        // Initialize anchor_point based on anchor using switch-case
+        switch (m_eviction_config.anchor_point) {
+            case AnchorPoints::RANDOM:
+            std::generate(anchor_point.begin(), anchor_point.end(), []() { return rand() % 2; });
```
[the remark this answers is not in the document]
gopikrishnajha: done, please verify once.

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
...
+            }
+            break;
+            }
+            case AnchorPoints::ALTERNATE:
+            for (size_t i = 0; i < m_block_size; ++i) {
+                anchor_point[i] = i % 2;
+            }
+            break;
+            default:
+            throw std::invalid_argument("Invalid anchor point type");
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -25,6 +25,106 @@ namespace ov::genai {
...
+        }
+        
+        // Step 3: Calculate Hamming distances between anchor point and each block
+        size_t num_blocks = k / m_block_size;
```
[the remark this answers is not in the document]
gopikrishnajha: we need all the blocks, as we are finding out the distances w.r.t. all blocks and then considering the ones (by filtering) that are evicted based on low scores.

src/cpp/include/openvino/genai/cache_eviction.hpp
```diff
@@ -55,6 +64,12 @@ class CacheEvictionConfig {
         return m_max_cache_size;
     }
 
+    /** @return cache budget (number of tokens) used by KVCrush for representative tokens - 
+    on top  of tokens with high attention scores*/
+     std::size_t get_kvcrush_budget() const {
+        return m_cl_size;
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -51,6 +150,36 @@ namespace ov::genai {
             size_t num_blocks_to_evict = get_num_blocks_to_evict(decoder_layer_idx);
             auto evicted_block_indices = get_indices_of_blocks_to_evict(scores_for_all_evictable_blocks, num_blocks_to_evict);
 
+            //KVCrush: start 
+            bool is_kvcrush_enabled = (m_eviction_config.get_kvcrush_budget() > 0) && (std::ceil(static_cast<double>(evicted_block_indices.size())) >= std::floor(static_cast<double>(m_eviction_config.get_kvcrush_budget()) / m_block_size));
```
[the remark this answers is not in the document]
gopikrishnajha: done

src/cpp/include/openvino/genai/cache_eviction.hpp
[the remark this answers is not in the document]
vshampor: Why not provide default values then?

src/cpp/include/openvino/genai/cache_eviction.hpp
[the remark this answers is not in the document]
vshampor: The behaviour you describe would be confusing for the user since incrementing or decrementing the budget by values less than the block size will lead to no observable effect. I think that this would be frustrating for the user, or at the very least be bad UX. Please implement the restriction on the budget size - the added bonus would be that you won't need to round the budget up in downstream code.

src/cpp/src/continuous_batching/cache_eviction.cpp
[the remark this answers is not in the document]
vshampor: Please extract the seed as a parameter for the KVCrush algorithm part and in the entire pipeline I suggest reusing the `GenerationConfig` [field](https://github.com/openvinotoolkit/openvino.genai/blob/34019312e65770ff4431e854f5407c582e2b31c4/src/cpp/include/openvino/genai/generation_config.hpp#L124) to initialize this parameter.

src/cpp/src/continuous_batching/cache_eviction.cpp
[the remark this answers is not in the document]
vshampor: When you say "all the blocks", do you mean the number of full blocks from `num_tokens_in_evictable_blocks` tokens (excluding the unfilled block if it is present) or the number of affected blocks (including the unfilled block)?

tests/cpp/cache_eviction.cpp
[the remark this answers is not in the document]
vshampor: I was suggesting that instead of adding an `if` branch in the already existing suite you first extract the code in [these lines](https://github.com/openvinotoolkit/openvino.genai/blob/34019312e65770ff4431e854f5407c582e2b31c4/tests/cpp/cache_eviction.cpp#L209-L228) into a GTest fixture, refactoring the existing `CacheEvictionLowScoreBlocksParameterizedTest` to use that, then adding your new logic in the KVCrush-specific parameterized test such as `CacheEvictionLowScoreBlocksKVCrushParameterizedTest` that would reuse the same fixture. This would have led to more granular test cases without code duplication.

But I have looked at your code again and it seems that the test case should be made tighter - I don't think that your subset-checking approach is the right way here since it would not indicate a failure if your algo wasn't working at all, i.e. returned the entire `test_evicted_blocks` subset without filtering, since that would also be a subset of itself. I suggest that you do not modify the test code here at all and instead set up separate cases for KVCrush-enabled scenarios, with references tailored to the KVCrush specifics. This means that `DEFAULT_CACHE_EVICTION_CONFIG` and `SHORT_CACHE_EVICTION_CONFIG` should correspond to *disabled* KVCrush (as they were before), and you could introduce another `KVCRUSH_CACHE_EVICTION_CONFIG` where KVCrush would be enabled; then add corresponding entries to `LOW_SCORE_BLOCK_EVICTION_TEST_CASES` container that specify that new `KVCRUSH_CACHE_EVICTION_CONFIG` to be used. You would be completely reusing existing test code and there would be no need for refactoring.

src/cpp/src/continuous_batching/cache_eviction.cpp
vshampor: In terms of hierarchy, the new class itself would be more or less independent. If you wish or need to, you could create separate .hpp/.cpp files for the new class at the same level we have `cache_eviction.hpp/.cpp`; otherwise, put declaration and implementations into the existing `cache_eviction.hpp/.cpp`. The `CacheEvictionAlgorithm` would be instantiating the new class and using its methods at the same point where you have your KVCrush code starting now, and you will also be able to test the new class independently from the rest of the eviction code. I think that the file containing tests for the new class would have to be separated from the rest of the eviction tests in any case.

vshampor (changes requested):
The PR has become much better, and there could be a compromise on the parameter order, but I think that the separation of the KVCrush algo into a separate class is a must right now. In the current PR there is no unit-level coverage for the new logic at all - the proposed test approach rather checks the entire eviction algorithm, and although *those* cases should also be extended to check holistic functioning with KVCrush enabled, there would have to be tests specifically for the KVCrush logic and that logic only. The only way to achieve this is by extracting the KVCrush logic into a separate class object, as I have suggested in one of the comments. Right now there is no guarantee that the new logic works as expected at all, and there is a chance that you would uncover and fix existing hidden bugs in your new logic as you are implementing the proper unit tests.

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -3,19 +3,28 @@
...
-    }
+CacheEvictionAlgorithm::CacheEvictionAlgorithm(const CacheEvictionConfig& eviction_config,
+                                               size_t block_size,
+                                               size_t num_decoder_layers)
+    : m_eviction_config(eviction_config),
+      m_block_size(block_size),
+      m_num_decoder_layers(num_decoder_layers),
+      m_cache_counter(num_decoder_layers),
+      m_scores(num_decoder_layers),
+      m_kvcrush_algo(KVCrushAlgorithm(eviction_config.get_kvcrush_config())) {
```
vshampor, line 16: There is an extra object creation and copy on this line, you should be passing constructor parameters to the field name directly without class name being involved.

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -51,6 +150,36 @@ namespace ov::genai {
...
+            bool is_kvcrush_enabled = (m_eviction_config.get_kvcrush_budget() > 0) && (std::ceil(static_cast<double>(evicted_block_indices.size())) >= std::floor(static_cast<double>(m_eviction_config.get_kvcrush_budget()) / m_block_size));
+            if(is_kvcrush_enabled)
+            {
+                size_t k = scores_for_all_evictable_blocks.size() * m_block_size;
+                auto kvcrush_retained_block_indices = get_indices_of_blocks_to_retain_using_kvcrush(decoder_layer_idx, k, evicted_block_indices);
+
+                // Remove the indcies in kvcrush_retained_block_indices from evicted_block_indices
+                if (!kvcrush_retained_block_indices.empty()) {
+                    // Convert both vectors to sets for efficient operations
+                    std::unordered_set<std::size_t> retained_set(
```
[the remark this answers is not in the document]
vshampor: Waiting for a fix here

src/cpp/src/continuous_batching/kvcrush.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include "continuous_batching/attention_output.hpp"
+#include "openvino/genai/cache_eviction.hpp"
+#include "openvino/openvino.hpp"
+
+namespace ov::genai {
+
+class KVCrushAlgorithm {
+public:
+    KVCrushAlgorithm() =
+        default;  // needed only to satisfy DefaultConstructible so that algo objects may be used as values in std::map
```
vshampor, line 19: Do you use `KVCrushAlgorithm` in an `std::map` anywhere in the code?

src/cpp/src/continuous_batching/kvcrush.hpp
```diff
@@ -0,0 +1,59 @@
...
+    explicit KVCrushAlgorithm(const KVCrushConfig& eviction_config);
+
+    /** @return A vector of size kvcrush_budget, where each element contains the index of a
+     * representative block selected using KVCrush algorithm */
+    std::vector<std::size_t> get_indices_of_blocks_to_retain_using_kvcrush(
+        size_t decoder_layer_idx,
+        size_t block_size,
```
vshampor, line 26: `block_size` is better suited to be a class field initialized via constructor argument since it does not change during algorithm object's lifetime

src/cpp/src/continuous_batching/kvcrush.cpp
```diff
@@ -0,0 +1,173 @@
...
+// step 2: create_anchor_point_kvcrush()
+std::vector<size_t> KVCrushAlgorithm::create_anchor_point_kvcrush(size_t num_tokens_in_evictable_blocks,
+                                                                  size_t block_size,
+                                                                  std::vector<size_t>& indicators) {
+    // Step 2: Create a binary vector of size block_size as anchor point
+    std::vector<size_t> anchor_point(block_size);
+    // Initialize anchor_point based on anchor using switch-case
+    switch (m_kvcrush_config.anchor_point_mode) {
+    case KVCrushAnchorPointMode::RANDOM: {
+        static std::mt19937 rng(m_kvcrush_config.rng_seed);
```
vshampor, line 42: `rng` should be a private field in the class.

src/cpp/include/openvino/genai/kvcrush.hpp
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/**
+ * @brief Configuration struct for the cache eviction algorithm.
+ */
+class KVCrushConfig {
+public:
+    KVCrushConfig() = default;
+
+    KVCrushConfig(size_t budget_, KVCrushAnchorPointMode anchor_point_mode_, size_t rng_seed_ = 0)
```
vshampor, line 30: Please document the class entirely (each constructor except for the `default`ed, each function argument and return value.

tests/cpp/cache_eviction.cpp
```diff
@@ -227,11 +279,28 @@ TEST_P(CacheEvictionLowScoreBlocksParameterizedTest, EvictsLowestScoredBlocks) {
...
+             (test_struct.eviction_config.get_kvcrush_budget()));
+        if (should_apply_kvcrush) {
+            // check test_evicted_blocks is a subset of ref_evicted_blocks
+            EXPECT_TRUE(std::includes(ref_evicted_blocks[layer_idx].begin(),
+                                      ref_evicted_blocks[layer_idx].end(),
+                                      test_evicted_blocks[layer_idx].begin(),
+                                      test_evicted_blocks[layer_idx].end()));
+            EXPECT_EQ(static_cast<int>(test_evicted_blocks[layer_idx].size()),
+                      static_cast<int>(ref_evicted_blocks[layer_idx].size()) -
+                          static_cast<int>(test_struct.eviction_config.get_kvcrush_budget()));
```
vshampor, line 295: The problem with these test requirements remains that these are potentially flaky (run-to-run unstable) because of the RNG involved. However, now that you have extracted the RNG controls and able to set a specific constant seed for the sake of testing - instead of these `EXPECT_` conditions, compare exact evicted block indices like it is done in the rest of the cases in this test. To learn which blocks get evicted, please run your new KVCrush test cases in the debugger or use print statements - you will also have an opportunity to mentally check if the outputs are correct before commiting them as references to the test cases.

tests/cpp/cache_eviction.cpp
```diff
@@ -227,11 +279,28 @@ TEST_P(CacheEvictionLowScoreBlocksParameterizedTest, EvictsLowestScoredBlocks) {
     auto test_evicted_blocks = algo.evict_logical_blocks();
     auto ref_evicted_blocks = test_struct.ref_evicted_blocks;
     for (size_t layer_idx = 0; layer_idx < num_decoder_layers; layer_idx++) {
-        EXPECT_EQ(test_evicted_blocks[layer_idx], ref_evicted_blocks[layer_idx]);
+        // if kvcrush is enabled
```
vshampor, line 282: Unnecessary comment

tests/cpp/cache_eviction.cpp
```diff
@@ -202,6 +210,50 @@ const std::vector<LowScoreBlocksTestStruct> LOW_SCORE_BLOCK_EVICTION_TEST_CASES
                 {{0, 9, 10, 11, 13}, {12, 11, 8, 9, 17}},
                 {{8, 9, 10, 11, 13}, {8, 9, 10, 11, 12}}
         },
+
+        // kvcrush enabled cases
+        //low-scored blocks in evictable area
```
vshampor, line 215: Please add a tag to each KVCrush-enabled test case comment instead of area effect comments, e.g. `// low-scored blocks in evictable area - with KVCrush`.
Also please remove the test cases for which it is obvious that they will never pass the `bool should_apply_kvcrush` condition in the test case body.

tests/cpp/cache_eviction.cpp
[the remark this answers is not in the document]
vshampor: Did I convince you to switch to the necessity of block-aligned budget specification?

vshampor (changes requested):
The PR has definitiely gone in the right direction, good job. Still some work to do on top of the new comments:
1. I don't see unit tests for the new `KVCrushAlgorithm` class, perhaps you forgot to add the new file to the git commit?
2. Within the scope of this PR, please also add the extended documentation for the new algorithm in https://github.com/openvinotoolkit/openvino.genai/blob/master/site/docs/concepts/optimization-techniques/kvcache-eviction-algorithm.md. You may add SVG images if you need illustrations.
3. Could you post the `similarity` metric results of the `test_optimized_generation_longbench` with the KVCrush enabled vs. disabled (and other necessary params), so that we could see the scale of improvements and whether it aligns with the Python POC?

src/cpp/include/openvino/genai/kvcrush.hpp
```diff
@@ -0,0 +1,60 @@
...
+ */
+/**
+ * @class KVCrushConfig
+ * @brief Configuration class for KVCrush cache mechanism.
+ *
+ * This class encapsulates the configuration parameters for the KVCrush cache,
+ * including cache budget, anchor point mode, and random seed.
+ */
+
+/**
```
vshampor, line 34: This doc block should be placed above ctor, not outside the class

src/cpp/include/openvino/genai/kvcrush.hpp
```diff
@@ -0,0 +1,60 @@
+// Copyright (C) 2023-2025 Intel Corporation
```
vshampor, line 1: Please keep the contents of this file withing the `src/cpp/include/openvino/genai/cache_eviction.hpp`. KVCrush conceptually belongs to the cache eviction mechanism and does not warrant an own public header file.

src/cpp/src/continuous_batching/cache_eviction.cpp
```diff
@@ -3,19 +3,28 @@
...
+                    "CacheEvictionConfig.start_size in tokens must be a multiple of block size ",
+                    m_block_size);
+    OPENVINO_ASSERT(!(m_eviction_config.get_recent_size() % m_block_size),
+                    "CacheEvictionConfig.recent_size in tokens must be a multiple of block size ",
+                    m_block_size);
+    OPENVINO_ASSERT(!(m_eviction_config.get_max_cache_size() % m_block_size),
+                    "CacheEvictionConfig.max_cache_size in tokens must be a multiple of block size ",
+                    m_block_size);
+    OPENVINO_ASSERT(m_num_decoder_layers, "num_decoder_layers must be non-zero");
+}
```
vshampor, line 27: Please avoid reformatting the code that does not belong to this PR

tests/python_tests/test_kvcrush.py
```diff
@@ -0,0 +1,212 @@
+# Copyright (C) 2023-2025 Intel Corporation
```
vshampor, line 1: How is the code in this file different from the existing cases in `test_kv_cache_eviction.py` except for prints and minor parameterization details? Please keep the KVCrush test cases in the `test_kv_cache_eviction.py` and properly reuse code from the already existing cases to avoid copy-paste.

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+// Test create_anchor_point_kvcrush with RANDOM mode
+TEST_F(KVCrushAlgorithmTest, CreateAnchorPointRandomTest) {
+    // Setup
+    size_t num_tokens_in_evictable_blocks = 16;
+    std::vector<size_t> indicators(num_tokens_in_evictable_blocks, 1);  // All indicators set to 1
+
+    m_kvcrush_config.anchor_point_mode = KVCrushAnchorPointMode::RANDOM;
+    m_kvcrush_algo = KVCrushAlgorithm(m_kvcrush_config, m_block_size);
+    auto anchor_point = m_kvcrush_algo.create_anchor_point_kvcrush(num_tokens_in_evictable_blocks, indicators);
```
vshampor, line 86: This part seems to be identical in all of your `CreateAnchorPoint...Test` cases. You could extract it into a `setup_anchor_points` function of a fixture class derived from your existing base fixture and call the `setup_anchor_points` function manually with anchor point mode as a parameter before doing case-specific verification, such as in this [SO question](https://stackoverflow.com/questions/38207346/specify-constructor-arguments-for-a-google-test-fixture)

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    // Since we're using RANDOM mode with a fixed seed, the results should be deterministic
+    // Count non-zero elements (should be distributed randomly)
```
vshampor, line 89: I suggest that you additionally check that the values are less than 1, and that they are all bitwise different, following the initial premise of being randomly generated.

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+    auto anchor_point = algo.create_anchor_point_kvcrush(num_tokens_in_evictable_blocks, indicators);
+
+    // Verify
+    EXPECT_EQ(anchor_point.size(), m_block_size);
+
+    // In ALTERNATE mode, we expect alternating blocks to be selected
+    // We should have non-zero values at regular intervals
+    bool found_pattern = false;
+    for (size_t i = 0; i < anchor_point.size(); i += m_block_size) {
+        if (i + m_block_size <= anchor_point.size()) {
```
vshampor, line 122: This condition is always true and is unnecessary

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // Execute
+    auto anchor_point = algo.create_anchor_point_kvcrush(num_tokens_in_evictable_blocks, indicators);
+
+    // Verify
+    EXPECT_EQ(anchor_point.size(), m_block_size);
+
+    // In ALTERNATE mode, we expect alternating blocks to be selected
+    // We should have non-zero values at regular intervals
+    bool found_pattern = false;
+    for (size_t i = 0; i < anchor_point.size(); i += m_block_size) {
```
vshampor, line 121: What's the point iterating here if you require `anchor_point` to have a size of exactly 1 block?

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+    auto block_distances =
+        m_kvcrush_algo.calculate_hamming_distance_kvcrush(num_tokens_in_evictable_blocks, indicators, anchor_point);
+
+    // Verify
+    size_t num_blocks = num_tokens_in_evictable_blocks / m_block_size;
+    EXPECT_EQ(block_distances.size(), num_blocks);
+
+    // For this test case with alternating patterns, each block should have maximum Hamming distance
+    for (const auto& block_dist : block_distances) {
+        EXPECT_EQ(block_dist.first, m_block_size);  // Distance should be m_block_size
```
vshampor, line 227: Please parameterize this test with multiple input pairs of patterns and reference hamming distances instead of just 1 case.

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // Execute
+    auto representative_blocks =
+        algo.get_representative_blocks_kvcrush(num_tokens_in_evictable_blocks, block_distances, keep_clus_eligible);
+
+    // Verify
+    EXPECT_EQ(representative_blocks.size(), 2);  // Should select 2 blocks (budget=2)
+
+    // With sorted distances, it should follow uniform spacing
+    EXPECT_EQ(representative_blocks[0], 0);
+    EXPECT_EQ(representative_blocks[1], 2);
```
vshampor, line 258: Same for this test, parameterize with multiple representative cases.

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+    // The retained indices should be a subset of the original evicted_block_indices
+    for (auto retained_idx : retained_indices) {
+        bool found = false;
+        for (auto evicted_idx : evicted_block_indices) {
+            if (retained_idx == evicted_idx) {
+                found = true;
+                break;
+            }
+        }
+        EXPECT_TRUE(found);
```
vshampor, line 290: Same, plus I think I asked you to use explicit index comparison instead of checking for subset-ness property.

tests/cpp/kvcrush.cpp
```diff
@@ -0,0 +1,321 @@
...
+    }
+}
+
+// Test edge case: budget larger than available blocks
```
vshampor, line 294: Could be included in the parameterization of the test case above.

src/cpp/src/continuous_batching/kvcrush.cpp
```diff
@@ -0,0 +1,178 @@
...
+#include "continuous_batching/kvcrush.hpp"
+
+#include <random>
+namespace ov::genai {
+std::mt19937 KVCrushAlgorithm::rng;
+
+KVCrushAlgorithm::KVCrushAlgorithm(const KVCrushConfig& kvcrush_config, size_t block_size)
+    : m_kvcrush_config(kvcrush_config),
+      m_block_size(block_size) {
+    rng = std::mt19937(m_kvcrush_config.rng_seed);
```
vshampor, line 13: This can be part of the constructor initialization list along with `m_kvcrush_config` and `m_block_size`

src/cpp/src/continuous_batching/kvcrush.hpp
```diff
@@ -0,0 +1,62 @@
...
+class KVCrushAlgorithm {
+private:
+    /**
+     * @brief Random number generator used for stochastic operations within the class.
+     *
+     * This static instance of std::mt19937 is initialized with the seed specified in
+     * m_kvcrush_config.rng_seed, ensuring reproducible random sequences for operations
+     * that require randomness, such as shuffling or sampling.
+     */
+    static std::mt19937 rng;
```
vshampor, line 26: Why do you insist on this being `static`?

src/cpp/include/openvino/genai/kvcrush.hpp
```diff
@@ -0,0 +1,60 @@
...
+class KVCrushConfig {
+public:
+    KVCrushConfig() = default;
+
+    KVCrushConfig(size_t budget_, KVCrushAnchorPointMode anchor_point_mode_, size_t rng_seed_ = 0)
+        : budget(budget_),
+          anchor_point_mode(anchor_point_mode_),
+          rng_seed(rng_seed_) {}
+
+    /*KVCrush Cache budget - number of tokens*/
```
vshampor, line 50: So is it tokens or blocks now? By the way, we have different block sizes on different devices (CPU vs GPU), so the user may end up with different KVCrush behaviour if they specify same block-wise budget on CPU vs GPU. Is this the behaviour you expect?

gopikrishnajha: Yes, It is blocks now. Behavior is expected.

src/cpp/src/continuous_batching/cache_eviction.cpp
gopikrishnajha: Noted. clang formatting changed the entire line where change was made. Now I have avoided changing the lines which are not freshly introduced by me.

src/cpp/src/continuous_batching/cache_eviction.cpp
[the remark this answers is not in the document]
gopikrishnajha: These are vectors and not sets, I have checked the return types.

tests/cpp/cache_eviction.cpp
gopikrishnajha: Writing test cases for exactly which blocks will be evicted is non trivial from end to end KVCrush pipeline perspective. Reason for that is evict_logical_blocks() calls kvcrush technique end to end, not partially, so exactly knowing how the scores should be set, such that indicators will be composed of certain bits, and their distance from anchor, and uniform sampling etc. is non trivial.

We have however written test cases for each individual function of kvcrush in tests/cpp/kvcrush.cpp. So at a function level this is validated already.

tests/python_tests/test_kvcrush.py
gopikrishnajha: The way testing is being done is different for KVCrush. For KVCrush the tests check if the KVCrush outperforms snapkv accuracy using the same budget.

But going by your suggestions, I have removed this file and added this test case in the ```test_kv_cache_eviction.py```

src/cpp/src/continuous_batching/kvcrush.hpp
gopikrishnajha: Have removed the comment. However, CacheEvictionAlgorithm has a default constructor, so it's member KVCrushAlgorithm must have a default constructor to compile.

Wovchena:
Moved to https://github.com/openvinotoolkit/openvino.genai/pull/2523

PR #2212, opened 2025-05-14
Added cache_dir argument

cc-crisan:
Added cache_dir argument for re-using local cache when possible.

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/model_loaders.py
+++ b/tools/who_what_benchmark/whowhatbench/model_loaders.py
@@ -85,11 +85,11 @@ def load_text_llamacpp_pipeline(model_dir):
     return model
 
 
-def load_text_hf_pipeline(model_id, device):
+def load_text_hf_pipeline(model_id, device, cache_dir=None):
     model_kwargs = {}
 
     if not torch.cuda.is_available or device.lower() == "cpu":
-        config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
+        config = AutoConfig.from_pretrained(model_id, trust_remote_code=True, cache_dir=cache_dir)
         is_gptq = False
         is_awq = False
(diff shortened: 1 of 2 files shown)


nikita-savelyevv (changes requested):
As far as I know caching of HF models is enabled by default at `/home/<username>/.cache/huggingface/hub`. The default model cache location can be changed with a `HF_HUB_CACHE` env. variable. More info here: https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache .

So if the default cache location doesn't suit you for some reason, I believe you can run wwb like this:
```
$ HF_HUB_CACHE=<some_path> wwb ...
```

@cc-crisan Have you tried the approach above? If so, what are the reasons the additional argument is still required?

cc-crisan:
> As far as I know caching of HF models is enabled by default at `/home/<username>/.cache/huggingface/hub`. The default model cache location can be changed with a `HF_HUB_CACHE` env. variable. More info here: https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache .
> 
> So if the default cache location doesn't suit you for some reason, I believe you can run wwb like this:
> 
> ```
> $ HF_HUB_CACHE=<some_path> wwb ...
> ```
> 
> @cc-crisan Have you tried the approach above? If so, what are the reasons the additional argument is still required?

I tried it in a different context  and it did not seem to be working. But I was expecting it to work with HF_HUB_OFFLINE=true, and in that context, it wouldn't get the revision (commit_hash) from the huggingface website (default it 'main') so it failed a regex match looking for a 40-characters string. I will re-test it and let you know.

cc-crisan:
My colleague, @rvuia, found a workaround by specifying a direct path to the cached model. That plus HF_HUB_OFFLINE set on true seem to be working. I'll close this PR as there is no need to change wwb.py.

PR #2213, opened 2025-05-14
Remove forced_decoder_ids workaround

as-suvorov:


as-suvorov:
After https://github.com/huggingface/optimum-intel/pull/1307 and https://github.com/huggingface/optimum-intel/pull/1301. 
Removal of `pipeline.model.config.forced_decoder_ids = None` on genai side fixes most of the tests. But it seems converted whisper-tiny model cannot detect language by default.
Leaving optimum version fixed before https://github.com/huggingface/optimum-intel/pull/1301.

PR #2218, opened 2025-05-15
[GGUF] Serialize Generated OV Model for Faster LLMPipeline Init

sammysun0711:
**Details:**
This PR aim to cache generated OV model from GGUF model in disk for faster subsequent LLMPipeline initialization w/ OpenVINO model cache.
- Serialize generated OV model from GGUF model w/ GGUF Reader with properties `ov::genai::enable_save_ov_model` (default value is `false`)
- User can check if OV model exists in same folder of GGUF model, load OV model directly instead of creating GGUF model w/ GGUF Reader.
- If GGUF model updated, user need to take the responsibility for cache invalidation and re-generate OV model with GGUF Reader.
- Use `OPENVINO_LOG_LEVEL` environment variable to control the verbose of GGUF related debug information, details please refer to [DEBUG_LOG.md](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/docs/DEBUG_LOG.md)

**Expected behavior:** 
- Set environment variable: `export OPENVINO_LOG_LEVEL=3`
-  First run w/ GGUF model:  
    - `build/samples/cpp/text_generation/greedy_causal_lm  gguf_models/qwen2.5-0.5b-instruct-q4_0.gguf "Who are you?" `

   > [GGUF Reader]: Loading and unpacking model from: gguf_models/qwen2.5-0.5b-instruct-q4_0.gguf
[GGUF Reader]: Loading and unpacking model done. Time: 196ms
[GGUF Reader]: Start generating OpenVINO model...
[GGUF Reader]: Save generated OpenVINO model to: gguf_models/openvino_model.xml done. Time: 466 ms
[GGUF Reader]: Model generation done. Time: 757ms
I am Qwen, a large language model created by Alibaba Cloud. I am a language model designed to assist users in generating human-like text, such as writing articles, stories, and even writing books. I am trained on a vast corpus of text data, including books, articles, and other written works. I am also trained on a large corpus of human language data, including written and spoken language. I am designed to provide information and insights to users, and to assist them in their tasks and

- 2nd run w/ OV model: 
  - `build/samples/cpp/text_generation/greedy_causal_lm  gguf_models "Who are you?"`
  > I am Qwen, a large language model created by Alibaba Cloud. I am a language model designed to assist users in generating human-like text, such as writing articles, stories, and even writing books. I am trained on a vast corpus of text data, including books, articles, and other written works. I am also trained on a large corpus of human language data, including written and spoken language. I am designed to provide information and insights to users, and to assist them in their tasks and


as-suvorov:
@sammysun0711 
Proposal looks like implicit model caching. OpenVINO model is being implicitly serialized on disk. And if gguf model has changed it seems outdated serialized OpenVINO model will be loaded.
Can we reuse OpenVINO `cache_dir` property for such scenario?

Form the logs I see gguf model load time is 245ms. Could you please also provide loading time for OpenVINO serialized model?

sammysun0711:
Hi @as-suvorov, thanks for your suggestion. 
> Can we reuse OpenVINO cache_dir property for such scenario?

Do you means if `cache_dir` property set in property, then we can enable save generated OV model in `cache_dir` for re-use? I think it is a good proposal that user can control from application whether to serialize OV model in disk explicitly.

> Could you please also provide loading time for OpenVINO serialized model?

Sure, I will add loading time for OpenVINO serialized model

sammysun0711:
- Add support to save OV model based on ov::cache_dir properties explicit
- Add time measurement for loading OV model.

**1st Run w/o model cache:**
> Loading and unpacking model from: gguf_models/qwen2.5-0.5b-instruct-q4_0.gguf
Loading and unpacking model done. Time: 202ms
Start generating OpenVINO model...
Model generation done. Time: 292ms

**2nd Run w/ model cache + serialize OV model:**
> Loading and unpacking model from: gguf_models/qwen2.5-0.5b-instruct-q4_0.gguf
Loading and unpacking model done. Time: 189ms
Start generating OpenVINO model...
Save generated OpenVINO model to: model_cache/qwen2.5-0.5b-instruct-q4_0_openvino_model.xml done. Time: 379 ms
Model generation done. Time: 647ms

**3rd Run w/ model cache using generated OV model**
> Found generated OpenVINO model: model_cache/qwen2.5-0.5b-instruct-q4_0_openvino_model.xml, skip creating from GGUF model.
Loading OpenVINO model done. Time: 71ms

> check if and how OpenVINO invalidates cached model and implement same approach for gguf format

Current naive method use GGUF file name to check if new OV model need to be regenerated, OV invalidate outdated model cache via hash calculate by [compute_hash.hpp](https://github.com/openvinotoolkit/openvino/blob/master/src/core/dev_api/openvino/runtime/compute_hash.hpp), but it seems compute_hash.hpp belong to dev_api, which is not accessible unless openvino.genai static link to openvino.  @as-suvorov, may I know if you have an suggestions?

sammysun0711:
I added initial hash function in, while found that hash function overhead is higher than model construction method. 
> Loading and unpacking model from: gguf_models/qwen2.5-0.5b-instruct-q4_0.gguf
Loading and unpacking model done. Time: 185ms
Compute GGUF hash with SHA1 done. Time: **856ms**
Start generating OpenVINO model...
Save generated OpenVINO model to: model_cache/qwen2.5-0.5b-instruct-q4_0_openvino_model.xml done. Time: 567 ms
Model generation done. Time: 1702ms

Although can try to optimize hash method with parallelism, the hash function will be called for gguf load, which is not idea if we would like to skip gguf model load & unpack process if OV model existed in model cache. 

I would like to suggest to serialize OV model in cache_dir when user set cache_dir property explicitly, and add verbose that existed cache OV model will skip GGUF->OV conversion w/o cache invalidate. User may need to clean up model cache and re-generate OV model if gguf model updated. 

We can add hash function check after [compute_hash.hpp](https://github.com/openvinotoolkit/openvino/blob/master/src/core/dev_api/openvino/runtime/compute_hash.hpp) expose as public API for re-use.

as-suvorov:
> I added initial hash function in [test branch](https://github.com/sammysun0711/openvino.genai/tree/gguf_model_cache_hash), while found that hash function overhead is higher than model construction method.
> 
> > Loading and unpacking model from: gguf_models/qwen2.5-0.5b-instruct-q4_0.gguf
> > Loading and unpacking model done. Time: 185ms
> > Compute GGUF hash with SHA1 done. Time: **856ms**
> > Start generating OpenVINO model...
> > Save generated OpenVINO model to: model_cache/qwen2.5-0.5b-instruct-q4_0_openvino_model.xml done. Time: 567 ms
> > Model generation done. Time: 1702ms
> 
> Although can try to optimize hash method with parallelism, the hash function will be called for gguf load, which is not idea if we would like to skip gguf model load & unpack process if OV model existed in model cache.
> 
> I would like to suggest to serialize OV model in cache_dir when user set cache_dir property explicitly, and add verbose that existed cache OV model will skip GGUF->OV conversion w/o cache invalidate. User may need to clean up model cache and re-generate OV model if gguf model updated.
> 
> We can add hash function check after [compute_hash.hpp](https://github.com/openvinotoolkit/openvino/blob/master/src/core/dev_api/openvino/runtime/compute_hash.hpp) expose as public API for re-use.

@Wovchena What do you think?

src/cpp/src/utils.cpp
```diff
@@ -292,10 +292,27 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
     if (is_gguf_model(model_dir)) {
 #ifdef ENABLE_GGUF
-        return create_from_gguf(model_dir.string());
+        if (properties.find(ov::cache_dir.name()) != properties.end()) {
+            std::string cache_dir = properties.at(ov::cache_dir.name()).as<std::string>();
+            if (!cache_dir.empty()) {
+                std::filesystem::path model_cache_dir(cache_dir);
+                std::filesystem::path model_path = model_cache_dir / (model_dir.stem().string() + "_openvino_model.xml");
+                if (std::filesystem::exists(model_path)) {
+                    std::cout << "Found generated OpenVINO model: " << model_path.string() << ", skip creating from GGUF model.\n";
```
Wovchena, line 304: After the model is saved, the next `LLMPipeline()` constructor can read the folder as usual ignoring GGUF files. That's not easy to implement in a sample to make everything short, readable and explicit, but in general case users can trigger GGUF branch for the fist constructor and OV branch for the next constructors. That would shift cache invalidation responsibility to the user. Would that work for you?

sammysun0711: Hi @Wovchena, thanks for your feedback. May I double check that your suggest not to use `cache_dir` property to control whether save OV model explictly, instead we implicit save OV model for first run (GGUF branch), and re-use OV model for 2nd run (OV branch). Right?

Wovchena: Correct. This is also possible flip the implicitness: GGUF doesn't save the model by default, but if the a special property is given, the model is saved. Implicitness or explicitness doesn't affect the general idea.

Wovchena: There also should be a way to disable model saving. For example if someone doesn't have enough space or GGUF model is on a share mounted as read only

sammysun0711:
Hi @Wovchena, I update my implementation as follow: 
1. Separate OV and GGUF branches to move the cache invalidate responsibility to user instead of GenAI
2. Add special property `ENABLE_SAVE_OV_MODEL` to control whether save OV model by user
3. Verified with attached sample code for following case [greedy_causal_lm.cpp.txt](https://github.com/user-attachments/files/20344504/greedy_causal_lm.cpp.txt)
   - ENABLE_SAVE_OV_MODEL=true  + ATTENTION_BACKEND='PA'
   - ENABLE_SAVE_OV_MODEL=false + ATTENTION_BACKEND='SDPA'
   - ENABLE_SAVE_OV_MODEL=true  + ATTENTION_BACKEND='PA'
   - ENABLE_SAVE_OV_MODEL=false + ATTENTION_BACKEND='SDPA'

src/cpp/src/utils.cpp
sammysun0711: Thanks, I've update my implementation to address your comments.

src/cpp/src/gguf_utils/gguf_modeling.cpp
```diff
@@ -152,24 +152,43 @@ std::shared_ptr<ov::Model> create_language_model(
...
+void save_openvino_model(const std::shared_ptr<ov::Model>& model, const std::string& save_path, bool compress_to_fp16) {
+    try {
+        auto serialize_start_time = std::chrono::high_resolution_clock::now();
+        ov::save_model(model, save_path, compress_to_fp16);
+        auto serialize_finish_time = std::chrono::high_resolution_clock::now();
+        auto serialize_duration = std::chrono::duration_cast<std::chrono::milliseconds>(serialize_finish_time - serialize_start_time).count();
+        std::cout << "Save generated OpenVINO model to: " << save_path << " done. Time: " << serialize_duration << " ms\n";
+    }
+    catch (const ov::Exception& e) {
+        std::cerr << "[Warning] Exception during model serialization: " << e.what() << std::endl;
```
Wovchena, line 164: The warning should become an error now. If it doesn't work for someone, they can disable it with `ENABLE_SAVE_OV_MODEL`.

src/cpp/src/utils.cpp
```diff
@@ -301,17 +302,30 @@ ov::Core singleton_core() {
...
 }
 
 } // namespace
 
-std::shared_ptr<ov::Model> read_model(const std::filesystem::path& model_dir,  const ov::AnyMap& config) {
+std::pair<ov::AnyMap, bool> extract_gguf_properties(const ov::AnyMap& external_properties) {
+    bool enable_save_ov_model = ENABLE_SAVE_OV_MODEL;
+    ov::AnyMap properties = external_properties;
+
+    auto it = properties.find("ENABLE_SAVE_OV_MODEL");
```
Wovchena, line 315: Can you document is somewhere?

src/cpp/src/gguf_utils/gguf_modeling.cpp
Wovchena, line 164: Please add a test to `tests/python_tests/test_llm_pipeline.py` to ensure we don't break the feature accidentally later

Wovchena (changes requested):

src/cpp/src/gguf_utils/gguf_modeling.cpp
sammysun0711: Added test case

src/cpp/src/gguf_utils/gguf_modeling.cpp
```diff
@@ -152,24 +152,43 @@ std::shared_ptr<ov::Model> create_language_model(
...
 } // namespace
 
-std::shared_ptr<ov::Model> create_from_gguf(const std::string& model_path) {
+void save_openvino_model(const std::shared_ptr<ov::Model>& model, const std::string& save_path, bool compress_to_fp16) {
+    try {
+        auto serialize_start_time = std::chrono::high_resolution_clock::now();
+        ov::save_model(model, save_path, compress_to_fp16);
+        auto serialize_finish_time = std::chrono::high_resolution_clock::now();
+        auto serialize_duration = std::chrono::duration_cast<std::chrono::milliseconds>(serialize_finish_time - serialize_start_time).count();
+        std::cout << "Save generated OpenVINO model to: " << save_path << " done. Time: " << serialize_duration << " ms\n";
```
Wovchena, line 161: I agree that `save_path` print is helpful, but it still would be strange to see this log on every GGU pipeline construction. That's the documentation is for. Add an entry to https://github.com/openvinotoolkit/openvino.genai/blob/dad32c5adb8ac396d12a05831c34f2f5e77d8694/src/cpp/include/openvino/genai/llm_pipeline.hpp#L329
and remove this print. I think `extern const bool ENABLE_SAVE_OV_MODEL;` from utils.hpp can be removed after that.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -842,3 +842,50 @@ def test_pipelines_with_gguf_generate(pipeline_type, model_ids):
...
+    ov.save_model(ov_tokenizer, ov_tokenizer_path)
+    ov.save_model(ov_detokenizer, ov_detokenizer_path)
+    ov_pipe_gguf = create_ov_gguf_pipeline(gguf_full_path, pipeline_type=pipeline_type, enable_save_ov_model=enable_save_ov_model)
+    encoded_result  = ov_pipe_gguf.generate(ov.Tensor(input_ids.numpy()), generation_config=ov_generation_config)
+    res_string_input_2 = hf_tokenizer.batch_decode([encoded_result.tokens[0]], skip_special_tokens=True)[0]
+
+    ov_model_saved_on_disk = False
+    if Path(gguf_full_path / "openvino_model.xml").exists():
+        ov_model_saved_on_disk = True
+    assert enable_save_ov_model == ov_model_saved_on_disk
```
Wovchena, line 891: I would propose to test the following scenario:
Construct GGUF pipeline. Construct another pipeline after that from `.xml`

That's the user scenario you are trying to cover with this PR, right?

Wovchena (changes requested):

src/cpp/src/gguf_utils/gguf_modeling.cpp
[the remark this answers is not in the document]
sammysun0711: Move all GGUF Reader related message under `OPENVINO_LOG_LEVEL ` and add documents in `DEBUG_LOG.md`

tests/python_tests/test_llm_pipeline.py
sammysun0711: Update test case to construct two pipeline 
1.  Construct GGUF pipeline and save generated ov model
2.  Construct OV pipeline from generated ov model
3.  Compare pipeline 1 & pipeline 2 generated results.

Wovchena (approved):
FYI, our CI is broken, so some failures are OK

tests/python_tests/test_llm_pipeline.py
```diff
@@ -879,3 +879,36 @@ def test_full_gguf_pipeline(pipeline_type, model_ids):
...
+    gguf_full_path = download_gguf_model(gguf_model_id, gguf_filename)
+    gguf_full_path = Path(gguf_full_path)
+
+    ov_pipe_gguf = create_ov_pipeline(gguf_full_path, pipeline_type=pipeline_type, enable_save_ov_model=enable_save_ov_model)
+    res_string_input_1  = ov_pipe_gguf.generate(prompt, generation_config=ov_generation_config)
+
+    ov_pipe_native = create_ov_pipeline(gguf_full_path.parent, pipeline_type=pipeline_type)
+    res_string_input_2  = ov_pipe_native.generate(prompt, generation_config=ov_generation_config)
+
+    assert res_string_input_1 == res_string_input_2
```
rkazants, line 914: probably you can extend existing `test_full_gguf_pipeline` test by adding new parameter to `@pytest.mark.parametrize` enable_save_ov with two values [True, False]

tests/python_tests/test_llm_pipeline.py
```diff
@@ -846,8 +846,9 @@ def test_pipelines_with_gguf_generate(pipeline_type, model_ids):
 
 @pytest.mark.parametrize("pipeline_type", get_gguf_pipeline_types())
 @pytest.mark.parametrize("model_ids", get_gguf_model_list())
+@pytest.mark.parametrize("enable_save_ov_model", get_gguf_enable_save_ov_model_property_list())
```
rkazants, line 849:
```suggestion
@pytest.mark.parametrize("enable_save_ov_model", [True, False])
```
and that function can be just removed:)

sammysun0711: Done

tests/python_tests/utils/ov_genai_pipelines.py
```diff
@@ -96,6 +102,24 @@ def create_ov_pipeline(models_path: Path,
...
+        return LLMPipeline(models_path, Tokenizer(models_path.parent), device, local_ov_config)
+    elif pipeline_type == PipelineType.STATEFUL:
+        local_ov_config["enable_save_ov_model"] = enable_save_ov_model
+        return LLMPipeline(models_path, Tokenizer(models_path.parent), device, local_ov_config, ATTENTION_BACKEND="SDPA")
+    elif pipeline_type == PipelineType.PAGED_ATTENTION:
+        local_ov_config["enable_save_ov_model"] = enable_save_ov_model
+        return LLMPipeline(models_path, Tokenizer(models_path.parent), device, local_ov_config, scheduler_config=scheduler_config, ATTENTION_BACKEND="PA")
+    else:
+        raise Exception(f"Unsupported pipeline type: {pipeline_type}")
+
```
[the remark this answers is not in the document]
sammysun0711: Removed

tests/python_tests/test_llm_pipeline.py
```diff
@@ -16,7 +16,7 @@
 from utils.hugging_face import generation_config_to_hf, download_and_convert_model, download_gguf_model
 # model_tmp_path fixture import required
 from utils.tokenizers import delete_rt_info, model_tmp_path
-from utils.ov_genai_pipelines import create_ov_pipeline, generate_and_compare, get_main_pipeline_types, PipelineType, get_gguf_pipeline_types
+from utils.ov_genai_pipelines import create_ov_pipeline, generate_and_compare, get_main_pipeline_types, PipelineType, get_gguf_pipeline_types, get_gguf_enable_save_ov_model_property_list
```
rkazants, line 19:
```suggestion
from utils.ov_genai_pipelines import create_ov_pipeline, generate_and_compare, get_main_pipeline_types, PipelineType, get_gguf_pipeline_types
```

rkazants (approved):

tests/python_tests/test_llm_pipeline.py
```diff
@@ -846,8 +846,9 @@ def test_pipelines_with_gguf_generate(pipeline_type, model_ids):
 
 @pytest.mark.parametrize("pipeline_type", get_gguf_pipeline_types())
 @pytest.mark.parametrize("model_ids", get_gguf_model_list())
+@pytest.mark.parametrize("enable_save_ov_model", [False])
```
rkazants: why?

sammysun0711: Test purpose only, try to figure out why MacOS test failed (segment fault): **https://github.com/openvinotoolkit/openvino.genai/actions/runs/15420607456/job/43434100193**

It seems [macos-13](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md) has smaller memory size: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Does it make sense to create a separate test case, instead of check 3 pipeline test (HF/GGUF/OV native) to only 2 pipeline test (GGUF/OV native) to save memory usage?

rkazants: @mryzhov, please advice other MacOS runner with more memory

sammysun0711: @mryzhov, can we try `macos-13-large`  (30GB memory) for debug purpose first? : https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/running-jobs-on-larger-runners?platform=mac#available-macos-larger-runners

sammysun0711: test results shows that not related to insufficient memory, https://github.com/openvinotoolkit/openvino.genai/actions/runs/15436902324/job/43447403549?pr=2218, will revert back to `macos-13` and continue to investigation.

sammysun0711: @akladiev provides my a local macos machine w/ macos-12 + 64GB memory (during test the memory usage ~30GB + 15GB Swap usage), cannot reproduce the same segment fault issue w/ GGUF related test in test_llm_pipeline.py
 
I found the same issue in other PR：[Bump actions/download-artifact from 4.1.9 to 4.3.0 · openvinotoolkit/openvino.genai@c3fe438](https://github.com/openvinotoolkit/openvino.genai/actions/runs/15438377512/job/43451509230?pr=2315), so I think it not introduced by my PR, could you please help to further investigate this issue?

sammysun0711: I try to separate `enable_save_ov_model` related gguf test from original test case to reduce memory usage. 
Still the macos test failed with https://github.com/openvinotoolkit/openvino.genai/blob/3c232ac2761c8068d1a45892df3db8b69802fa47/tests/python_tests/test_llm_pipeline.py#L909
GHA action: https://github.com/openvinotoolkit/openvino.genai/actions/runs/15485048274/job/43598941161

sammysun0711: Temporal disable gguf related test on MacOS due to sporadic failure: https://github.com/openvinotoolkit/openvino.genai/pull/2328

sammysun0711:
> need review again

Refactor test and all tests passed, please kindly review again, thanks!

PR #2232, opened 2025-05-19
Fix Coverity issues

olpipi:
Fix Coverity issues
CVS-165984


src/cpp/src/generation_config.cpp
```diff
@@ -222,7 +222,6 @@ void GenerationConfig::validate() const {
     }
 
     if (is_multinomial()) {
-        OPENVINO_ASSERT(top_k >= 0, "When 'do_sample' is true, top_k must be a non-negative, but got ", top_k);
```
Wovchena, line 225: Please, change `static constexpr ov::Property<int> top_k{"top_k"};` to `size_t` to align them

samples/cpp/whisper_speech_recognition/audio_utils.cpp
```diff
@@ -87,7 +87,7 @@ ov::genai::RawSpeechInput read_wav(const std::string& filename) {
     }
 
     const uint64_t n =
-        wav_data.empty() ? wav.totalPCMFrameCount : wav_data.size() / (wav.channels * wav.bitsPerSample / 8);
+        wav_data.empty() ? wav.totalPCMFrameCount : wav_data.size() / (static_cast<uint64_t>(wav.channels) * wav.bitsPerSample / 8);
```
popovaan, line 90: Do we need cast to uint64_t for `wav.bitsPerSample`? 
It has `drwav_uint16` type as well.

src/cpp/src/generation_config.cpp
olpipi: done

samples/cpp/whisper_speech_recognition/audio_utils.cpp
olpipi: the point is `/ 8`. 8 is a int literal. So the whole expression is converted to int. 
Changed 8 literal type to unsigned long.

Wovchena (approved):

PR #2235, opened 2025-05-19
[JS] Add an interrupt option for LLMPipeline

Retribution98:
## Details
- Add support for StreamingStatus
- Fix converting for GenerationConfig
- Add some types

## Tickets
- CVS-167755


src/js/include/llm_pipeline/llm_pipeline_wrapper.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <future>
```
almilosz, line 3: I think it's better to keep in header only necessary files and move it down into the source file

src/js/src/helper.cpp
```diff
@@ -43,27 +43,49 @@ ov::Any js_to_cpp<ov::Any>(const Napi::Env& env, const Napi::Value& value) {
         }
     } else if (value.IsBoolean()) {
         return ov::Any(static_cast<bool>(value.ToBoolean()));
-    } else {
-        OPENVINO_THROW("Cannot convert to ov::Any");
+    } else if (value.IsArray()) {
+        return ov::Any(js_to_cpp<std::vector<std::string>>(env, value));
+    } else if (value.IsObject()) {
+        if (value.ToString().Utf8Value() == "[object Set]") {
```
almilosz, line 49: What js type is needed here? Maybe [this](https://github.com/openvinotoolkit/openvino/blob/5561c207aba860abeca4b383a7976f61433d8e1f/src/bindings/js/node/src/helper.cpp#L539) approach will be better?

src/js/include/llm_pipeline/llm_pipeline_wrapper.hpp
Wovchena, line 3: Adding a header without changing any code in the file is strange. Remove it

src/js/src/helper.cpp
Retribution98: We check the type of Set here.
I tried to call `inferenceOf` for Set from global environment, but unfortunately that didn't work. The only reason I can see for the failure to check is that the object was created in a different global environment.
I don't really like the current approach, but for now  it can be used as a workaround.

src/js/include/llm_pipeline/llm_pipeline_wrapper.hpp
Retribution98: Sorry, I thought this include should be set in the header file, not in the source file. 
I have fixed that.

Retribution98: Thank you, It has fixed

Wovchena (approved):

src/js/lib/utils.ts
```diff
@@ -0,0 +1,110 @@
+/* eslint-disable max-len, no-unused-vars */
+export enum StreamingStatus {
+    RUNNING,
+    STOP,
+    CANCEL,
+}
+
+/** controls the stopping condition for grouped beam search. The following values are possible?:
```
almilosz, line 8:
```suggestion
/** controls the stopping condition for grouped beam search. The following values are possible:
```

PR #2236, opened 2025-05-20
Switching to coverity entrerprise (prod3)

jszczepa:


.github/workflows/coverity.yml
Wovchena, line 1: You can temporary change the trigger to pull_request on top of the file to test the workflow

PR #2244, opened 2025-05-21
[NPU] Apply default STATIC_PIPELINE and memory-safe config for NPU backend

ohjaesik:


ohjaesik:
@openvinotoolkit/maintainers Could someone please review and approve this PR? It contains NPU optimization fixes.

src/cpp/src/llm/pipeline.cpp
```diff
@@ -15,6 +15,30 @@
...
+    if (!properties.count("MAX_PROMPT_LEN"))
+        properties["MAX_PROMPT_LEN"] = 64;
+
+    // Set minimum number of tokens to generate in the response
+    if (!properties.count("MIN_RESPONSE_LEN"))
+        properties["MIN_RESPONSE_LEN"] = 64;
+
+    // Set cache directory to store compiled pipeline
+    if (!properties.count("CACHE_DIR"))
+        properties["CACHE_DIR"] = "./ov_cache";
```
as-suvorov, line 36: `CACHE_DIR` is optional and disabled by default in OpenVINO. I think GenAI default should be aligned.

src/cpp/src/llm/pipeline.cpp
```diff
@@ -15,6 +15,30 @@
...
+
+    // Limit to single stream to avoid parallel execution issues on NPU
+    if (!properties.count("NUM_STREAMS"))
+        properties["NUM_STREAMS"] = 1;
```
as-suvorov, line 40: Same for `NUM_STREAMS`.

src/cpp/src/llm/pipeline.cpp
```diff
@@ -15,6 +15,30 @@
 #include "speculative_decoding/speculative_decoding_impl.hpp"
 #include "utils.hpp"
 
+// Patch properties with safe default values for NPU inference
+// This ensures stable behavior by enabling static pipeline execution
+// and reducing prompt/response sizes to prevent memory issues.
+void patch_npu_properties(ov::AnyMap& properties) {
+    // Enable static pipeline (required for NPU)
+    if (!properties.count("STATIC_PIPELINE"))
+        properties["STATIC_PIPELINE"] = true;
```
as-suvorov, line 24: What are the instability issues with `StatefulPipeline`?

src/cpp/src/llm/pipeline.cpp
ohjaesik: remove this code?

src/cpp/src/llm/pipeline.cpp
AsyaPronina: Exactly the unified and optimized configuration is chosen by default, so `STATIC_PIPELINE` is not used on purpose.
Please attach numbers that shows the difference between `STATIC_PIPELINE` and not if you found the difference.

as-suvorov (changes requested):

ohjaesik:
nothing

PR #2254, opened 2025-05-22
Support for mmlu dataset and multiple choice evaluation

sstrehlk:


tools/who_what_benchmark/whowhatbench/wwb.py
Wovchena, line 1: Why is `mmlu` needed?

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -200,8 +200,22 @@ def load_prompts(args):
         name = None
     data = load_dataset(path=path, name=name, split=split)
 
-    res = data[args.dataset_field]
-    res = {"prompts": list(res)}
+    if "mmlu" in path:
```
sbalandi, line 203: Can only this dataset be used for this case ? 
I would prefer to manage cases via cmd keys, not dataset, maybe add a separate key for example `--metrics`, which can be the list with pre-defined values and set as `similarity` be default. If `choice_similarity` in metrics , then make `mmlu` dataset the as default.

PR #2242, opened 2025-05-20
[NPUW] Re-fixed issue with long prompt for NPU

AsyaPronina:

PR #2256, opened 2025-05-22
Added tests and refactored fix for long prompt issue

AsyaPronina:
Follow-up for https://github.com/openvinotoolkit/openvino.genai/pull/2242

### DIFF
--- a/tests/python_tests/test_llm_pipeline_npu.py
+++ b/tests/python_tests/test_llm_pipeline_npu.py
@@ -1,12 +1,14 @@
 # Copyright (C) 2024-2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-from openvino_genai import GenerationConfig, Tokenizer, LLMPipeline, StreamerBase
+from openvino_genai import GenerationConfig, Tokenizer, LLMPipeline, StreamerBase, TokenizedInputs
+from openvino import Tensor
 import os
 
 import pytest
 import platform
 import sys
+import re
 
 from utils.constants import get_default_llm_properties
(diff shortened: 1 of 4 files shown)


tests/python_tests/test_llm_pipeline_npu.py
```diff
@@ -287,3 +289,150 @@ def test_chat_generation(config, model_id):
...
+
+    pipe.start_chat()
+    pipe.generate(prompt_1, max_new_tokens=64, ignore_eos=True)
+
+    exception_pattern = r'Stateful LLM pipeline on NPU may only process prompts.*up to 64 ' +\
+                        r'tokens\..*[\n]?Set the "MAX_PROMPT_LEN" config option to increase ' +\
+                        r'the limit\.\n'
+    with pytest.raises(RuntimeError, match=exception_pattern):
+        pipe.generate(prompt_2, max_new_tokens=2, ignore_eos=True)
+    # pipe.finish_chat()
```
Wovchena, line 438: It's up to you to call `finish_chat()` because `pipe` is destroyed after. But don't introduce commented code

tests/python_tests/test_llm_pipeline_npu.py
Wovchena, line 1: Is it really worth renaming? I stateful still can be used for other devices. We just don't use it by default. If the renaming is needed, fix the labler because it affects smart CI: https://github.com/openvinotoolkit/openvino.genai/blob/b7e8993565358a9e2d8f69bc915bfbf8e5ea9ed4/.github/labeler.yml#L27

as-suvorov:
@AsyaPronina do you work on this PR?

PR #2261, opened 2025-05-23
Show how to run from PR

Wovchena:


mryzhov:
Maybe it is better to improve action documentation instead?
https://github.com/openvinotoolkit/openvino/blob/manifest_extensions/.github/actions/openvino_provider/action.yml#L12

Wovchena:
The docs seem to be clear. It's just us not reading them

as-suvorov (approved):

PR #2270, opened 2025-05-26
[wwb]: load transformers model first, then only trust_remote_code

eaidova:
General huggingface recommendation is to use `trust_remote_code` only if it is necessary, andthe  model requires it, as potentially remote code may try to run arbitrary insecure code.
Besides that some models (e.g. phi-3) may have outdated remote code version kept for backward compatibility and non-maintained by its authors as the model is already integrated into transformers.

Suggestion is try to load model without remote code and only if it is failed use remote code for resolve such issues

CVS-168307

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/wwb.py
+++ b/tools/who_what_benchmark/whowhatbench/wwb.py
@@ -258,7 +258,7 @@ def load_processor(args):
     else:
         preprocessor_id = model_id
 
-    preprocessor = AutoProcessor.from_pretrained(preprocessor_id, trust_remote_code=trust_remote_code) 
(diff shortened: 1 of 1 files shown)


tools/who_what_benchmark/examples/openvino_batched_eval.py
```diff
@@ -16,17 +16,14 @@
 
 from openvino_tokenizers import convert_tokenizer
 from openvino import serialize
-from transformers import AutoTokenizer
+from transformers import AutoTokenizer, AutoConfig
```
eaidova, line 19:
```suggestion
from transformers import AutoTokenizer
```

andreyanufr (approved):

PR #2271, opened 2025-05-26
Force the PA implementation in the llm-bench by default

sbalandi:
[CVS-168067](https://jira.devtools.intel.com/browse/CVS-168067)

### DIFF
--- a/samples/deployment-requirements.txt
+++ b/samples/deployment-requirements.txt
@@ -1,6 +1,6 @@
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/pre-release
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
-openvino_genai~=2025.2.0.0.dev
+openvino_genai~=2025.3.0.0.dev
 librosa==0.10.2.post1  # For Whisper
 pillow==11.2.1  # Image processing for VLMs
 json5==0.12.0  # For ReAct
--- a/samples/export-requirements.txt
+++ b/samples/export-requirements.txt
@@ -1,7 +1,7 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/pre-release
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
-openvino-tokenizers~=2025.2.0.0.dev
+openvino-tokenizers~=2025.3.0.0.dev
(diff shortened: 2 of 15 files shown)


tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -181,18 +181,26 @@ def analyze_args(args):
             model_args['config'] = config
     if model_framework == 'ov':
         set_default_param_for_ov_config(model_args['config'])
+        if 'ATTENTION_BACKEND' not in model_args['config'] and use_case in ['text_gen', 'vlm'] and args.device != "NPU":
+            model_args['config']['ATTENTION_BACKEND'] = PA_ATTENTION_BACKEND
+        if model_args['config'].get('ATTENTION_BACKEND', '') == PA_ATTENTION_BACKEND and args.device == "NPU":
```
eaidova, line 186: I think, this is unnecessary check,  general rule is that the user is always right about what he wants to use. Just let pipeline decide how to handle this request - fail with mistake or switch back to sdpa(in case if npu surprisingly  starts supporting it, you will not need to update llm_bench)

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -181,18 +181,26 @@ def analyze_args(args):
...
+            model_args['config']['ATTENTION_BACKEND'] = SDPA_ATTENTION_BACKEND
+            log.warning("Continuous Batching, Speculative decoding and Prompt Lookup decoding is not supported for NPU device")
         log.info(f"OV Config={model_args['config']}")
     elif model_framework == 'pt':
         log.info(f"PT Config={model_args['config']}")
     model_args['model_type'] = get_model_type(model_name, use_case, model_framework)
     model_args['model_name'] = model_name
 
-    if use_cb and optimum:
+    if model_args['config'].get('ATTENTION_BACKEND', '') == PA_ATTENTION_BACKEND and optimum:
```
eaidova, line 195: probably now you need an additional check that if SDPA in attention backed and speculative decoding selected

tools/llm_bench/llm_bench_utils/model_utils.py
sbalandi: removed

tools/llm_bench/llm_bench_utils/model_utils.py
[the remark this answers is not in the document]
sbalandi: thank you ! removed this condition

PR #2283, opened 2025-05-28
[GHA]  Coverity pipeline fixes

mryzhov:
- fixed build command line (-j causes memory consumption)
- Improved coverity pipeline to test it without results analyzing

### DIFF
--- a/tools/who_what_benchmark/examples/openvino_batched_eval.py
+++ b/tools/who_what_benchmark/examples/openvino_batched_eval.py
@@ -23,10 +23,7 @@ MAX_NEW_TOKENS = 128
 SEQS_PER_REQUEST = 5
 MAX_SEQUENCES = 100
 
-
-model = OVModelForCausalLM.from_pretrained(
-    model_id, export=True, trust_remote_code=True
-)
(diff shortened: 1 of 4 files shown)


.github/workflows/coverity.yml
```diff
@@ -92,8 +92,17 @@ jobs:
...
           path: ${{ env.OV_INSTALL_DIR }}
           merge-multiple: true
+          
+      - name: Cache Coverity Tool
+        id: cache-coverity
+        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
+        with:
+          path: coverity_tool
+          key: coverity-${{ runner.os }}-${{ github.sha }}
+          restore-keys: coverity-${{ runner.os }}
```
akashchi, line 102: I think it should be both before and after the `Download coverity tool` step.
```suggestion
      - name: Cache Coverity Tool
        id: cache-coverity
        uses: actions/cache@5a3ec84eff668545956fd18022155c47e93e2684 # v4.2.3
        with:
          path: coverity_tool.tgz
          key: coverity-${{ runner.os }}-${{ github.sha }}
          restore-keys: coverity-${{ runner.os }}
```

PR #2295, opened 2025-05-30
Structured Output generation with `XGrammar`

pavel-esir:
### This PR introduces support for guided text generation using the `XGrammar` backend and allows extending it to other guided/structured output generation backends

- It utilizes a logit transformer to nullify logits for tokens that are not allowed by a JSON schema grammar or any grammar expressed in EBNF format.
- We introduce the `StructuredOutputController`, which orchestrates structured output generation.
- `IStructuredOutputImpl` is a helper interface that creates instances of `XGrammarLogitsTransformer`. Previously, to create a `XGrammarLogitsTransformer` (or any other structured output logit transformer), the `StructuredOutputController` would have needed to depend directly on the specific implementation. By introducing this interface, this dependency is avoided, keeping the logic separate and encapsulated within each `IStructuredOutputImpl` implementation. This design also simplifies the process of extending support for new backends.
- `XGrammarStructuredOutput` inherits from `IStructuredOutputImpl` and acts as the logit transformer builder for the XGrammar backend.
- `XGrammarLogitsTransformer` is the final component produced by this pipeline. It encapsulates the XGrammar backend implementation and exposes a public `apply` method, which applies grammar constraints to the logits each time a new token is generated.
- We have also introduced `IStatefulLogitTransformer`, which inherits from the base logit transformer class and provides the `accept_tokens` method to update its state based on newly generated tokens.

- Class dependencies are shown below:
image

This PR introduces support for guided text generation using XGrammar.

Example of usage 
```
import openvino_genai
from pydantic import BaseModel

class Person(BaseModel):
    given_name: str
    surname: str
    age: int
    city: Literal["Dublin", "Dubai", "Munich"]

pipe = openvino_genai.LLMPipeline(model_dir, device)

config = openvino_genai.GenerationConfig()
config.max_new_tokens = 100
config.do_sample = True

structured_output_config = openvino_genai.StructuredOutputConfig()
structured_output_config.json_schema = json.dumps(Person.model_json_schema())
config.structured_output_config = structured_output_config

print(pipe.generate("Generate a json about a person.", config))
```

CVS-166690 CVS-166891


samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,40 @@
...
+    args = parser.parse_args()
+
+    device = 'CPU'  # GPU can be used as well
+    pipe = openvino_genai.LLMPipeline(args.model_dir, device)
+
+    config = openvino_genai.GenerationConfig()
+    config.max_new_tokens = 100
+    config.repetition_penalty = 2
+    config.do_sample = True
+    config.json = json.dumps(Person.model_json_schema())
```
as-suvorov:
```suggestion
    config.json_schema = json.dumps(Person.model_json_schema())
```

src/cpp/src/sampling/sampler.cpp
```diff
@@ -920,8 +933,18 @@ SamplerOutput Sampler::sample(const std::vector<SequenceGroup::Ptr> & sequence_g
...
+        #ifdef ENABLE_XGRAMMAR
+        if (!xgrammar_registered) {
+            xgrammar_registered = registerXGrammarBackend();
+        }
+        #endif
         if (!m_logit_processors.count(request_id)) {
-            m_logit_processors.insert({request_id, LogitProcessor(sampling_params, sequence_group->get_prompt_ids())});
+            m_logit_processors.insert({request_id, LogitProcessor(sampling_params, sequence_group->get_prompt_ids()
+            #ifdef ENABLE_XGRAMMAR
+                                                                 , m_structured_output_controller
```
dtrawins, line 944:
```suggestion
```
duplicated argument

src/cpp/src/sampling/sampler.cpp
```diff
@@ -920,8 +933,18 @@ SamplerOutput Sampler::sample(const std::vector<SequenceGroup::Ptr> & sequence_g
...
+                                                                  , m_structured_output_controller, std::make_shared<StructuredOutputController>(m_tokenizer, m_tokenizer.get_vocab_vector().size())
```
dtrawins:
```suggestion
, std::make_shared<StructuredOutputController>(m_tokenizer, m_tokenizer.get_vocab_vector().size())
```

dtrawins:
getting seq fault when stop string '#' is not included in sampling params:
.build_tmpdir/_deps/xgrammar-src/cpp/grammar_matcher.cc:496: Check failed: (!IsStopTokenAccepted()) is false: GrammarMatcher has terminated after accepting the stop token, but is trying to find the next token mask
.
Segmentation fault (core dumped)

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,40 @@
...
+    pprint(config.json)
```
as-suvorov: Looks like a debug print, should be removed?

src/cpp/CMakeLists.txt
```diff
@@ -88,6 +121,14 @@ if(NOT ENABLE_GGUF)
     list(REMOVE_ITEM SOURCE_FILES ${GGUF_SOURCES})
 endif()
 
+if(NOT ENABLE_XGRAMMAR)
+    set(XGRAMMAR_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/xgrammar_backend.hpp
+                         ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/xgrammar_backend.cpp
+                         ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/structured_output_controller.hpp
+                         ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/structured_output_controller.cpp)
+    list(REMOVE_ITEM SOURCE_FILES ${XGRAMMAR_SOURCES})
```
as-suvorov, line 132: Did you consider dynamic linking option?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -78,6 +96,7 @@ enum class StopCriteria { EARLY, HEURISTIC, NEVER };
  * @param num_assistant_tokens the defined candidates number to be generated by draft model/prompt lookup in case of static strategy candidates number update.
  * @param max_ngram_size is maximum ngram to use when looking for matches in the prompt.
  *
+
```
as-suvorov:
```suggestion
    * @param guided_generation_config ...
```
Can be done later

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -128,6 +147,11 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
     size_t num_assistant_tokens = 0;
     size_t max_ngram_size = 0;
 
+    // Structured output parameters
+    std::optional<GuidedGenerationConfig> guided_generation_config;
+    // todo: add structural tag for function calling
+    // todo: add structured_output_backend
```
as-suvorov: Let's maybe create a tickets for this? Not sure if we want todos in a public header.

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -195,6 +220,8 @@ static constexpr ov::Property<float> assistant_confidence_threshold{"assistant_c
 static constexpr ov::Property<size_t> num_assistant_tokens{"num_assistant_tokens"};
 static constexpr ov::Property<size_t> max_ngram_size{"max_ngram_size"};
 
+static constexpr ov::Property<GuidedGenerationConfig> guided_generation_config{"guided_generation_config"};
```
as-suvorov: Do we need guided generation config standalone properties like `json_schema`?

src/cpp/src/sampling/logit_processor.hpp
```diff
@@ -404,3 +125,5 @@ class LogitProcessor {
     }
 
 };
+
+} // namespace ov::genai
```
as-suvorov:
```suggestion
} // namespace ov::genai

```

src/cpp/src/sampling/sampler.cpp
```diff
@@ -994,10 +1017,25 @@ LogitProcessor& Sampler::get_logit_processor(uint64_t request_id) {
 
 
 void Sampler::create_logit_processor(uint64_t request_id, const GenerationConfig& sampling_params, const TokenIds& prompt) {
-    m_logit_processors.insert({request_id, LogitProcessor(sampling_params, prompt)});
+    std::cerr << "Creating logit processor for request id: " << request_id << std::endl;
```
as-suvorov: Debug print?

src/cpp/src/sampling/sampler.cpp
```diff
@@ -994,10 +1017,25 @@ LogitProcessor& Sampler::get_logit_processor(uint64_t request_id) {
...
+    #ifdef ENABLE_XGRAMMAR
+    if (sampling_params.is_guided_generation() && !m_structured_output_controller) {
+        std::cerr << "Structured output controller is created" << std::endl;
```
as-suvorov: Here as well

tests/python_tests/test_llm_pipeline.py
```diff
@@ -878,4 +927,4 @@ def test_full_gguf_pipeline(pipeline_type, model_ids):
     ov_pipe_gguf = create_ov_pipeline(gguf_full_path, pipeline_type=pipeline_type)
     res_string_input_2 = ov_pipe_gguf.generate(prompt, generation_config=ov_generation_config)
 
-    assert res_string_input_1 == res_string_input_2
+    assert res_string_input_1 == res_string_input_2
```
as-suvorov:
```suggestion
    assert res_string_input_1 == res_string_input_2

```

tests/python_tests/data/models.py
```diff
@@ -8,7 +8,8 @@
 
 def get_models_list():
     precommit_models = [
-        "katuni4ka/tiny-random-phi3",
+        "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+        # "katuni4ka/tiny-random-phi3",
```
as-suvorov: Please revert

src/python/py_generation_config.cpp
```diff
@@ -27,6 +28,19 @@ auto stop_criteria_docstring =  R"(
         "openvino_genai.StopCriteria.NEVER" stops when there cannot be better candidates.
 )";
 
+auto guided_generation_config_docstring = R"(
+    Structure to keep generation config parameters for guided generation.
+    It is used to store the configuration for guided generation, which includes
+    the JSON schema and other related parameters.
+
+    Structured output parameters:
+    json:           if set, the output will be a JSON string constraint by the specified json-schema.
```
as-suvorov:
```suggestion
    json_schema:           if set, the output will be a JSON string constraint by the specified json-schema.
```

src/cpp/src/sampling/sampler.cpp
```diff
@@ -851,6 +852,18 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
                     }
                 }
                 register_new_token(sampled_token, running_sequences[running_sequence_id], logit_processor, is_extend_sequence, is_validation_mode_enabled);
+                
+                #ifdef ENABLE_XGRAMMAR
+                // add tokens
+                for (auto& transformer: logit_processor.m_logit_transformers) {
+                    // if its a structured output transformer, we need to add the token to the transformer
+                    auto gram_transformer = std::dynamic_pointer_cast<LogitTransformers::XGrammarLogitsTransformer>(transformer);
```
as-suvorov: Should `Sampler` be aware about structured output implementation?
Can `StructuredOutputController` member handle this logic?

src/cpp/src/sampling/logit_processor.hpp
```diff
@@ -332,9 +45,17 @@ class LogitProcessor {
             );
         }
 
+        #ifdef ENABLE_XGRAMMAR
+        OPENVINO_ASSERT(structured_output_controller != nullptr, "Structured output controller is not initialized");
```
as-suvorov: What do you think if we incapsulate this into `StructuredOutputController`?

as-suvorov: Is it possible to move all `ifdef ENABLE_XGRAMMAR` into `StructuredOutputController`? And preserve `StructuredOutputController` unconditionally?

src/cpp/src/sampling/sampler.cpp
pavel-esir: removed

rkazants:
Can you please provide a good PR description with example?

src/cpp/src/generation_config.cpp
```diff
@@ -264,6 +268,25 @@ void GenerationConfig::validate() const {
...
+        #else
+           (*guided_generation_config).validate();
+        #endif
+    }
+}
+
+void GuidedGenerationConfig::validate() const {
+    OPENVINO_ASSERT(
+        (json_schema.has_value() + regex.has_value() + choices.has_value() + grammar.has_value()) == 1,
+        "Only one of json, regex, choices or grammar shoud be set in GuidedGenerationConfig, but got: ",
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/sampling/sampler.cpp
pavel-esir: done

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

src/cpp/include/openvino/genai/generation_config.hpp
pavel-esir: done

src/cpp/src/sampling/logit_processor.hpp
pavel-esir: done

tests/python_tests/test_llm_pipeline.py
pavel-esir: done

tests/python_tests/data/models.py
pavel-esir: reverted

src/python/py_generation_config.cpp
pavel-esir: done

src/cpp/src/sampling/sampler.cpp
pavel-esir: You are right. `Sampler` should not aware of structured output backend implementation. @apaniukov initially implemented it right with `IStructuredOutputImpl`. When i removed it i worsened situation and made `Sampler` dependent from the implementation backend. 

Now i returned it back, it's independent from backend, and also removed includes and macros from here.

src/cpp/src/sampling/logit_processor.hpp
pavel-esir: You are right. It's possible. Done. Just left `#ifdef ENABLE_XGRAMMAR` in xgrammar_backedn.[hpp|cpp].

src/cpp/src/sampling/sampler.cpp
pavel-esir: fixed

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,39 @@
...
+    parser.add_argument('prompt', nargs='?', help="Prompt to generate structured output. If not provided, a default prompt will be used.")
+    args = parser.parse_args()
+
+    device = 'CPU'  # GPU can be used as well
+    pipe = openvino_genai.LLMPipeline(args.model_dir, device)
+
+    config = openvino_genai.GenerationConfig()
+    structured_output_config = openvino_genai.StructuredOutputConfig()
+    structured_output_config.json_schema = json.dumps(Person.model_json_schema())
+    breakpoint()
```
apaniukov: del

pavel-esir: done

samples/python/text_generation/structured_output_generation.py
pavel-esir: removed

pavel-esir:
> Can you please provide a good PR description with example?

Done. Please review

pavel-esir:
> getting seq fault when stop string '#' is not included in sampling params: .build_tmpdir/_deps/xgrammar-src/cpp/grammar_matcher.cc:496: Check failed: (!IsStopTokenAccepted()) is false: GrammarMatcher has terminated after accepting the stop token, but is trying to find the next token mask . Segmentation fault (core dumped)

this issue has been fixed

src/cpp/CMakeLists.txt
pavel-esir: We discussed with @apaniukov that. But if we have xgrammar dll/so separately and load it in runtime and enable/disable structured output depending on whether this file exists or not (as it's done for cpu/gpu/npy plugins) makes genai code quite difficult. We would have to use `dlopen`, `GetModuleHandle` / `GetProcAddress` which is quite cumbersome.

So we decided to use at the moment the same approach as for other thirdpary libs.

src/cpp/include/openvino/genai/generation_config.hpp
pavel-esir: Removed this todos. Tickets for that have been created

src/cpp/src/sampling/sampler.cpp
```diff
@@ -994,10 +998,15 @@ LogitProcessor& Sampler::get_logit_processor(uint64_t request_id) {
 
 
 void Sampler::create_logit_processor(uint64_t request_id, const GenerationConfig& sampling_params, const TokenIds& prompt) {
-    m_logit_processors.insert({request_id, LogitProcessor(sampling_params, prompt)});
+    if (!m_structured_output_controller) {
+        // TODO: set valid vocab size
+        m_structured_output_controller = std::make_shared<StructuredOutputController>(m_tokenizer, m_tokenizer.get_vocab_vector().size());
```
pavel-esir: resolving this todo is in progress

src/cpp/CMakeLists.txt
```diff
@@ -88,6 +121,14 @@ if(NOT ENABLE_GGUF)
     list(REMOVE_ITEM SOURCE_FILES ${GGUF_SOURCES})
 endif()
 
+if(NOT ENABLE_XGRAMMAR)
+    set(XGRAMMAR_SOURCES ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/xgrammar_backend.hpp
+                         ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/xgrammar_backend.cpp
+                         ${CMAKE_CURRENT_SOURCE_DIR}/src/structured_output/structured_output_controller.hpp
```
apaniukov: I think we can exclude the structured output controller from here.

apaniukov: This condition is also allows to remove `#ifdef ENABLE_XGRAMMAR` from `xgrammar_backend.hpp/cpp`

src/cpp/src/generation_config.cpp
```diff
@@ -264,6 +268,25 @@ void GenerationConfig::validate() const {
     if (num_assistant_tokens == 0) {
         OPENVINO_ASSERT(max_ngram_size == 0, "'max_ngram_size' should be set to default value 0 when prompt lookup is disabled");
     }
+
+    if(is_structured_output_generation()) {
+        #ifndef ENABLE_XGRAMMAR
+            OPENVINO_THROW("Structured output is not supported in this build. Please, recompile with -DENABLE_XGRAMMAR=ON");
+        #else
+           (*structured_output_config).validate();
+        #endif
```
apaniukov: We can check if there is any backend registered here or in the `validate` method.

src/cpp/CMakeLists.txt
pavel-esir: done

pavel-esir: done

src/cpp/src/generation_config.cpp
pavel-esir: Done. Removed this last macros. Now we don't have any xgrammar macros in the sources.

src/cpp/src/sampling/sampler.cpp
pavel-esir: done

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,100 @@
...
+        print(prompt)
+        res = pipe.generate(prompt, config)
+        pipe.finish_chat()
+
+        config.do_sample = True
+        config.temperature = 0.8
+
+        pipe.start_chat()
+        generate_has_run = False
+        print(res)
```
pavel-esir:
```suggestion
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,100 @@
...
+    print("I'm a smart assistant that generates structured output in JSON format.")
+    print("You can ask me to generate information about a person, car, or transaction.")
+    print("For example, you can ask me: \"Please give a jsons of 3 persons and 1 transaction\".")
+
+    while True:
+        prompt = input('> ')
+        pipe.start_chat(sys_message)
+        structured_output_config.json_schema = json.dumps(ItemQuantities.model_json_schema())
+        config.structured_output_config = structured_output_config
+        print(prompt)
```
pavel-esir:
```suggestion
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,100 @@
...
+    parser.add_argument('model_dir', help="Path to the model directory. It should contain the OpenVINO model files.")
+    args = parser.parse_args()
+
+    device = 'CPU'  # GPU can be used as well
+    pipe = openvino_genai.LLMPipeline(args.model_dir, device)
+
+    config = openvino_genai.GenerationConfig()
+    structured_output_config = openvino_genai.StructuredOutputConfig()
+    config.max_new_tokens = 300
+    config.do_sample = False
```
pavel-esir:
```suggestion
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,100 @@
...
+    #  Persons surnames should be Karter
+    print("I'm a smart assistant that generates structured output in JSON format.")
+    print("You can ask me to generate information about a person, car, or transaction.")
+    print("For example, you can ask me: \"Please give a jsons of 3 persons and 1 transaction\".")
+
+    while True:
+        prompt = input('> ')
+        pipe.start_chat(sys_message)
+        structured_output_config.json_schema = json.dumps(ItemQuantities.model_json_schema())
+        config.structured_output_config = structured_output_config
```
pavel-esir:
```suggestion
        config.structured_output_config = structured_output_config
        config.do_sample = False
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,100 @@
...
+    pipe = openvino_genai.LLMPipeline(args.model_dir, device)
+
+    config = openvino_genai.GenerationConfig()
+    structured_output_config = openvino_genai.StructuredOutputConfig()
+    config.max_new_tokens = 300
+    config.do_sample = False
+
+    #  Persons surnames should be Karter
+    print("I'm a smart assistant that generates structured output in JSON format.")
+    print("You can ask me to generate information about a person, car, or transaction.")
```
apaniukov:
```suggestion
    print("You can ask me to generate information about a person, car, or bank transaction.")
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,100 @@
...
+        config.do_sample = True
+        config.temperature = 0.8
+
+        pipe.start_chat()
+        generate_has_run = False
+        print(res)
+        for item, quantity in json.loads(res).items():
+            for _ in range(quantity):
+                generate_has_run = True
+                config.structured_output_config.json_schema = json.dumps(items_map[item].model_json_schema())
```
apaniukov:
```suggestion
            config.structured_output_config.json_schema = json.dumps(items_map[item].model_json_schema())
            for _ in range(quantity):
                generate_has_run = True
```

samples/python/text_generation/structured_output_generation.py
apaniukov:
```suggestion
        print(f"Structured user input: {res}")
```

apaniukov (approved):

.github/labeler.yml
```diff
@@ -109,6 +109,12 @@
 - 'samples/python/rag/**/*'
 - 'tests/python_tests/samples/test_text_embedding_pipeline.py'
 
+'category: structured output generation':
+- 'src/cpp/src/sampling/structured_output/xgrammar_backend.cpp'
+- 'src/cpp/src/sampling/structured_output/xgrammar_backend.hpp'
+- 'src/cpp/src/sampling/structured_output/structured_output_controller.cpp'
+- 'src/cpp/src/sampling/structured_output/structured_output_controller.hpp'
```
Wovchena:
```suggestion
- 'src/cpp/src/sampling/structured_output/*'
```

samples/export-requirements.txt
```diff
@@ -2,7 +2,7 @@
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/pre-release
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
 openvino-tokenizers[transformers]~=2025.3.0.0.dev
-optimum-intel[nncf] @ git+https://github.com/huggingface/optimum-intel.git@main
+optimum-intel[nncf] @ git+https://github.com/huggingface/optimum-intel.git@3b2cffb40c3de6bd3bf90c1e8a98d902e87db80d
```
Wovchena: Can this be reverted? Same for tests/python_tests/requirements.txt

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,98 @@
...
+    args = parser.parse_args()
+
+    device = 'CPU'  # GPU can be used as well
+    pipe = openvino_genai.LLMPipeline(args.model_dir, device)
+
+    config = openvino_genai.GenerationConfig()
+    structured_output_config = openvino_genai.StructuredOutputConfig()
+    config.max_new_tokens = 300
+
+    #  Persons surnames should be Karter
```
Wovchena: What does this comment mean?

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,98 @@
...
+    print("I'm a smart assistant that generates structured output in JSON format.")
+    print("You can ask me to generate information about a person, car, or transaction.")
```
Wovchena: Is this chunk of prints needed? This text isn't visible to the model

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,98 @@
...
+    print("For example, you can ask me: \"Please give a jsons of 3 persons and 1 transaction\".")
+
+    while True:
+        prompt = input('> ')
```
Wovchena: What is supposed to come here? You already have `prompt_map` with predefined prompts

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,98 @@
...
+        pipe.start_chat(sys_message)
+        structured_output_config.json_schema = json.dumps(ItemQuantities.model_json_schema())
+        config.structured_output_config = structured_output_config
+        config.do_sample = False
+        res = pipe.generate(prompt, config)
+        pipe.finish_chat()
```
Wovchena, line 76: start/finish_chat are called for a single generate(). They can be removed

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,98 @@
...
+    print("You can ask me to generate information about a person, car, or transaction.")
+    print("For example, you can ask me: \"Please give a jsons of 3 persons and 1 transaction\".")
+
+    while True:
+        prompt = input('> ')
+        pipe.start_chat(sys_message)
+        structured_output_config.json_schema = json.dumps(ItemQuantities.model_json_schema())
+        config.structured_output_config = structured_output_config
+        config.do_sample = False
+        res = pipe.generate(prompt, config)
```
Wovchena: `res` isn't printed. Is this intentional? Why so?

samples/python/text_generation/structured_output_generation.py
Wovchena, line 1: Add a test for the sample
C++ sample is missing. I think C++ could use hardcoded result of `model_json_schema()`

tests/cpp/CMakeLists.txt
```diff
@@ -26,6 +26,12 @@ set(TEST_TARGET_NAME "tests_continuous_batching")
 
 add_executable(${TEST_TARGET_NAME} ${tests_src} $<TARGET_OBJECTS:openvino_genai_obj>)
 
+if(ENABLE_XGRAMMAR)
+    target_link_libraries(${TEST_TARGET_NAME} PRIVATE xgrammar)
+    target_compile_definitions(${TEST_TARGET_NAME} PRIVATE ENABLE_XGRAMMAR)
+    target_include_directories(${TEST_TARGET_NAME} PRIVATE "${XGRAMMAR_DIR}/include")
+endif()
```
Wovchena: I don't see C++ tests added for xgrammar, but their cmake is patched

tests/cpp/logit_filtering.cpp
```diff
@@ -6,7 +6,8 @@
 
 #include "sampling/logit_processor.hpp"
 
-using namespace LogitTransformers;
+using namespace ov::genai;
+using namespace ov::genai::LogitTransformers;
```
Wovchena, line 10: It's strange to change only `using` in a cpp file. There should be some usage of this change

tests/python_tests/test_structured_output.py
```diff
@@ -0,0 +1,61 @@
...
+class RESTAPIResponse(BaseModel):
+    status: Literal["success", "error"]
+    data: str
+
+@pytest.mark.precommit
+@pytest.mark.nightly
+@pytest.mark.parametrize("model_id", [
+    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+    # 'katuni4ka/tiny-random-phi3',  # model itself generate incorrect json
+    # "meta-llama/Llama-3.1-8B-Instruct",  # too heavy for CI
```
Wovchena:
```suggestion
```
It's not allowed to use it anyway because of access requesting

tests/python_tests/test_structured_output.py
```diff
@@ -0,0 +1,61 @@
...
+
+class RESTAPIResponse(BaseModel):
+    status: Literal["success", "error"]
+    data: str
+
+@pytest.mark.precommit
+@pytest.mark.nightly
+@pytest.mark.parametrize("model_id", [
+    "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+    # 'katuni4ka/tiny-random-phi3',  # model itself generate incorrect json
```
Wovchena: Isn't xgrammar supposed to enforce a valid json no matter what the model predicts?

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1789,6 +1790,56 @@ class StreamingStatus:
     @property
     def value(self) -> int:
         ...
+class StructuredOutputConfig:
+    """
+    
+        Structure to keep generation config parameters for structured output generation.
+        It is used to store the configuration for structured generation, which includes
+        the JSON schema and other related parameters.
```
Wovchena, line 1798: Describe how caching behaves

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -1789,6 +1790,56 @@ class StreamingStatus:
...
+    
+        Structured output parameters:
+        json_schema:           if set, the output will be a JSON string constrained by the specified json-schema.
+        regex:          if set, the output will be constrained by specified regex.
+        choices:        if set, the output will be one of specified strings.
+        grammar:        if set, the output will be constrained by specified grammar.
```
Wovchena: Define what happens if the parameters are mixed

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -861,6 +861,11 @@ Vocab Tokenizer::get_vocab() const {
     return vocab;
 }
 
+const std::vector<std::string>& Tokenizer::get_vocab_vector() const {
+    OPENVINO_ASSERT(!m_pimpl->m_vocab.empty(), "Tokenizer vocab is empty. Please check if the detokenizer model was provided and loaded correctly.");
```
Wovchena, line 879: Reroute `get_vocab()` to call `get_vocab_vector(). That would remove assert duplication

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -287,6 +287,20 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
...
+     * This function retrieves the vocabulary from the detokenizer, which maps
+     * token indices to their corresponding string. Note that some token strings
+     * may not be valid UTF-8 encoded. The resulting vocabulary may differ from the
+     * original tokenizer's vocabulary due to optimizations during conversion (space symbol
+     * swaps, byte fallback reverse, and other preprocessing changes).
+     *
+     * @return A vector of string tokens.
+     * @throws Exception if the detokenizer is not available.
+     */
+    const std::vector<std::string>& get_vocab_vector() const;
```
Wovchena, line 302: Python binding is missing

src/cpp/CMakeLists.txt
```diff
@@ -103,6 +142,12 @@ if(ENABLE_GGUF)
     target_compile_definitions(${TARGET_NAME_OBJ} PRIVATE ENABLE_GGUF)
 endif()
 
+if(ENABLE_XGRAMMAR)
+    target_link_libraries(${TARGET_NAME_OBJ} PRIVATE xgrammar)
+    target_compile_definitions(${TARGET_NAME_OBJ} PRIVATE ENABLE_XGRAMMAR)
+    target_include_directories(${TARGET_NAME_OBJ} PRIVATE "${XGRAMMAR_DIR}/include")
```
Wovchena: It's strange that you need to specify includes. Can `xgrammar` be updated to do that automatically instead?

src/cpp/CMakeLists.txt
```diff
@@ -67,6 +67,39 @@ if(ENABLE_GGUF)
     target_include_directories(gguflib PUBLIC "${gguflib_SOURCE_DIR}")
 endif()
 
+if(ENABLE_XGRAMMAR)
```
Wovchena, line 75: Add `ENABLE_XGRAMMAR` to `features.cmake`.
I thought it should be enabled by default, but it's off. Is this intentional? If the default is flipped, we will need a test build with ENABLE_XGRAMMAR=OFF

.github/labeler.yml
```diff
@@ -109,6 +109,12 @@
 - 'samples/python/rag/**/*'
 - 'tests/python_tests/samples/test_text_embedding_pipeline.py'
 
+'category: structured output generation':
```
Wovchena, line 112: Add it to `.github/components.yml` and smart CI

src/cpp/CMakeLists.txt
```diff
@@ -122,6 +167,9 @@ target_include_directories(${TARGET_NAME} INTERFACE "$<INSTALL_INTERFACE:runtime
                                                     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
                                                     "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>")
 
+if(ENABLE_XGRAMMAR)
+    target_link_libraries(${TARGET_NAME} PRIVATE xgrammar)
```
Wovchena, line 174: Is it really needed to link xgrammar to `TARGET_NAME_OBJ` and `TARGET_NAME`? I'd expect `TARGET_NAME_OBJ` to be enough

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -25,6 +25,25 @@ namespace genai {
...
+ * @param regex if set, the output will be constraint by specified regex.
+ * @param choices if set, the output will be one of specified strings.
+ * @param grammar if set, the output will be constraint by specified EBNF grammar.
+ */
+public:
+    StructuredOutputConfig() = default;
+    std::optional<std::string> json_schema;
+    std::optional<std::string> regex;
+    std::optional<std::vector<std::string>> choices;
+    std::optional<std::string> grammar;
```
Wovchena, line 50: Would it be better if the actual uses case was represented with a type, so `StructuredOutputConfig` becomes a variant of all the types + `backend` member. That way API would reflect, that only one of them can be specified at a time.

src/cpp/CMakeLists.txt
```diff
@@ -103,6 +142,12 @@ if(ENABLE_GGUF)
     target_compile_definitions(${TARGET_NAME_OBJ} PRIVATE ENABLE_GGUF)
 endif()
 
+if(ENABLE_XGRAMMAR)
+    target_link_libraries(${TARGET_NAME_OBJ} PRIVATE xgrammar)
+    target_compile_definitions(${TARGET_NAME_OBJ} PRIVATE ENABLE_XGRAMMAR)
```
Wovchena:
```suggestion
```
The absence of backend is defined by the presence of `xgrammar_backend` files

samples/python/text_generation/structured_output_generation.py
Wovchena, line 1: Update docs. Sample's readme. https://openvinotoolkit.github.io/openvino.genai/ at leas should be extended using PR description.

Wovchena (changes requested):

tests/python_tests/test_structured_output.py
apaniukov: A random model is too random to correctly finish the required fields. But if we restrict string sizes it might work.

Wovchena: It's the point of structured output to suppress the randomness

apaniukov: You can easily run out of the token budget before finishing the string value, resulting in an invalid JSON string.

tests/python_tests/test_structured_output.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.mark.nightly
+@pytest.mark.parametrize("model_id", structured_id_models)
+@pytest.mark.parametrize("prompt_and_scheme", [
+    ("Generate a json about a person.", Person), 
+    ("Generate a json about a transaction.", Transaction),
+    ("Generate a json about a REST API response.", RESTAPIResponse)
+])
+def test_structured_output_generation(model_id, prompt_and_scheme):
+    prompt, SchemeType = prompt_and_scheme
+    opt_model, hf_tokenizer, models_path  = download_and_convert_model(model_id)
```
apaniukov:
```suggestion
    _, _, models_path  = download_and_convert_model(model_id)
```
```suggestion
    opt_model, hf_tokenizer, models_path  = download_and_convert_model(model_id)
```

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

samples/python/text_generation/structured_output_generation.py
pavel-esir: done, but with `print(f"Generated JSON with item quantities: {res}")`

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

.github/labeler.yml
pavel-esir: done

samples/export-requirements.txt
pavel-esir: reverted to the main. at that time test were failing because of the optimum, as a temporary solution i just pinned the wroking hash. but not it's good with `main`

samples/python/text_generation/structured_output_generation.py
pavel-esir: oups, it was left by accident when i forced to jenerate json with specific names. removed it

samples/python/text_generation/structured_output_generation.py
pavel-esir: Replaced with one print, but text mainly remains the same. It's a help for the user which describes what sample does, and what type of prompts it accepts with an example

samples/python/text_generation/structured_output_generation.py
pavel-esir: Removed `prompt_map` with predefined prompts. now the exact user prompt without modifications is used twice. First time to generate number of items for each object type, and after that we use the same prompt for generating jsons with items.
1st stge
prompt = 'give me json of 2 person, and 1 car'
generate output= `{"person": 2, "car": 1, "transaction": 0}`

2nd stage calling generate for each item, genrate will be called 3 times, for each item with specific json schema, for the Car with car, for Transaction with transaction and so on.
prompt remain the same
generate(prompt, ...json_schema_for_car) - `{'name': 'John', 'surname': 'Doe', 'age': 40, 'city': 'Munich'}`
generate(prompt, ...json_schema_for_person) - `{'name': 'John', 'surname': 'Doe', 'age': 40, 'city': 'Munich'}`
generate(prompt, ...json_schema_for_person) - `{'name': 'John', 'surname': 'Smith', 'age': 21, 'city': 'Dublin'}`

I described this in Readme.

samples/python/text_generation/structured_output_generation.py
pavel-esir: We used it for setting system prompt. In the latest sample both start_chats set their own sys_prompt.

samples/python/text_generation/structured_output_generation.py
pavel-esir: now it's printed

.github/labeler.yml
pavel-esir: done

tests/cpp/logit_filtering.cpp
pavel-esir: Since we refactored and moved LogitTransformers from the `logit_processor.hpp` header into another header and put them inside a namespace some classes here are unaccecable. Hence here is this new using.

tests/python_tests/test_structured_output.py
pavel-esir: Made rules for json generatino more stric, so now even tiny models fit into max_new_tokens with correct json. Uncommented this tiny model

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: done

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: added

.github/labeler.yml
pavel-esir: done

tests/python_tests/test_structured_output.py
pavel-esir: not relevand anymore. changed this lines completely with the usage of the new fixture

tests/python_tests/test_structured_output.py
pavel-esir: removed

src/cpp/CMakeLists.txt
pavel-esir: It should be OFF by default because of the large xgrammar binary size. I agree, if we opt to enable it by default, we would need to test separately `ENABLE_XGRAMMAR=OFF` as well.

src/cpp/CMakeLists.txt
pavel-esir: It didn't link at the end when i likned only to `TARGET_NAME_OBJ`, i don't remember why. Let me try to remove and check the result

samples/python/text_generation/structured_output_generation.py
pavel-esir: We would have to parse json after the first generate was. To do that we would need to use json parsing thirdparty libs. Can we merge this PR without C++ sample and do it as a separate task?

src/cpp/CMakeLists.txt
pavel-esir: yes, it's necessary here, without it, it fails
```
> const&, bool, std::optional<int>, std::optional<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > >, bool, bool)'
[build] /usr/bin/ld: openvino_genai/libopenvino_genai.so.2025.3.0.0: undefined reference to `xgrammar::GrammarMatcher::AcceptToken(int, bool)'
[build] collect2: error: ld returned 1 exit status
```

src/cpp/include/openvino/genai/generation_config.hpp
pavel-esir: This is done this way because of simplicity and also is similar to vllm.

```
sampling_params = SamplingParams(guided_decoding=guided_decoding_params, stop=["\n"])
```
https://docs.vllm.ai/en/v0.8.2/getting_started/examples/structured_outputs.html

Also the current approach seemed appropriate since we also have sometimes conflicting fields in GenerationConfig which are checked inside `GenerationConfig::validate()`

src/cpp/CMakeLists.txt
pavel-esir: removed

src/cpp/CMakeLists.txt
pavel-esir: removed

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

src/python/openvino_genai/py_openvino_genai.pyi
pavel-esir: done

tests/cpp/CMakeLists.txt
pavel-esir: removed that

samples/python/text_generation/structured_output_generation.py
pavel-esir: Discussed offline that c++ smaple will be added separately

src/python/openvino_genai/py_openvino_genai.pyi
pavel-esir: Didn't quite understand what caching you mean here?

src/cpp/CMakeLists.txt
Wovchena: It was expected to add about 17 Mb which was considered fine. Did that change?

samples/python/text_generation/README.md
```diff
@@ -171,6 +171,33 @@ Recommended models: Qwen/Qwen2.5-3B-Instruct, Qwen/Qwen2.5-7B-Instruct
   ```
 
 
+### 11. Structured Output Sample (`structured_output_sample`)
+- **Description:**
+This sample demonstrates how to use OpenVINO GenAI to generate structured outputs—such as JSON or other formats—from text prompts. JSON generation process is split into multiple steps to mitigate generating complex, variadic JSON structures in a single pass. This is done because not all models are able to generate a complex JSON, with a variadic number of elements, in one shot, especially if the model is small and not fine-tuned for this task. By separating the tasks, it becomes possible to use smaller models and still achieve good quality generated JSON.
```
Wovchena:
```suggestion
This sample demonstrates how to use OpenVINO GenAI to generate structured outputs such as JSON from text prompts. JSON generation process is split into multiple steps to mitigate generating complex, variadic JSON structures in a single pass. This is done because not all models are able to generate a complex JSON, with a variadic number of elements, in one shot, especially if the model is small and not fine-tuned for this task. By separating the tasks, it becomes possible to use smaller models and still achieve good quality generated JSON. 
```
I guess you can keep `-` if you want, but surround it with spaces

Wovchena: It's not clear what "generation process split" refers to. Is it the internal xgrammar implementation or the prompt that asks to generate smaller jsons

Wovchena: The sample highly depends on the model quality. Add a recommendation which model to use similar to `LLM ReAct Agent Sample`

samples/python/text_generation/README.md
```diff
@@ -171,6 +171,33 @@ Recommended models: Qwen/Qwen2.5-3B-Instruct, Qwen/Qwen2.5-7B-Instruct
...
 
 
+### 11. Structured Output Sample (`structured_output_sample`)
+- **Description:**
+This sample demonstrates how to use OpenVINO GenAI to generate structured outputs—such as JSON or other formats—from text prompts. JSON generation process is split into multiple steps to mitigate generating complex, variadic JSON structures in a single pass. This is done because not all models are able to generate a complex JSON, with a variadic number of elements, in one shot, especially if the model is small and not fine-tuned for this task. By separating the tasks, it becomes possible to use smaller models and still achieve good quality generated JSON. 
+- **Run Command:**
+  ```bash
+  python structured_output_generation.py model_dir
+  ```
+  After running the command, an interactive dialog will start. You can enter a prompt and receive a structured output in response. The generation process is divided into two stages:
```
Wovchena:
```suggestion
  After running the command, an interactive dialog starts. Enter a prompt and receive a structured output in response. The generation process is divided into two stages:
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    car: int = Field(ge=0, le=100)
+    transaction: int = Field(ge=0, le=100)
+
+
+items_map = {"person": Person, "car": Car, "transaction": Transaction}
+
+sys_message = (
+    "You generate jsons based on the user's request. You can generate jsons with different types of objects: person, car, transaction. "
+    "If user requested different type json fields should remain zero. "
+    "Please not that words 'individual', 'person', 'man', 'human', 'woman', 'people', 'inhabitant', 'citizen' are synonyms and can be used interchangeably. "
```
Wovchena:
```suggestion
    "Please note that words 'individual', 'person', 'man', 'human', 'woman', 'people', 'inhabitant', 'citizen' are synonyms and can be used interchangeably. "
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    "E.g. if user wants 5 houses, then json must be {\"person\": 0, \"car\": 0, \"transactions\": 0}, "
+    "if user wand 3 person and 1 house then json must be {\"person\": 3, \"car\": 0, \"transaction\": 0}. "
```
Wovchena:
```suggestion
    "if user wants 3 people and 1 house then json must be {\"person\": 3, \"car\": 0, \"transaction\": 0}. "
```
(I didn't verified if the model still works correctly after that change)

Wovchena (changes requested):

samples/python/text_generation/README.md
pavel-esir: accepted your changes

pavel-esir: added recommended models

samples/python/text_generation/README.md
pavel-esir: applied

samples/python/text_generation/structured_output_generation.py
pavel-esir: Done. Model still works. It should work even better after your grammar correction.

samples/python/text_generation/structured_output_generation.py
pavel-esir: applied

samples/python/text_generation/README.md
pavel-esir: It's the second. Added explicit clarification that 2 stages are separate callings of "generate" to get smaller jsons.

src/cpp/CMakeLists.txt
pavel-esir: @apaniukov what do you think? I check and yes, it will take ~14 Mb extra to the existing 15 Mb.

apaniukov: Let's make it `ON` by default.

pavel-esir: Turned ON by default. And i'm also testing that with `OFF` the main llm tests still work

.github/workflows/linux.yml
```diff
@@ -676,7 +676,7 @@ jobs:
 
       - name: Test Samples (Python and C++)
         if: ${{ matrix.test.run_condition }}
-        run: python -m pytest -vvs ${{ env.SRC_DIR }}/${{ matrix.test.cmd }} -m "${{ matrix.test.marker }}"
+        run: python -m pytest -vvs ${{ env.SRC_DIR }}/${{ matrix.test.cmd }} -m "${{ matrix.test.marker }}"  # strctured output will be run separately
```
Wovchena: i can't find that separate place. What's the reason to run it separately and not in matrix?

samples/python/text_generation/README.md
```diff
@@ -171,6 +171,35 @@ Recommended models: Qwen/Qwen2.5-3B-Instruct, Qwen/Qwen2.5-7B-Instruct
   ```
 
 
+### 11. Structured Output Sample (`structured_output_sample`)
+- **Description:**
+This sample demonstrates how to use OpenVINO GenAI to generate structured outputs such as JSON from text prompts. In the sample process is split into "generate" calls to mitigate generating complex, variadic JSON structures in a single pass. This is done because not all models are able to generate a complex JSON, with a variadic number of elements in one shot, especially if the model is small and not fine-tuned for this task. By separating the task into two stage, it becomes possible to use smaller models and still achieve good quality generated JSON.
```
Wovchena:
```suggestion
This sample demonstrates how to use OpenVINO GenAI to generate structured outputs such as JSON from text prompts. This sample implementation is split into multiple "generate" calls to mitigate generating complex, variadic JSON structures in a single pass. This is done because not all models are able to generate a complex JSON, with a variadic number of elements in one shot, especially if the model is small and not fine-tuned for this task. By separating the task into two stages, it becomes possible to use smaller models and still achieve generated JSON good quality.
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    person: int = Field(ge=0, le=100)
+    car: int = Field(ge=0, le=100)
+    transaction: int = Field(ge=0, le=100)
+
+
+items_map = {"person": Person, "car": Car, "transaction": Transaction}
+
+sys_message = (
+    "You generate jsons based on the user's request. You can generate jsons with different types of objects: person, car, transaction. "
+    "If user requested different type json fields should remain zero. "
```
Wovchena:
```suggestion
    "If the user requests a different type, json fields should remain zero. "
```
0 conditional

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    "Please note that words 'individual', 'person', 'people', 'man', 'human', 'woman', 'people', 'inhabitant', 'citizen' are synonyms and can be used interchangeably. "
+    "E.g. if user wants 5 houses, then json must be {\"person\": 0, \"car\": 0, \"transactions\": 0}, "
+    "if user wants 3 people and 1 house then json must be {\"person\": 3, \"car\": 0, \"transaction\": 0}. "
+    "Make sure that json contans numbers that user requested. If user asks specifi attributes, like 'surname', 'model', etc. "
```
Wovchena:
```suggestion
    "Make sure that json contans numbers that user requested. If user asks to specify attributes, like 'surname', 'model', etc. "
```
It's worth asking an assistant to refine this message even further :) Maybe it finds other mistakes

samples/requirements.txt
```diff
@@ -1,2 +1,3 @@
 -r ./deployment-requirements.txt
 -r ./export-requirements.txt
+pydantic
```
Wovchena, line 3: Move to `deployment-requirements.txt` and a comment what it is for

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    args = parser.parse_args()
+
+    device = 'CPU'  # GPU can be used as well
+    pipe = LLMPipeline(args.model_dir, device)
+
+    config = GenerationConfig()
+    config.max_new_tokens = 300
+
+    print("I'm a smart assistant that generates structured output in JSON format."
+          "You can ask me to generate information about a person, car, or bank transaction."
```
Wovchena:
```suggestion
    print("This is a smart assistant that generates structured output in JSON format."
          "You can ask to generate information about a person, car, or bank transaction."
```
No reason for a print to pretend assistant

src/README.md
```diff
@@ -397,6 +397,49 @@ print(f'Median from token to token duration: {np.median(durations):.2f} ms')
 
 For more examples of how metrics are used, please refer to the Python [benchmark_genai.py](../samples/python/text_generation/README.md) and C++ [benchmark_genai](../samples/cpp/text_generation/README.md) samples.
 
+
+### Structured Output generation
+OpenVINO™ GenAI supports structured output generation, which allows you to generate outputs in a structured format such as JSON, regex, or accoring to EBNF grammar.
```
Wovchena: 1. Add what `EBNF` stands for to simplify googling it
2. JSON is sometimes written in lower case. Make it consistent

src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
```diff
@@ -0,0 +1,113 @@
...
+XGrammarStructuredOutput::get_logits_transformer(const GenerationConfig& sampling_parameters) {
+    OPENVINO_ASSERT(sampling_parameters.is_structured_output_generation(),
+                   "XGrammarStructuredOutput can only be used for structured output generation");
+    
+    auto& guided_gen_config = *sampling_parameters.structured_output_config;
+    guided_gen_config.validate();
+
+    xgrammar::Grammar grammar;
+    if (guided_gen_config.json_schema.has_value()) {
+        // std::cout << *guided_generation_config.json_schema << std::endl;
```
Wovchena:
```suggestion
```

tests/cpp/CMakeLists.txt
```diff
@@ -26,6 +26,11 @@ set(TEST_TARGET_NAME "tests_continuous_batching")
 
 add_executable(${TEST_TARGET_NAME} ${tests_src} $<TARGET_OBJECTS:openvino_genai_obj>)
 
+if(ENABLE_XGRAMMAR)
+    target_link_libraries(${TEST_TARGET_NAME} PRIVATE xgrammar)
+    target_compile_definitions(${TEST_TARGET_NAME} PRIVATE ENABLE_XGRAMMAR)
+endif()
+
```
Wovchena:
```suggestion
```

tests/python_tests/samples/test_structured_output_sample.py
```diff
@@ -0,0 +1,90 @@
...
+        if not items_generated and line.startswith("> Generated JSON with item quantities:"):
+            item_quantities = line.split(":", 1)[1].strip()
+            items_generated = True
+        elif items_generated and line.startswith('{'):
+            items.append(line.strip())
+
+    try:
+        data = json.loads(item_quantities)
+    except Exception as e:
+        pytest.fail(f"Output is not valid JSON: {e}")
```
Wovchena: Why do you convert an Exception into pytest's handling. I think pytest already forwards exceptions. Same below

tests/python_tests/test_structured_output.py
```diff
@@ -0,0 +1,112 @@
...
+    # If it's invalid it will raise an error.
+    SchemeType.model_validate_json(res_str)
+
+
+@pytest.mark.precommit
+@pytest.mark.nightly
+@pytest.mark.parametrize("ov_pipe", structured_id_models, indirect=True)
+@pytest.mark.parametrize("prompt_and_regex", [
+    ("Generate a json about a person.", r'^\{"city":"(Dublin|Dubai|Munich)"\}$'),
+    # without regex constraint it will generate an email letter, but with the regex it will generate an email address string
```
Wovchena:
```suggestion
    # without regex constraint it generates email letter content, but with the regex it generates an email address string
```

tests/python_tests/test_structured_output.py
```diff
@@ -0,0 +1,112 @@
...
+    gen_config.max_new_tokens = 100
+    gen_config.structured_output_config = structured_output_config
+
+    res_str = ov_pipe.generate(prompt, generation_config=gen_config)
+
+    # Basic checks for the generated format
+    # Currently there is not general way to validate EBNF output,
+    # so we will just check if it matches the expected date format.
+    if "date" in prompt:
+        assert re.match(r"^\d{4}-\d{2}-\d{2}$", res_str), f"Output {res_str} does not match date format"
```
Wovchena:
```suggestion
    assert re.match(r"^\d{4}-\d{2}-\d{2}$", res_str), f"Output {res_str} does not match date format"
```
date is always expected because it's enforced by `ebnf_grammar`

src/cpp/src/sampling/structured_output/structured_output_controller.hpp
```diff
@@ -0,0 +1,51 @@
...
+namespace genai {
+
+class IStructuredOutputImpl {
+public:
+    virtual ~IStructuredOutputImpl() = default;
+    virtual std::shared_ptr<ov::genai::LogitTransformers::ILogitTransformer>
+        get_logits_transformer(const ov::genai::GenerationConfig& sampling_parameters) = 0;
+};
+
+class StructuredOutputController {
```
Wovchena, line 62: It would be a shame to loose such a detailed PR description. I still suggests converting it into one of the Concepts pages. Alternatively you can add comments to the corresponding classes with decisions explanations about your design

src/cpp/src/sampling/sampler.hpp
```diff
@@ -97,7 +99,7 @@ class Sampler {
     Tokenizer m_tokenizer;
 
     ThreadPool m_thread_pool;
-
+    std::shared_ptr<ov::genai::StructuredOutputController> m_structured_output_controller;
```
Wovchena, line 102: Why is it shared_ptr?

src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
```diff
@@ -0,0 +1,113 @@
...
+    } else if (guided_gen_config.regex.has_value()) {
+        grammar = xgrammar::Grammar::FromRegex(*guided_gen_config.regex);
+    } else if (guided_gen_config.grammar.has_value()) {
+        grammar = xgrammar::Grammar::FromEBNF(*guided_gen_config.grammar);
+    }
+
+    auto compiled_grammar = m_grammar_compiler->CompileGrammar(grammar);
+    std::vector<int> override_stop_tokens(sampling_parameters.stop_token_ids.begin(), sampling_parameters.stop_token_ids.end());
+    
+    return std::make_shared<LogitTransformers::XGrammarLogitsTransformer>(std::move(compiled_grammar), override_stop_tokens);
```
Wovchena, line 46: Is compiled grammar caching implemented?

Wovchena (changes requested):

.github/workflows/linux.yml
pavel-esir: oups, this is a leftover comment. I should've removed it when enable structured output by default in 81e02e5

src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
apaniukov: It is on by default in the grammar compiler.

Wovchena: Explain that in docstrings. To avoid surprises when the next `generate()` with the identical schema string representation but different `StructuredOutputConfig` takes different time
Should cache settings be exposed?

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

samples/requirements.txt
pavel-esir: done

src/README.md
pavel-esir: 1. revealed in brackets what abbreviation `EBNF` stands for
2. in all md docs in this PR capitalized to `JSON`

.github/workflows/linux.yml
pavel-esir: removed that misguiding leftover comment

tests/python_tests/samples/test_structured_output_sample.py
pavel-esir: removed both try blocks because of their redundancy

src/cpp/src/sampling/structured_output/structured_output_controller.hpp
pavel-esir: Good point! I'll do that

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    args = parser.parse_args()
+
+    device = 'CPU'  # GPU can be used as well
+    pipe = LLMPipeline(args.model_dir, device)
+
+    config = GenerationConfig()
+    config.max_new_tokens = 300
+
+    print("This is a smart assistant that generates structured output in JSON format."
+          "You can ask to generate information about a person, car, or bank transaction."
```
Wovchena:
```suggestion
    print("This is a smart assistant that generates structured output in JSON format. "
          "You can ask to generate information about a person, car, or bank transaction. "
```

samples/python/text_generation/structured_output_generation.py
```diff
@@ -0,0 +1,95 @@
...
+        try:
+            prompt = input('> ')
+        except EOFError:
+            break
+        pipe.start_chat(sys_message)
+        config.structured_output_config = StructuredOutputConfig(json_schema = json.dumps(ItemQuantities.model_json_schema()))
+        config.do_sample = False
+        res = pipe.generate(prompt, config)
+        pipe.finish_chat()
+        print(f"Generated JSON with item quantities: {res}")
```
Wovchena, line 77: Wrap the printed value with `json.loads()` how it's done below to align the print style

src/cpp/src/sampling/structured_output/structured_output_controller.hpp
pavel-esir: - Added docstring for every new introduced class with description of the design.
- I even included ascii diagram of class dependencies :)

src/cpp/src/sampling/sampler.hpp
pavel-esir: Mainly it's introduces so that we can distinguish if the field was initialized or not by comparing with `nullptr`.

src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
pavel-esir: Done. Please take a look

Wovchena:
Looking good. Just address the last comments

src/cpp/src/sampling/logit_processor.hpp
```diff
@@ -396,6 +110,11 @@ class LogitProcessor {
         } else {
             it->second++;
         }
+        for (const auto& transformer : m_stateful_logit_transformers) {
+            if (transformer->is_applicable(m_generated_tokens)) {
+                transformer->accept_tokens({new_token_id});
```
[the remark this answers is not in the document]
pavel-esir: copilot, you were wrong `new_token_id` is well defined as an input argument

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -25,6 +25,34 @@ namespace genai {
...
 
+
+/* 
+* Structured output parameters:
+* @param json_schema if set, the output will be a JSON string constrained by the specified json_schema.
+* @param regex if set, the output will be constrained by specified regex.
+* @param grammar if set, the output will be constrained by specified EBNF grammar.
+* @param backend if set, the structured output generation will use specified backend, currently only "xgrammar" is supported.
+* 
+* If several several parameters are set, e.g. json_schema and regex, then an error will be thrown when validating the configuration.
```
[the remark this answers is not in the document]
pavel-esir: You are also wrong here. Word `several` is crucial her.

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

samples/python/text_generation/structured_output_generation.py
pavel-esir: done

pavel-esir:
> Looking good. Just address the last comments

Applied you suggestions. Copilots suggestions were bad, didn't apply them, in comment answered why they are bad.

samples/python/text_generation/structured_output_generation.py
Wovchena: I guess I picked a wrong line for that comment. Wrap the result of `pipe.generate(prompt, config)` to json so it's done only once

src/cpp/CMakeLists.txt
```diff
@@ -67,6 +67,11 @@ if(ENABLE_GGUF)
     target_include_directories(gguflib PUBLIC "${gguflib_SOURCE_DIR}")
 endif()
 
+# Check if it's Debug and Windows set ENABLE_XGRAMMAR=OFF forcefully
+if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND WIN32)
```
pavel-esir, line 71: this is a temporary solution. need to fix Debug build on win

samples/python/text_generation/structured_output_generation.py
Wovchena: Why is there `res.replace("'", '"')`? Should you ask the model to use specific quotes?

src/cpp/include/openvino/genai/generation_config.hpp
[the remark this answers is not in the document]
Wovchena: Can you explain?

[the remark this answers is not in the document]
pavel-esir: If i remove the word 'several' then it will mean that "if parameters are set then error will be thrown" which is not true. Error is thrown only if several regex, json_shema, etc. are set a once.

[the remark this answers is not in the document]
Wovchena: Both `several` were removed in "if parameters are set then error will be thrown", but the suggestion is to remove one of them.

Would multiple be a better word?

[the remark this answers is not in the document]
pavel-esir: 🤦i notices only now i used the word 2 times. will fix it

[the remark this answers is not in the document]
pavel-esir: fixed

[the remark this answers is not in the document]
pavel-esir: Thanks a lot got attentive review. Not noticing 2 words even after i was pointed to it was by bad of which i'm greatly shamed of.

samples/python/text_generation/structured_output_generation.py
pavel-esir: Done/ Asked mode to generate `"` and it was enough. And also now i call `json.loads()` only once.

Wovchena (approved):

src/README.md
```diff
@@ -397,6 +397,49 @@ print(f'Median from token to token duration: {np.median(durations):.2f} ms')
...
+    surname: str = Field(pattern=r"^[A-Z][a-z]{1,20}$")
+    age: int
+    city: str
+
+pipe = LLMPipeline(models_path, "CPU")
+
+config = GenerationConfig()
+config.max_new_tokens = 100
+# If backend is not specified, it will use the default backend which is "xgrammar" for the moment.
+config.structured_output_config = StructuredOutputConfig(json_schema=json.dumps(Person.model_json_schema()), backed="xgrammar")
```
Wovchena:
```suggestion
config.structured_output_config = StructuredOutputConfig(json_schema=json.dumps(Person.model_json_schema()), backend="xgrammar")
```

pavel-esir:
Fails are caused by https://github.com/huggingface/datasets/issues/7647 not by the PR changes

PR #2296, opened 2025-05-30
Switch to SDPA for VLMs

yatarkan:


andrei-kochin (approved):

mzegla:
In OVMS we rely on that constructor and with that change default setting (using paged attention) is changed so CB adapter is not invoked. In such case we end up with different kind of pipeline which is not expected in that flow in OVMS.

If we explicitly add {"ATTENTION_BACKEND": "PA"}, would that be sufficient workaround?

yatarkan:
@mzegla Please re-review current changes. The change affects only VLM pipeline now. 

> If we explicitly add {"ATTENTION_BACKEND": "PA"} property, would that be sufficient workaround?

Yes

andrei-kochin (approved):

PR #2300, opened 2025-05-30
fix supported models link in TTS samples

eaidova:


rkazants (approved):

as-suvorov:
@akladiev we have an issue with smart CI on PR containing changes in .md files only (another example: https://github.com/openvinotoolkit/openvino.genai/pull/2287). Could you help with this issue?

PR #2320, opened 2025-06-06
Disabled crop of the prompt for minicpmv.

andreyanufr:
Disabled crop of the prompt for minicpmv.

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/wwb.py
+++ b/tools/who_what_benchmark/whowhatbench/wwb.py
@@ -366,6 +366,10 @@ def genai_gen_visual_text(model, prompt, image, processor, tokenizer, max_new_to
     return out.texts[0]
 
 
+def is_model_with_automatic_crop(config):
+    return "internvl" in config.model_type or "minicpmv" in config.model_type
+
+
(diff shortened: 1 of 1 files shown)


tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -415,7 +419,7 @@ def create_evaluator(base_model, args):
         elif task == "visual-text":
             tokenizer = load_tokenizer(args)
             processor, config = load_processor(args)
-            if config and "internvl" in config.model_type and args.hf:
+            if config and is_model_with_automatic_crop(config) and args.hf:
                 crop_question = False
```
Wovchena, line 423: I assume this improves the similarity score. Why isn't it disable for every model?

andreyanufr: Because this is not the default behavior for all models. For example microsoft/Phi-4-multimodal-instruct returns prompt+generation, but openbmb/MiniCPM-V-2_6 returns only generation.

Wovchena (approved):

PR #2321, opened 2025-06-06
Add extended perf metrics for speculative decoding

sbalandi:


src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include "openvino/genai/perf_metrics.hpp"
+#include "openvino/genai/visibility.hpp"
+
+
+namespace ov::genai {
+
+struct OPENVINO_GENAI_EXPORTS SDPerfMetrics : public ov::genai::PerfMetrics {
+    ov::genai::MeanStdPair ttst;  // Time to the second token (in ms).
+    ov::genai::MeanStdPair ttl;  // latency from the third token (in ms).
+    size_t num_accepted_tokens; // num tokens, which was accepted but main model
```
Wovchena:
```suggestion
    size_t num_accepted_tokens; // num tokens, which was accepted by main model
```
Are the number of generated tokens and num_accepted_tokens the same?
Is this values the same for `main_model_metrics` and `draft_model_metrics`?
Add the answers to docstrings

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -36,13 +36,15 @@ using StringInputs = std::variant<std::string, std::vector<std::string>>;
...
-* @param metrics performance metrics with tpot, ttft, etc. of type ov::genai::PerfMetrics
+* @param perf_metrics performance metrics with tpot, ttft, etc. of type ov::genai::PerfMetrics
+* @param extended_perf_metrics pipeline specific performance metrics etc. of type ov::genai::PerfMetrics
 */
 class EncodedResults {
 public:
     std::vector<std::vector<int64_t>> tokens;
     std::vector<float> scores;
     PerfMetrics perf_metrics;
+    std::shared_ptr<PerfMetrics> extended_perf_metrics;
```
Wovchena: Should python bindings be updated?
In anyway, add a test for the member

src/cpp/src/speculative_decoding/speculative_decoding_impl.cpp
```diff
@@ -204,32 +207,41 @@ void ContinuousBatchingPipeline::SpeculativeDecodingImpl::step() {
...
-    
         raw_perf_counters.m_token_infer_durations.emplace_back(infer_duration);
         raw_perf_counters.m_inference_durations[0] += MicroSeconds(infer_duration);
         raw_perf_counters.m_new_token_times.emplace_back(main_timer.get_end_time());
-
         raw_perf_counters.m_batch_sizes.emplace_back(num_generated_tokens);
+
+        auto main_model_gen_duration = main_timer.get_duration_microsec();
+        auto m_main_pipeline_metrics = m_main_pipeline->get_metrics();
+        main_raw_perf_counters.m_durations.emplace_back( MicroSeconds(main_model_gen_duration));
```
Wovchena:
```suggestion
        main_raw_perf_counters.m_durations.push_back(MicroSeconds(main_model_gen_duration));
```

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include "openvino/genai/perf_metrics.hpp"
+#include "openvino/genai/visibility.hpp"
+
+
+namespace ov::genai {
+
+struct OPENVINO_GENAI_EXPORTS SDPerfMetrics : public ov::genai::PerfMetrics {
+    ov::genai::MeanStdPair ttst;  // Time to the second token (in ms).
+    ov::genai::MeanStdPair ttl;  // latency from the third token (in ms).
```
Wovchena: Add a comment why the third token is introduced.

Is there the same strangeness for the second token for greedy?
Does it hold for CPU and GPU?
You'll probably need to summon the plugins once you answerer these questions.

src/cpp/src/speculative_decoding/speculative_decoding_perf_metrics.cpp
```diff
@@ -0,0 +1,112 @@
...
+        raw_metrics.m_times_to_first_token.clear();
+        raw_metrics.m_times_to_first_token.emplace_back(durations[0]);
+
+        num_generated_tokens = batch_sizes[0];
+        num_accepted_tokens = 0;
+
+        raw_metrics.generate_durations.clear();
+        raw_metrics.generate_durations.emplace_back(durations[0]);
+
+        num_iterations = durations.size();
```
Wovchena: Is `num_iterations` really needed as a member? It's ` raw_metrics.m_durations.size()`. You can update  a docstring instead with a hint how to obtain the value. If it's required to have a separated entity for that value, it should be a getter function to avoid values missalinment.

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
```diff
@@ -0,0 +1,38 @@
...
+    size_t num_accepted_tokens; // num tokens, which was accepted but main model
+    size_t num_iterations; // num inference
+
+    ov::genai::MeanStdPair get_ttst();
+    ov::genai::MeanStdPair get_latency();
+
+    SDPerfMetrics() = default;
```
Wovchena:
```suggestion
```
What is it for? `PerfMetrics` doesn't have it. Same for `SDPerModelsPerfMetrics()`: you can specify the defaults inplace

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
Wovchena, line 1: Update speculative_decoding_lm samples to show how to use it and add docstrings explaining when it's OK to cast `extended_perf_metrics` to `SDPerModelsPerfMetrics`

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
```diff
@@ -0,0 +1,38 @@
...
+#pragma once
+
+#include "openvino/genai/perf_metrics.hpp"
+#include "openvino/genai/visibility.hpp"
+
+
+namespace ov::genai {
+
+struct OPENVINO_GENAI_EXPORTS SDPerfMetrics : public ov::genai::PerfMetrics {
+    ov::genai::MeanStdPair ttst;  // Time to the second token (in ms).
```
Wovchena:
```suggestion
    ov::genai::MeanStdPair ttst;  // Time from the first to the second token (in ms).
```

src/cpp/src/speculative_decoding/speculative_decoding_perf_metrics.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        for (size_t i = 1; i < durations.size(); ++i) {
+            if (i == 1) {
+                // TODO: ttst should be based on time per token or latency?
```
Wovchena: I think time per token and latency are the same things here because there's just one token

Wovchena (changes requested):

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
sbalandi: Are the number of generated tokens and num_accepted_tokens the same? - no, each step produces (number of accepted tokens + 1) tokens
Is this values the same for main_model_metrics and draft_model_metrics? - it has meaning only for main model, it looks like value can be moved to SDPerModelsPerfMetrics class

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
sbalandi: removed

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
sbalandi: I have change to `Time of the generation of the second token` 
if I leave this wording here, then `Latency from the third token` should be moved to from the second, but it doesn't sound right, it sounds like time for AVG(2th, 3th .... )

src/cpp/include/openvino/genai/llm_pipeline.hpp
sbalandi: yes, I will add it

src/python/py_perf_metrics.cpp
Wovchena, line 1: py_openvino_genai.pyi is outdated and need to be regenerated with pybind11-stubgen

samples/cpp/text_generation/speculative_decoding_lm.cpp
Wovchena, line 1: Revert this file style to keep it aligned with other samples

samples/cpp/text_generation/speculative_decoding_lm.cpp
```diff
@@ -38,7 +39,31 @@ int main(int argc, char* argv[]) try {
 
     // Since the streamer is set, the results will
     // be printed each time a new token is generated.
-    pipe.generate(prompt, config, streamer);
+    auto result = pipe.generate(prompt, config, streamer);
+
+    if (result.extended_perf_metrics) {
+        auto sd_perf_metrics = std::dynamic_pointer_cast<ov::genai::SDPerModelsPerfMetrics>(result.extended_perf_metrics);
```
Wovchena:
```suggestion
    auto sd_perf_metrics = std::dynamic_pointer_cast<ov::genai::SDPerModelsPerfMetrics>(result.extended_perf_metrics);
    if (sd_perf_metrics) {
```

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -36,13 +36,17 @@ using StringInputs = std::variant<std::string, std::vector<std::string>>;
 *
 * @param tokens sequence of resulting tokens
 * @param scores sum of logarithmic probabilities of all tokens in the sequence
-* @param metrics performance metrics with tpot, ttft, etc. of type ov::genai::PerfMetrics
+* @param perf_metrics performance metrics with tpot, ttft, etc. of type ov::genai::PerfMetrics
+* @param extended_perf_metrics pipeline specific performance metrics etc. of type ov::genai::PerfMetrics.
+*        Applicable for pipelines with implemented extended metrics: SpeculativeDecoding Pipeline.
+*        To get metrics, it should be cast to corresponding class for extended perf metrics from pipeline.
```
Wovchena, line 42:
```suggestion
*        To get metrics, it should be cast to corresponding class for extended perf metrics from pipeline.
*        Cast to SDPerModelsPerfMetrics for SpeculativeDecoding.
```

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include "openvino/genai/visibility.hpp"
+
+
+namespace ov::genai {
+
+/**
+ * @brief Performance metrics for each generate call of Speculative Decoding Pipeline.
+ * It is extended version of PerfMetrics.
+ * 
+ * @param get_ttst Returns the mean and standard deviation of TTST(Time to the second token).
```
Wovchena: I suggest adding answers to the questions from the ticket to docstrings as well. https://github.com/openvinotoolkit/openvino.genai/pull/2321#discussion_r2150704824 is also useful to 

We missed that when `PerfMetrics`, but the docstrings of methods should be near these methods similar to `LLMPipleine`. It was there before, so it's optional

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
```diff
@@ -0,0 +1,62 @@
...
+ * @param get_latency Returns the mean and standard deviation of the latency(avg_latency).
+ *
+ * Cached mean and standard deviations.
+ * @param ttst Mean and standard deviation of Time to the Second Token (TTST) in milliseconds.
+ * @param avg_latency Mean and standard deviation of the latency from the third token in milliseconds per inference, which includs also prev and post processing.
+ * First and second token time is presented separately as ttft and ttst.
```
Wovchena: Add the explanation from the comment under ticket 169333. It's weird to have third token metric, so it needs to be explained.

Since this holds for greedy as well, it may be that we will be requested to add third token metric for greedy as well... I guess we could move `SDPerfMetrics`'s members to `PerfMetrics`. This is just FYI

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp
```diff
@@ -33,6 +33,8 @@ class ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl :
 
     UpdateRequestResult init_request_by_candidate(uint64_t request_id, const GeneratedSequences& candidates);
 
+    RawPerfMetrics raw_perf_metrics;
```
Wovchena, line 36: Is it allowed to run speculative decoding in multiple threads similar to greedy generation with `add_request()`? If so, it may be a bad idea to have this state

samples/cpp/text_generation/speculative_decoding_lm.cpp
sbalandi: updated

src/cpp/include/openvino/genai/llm_pipeline.hpp
sbalandi: added

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
sbalandi: comments added

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp
sbalandi: I haven't found method to get perf_metrcis in case of using add_request. Why having this state may be a bad idea ?

src/cpp/include/openvino/genai/speculative_decoding/perf_metrics.hpp
Wovchena: For GPU
> Async compilation of some opt kernels can be completed after second token. Also, additional memory manipulation can happen at second token time.

Is a nice explanation so the statment
> The second token is presented separately as it shows more time than the subsequent tokens

doesn't look out of blue. Is there a reason not to have it?

PR #2329, opened 2025-06-10
[Test][GGUF] Add DeepSeek-R1-Distill-Qwen GGUF in CI

sammysun0711:
**Details**
Add tiny dummy gguf model as proxy model for DeepSeek-R1-Distill-Qwen GGUF model in pre-commit test


tests/python_tests/data/models.py
```diff
@@ -80,4 +80,8 @@ def get_gguf_model_list():
             "gguf_model_id": "Qwen/Qwen2.5-0.5B-Instruct-GGUF",
             "gguf_filename": "qwen2.5-0.5b-instruct-q4_0.gguf"
         },
+        {
```
rkazants, line 83: Its weight is ~1.9GB. May be we should add rather to nightly validation? @Wovchena, please advice.
Or create a tiny version of it if possible?

sammysun0711: @rkazants, thanks for your feedback. 
For GGUF Q4_K_M model with both Q4_K and Q6_K tensor, we have different handling method.
1. GGUF -> OV: dequantize Q6_K tensor to f16 (WA for GPU crash), while keep Q4_K as INT4 
2. GGUF->Pytorch: dequantize both Q4_K and Q6_K tensor to f16. 
Therefore we will not get same results compare with 1 & 2 w/ Q4_K_M model. 

So the minimal model to used in CI is Q8_0 model. I agree that model size quite large and we do not host any dummy model in HF. So we can add to nightly validation if possible.

Wovchena: There's no nightly validation. The main reason is that it's hard to enforce regular nightly results monitoring.

Would it be possible to generate the model with random weights on the flight - that would be even better then downloading a dummy model, because that removes Internet connection issues.
You can generate and publish the model yourself. That's how other tiny random models were obtained.

sammysun0711: Hi @Wovchena, thanks for your suggestion, create a dummy model on-the-fly is good idea, while not easy to figure out how to set proper gguf tensor and metadata per model. 

Instead, I create a tiny dummy gguf model w/ single transformer block with minimal size (reduce from 1.89GB to 552MB). Now I update the CI test with dummy model.

PR #2331, opened 2025-06-10
Fix multinomial sampling for PromptLookupDecoding

sbalandi:
[CVS-169030](https://jira.devtools.intel.com/browse/CVS-169030)

Way of sampling in multinomial scenario made according to article [Fast Inference from Transformers via Speculative Decoding](https://arxiv.org/pdf/2211.17192) and [HF](https://github.com/huggingface/transformers/blob/626c610a4d9d36427d392e0ed70a5c7018900eba/src/transformers/generation/utils.py#L4271) and assumes that candidates have real probabilities. But generated candidates have [synthetic highest probability](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/prompt_lookup/continuous_batching_for_prompt_lookup.cpp#L77) in case of prompt lookup decoding and will always be accepted by [validation_candidates(...)](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/sampling/sampler.cpp#L696). So for this case we don't need this logic.


sbalandi:
Initial changes https://github.com/openvinotoolkit/openvino.genai/pull/1092 and https://github.com/openvinotoolkit/openvino.genai/pull/1068 .

Wovchena (approved):

PR #2336, opened 2025-06-11
[llm bench] Avoid using not supported by beam_search parameters for beam_search case

sbalandi:
[CVS-168816](https://jira.devtools.intel.com/browse/CVS-168816)

### DIFF
--- a/tools/llm_bench/task/text_generation.py
+++ b/tools/llm_bench/task/text_generation.py
@@ -248,6 +248,10 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
     gen_config.rng_seed = args["seed"]
     gen_config.num_beams = args["num_beams"]
     gen_config.do_sample = False
+    if gen_config.num_beams > 1:
+        gen_config.frequency_penalty = 0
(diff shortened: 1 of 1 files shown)


tools/llm_bench/task/text_generation.py
```diff
@@ -248,6 +248,10 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
     gen_config.rng_seed = args["seed"]
     gen_config.num_beams = args["num_beams"]
     gen_config.do_sample = False
+    if gen_config.num_beams > 1:
+        gen_config.frequency_penalty = 0
+        gen_config.presence_penalty = 0
+        gen_config.repetition_penalty = 1
```
Wovchena, line 254: These are already the defaults of `GenerationConfig`. Why is it needed?

sbalandi: llm_bench doesn't use the default generation config, llm_bench gets it from the model `gen_config = model.get_generation_config()` , baichuan2-13b-chat's generation config includes not suitable for beam search values of this parameters

Wovchena (approved):

PR #2337, opened 2025-06-11
Fix paths with unicode for tokenizers

yatarkan:
Ticket: CVS-169069


Wovchena:
For your reference. this is how it was done for argv in samples: https://github.com/openvinotoolkit/openvino/blob/c168ebba59ec796e69f0b2b927cdb28bb8d9960c/samples/cpp/common/utils/include/samples/common.hpp#L47

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -38,9 +38,17 @@ constexpr char pad_token_key_name[] = "pad_token";
 
 ov::Core core_with_extension() {
     ov::Core core;
+
+#ifdef _WIN32
+    const wchar_t* ov_tokenizer_path_w = _wgetenv(ScopedVar::ENVIRONMENT_VARIABLE_NAME_W);
```
yangwang201911, line 43: We can replace `wchat_t *` and `char *` with `std::filesystem::path` after the [PR#30938](https://github.com/openvinotoolkit/openvino/pull/30938) is merged into OpenVINO. This PR enables the `core.add_extension()` to support Unicode path via `std::filesystem::path`.

Wovchena (approved):

src/cpp/src/tokenizer/tokenizer.cpp
Wovchena: I think it's still required to specify the way to read the env var so it doesn't really matter if it's casted to wsting or path later. Feel free to submit a patch after OV's PR is merged.

PR #2341, opened 2025-06-12
[llm_bench] Remove default scheduler config

sbalandi:
Scheduler config was added initially to initialize continues batching pipeline via LLMPipeline, then it was kept for the sake of cache_size, as dynamic allocation was expected to impact performance. But It is not needed for cb pipeline now and dynamic allocation shows better performance .

### DIFF
--- a/tools/llm_bench/llm_bench_utils/ov_utils.py
+++ b/tools/llm_bench/llm_bench_utils/ov_utils.py
@@ -6,7 +6,6 @@ from transformers import AutoConfig, AutoProcessor, AutoTokenizer
 from openvino import Core
 import openvino as ov
 import logging as log
-import sys
 import time
 import json
 import types
@@ -19,8 +18,7 @@ from llm_bench_utils.config_class import (
     DEFAULT_MODEL_CLASSES,
     IMAGE_GEN_CLS,
     INPAINTING_IMAGE_GEN_CLS,
-    IMAGE_TO_IMAGE_GEN_CLS,
-    PA_ATTENTION_BACKEND
+    IMAGE_TO_IMAGE_GEN_CLS
 )
 from transformers import pipeline
 import queue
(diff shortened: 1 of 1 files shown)


tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -603,10 +597,8 @@ def create_genai_image_text_gen_model(model_path, device, ov_config, memory_moni
 
     processor_config = get_vlm_processor(model_path)
 
-    cb = ov_config.get('ATTENTION_BACKEND', '') == PA_ATTENTION_BACKEND
     cb_config = kwargs.get("cb_config")
-    if cb or cb_config is not None:
-        log.info("Continuous Batching mode activated")
+    if cb_config is not None:
```
sammysun0711, line 601: What if we would like to specify `enable_prefix_cache=False` via cb_config in llm_bench?

sammysun0711: Or we can use `--disable_prompt_permutation` option 
https://github.com/openvinotoolkit/openvino.genai/blob/23bf529766c52a05de074fd53637c321271d821b/tools/llm_bench/benchmark.py#L192

in `visual_language_generation`  task: https://github.com/openvinotoolkit/openvino.genai/blob/23bf529766c52a05de074fd53637c321271d821b/tools/llm_bench/task/visual_language_generation.py#L185

Similar to `text_generation` task
https://github.com/openvinotoolkit/openvino.genai/blob/23bf529766c52a05de074fd53637c321271d821b/tools/llm_bench/task/text_generation.py#L218

sbalandi: I don't get the question. User still can specify scheduler config for pipeline, I have deleted just default config here.

sammysun0711: Sorry, I just misunderstood that we can not set cb_config in llm_bench.

Wovchena (approved):

Issue #2333, opened 2025-06-10
Text-to-Speech sample fails with missing tokenzier error


RyanMetcalfeInt8:
Using OpenVINO GenAI release 2025.2 branch (releases/2025/2), I tried following the steps [here](https://github.com/openvinotoolkit/openvino.genai/tree/releases/2025/2/samples/cpp/speech_generation) to run the new text-to-speech sample.

But, it seems like the tokenizer was not produced during model conversion and so I received the following error:
```
>text2speech.exe speecht5_tts "Hello OpenVINO GenAI"
Check 'm_ireq_queue_tokenizer' failed at C:\Users\LNL\Workspace\GENAI\2025.2\openvino.genai\src\cpp\src\tokenizer\tokenizer.cpp:494:
Either openvino_tokenizer.xml was not provided or it was not loaded correctly. Tokenizer::encode is not available
```

Here is the output from the conversion step:
```
>optimum-cli export openvino --model microsoft/speecht5_tts --model-kwargs "{\"vocoder\": \"microsoft/speecht5_hifigan\"}" speecht5_tts
Using a slow image processor as `use_fast` is unset and a slow processor was saved with this model. `use_fast=True` will be the default behavior in v4.52, even if the model was saved with a slow processor. This will result in minor differences in outputs. You'll still be able to use a slow processor with `use_fast=False`.
Moving the following attributes in the config to the generation config: {'max_length': 1876}. You are seeing this warning because you've set generation parameters in the model config, as opposed to in the generation config.
`loss_type=None` was set in the config but it is unrecognised.Using the default loss: `ForCausalLMLoss`.
C:\Users\LNL\Workspace\GENAI\2025.2\openvino.genai-build\samples\cpp\speech_generation\Release\my_env\lib\site-packages\transformers\models\speecht5\modeling_speecht5.py:947: TracerWarning: Converting a tensor to a Python boolean might cause the trace to be incorrect. We can't record the data flow of Python values, so this value will be treated as a constant in the future. This means that the trace might not generalize to other inputs!
  if attn_weights.size() != (bsz * self.num_heads, tgt_len, src_len):
[...]

PR #2342, opened 2025-06-12
Update Export Requirements

apaniukov:
Export requirements missing sentencepiece library, that are required to load and convert original model tokenizers.
Fixes https://github.com/openvinotoolkit/openvino.genai/issues/2333

### DIFF
--- a/samples/export-requirements.txt
+++ b/samples/export-requirements.txt
@@ -1,7 +1,7 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/pre-release
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
-openvino-tokenizers~=2025.3.0.0.dev
+openvino-tokenizers[transformers]~=2025.3.0.0.dev
 optimum-intel[nncf] @ git+https://github.com/huggingface/optimum-intel.git@main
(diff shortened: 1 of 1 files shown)


Wovchena (approved):
Should the tokenizers report the issue?

apaniukov:
> Should the tokenizers report the issue?

OpenVINO Tokenziers can report the issue when used directly (the `convert_tokenizer` cli command will trigger the appropriate transformers error). The issue is in the optimum`s list of processors, which [should contain the tokenizer](https://github.com/huggingface/optimum-intel/blob/b5c35e3b0e2e312038784176537075f8581f552b/optimum/exporters/openvino/__main__.py#L557), but cannot do that because of a missing dependency.

Wovchena:
Why isn't there a warning? The code is already wrapped with try except and if else to log the problems

apaniukov:
> Why isn't there a warning? The code is already wrapped with try except and if else to log the problems

We don't know the reason why there is no tokenizer object in the processors. We can log that there is no tokenizer, but it doesn't help to debug, because optimum just silence the exception: https://github.com/huggingface/optimum/blob/d4ad6ec3c921803cf5578a253875d719626ec53b/optimum/utils/save_utils.py#L27

Wovchena:
So is it optimum or optimum-intel issue? In any case, please submit a request to them

PR #2197, opened 2025-05-12
Support multi images for vlm test

wgzintel:

PR #2344, opened 2025-06-13
vlm benchmarking multi images support for both python and cpp

dan20210809:
Add multi images input support for benchmark_vlm.py and benchmark_vlm.cpp, which is based on the existing implementation in visual_language_chat.py and visual_language_chat.cpp.


Wovchena (changes requested):
Looks like a duplicate of https://github.com/openvinotoolkit/openvino.genai/pull/2197. Can you review it instead?

dan20210809:
> Looks like a duplicate of #2197. Can you review it instead?

Sure, I am looking into #2197, seems my pr is duplicated. Let me review it thanks.

PR #2356, opened 2025-06-17
Add C API for text-to-speech pipeline with config, metrics, and test

pranaytg:
## Summary

This PR adds the initial C API for the text-to-speech (TTS) pipeline in OpenVINO GenAI.

### Key Components

- **Header Files**
  - `text2speech_pipeline.h` — main pipeline interface
  - `speech_generation_config.h` — configuration struct and setters/getters
  - `speech_generation_perf_metrics.h` — performance metric APIs

- **Source Files**
  - Implementations for each header above.

- **Tests**
  - Added test file for `text2speech_pipeline` under `tests/cpp`.

### Features

- Pipeline creation, generation, destruction
- Batch inference support
- Configurable decoding parameters (min/max length ratio, threshold)
- Optional speaker embeddings
- Result output via `text2speech_decoded_results_t`

Let me know if any adjustments are needed!

### DIFF
--- a/src/c/include/openvino/genai/c/speech_generation_config.h
+++ b/src/c/include/openvino/genai/c/speech_generation_config.h
@@ -1,0 +1,92 @@
+/**
+ * @brief C API for speech generation configuration
+ * @file speech_generation_config.h
+ */
+
+#pragma once
+
+#include "visibility.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Opaque handle for speech generation configuration
+ */
+typedef struct speech_generation_config_t* speech_generation_config_handle_t;
+
+/**
+ * @brief Creates a speech generation configuration
+ * @param[out] config Pointer to store the created configuration handle
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_create(speech_generation_config_handle_t* config);
+
+/**
+ * @brief Creates a speech generation configuration from JSON file
+ * @param[out] config Pointer to store the created configuration handle
+ * @param[in] json_path Path to the JSON configuration file
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_create_from_json(speech_generation_config_handle_t* config,
+                                                                  const char* json_path);
+
+/**
+ * @brief Destroys a speech generation configuration
+ * @param[in] config Configuration handle to destroy
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_destroy(speech_generation_config_handle_t config);
+
+/**
+ * @brief Sets the minimum length ratio
+ * @param[in] config Configuration handle
+ * @param[in] value Minimum ratio of output length to input text length
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_set_minlenratio(speech_generation_config_handle_t config, float value);
+
+/**
+ * @brief Gets the minimum length ratio
+ * @param[in] config Configuration handle
+ * @param[out] value Pointer to store the minimum length ratio
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_get_minlenratio(speech_generation_config_handle_t config, float* value);
+
+/**
+ * @brief Sets the maximum length ratio
+ * @param[in] config Configuration handle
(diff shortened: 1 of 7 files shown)


pranaytg:
Review yourself because AI just cant do anything

src/c/include/openvino/genai/c/speech_generation_config.h
```diff
@@ -0,0 +1,92 @@
...
+ * @brief Opaque handle for speech generation configuration
+ */
+typedef struct speech_generation_config_t* speech_generation_config_handle_t;
+
+/**
+ * @brief Creates a speech generation configuration
+ * @param[out] config Pointer to store the created configuration handle
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_create(speech_generation_config_handle_t* config);
```
apinge, line 24: I think it is better to use `OPENVINO_GENAI_C_EXPORTS ` 
e.g. https://github.com/openvinotoolkit/openvino.genai/blob/master/src/c/include/openvino/genai/c/llm_pipeline.h#L26
and in this way L10-L12 are not needed. 
```
// ifdef __cplusplus endif are not needed.
#ifdef __cplusplus
extern "C" {
#endif
```

src/c/src/speech_generation_config.cpp
```diff
@@ -0,0 +1,130 @@
+// Copyright (C) 2023-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "openvino/genai/c/speech_generation_config.h"
+#include "openvino/genai/speech_generation/speech_generation_config.hpp"
+
+#include <filesystem>
+#include <stdexcept>
+
+struct speech_generation_config_t {
```
apinge, line 10: I think it is better to add this declaration to the file `src/c/src/types_c.h` (and the cpp file add `#include "types_c.h"` )
to align with the other C classes.

And Also please follow the usage of opaque pointer.
For example in `type_c.h` add
```C
struct ov_genai_speech_generation_config_opaque {
    std::shared_ptr<ov::genai::SpeechGenerationConfig> object;
};
```
In the cpp file add
```C++
typedef struct ov_genai_speech_generation_config_opaque ov_genai_speech_generation_config;
```

src/c/include/openvino/genai/c/speech_generation_config.h
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * @brief Creates a speech generation configuration from JSON file
+ * @param[out] config Pointer to store the created configuration handle
+ * @param[in] json_path Path to the JSON configuration file
+ * @return 0 if successful, non-zero otherwise
+ */
+OPENVINO_GENAI_C_API int speech_generation_config_create_from_json(speech_generation_config_handle_t* config,
```
apinge, line 32: The return value should be `ov_status_e` 
Please find this definition from
https://github.com/openvinotoolkit/openvino/blob/d9c2aee59bc059797f20c26f32e3bd70b2e1ce2f/src/bindings/c/include/openvino/c/ov_common.h#L135-L163 

Also please update all the function and struct name as `ov_genai_speech_generation_config_create_from_json` (to algin with the namespace `ov::genai::`)

src/c/src/speech_generation_config.cpp
```diff
@@ -0,0 +1,130 @@
...
+        return 1;  // Invalid argument
+    }
+
+    try {
+        *config = new speech_generation_config_t{ov::genai::SpeechGenerationConfig()};
+        return 0;
+    } catch (const std::exception&) {
+        return 1;
+    }
+}
```
apinge, line 25: Please update the return values, allocation methods and function names  like this 
```C++
ov_status_e ov_genai_speech_generation_config_create(ov_genai_speech_generation_config** config) {
    if (!config) {
        return ov_status_e::INVALID_C_PARAM;  // Invalid argument
    }

    try {
        std::unique_ptr<ov_genai_speech_generation_config> _config = std::make_unique<ov_genai_speech_generation_config>();
        _config->object = std::make_shared<ov::genai::SpeechGenerationConfig>();
        *config = _config.release();

        return ov_status_e::UNKNOW_EXCEPTION;
    } catch (const std::exception&) {
        return ov_status_e::OK;
    }
}
```
Use `std::unique_ptr` and `std::make_unique` to ensure safe allocation.

src/c/include/openvino/genai/c/speech_generation_config.h
```diff
@@ -0,0 +1,92 @@
+/**
+ * @brief C API for speech generation configuration
+ * @file speech_generation_config.h
+ */
```
apinge, line 4: Please use the Copyright like
```C
// Copyright (C) 2025 Intel Corporation
```

PR #2358, opened 2025-06-18
Add MileBench validation for VLMs

l-bat:
Ticket: 161649

### DIFF
--- a/tests/python_tests/test_kv_cache_eviction.py
+++ b/tests/python_tests/test_kv_cache_eviction.py
@@ -4,17 +4,20 @@
 import sys
 import datasets
 import pytest
+import transformers
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
 from tqdm import tqdm
-
+from optimum.intel.openvino import OVModelForVisualCausalLM
 from openvino_genai import ContinuousBatchingPipeline, SchedulerConfig, GenerationConfig, CacheEvictionConfig, AggregationMode
 
 from utils.ov_genai_pipelines import PipelineType, generate_and_compare
 from utils.longbench import dataset2maxlen, evaluate, preprocess_prompt, post_process_pred
+from utils.milebench import MileBenchDataset, Eval
 from utils.constants import get_default_llm_properties
-from utils.hugging_face import download_and_convert_model
+from utils.hugging_face import download_and_convert_model, _download_and_convert_model
+from utils.network import retry_request
 from data.test_dataset import get_test_dataset
 
 
@@ -182,7 +185,7 @@ def test_dynamic_memory_allocation(params):
 
 
 @dataclass
-class LongBenchTestData:
+class BenchmarkTestData:
     subset: str
     threshold: float
     max_cache_usage_optimization_ratio: float
@@ -192,9 +195,9 @@ class LongBenchTestData:
 @pytest.mark.nightly
 @pytest.mark.parametrize("device", ["CPU", "GPU"])
 @pytest.mark.parametrize("test_struct", [
-    LongBenchTestData("samsum", 4, 1.6, 3.3),
-    LongBenchTestData("trec", 3.2, 2.0, 3.3),
-    LongBenchTestData("qasper", 5.8, 1.7, 3.6),
+    BenchmarkTestData("samsum", 4, 1.6, 3.3),
+    BenchmarkTestData("trec", 3.2, 2.0, 3.3),
+    BenchmarkTestData("qasper", 5.8, 1.7, 3.6),
 ])
 def test_optimized_generation_longbench(device, test_struct):
     seqs_per_request = 32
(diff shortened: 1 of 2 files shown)


vshampor (approved):

tests/python_tests/test_kv_cache_eviction.py
```diff
@@ -265,3 +267,98 @@ def test_optimized_generation_longbench(device, test_struct):
...
+@pytest.mark.parametrize("device", ["CPU", "GPU"])
+@pytest.mark.parametrize(
+    ("test_struct", "download_test_content"), [
+    (BenchmarkTestData("ALFRED", 0.011, 1.440, 1.574), "MileBench_part0.tar.gz"),
+    (BenchmarkTestData("MMCoQA", 0.032, 1.843, 1.620), "MileBench_part2.tar.gz"),
+    (BenchmarkTestData("WikiVQA", 0.032, 1.412, 1.527), "MileBench_part5.tar.gz"),
+    ],
+    indirect=["download_test_content"]
+)
+def test_optimized_generation_milebench(device, test_struct, download_test_content):
```
Wovchena: Should it be a part of WWB instead?

l-bat: No, because we’re measuring actual accuracy metrics (like ROUGE or multiple-choice accuracy) of the optimized model—not just its similarity to the original. This is necessary to align KVCache compression in GenAI with the POC. When the optimized model outperforms the original, WWB might still show a drop in similarity.

Wovchena: Here's another PR about multiple-choice: https://github.com/openvinotoolkit/openvino.genai/pull/2254

l-bat: The multiple-choice validation logic in PR #2254 differs from MileBench. In #2254, self.max_new_tokens = 1 assumes the correct answer is a single letter. However, MileBench extends this to match the answer under the letter to the full option text. 
I prefer to keep the original MileBench metric since we may use it in a future blog post. For now, the KVCrush blog post is in preparation and will include validation based on the LongBench evaluation (#2211)

Wovchena: The point isn't about specific metric choice but WWB being not limited to similarity score.

I remember that I approved adding this as a test, but I didn't think about WWB which looks like a more suitable place. Since you care about advertising, the following may be important for you. Features that aren't tested by validation team are treated as preview features. There are arguments between managers if preview features should be advertised at all. Adding the dataset to WWB would simplify validation enablement.

Note, you would need @sbalandi approval to extend WWB, if it's going to be done

l-bat: Do you have an example of how to measure a model's actual accuracy metrics independently, without comparing it to the original model?

Wovchena: No

alexsu52: @Wovchena, @l-bat WWB already supports VLM and does not require MileBench for VLMs validation. As far as I understand, this pr only introduces the regression test for the eviction based on milebench. There are no plans to use this in the regular validation like wwb. 

I do not support the idea of ​​adding metric calculations on the dataset in wwb, but if you are not satisfied with the fact that the code that calculates the metrics on milebench is in the tests, then you can move it to tool as a separate script and use it in the tests.

@Wovchena let's discuss it offline if you have other ideas.

Wovchena: Discussed. The current state of PR doesn't run `test_optimized_generation_milebench()`: https://github.com/openvinotoolkit/openvino.genai/actions/runs/16110319396/job/45453067484?pr=2358 I've removed `nightly` label: https://github.com/openvinotoolkit/openvino.genai/pull/2408

There are two possible ways to proceed:
1. Add full validation as a tool, collect approvals and require the validation team to run it.
2. Extend current precommit scope

Extending precommit seems to be more doable

Wovchena (approved):

Wovchena:
It's getting too slow.
https://github.com/openvinotoolkit/openvino.genai/actions/runs/16590194179/job/46927005815?pr=2358
Split the test running different parameters in different matrix elements: https://github.com/openvinotoolkit/openvino.genai/blob/7a3eb4004354e2711cd4f927f887410767032fbe/.github/workflows/linux.yml#L505

tests/python_tests/conftest.py
```diff
@@ -2,22 +2,55 @@
...
+    "3283_1447_000.tar.gz": "https://huggingface.co/datasets/facebook/multilingual_librispeech/resolve/main/data/mls_polish/train/audio/3283_1447_000.tar.gz",
+    "cmu_us_awb_arctic-wav-arctic_a0001.bin": "https://huggingface.co/datasets/Xenova/cmu-arctic-xvectors-extracted/resolve/main/cmu_us_awb_arctic-wav-arctic_a0001.bin",
+    "MileBench_part1.tar.gz": "https://huggingface.co/datasets/FreedomIntelligence/MileBench/resolve/main/MileBench_part1.tar.gz",
+    "MileBench_part2.tar.gz": "https://huggingface.co/datasets/FreedomIntelligence/MileBench/resolve/main/MileBench_part2.tar.gz",
+}
+
+
 @pytest.fixture(scope="session", autouse=True)
-def setup_and_teardown():
+def setup_and_teardown(request, tmp_path_factory):
```
mryzhov:
```suggestion
def setup_and_teardown(cache, tmp_path_factory):
```

tests/python_tests/conftest.py
```diff
@@ -2,22 +2,55 @@
...
+    hf_home = os.environ.get("HF_HOME", ov_cache)
+    test_data = os.path.normpath(os.path.join(hf_home, "test_data"))
+
+    logger.info(f"Creating directories: {models_dir} and {test_data}")
+    os.makedirs(models_dir, exist_ok=True)
+    os.makedirs(test_data, exist_ok=True)
+
+    request.config.cache.set("OV_CACHE", str(ov_cache))
+    request.config.cache.set("MODELS_DIR", str(models_dir))
+    request.config.cache.set("TEST_DATA", str(test_data))
```
mryzhov:
```suggestion
    cache.set("OV_CACHE", str(ov_cache))
    cache.set("MODELS_DIR", str(models_dir))
    cache.set("TEST_DATA", str(test_data))
```

mryzhov (approved):

tests/python_tests/conftest.py
l-bat: Failed: ScopeMismatch: You tried to access the function scoped fixture cache with a session scoped request object. Requesting fixture stack:
conftest.py:32:  def setup_and_teardown(cache, tmp_path_factory)

PR #2360, opened 2025-06-18
Align PromptLookupDecoding with greedy when dynamic_split_fuse works

sbalandi:
[CVS-169291](https://jira.devtools.intel.com/browse/CVS-169291)

when dynamic_split_fuse is true and max_num_batched_tokens less them prompt len, then PromptLookupDecoding Pipeline starts to generate candidates before full prompt will be put to cache, which causes to difference with greedy pipeline, let's avoid that

### DIFF
--- a/tests/python_tests/test_continuous_batching.py
+++ b/tests/python_tests/test_continuous_batching.py
@@ -4,6 +4,7 @@
 import os
 import pytest
 import math
+import sys
 
 from pathlib import Path
 from shutil import rmtree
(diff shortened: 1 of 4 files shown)


tests/python_tests/test_continuous_batching.py
```diff
@@ -414,6 +415,30 @@ def test_preemption_with_multinomial_n_seq(dynamic_split_fuse):
                          scheduler_config=scheduler_config)
 
 
+@pytest.mark.parametrize("pipeline_type", [PipelineType.PROMPT_LOOKUP_DECODING])
+@pytest.mark.precommit
+def test_dynamic_split_fuse_works(pipeline_type):
```
Wovchena:
```suggestion
def test_dynamic_split_fuse_doesnt_affect_generated_text(pipeline_type):
```

sbalandi: renamed

PR #2362, opened 2025-06-19
Remove misspelled CMAKE_CURRENT_SOUCE_DIR

Wovchena:


.github/workflows/windows.yml
```diff
@@ -23,7 +23,7 @@ env:
   CMAKE_C_COMPILER_LAUNCHER: ccache
   CCACHE_MAXSIZE: 500Mi
   HF_HOME: C:/mount/caches/huggingface/win
-  OV_CACHE: C:/mount/caches/huggingface/.ov_cache/win
+  OV_CACHE: C:/mount/caches/huggingface/.ov_cache/win/0acefcb26357d839d9cb6aae74a6e7c5ad6e18a4
```
rkazants, line 26: do we need this change? what is `0acefcb26357d839d9cb6aae74a6e7c5ad6e18a4`?

Wovchena: The old cache is broken. See the run for the previous commit. The used hash is arbitrary, its the hash of the failing commit

PR #2376, opened 2025-06-24
Fix for cloning NPU Image Generation pipelines

dkalinowski:
Pipeline cloning for multi-threaded usage had a few bugs:
- generation config was not copied to newly cloned instances, causing n/h/w/guidance_scale to not propagate further
- BS1 inference classes clone/copy had issue with default constructors which caused m_requests to duplicate and grow over time

This change was tested with SD v1.5 with `text2image_concurrent` sample, and this change will enable OpenVINO Model Server with NPU enablement for Image Generation use cases.

CVS-168932


samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -22,7 +22,23 @@ int32_t main(int32_t argc, char* argv[]) try {
...
     // Prepare initial pipeline and compiled models into device
-    pipelines.emplace_back(models_path, device);
+    ov::AnyMap properties;
+    if (device == "NPU") {
+        pipelines.emplace_back(models_path);
+        pipelines.back().reshape(/*N=*/1, /*H=*/512, /*W=*/512, /*guidance_scale=*/7.5f);
+        pipelines.back().compile(device);  // All models are compiled for NPU
+        // pipelines.back().compile("NPU", "NPU", "GPU");  // Compile for NPU and GPU, if needed
+
+        // We cannot specify N, H, W, and guidance_scale in the properties map
```
Wovchena:
```suggestion
        // Don't specify N, H, W, and guidance_scale in the properties map because they were made static
```

samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -22,7 +22,23 @@ int32_t main(int32_t argc, char* argv[]) try {
         prompts.push_back(argv[i]);
 
     // Prepare initial pipeline and compiled models into device
-    pipelines.emplace_back(models_path, device);
+    ov::AnyMap properties;
+    if (device == "NPU") {
+        pipelines.emplace_back(models_path);
+        pipelines.back().reshape(/*N=*/1, /*H=*/512, /*W=*/512, /*guidance_scale=*/7.5f);
```
Wovchena: define variables with the corresponding names. That way compiler can track them

samples/cpp/image_generation/text2image_concurrency.cpp
```diff
@@ -22,7 +22,23 @@ int32_t main(int32_t argc, char* argv[]) try {
         prompts.push_back(argv[i]);
 
     // Prepare initial pipeline and compiled models into device
-    pipelines.emplace_back(models_path, device);
+    ov::AnyMap properties;
+    if (device == "NPU") {
```
likholat, line 26: I think we need the same for img2img_concurrency sample

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: done

samples/cpp/image_generation/text2image_concurrency.cpp
dkalinowski: done

Wovchena (approved):

src/cpp/src/image_generation/models/unet_inference_static_bs1.hpp
```diff
@@ -15,14 +15,22 @@ namespace genai {
 // Static Batch-Size 1 variant of UNetInference
 class UNet2DConditionModel::UNetInferenceStaticBS1 : public UNet2DConditionModel::UNetInference {
 public:
+    UNetInferenceStaticBS1() : UNetInference(), m_native_batch_size(0) {}
+
+    UNetInferenceStaticBS1(const UNetInferenceStaticBS1&) = delete;
+    UNetInferenceStaticBS1(UNetInferenceStaticBS1&&) = delete;
+
+    UNetInferenceStaticBS1(UNetInferenceStaticBS1& other)
```
likholat: We should disable direct copying for user, the `clone()` method must be the only way to create a copy. 
Let's move this constructor to the private section

dkalinowski: done

samples/cpp/image_generation/image2image_concurrency.cpp
```diff
@@ -25,8 +25,28 @@ int32_t main(int32_t argc, char* argv[]) try {
...
+        //pipelines.back().compile("NPU", "NPU", "GPU");  // Compile for NPU and GPU, if needed
+
+        // Don't specify N, H, W, and guidance_scale in the properties map because they were made static
+        properties = ov::AnyMap{ov::genai::strength(0.8f),
+                                ov::genai::num_inference_steps(2)};
+    } else {
+        pipelines.emplace_back(models_path, device);
+
+        properties = ov::AnyMap{ov::genai::strength(0.8f),  // controls how initial image is noised after being converted to latent space. `1` means initial image is fully noised
+                                ov::genai::num_inference_steps(2)};
```
Wovchena: `num_inference_steps` was 4 for that sample (and 2 for samples/cpp/image_generation/text2image_concurrency.cpp). I'd propose to make it 4 for samples/cpp/image_generation/image2image_concurrency.cpp. Unless @likholat says the opposite

likholat: @Wovchena is right, we need to do at least 4 steps for FLUX.1-schnell to get a decent image, and more steps for other models

likholat (approved):

PR #2378, opened 2025-06-25
[Continuous Batching] Fix multi-threads safety issue about variable sharing & avoid the denominator is 0 issue in statistic part

zhaixuejun1993:
1) For the thread control variable "finishThread", using the standard method change the value to keep the safety;
2) For the total duration time in second will be return 0, when the execution is very small, which will make some unexpected behaviors (hang in std::cout). So will get the ms duration and than transform to s and get the final statistic results.

### DIFF
--- a/tools/continuous_batching/benchmark/continuous_batching_benchmark.cpp
+++ b/tools/continuous_batching/benchmark/continuous_batching_benchmark.cpp
@@ -300,7 +300,7 @@ public:
         mean_ttft /= generations_info.size();
         mean_tpot /= generations_info.size();
 
-        auto total_duration_second = total_duration.count() / 1000.0;
+        auto total_duration_second = std::chrono::duration<double>(total_duration).count();
         std::cout << "Benchmark duration: " << total_duration_second << " s" << std::endl;
         std::cout << "Total number of input tokens: " << total_input_len << std::endl;
         std::cout << "Total number of output tokens: " << total_output_len << std::endl;


tools/continuous_batching/benchmark/continuous_batching_benchmark.cpp
```diff
@@ -298,11 +299,13 @@ class GenerationInfoCollector {
         }
         mean_ttft /= generations_info.size();
         mean_tpot /= generations_info.size();
-        std::cout << "Benchmark duration: " << total_duration.count() << " s" << std::endl;
+
+        auto total_duration_second = total_duration.count() / 1000.0;
+        std::cout << "Benchmark duration: " << total_duration_second << " s" << std::endl;
```
Wovchena, line 304: How could `cout` hang in
> For the total duration time in second will be return 0, when the execution is very small, which will make some unexpected behaviors (hang in std::cout). So will get the ms duration and than transform to s and get the final statistic results.

zhaixuejun1993: u can try this line, which will cause some unexpected behavior
std::cout << "tmp = " << 10 / 0 << std::endl;

Wovchena: It's just division by zero. `cout` isn't involved. Copilot's suggestion with double should fix it. Please, apply it

tools/continuous_batching/benchmark/continuous_batching_benchmark.cpp
```diff
@@ -298,11 +299,13 @@ class GenerationInfoCollector {
         }
         mean_ttft /= generations_info.size();
         mean_tpot /= generations_info.size();
-        std::cout << "Benchmark duration: " << total_duration.count() << " s" << std::endl;
+
+        auto total_duration_second = std::chrono::duration<double>(total_duration).count();
+        std::cout << "Benchmark duration: " << total_duration_second << " s" << std::endl;
```
Wovchena, line 304: `total_duration` is a result of `std::chrono::duration_cast<std::chrono::milliseconds>()` so its type is `milliseconds`. `std::chrono::duration<double>(total_duration).count()` gives the number of milliseconds but it's reported as seconds

PR #2385, opened 2025-06-26
[JS] Implement TextEmbeddingPipeline in JS API

Retribution98:
## Details:
- Implemented TextEmbeddingPipeline for JS API
- Supported `Config` and `properties` for TextEmbeddingPipeline 
- Added `text_embeddings.js` sample
## Tickets:
- CVS-168622


Wovchena:
We'll need to wait for @as-suvorov to come back from his vacation

src/js/src/text_embedding_pipeline/pipeline_wrapper.cpp
```diff
@@ -0,0 +1,74 @@
...
+            InstanceMethod("embedDocumentsSync", &TextEmbeddingPipelineWrapper::embed_documents),
+            InstanceMethod("embedQuery", &TextEmbeddingPipelineWrapper::embed_query_async),
+            InstanceMethod("embedDocuments", &TextEmbeddingPipelineWrapper::embed_documents_async),
+        }
+    );
+}
+
+Napi::Value TextEmbeddingPipelineWrapper::init(const Napi::CallbackInfo& info) {
+    Napi::Env env = info.Env();
+    const std::string model_path = info[0].ToString();
```
almilosz, line 25: Where `info[0]` is validated? Is it somewhere in typescript layer?

samples/js/rag/text_embeddings.js
```diff
@@ -0,0 +1,29 @@
...
+main();
+
+async function main() {
+    const modelPath = process.argv[2];
+    const texts = process.argv.slice(3);
+
+    if (!modelPath) {
+        console.error('Please specify path to model directory\n'
+                    + `Run command must be: 'node ${basename(process.argv[1])} *path_to_model_dir* *prompt*'`);
+        process.exit(1);
```
as-suvorov: Is there a specific reason for using process.exit over throwing an error?
[Docs suggests](https://nodejs.org/api/process.html#process_process_exit_code):
> If it is necessary to terminate the Node.js process due to an error condition, throwing an uncaught error and allowing the process to terminate accordingly is safer than calling process.exit().

samples/js/rag/text_embeddings.js
```diff
@@ -0,0 +1,29 @@
...
+
+main();
+
+async function main() {
+    const modelPath = process.argv[2];
+    const texts = process.argv.slice(3);
+
+    if (!modelPath) {
+        console.error('Please specify path to model directory\n'
+                    + `Run command must be: 'node ${basename(process.argv[1])} *path_to_model_dir* *prompt*'`);
```
as-suvorov: Maybe make sense to align usage messages between samples
```suggestion
                    + `Usage: node ${basename(process.argv[1])} <MODEL_DIR> '<TEXT 1>' ['<TEXT 2>' ...]`);
```

samples/js/rag/text_embeddings.js
```diff
@@ -0,0 +1,29 @@
...
+        process.exit(1);
+    }
+    if (!texts) {
```
as-suvorov:
```suggestion
    if (!texts.length) {
```
Maybe it's enough just to check number of arguments like we do in `cpp` samples:
```js
if (!process.argv.length < 4) {
  throw new Error(`Usage: node ${basename(process.argv[1])} <MODEL_DIR> '<TEXT 1>' ['<TEXT 2>' ...]`)
}
```

src/js/CMakeLists.txt
```diff
@@ -44,6 +44,10 @@ add_library(${PROJECT_NAME} SHARED
     ${CMAKE_CURRENT_SOURCE_DIR}/src/llm_pipeline/finish_chat_worker.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/llm_pipeline/start_chat_worker.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/llm_pipeline/init_worker.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/text_embedding_pipeline/init_worker.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/text_embedding_pipeline/embed_query_worker.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/text_embedding_pipeline/embed_documents_worker.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/text_embedding_pipeline/pipeline_wrapper.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/helper.cpp
```
as-suvorov: Could you please try to use: `"${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp"`

src/js/lib/pipelines/textEmbeddingPipeline.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    constructor(modelPath: string, device: string, config?: TextEmbeddingConfig, ovProperties?: object) {
+        this.modelPath = modelPath;
+        this.device = device;
+        this.config = config || {};
+        this.ovProperties = ovProperties || {};
+    }
+
+    async init() {
+        if (this.isInitialized)
```
as-suvorov: `this.isInitialized` seems redundant
```suggestion
        if (this.pipeline)
```

src/js/lib/pipelines/textEmbeddingPipeline.ts
```diff
@@ -0,0 +1,67 @@
...
+        const initPromise = util.promisify(this.pipeline.init.bind(this.pipeline));
+        const result = await initPromise(this.modelPath, this.device, this.config, this.ovProperties);
+    
+        this.isInitialized = true;
+    
+        return result;
+    }
+
+    embedDocumentsSync(texts: string[]): EmbeddingResults {
+        if (this.pipeline === null ) throw new Error('Pipeline is not initialized');
```
as-suvorov: nitpick. Here and below
```suggestion
        if (!this.pipeline) throw new Error('Pipeline is not initialized');
```

src/js/lib/addon.ts
```diff
@@ -1,9 +1,39 @@
...
+    documents: string[],
+    callback: (err: NodeJS.ErrnoException | null, value: EmbeddingResults) => void,
+  ): void;
+  embedQuerySync(text: string): EmbeddingResult;
+  embedDocumentsSync(documents: string[]): EmbeddingResults;
+}
+
+interface OpenVINOGenAIAddon {
+  TextEmbeddingPipeline: TextEmbeddingPipelineWrapper,
+  LLMPipeline: any,
```
as-suvorov, line 54: Should we have an interface for `LLMPipeline`?

src/js/lib/utils.ts
```diff
@@ -108,3 +108,28 @@ export type GenerationConfig = GenericGenerationConfig
...
+/**
+     * Pooling strategy
+     */
+export enum PoolingType {
+    /** First token embeddings */
+    CLS,
+    /** The average of all token embeddings */
+    MEAN,
+}
+export type TextEmbeddingConfig = {
```
as-suvorov: Should we keep text embedding types near the pipeline itself in `textEmbeddingPipeline.ts`?

as-suvorov:
Linting is failing in CI. There is also an issue with samples test run: https://github.com/openvinotoolkit/openvino.genai/actions/runs/15899667454/job/44840048865?pr=2385#step:13:260.

src/js/src/text_embedding_pipeline/pipeline_wrapper.cpp
almilosz: Guess it'll be here:
https://github.com/Retribution98/openvino.genai/blob/23096cdb646692f88b616f7cc148fc3f56205777/src/js/lib/pipelines/textEmbeddingPipeline.ts#L33

samples/js/rag/text_embeddings.js
Retribution98: It's correct. In this case, there is no difference in behavior if it is changed to throw an exception. The `process.exit(1)` is used to make the results more readable for the user.

samples/js/rag/text_embeddings.js
Retribution98: Done

samples/js/rag/text_embeddings.js
Retribution98: These conditions are equivalent for JavaScript. I used your suggestion to make it more readable.

src/js/CMakeLists.txt
Retribution98: It was changed using GLOB_RECURSE

src/js/lib/pipelines/textEmbeddingPipeline.ts
Retribution98: I agree, it is fixed.

src/js/lib/pipelines/textEmbeddingPipeline.ts
Retribution98: Thank you! That will be stricter.

src/js/lib/addon.ts
Retribution98: Yes, but not in this PR :)

src/js/lib/utils.ts
Retribution98: These types cannot be placed in `textEmbeddingPipeline.ts` because it is used in `addon.ts`. However, I moved it into the last one.

.github/workflows/linux.yml
```diff
@@ -834,18 +834,6 @@ jobs:
         working-directory: ${{ env.SRC_DIR }}/src/js
         run: npm test
 
-      - name: Install openvino-genai-node samples dependencies
-        working-directory: ${{ env.SRC_DIR }}/samples/js/text_generation
-        run: |
-          npm install ${{ env.SRC_DIR }}/src/js
```
as-suvorov, line 840: `openvino-genai-node` is a dev dependency in `package.json`, I think it should be moved to a normal dependency. Let's do it in another PR.

samples/js/rag/text_embeddings.js
as-suvorov: I don't see how it makes the results more readable. Let's merge this PR as is anyway.

as-suvorov (approved):

as-suvorov:
Unresolved copilot comments are valid, could you please fix them?

samples/js/rag/text_embeddings.js
Retribution98: If an exception is thrown, an error message with a deep stack trace is written to the console. It also doesn't highlight anything in a different color. If exit is used, only a message about incorrect parameters is shown. In this case, the user won't see any unnecessary information.

as-suvorov: Thrown error message written to `stderr`. It should be highlighted the same as `console.error`.
Long stack trace looks not great, I agree. And `argparse` implements the same approach with `process.exit`. So let's keep it as is.

as-suvorov:
I've tried to run nodejs bindings. `npm install` fails with error:
```
npm error sha512-27yArdPDLSvoLR3vXxVRSXfoIOlmK70A/jlTngL6X/TVBEYk+OCEQc07BHCqAsVqgNLalxuStHE9PyD4HZtQjg== integrity checksum failed when using sha512: wanted sha512-27yArdPDLSvoLR3vXxVRSXfoIOlmK70A/jlTngL6X/TVBEYk+OCEQc07BHCqAsVqgNLalxuStHE9PyD4HZtQjg== but got sha512-nAsC/FHy0XRI7T0iPs+HWrvJkTVR69otRPd2OY+71o8Wf/ebGuVaNOWtSXb8UTnCTpvjijMWpAugGyZNLxqU7w==. (11728 bytes)
```

`package-lock.json` regeneration helps. Could you please check if we need to update lock file?

Retribution98:
> `package-lock.json` regeneration helps. Could you please check if we need to update lock file?

Yes, I see.  Thank you, I will update it in another PR

PR #2387, opened 2025-06-26
Fix missing use cases for inpainting models and defining use case with relative path

sbalandi:
Task CVS-169536


tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -260,6 +261,7 @@ def resolve_complex_model_types(config):
 def get_model_name(model_name_or_path):
     # try to get use_case from model name
     path = os.path.normpath(model_name_or_path)
+    path = os.path.abspath(path)
```
Wovchena: Does it make sense to have normpath and abspath? https://docs.python.org/3/library/os.path.html#os.path.abspath

sbalandi: no, removed, thank you

Wovchena (approved):

PR #2388, opened 2025-06-26
Fix test_vlm_npu_no_exception

AlexanderKalistratov:
* Previous implementation was running only single model (`model_ids[0]`)
* Increase max context size to pass tests for `phi-4`.
* Skip `internvl` as it fails

### DIFF
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -387,7 +387,6 @@ def test_perf_metrics(cache, backend):
 
     start_time = perf_counter_ns()
     pipe = VLMPipeline(models_path, "CPU", ATTENTION_BACKEND=backend)
-    
     start_generate = perf_counter_ns()
     result = pipe.generate(
         prompts[0],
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_vlm_pipeline.py
```diff
@@ -442,21 +441,30 @@ def test_perf_metrics(cache, backend):
...
 def test_vlm_npu_no_exception(model_id, backend):
-    models_path = get_ov_model(model_ids[0])
+    unsupported_models = model_ids = {
+        "katuni4ka/tiny-random-internvl2",
+    }
+
+    if model_id in unsupported_models:
+        pytest.skip(f"{model_id} is not supported")
+
+    models_path = get_ov_model(model_id)
```
AsyaPronina, line 458: 👍 Thanks!

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -442,21 +441,30 @@ def test_perf_metrics(cache, backend):
...
+@pytest.mark.parametrize("model_id", model_ids)
 @pytest.mark.parametrize("backend", attention_backend)
 @pytest.mark.skipif(
     sys.platform == "darwin" or platform.machine() in ["aarch64", "arm64", "ARM64"],
     reason="NPU plugin is available only on Linux and Windows x86_64",
 )
 def test_vlm_npu_no_exception(model_id, backend):
-    models_path = get_ov_model(model_ids[0])
+    unsupported_models = model_ids = {
+        "katuni4ka/tiny-random-internvl2",
```
AsyaPronina, line 452: what about ` katuni4ka/tiny-random-qwen2.5-vl` from the comment above?

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -442,21 +441,30 @@ def test_perf_metrics(cache, backend):
...
+    }
```
Wovchena, line 453: `model_ids` is overridden. Is it intentional?

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -442,21 +441,30 @@ def test_perf_metrics(cache, backend):
...
+
+    models_path = get_ov_model(model_id)
     properties = {
         "DEVICE_PROPERTIES": {
             "NPU": {"NPUW_DEVICES": "CPU", "NPUW_ONLINE_PIPELINE": "NONE"}
         }
     }
 
+    if "phi-4" in model_id:
+        properties["DEVICE_PROPERTIES"]["NPU"]["MAX_PROMPT_LEN"] = 2048
```
Wovchena: Would it be better to set it for all cases? That way it's easier to keep in mind what the test does because there would be no exceptions.

Add a comment that it's needed because katuni4ka/tiny-random-phi-4-multimodal generates more tokens per image

Wovchena (changes requested):

tests/python_tests/test_vlm_pipeline.py
AlexanderKalistratov: No, thanks!

tests/python_tests/test_vlm_pipeline.py
AlexanderKalistratov: I've run it several times and it is passed all the times.

tests/python_tests/test_vlm_pipeline.py
AlexanderKalistratov: `MAX_PROMPT_LEN` increases execution time, so I worried it may affect test run time. But I've done some measurements and didn't find significant effect of MAX_PROMPT_LEN on total time of `test_vlm_npu_no_exception`.
So I've set it to `2048` for all models.
Thanks!

Wovchena (approved):

PR #2392, opened 2025-06-30
[GGUF] Revert GGUF WA for GPU

sammysun0711:
**Details:**
Revert GGUF Reader WA for OV GPU plugin: https://github.com/openvinotoolkit/openvino.genai/pull/2110

**Ticket:** 
CVS-169891


Wovchena:
The same test that was skipped for Mac fails for Linux. Maybe resolving it for Linux fixes it for Mac as well...

sammysun0711:
> The same test that was skipped for Mac fails for Linux. Maybe resolving it for Linux fixes it for Mac as well...

Although same test `test_full_gguf_pipeline` failed on Mac and Linux. But the issue are different. 

- Skipped Mac test failed with segment fault with 2nd test model
https://github.com/openvinotoolkit/openvino.genai/blob/6a620fb83cda329a286efa184ab7239eb417de23/tests/python_tests/data/models.py#L80-L81 
- Current Linux test case failed with 3rd test model due to output mismatch w/ HF model
https://github.com/openvinotoolkit/openvino.genai/blob/6a620fb83cda329a286efa184ab7239eb417de23/tests/python_tests/data/models.py#L84-L85

Will take a look for proper fix the test case.

src/cpp/src/gguf_utils/gguf_modeling.cpp
```diff
@@ -143,9 +132,8 @@ std::shared_ptr<ov::Model> create_language_model(
         model->set_rt_info(ov::element::f16, {"runtime_options", ov::hint::kv_cache_precision.name()});
     }
     model->set_rt_info(8.0f, {"runtime_options", ov::hint::activations_scale_factor.name()});
-    // CVS-166554: Dynamic quatnization enabled by default with gourp size 32 on MTL platfrom cause the runtime issue
-    // Apply WA to disable dynamic quantization with rt_info to fix GPU plugin issue on MTL
-    model->set_rt_info(0, {"runtime_options", ov::hint::dynamic_quantization_group_size.name()});
+    // Dummy GGUF models in CI are sensitive to dynamic quantization, set default group size 64 for dynamic quantization as default value
+    model->set_rt_info(64, {"runtime_options", ov::hint::dynamic_quantization_group_size.name()});
```
Wovchena: Will that affect all models, not just dummy models from tests?

sammysun0711: Yes, it will affect all models, instead, we can modify python test to apply this configuration specific for dummy deepseek GGUF model used in CI in pipeline level. Do you think if it works better?

Wovchena: Yep. There are user ways to specify dynamic_quantization_group_size so we shouldn't override them in GenAI silently. Moving that to test resolves the issue.

sammysun0711: Update the CI gguf test to specify dynamic_quantization_group_size based on gguf model from pipeline level.

sammysun0711:
@Wovchena, updated GGUF CI tests and all test passed, local GPU test passed, can we merge this PR?

Wovchena (approved):

PR #2395, opened 2025-06-30
LoRA lm_head and embed_tokens constants support

likholat:
Continuation of [1928](https://github.com/openvinotoolkit/openvino.genai/pull/1928)
Ticket: [CVS-159931](https://jira.devtools.intel.com/browse/CVS-159931)

### DIFF
--- a/src/cpp/src/lora/adapter.cpp
+++ b/src/cpp/src/lora/adapter.cpp
@@ -1,6 +1,8 @@
 // Copyright (C) 2023-2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+//replace if to add ?
+
 #include <algorithm>
 #include <set>
 #include <map>
@@ -478,7 +480,7 @@ struct LoRAStateGetterForConst : public BaseStateGetter {
         if (auto params = getter(name)) {
             // FIXME: Potential name conflict if LoRA is applied multiple times by using this infrastructure independently each time (not a recommended approach).
             // TODO: Check for name collisions searching for existing variables with the same names.
-            std::string variable_id_prefix = "lora_constant_" + std::to_string(model->get_sinks().size()) + name;
+            std::string variable_id_name = "lora_constant_" + std::to_string(model->get_sinks().size()) + "_" +name;
             LoRAConstantNode result;
             ov::op::util::VariableInfo variable_info;
 
@@ -486,7 +488,7 @@ struct LoRAStateGetterForConst : public BaseStateGetter {
             variable_info = ov::op::util::VariableInfo{
                 params->tensor->get_output_shape(0),
                 params->tensor->get_output_element_type(0),
-                variable_id_prefix
+                variable_id_name
             };
             result.tensor = add_variable(variable_info);
             variable_ids.emplace(name, variable_info);
@@ -499,7 +501,7 @@ struct LoRAStateGetterForConst : public BaseStateGetter {
 
     NodePtr create_if_input() {
         auto variable_info = ov::op::util::VariableInfo{
-            ov::Shape{1},  // Will be used with transpose_b == true
+            ov::Shape{1},
             ov::element::Type_t::boolean,
             if_variable_id
         };
@@ -613,11 +615,20 @@ protected:
     bool apply(NodePtr node, const LoRAConstantNode& lora_weight) override {
         auto consumers = node->get_output_target_inputs(0);
 
+        // std::cout << lora_weight.tensor->get_output_shape(0) << std::endl;
+        // std::cout << node->get_output_shape(0) << std::endl;
+
         std::shared_ptr<ov::Model> then_body = std::make_shared<ov::Model>(lora_weight.tensor, ov::ParameterVector{}),
                                    else_body = std::make_shared<ov::Model>(node, ov::ParameterVector{});
         std::shared_ptr<ov::op::v8::If> if_node = std::make_shared<ov::op::v8::If>(if_input);
         if_node->set_then_body(then_body);
         if_node->set_else_body(else_body);
+        if_node->set_output(then_body->get_results()[0], else_body->get_results()[0]);
+        
+        // std::cout << if_node->outputs().size() << std::endl;
+
+        // std::cout << if_node->output(0).get_shape() << std::endl;
+
 
         for (auto& consumer : consumers) {
             consumer.replace_source_output(if_node->output(0));
@@ -1344,6 +1355,25 @@ struct AdapterControllerImpl {
             lora_indices.A = state_name_to_index.at(lora_var_ids.second.A.variable_id);
             lora_indices.B = state_name_to_index.at(lora_var_ids.second.B.variable_id);
             set_lora_tensors(state, lora_var_ids.first, lora_var_ids.second, lora_indices, weight_getters, alpha_only);
+        }
+
+        for (const auto& constant_variable_id : constant_variable_ids) {
+            auto b = constant_variable_id.second.variable_id;
+            // variable_info
+            if (constant_variable_id.first.find("const") != std::string::npos) {
+                auto a = std::vector<bool>{true};
+                ov::Tensor if_input_true = ov::Tensor(ov::element::boolean, ov::Shape(1), a);
+                // ov::Tensor if_input_true = ov::Tensor(ov::element::boolean, ov::Shape{1});
+                size_t lora_index = state_name_to_index.at(b);
+                // NodePtr node = const_getter[0](constant_variable_id.first)->tensor;
+                // // auto a = std::dynamic_pointer_cast<ov::op::v0::Constant>(node);
+                state[lora_index].set_state(if_input_true);
+            }
+            size_t lora_index = state_name_to_index.at(b);
+            NodePtr node = const_getter[0](constant_variable_id.first)->tensor;
+            auto a = std::dynamic_pointer_cast<ov::op::v0::Constant>(node);
+            state[lora_index].set_state(a->get_tensor_view());
+            // set_lora_tensors(state, constant_variable_id.first, constant_variable_id.second, lora_indices, weight_getters, alpha_only);
         }
     }
 
--- a/src/cpp/src/lora/adapter.cpp
+++ b/src/cpp/src/lora/adapter.cpp
@@ -1068,6 +1068,7 @@ public:
         : tensors(group_lora_tensors(safetensor_to_constant_map(safetensor), default_lora_patterns())) {}
 
     const LoRATensors& get_tensors() const override {
+        int i = 0;
         return tensors;
     }
 
@@ -1337,7 +1338,7 @@ struct AdapterControllerImpl {
         set_new_adapter_tensors(infer_request, /*alpha_only=*/true);
     }
 
-    void set_new_adapter_tensors (ov::InferRequest& infer_request, bool alpha_only = false) {
+    void set_new_adapter_tensors(ov::InferRequest& infer_request, bool alpha_only = false) {
         std::cout << "!!!!!!!!!!!!!!set_new_adapter_tensors"<< std::endl;
         if(current_config.get_mode() != AdapterConfig::MODE_AUTO && current_config.get_mode() != AdapterConfig::MODE_DYNAMIC && current_config.get_mode() != AdapterConfig::MODE_STATIC_RANK ) {
             return;
@@ -1409,7 +1410,7 @@ struct AdapterControllerImpl {
 
                         // Copy data
                         const_tensor = ov::Tensor(constant_node->get_element_type(), constant_node->get_shape());
-                        std::memcpy(const_tensor.data(), constant_node->get_data_ptr(), сonst_tensor.get_byte_size());
+                        std::memcpy(const_tensor.data(), constant_node->get_data_ptr(), const_tensor.get_byte_size());
                     } else {
                         OPENVINO_THROW("LoRA constant tensor not found for name: ", const_name);
                     }
@@ -1754,6 +1755,7 @@ AdapterConfig& AdapterConfig::add(const Adapter& adapter, float alpha) {
 
 
 AdapterConfig& AdapterConfig::add(const Adapter& adapter) {
+    int i = 0;
     return add(adapter, 1);
 }
 
--- a/src/cpp/src/lora/adapter.cpp
+++ b/src/cpp/src/lora/adapter.cpp
@@ -68,7 +68,7 @@ using ConstantVector = std::vector<std::shared_ptr<v0::Constant>>;
 
 // Holds usual LoRA parameters alpha, A and B of a given type.
 using LoRANode = LoRAParts<std::shared_ptr<ov::Node>>;
-using LoRAConstantNode = std::shared_ptr<ov::Node>;
+using NodePtr = std::shared_ptr<ov::Node>;
 using LoRAPartsParser = LoRAParts<std::function<std::optional<std::string>(const std::string& name)>>;
 
 // Converts Safetensors element type to OV element type. Only part of the types are supported.
@@ -172,7 +172,7 @@ LoRATensors group_lora_tensors(const ConstantMap& tensors, const LoRAPartsParser
     return result;
 }
 
-using LoRAConstantTensors = std::map<std::string, LoRAConstantNode>;
+using LoRAConstantTensors = std::map<std::string, NodePtr>;
 
 // Group constant tensors loaded from LoRA adapter file into constants
 LoRAConstantTensors group_lora_constant_tensors(const ConstantMap& tensors, const std::vector<RegexParser>& const_parsers) {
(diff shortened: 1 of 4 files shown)


src/cpp/src/lora/adapter.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright (C) 2023-2025 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+
```
Wovchena:
```suggestion
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -1452,6 +1729,7 @@ AdapterConfig& AdapterConfig::add(const Adapter& adapter, float alpha) {
 
 
 AdapterConfig& AdapterConfig::add(const Adapter& adapter) {
+    int i = 0;
```
Wovchena:
```suggestion
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -342,17 +377,45 @@ void deduce_input_output_dims(NodePtr node, ov::Dimension& input_dim, ov::Dimens
...
+        ov::Shape init_shape(shape.rank().get_length());
+        for(size_t i = 0; i < shape.size(); ++i) {
+            init_shape[i] = shape[i].get_min_length();
+        }
+        auto init = v0::Constant::create(type, init_shape, std::vector<float>(ov::shape_size(init_shape), 0));
+        auto read_value = std::make_shared<v6::ReadValue>(init, variable);
+        #else
+        auto read_value = std::make_shared<v6::ReadValue>(variable);
+        #endif
+        model->add_sinks({std::make_shared<v6::Assign>(read_value, variable)});  // FIXME: Required? -- Yes, create ticket against CPU
```
Wovchena, line 395: Do you know what ticket the comment refers to? Add it to the comment if yes

src/cpp/src/lora/adapter.cpp
```diff
@@ -342,17 +377,45 @@ void deduce_input_output_dims(NodePtr node, ov::Dimension& input_dim, ov::Dimens
...
     LoRAVarMap& variable_ids;
     // TODO: Use variable indices instead of variable_id for faster search for a state tensor
 
-    LoRAWeightStateGetter (const LoRAParametersGetter& params_getter, std::shared_ptr<ov::Model> model, LoRAVarMap& variable_ids) :
-        params_getter(params_getter), model(model), variable_ids(variable_ids) {}
+    LoRAWeightStateGetter (const LoRAParametersGetter& params_getter,
+                           std::shared_ptr<ov::Model> model,
+                           LoRAVarMap& variable_ids) :
+        params_getter(params_getter),
+        BaseStateGetter(model),
```
Wovchena:
```suggestion
        BaseStateGetter(model),
        params_getter(params_getter),
```
There's probably a compilation warning about that explaining the problem. Same for LoRAStateGetterForConst

src/cpp/src/lora/adapter.cpp
```diff
@@ -396,24 +459,53 @@ struct LoRAWeightStateGetter {
...
+        getter(getter), 
+        BaseStateGetter(model),
+        variable_ids(variable_ids) {}
+
+    std::optional<LoRAConstantNode> operator() (NodePtr node) const {
+        std::string name = node->get_friendly_name();
+        if (auto params = getter(name)) {
+            // FIXME: Potential name conflict if LoRA is applied multiple times by using this infrastructure independently each time (not a recommended approach).
+            // TODO: Check for name collisions searching for existing variables with the same names.
+            std::string variable_id_name = "lora_constant_" + std::to_string(model->get_sinks().size()) + "_" +name;
```
Wovchena:
```suggestion
            std::string variable_id_name = "lora_constant_" + std::to_string(model->get_sinks().size()) + "_" + name;
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -396,24 +459,53 @@ struct LoRAWeightStateGetter {
...
+            LoRAConstantNode result;
+            ov::op::util::VariableInfo variable_info;
+
+            // FIXME: No guarantees on ordering of state in InferRequest makes impossible using indices of variables later, forced to use variable_id instead
+            variable_info = ov::op::util::VariableInfo{
```
Wovchena:
```suggestion
            // FIXME: No guarantees on ordering of state in InferRequest makes impossible using indices of variables later, forced to use variable_id instead
            ov::op::util::VariableInfo variable_info = ov::op::util::VariableInfo{
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -396,24 +459,53 @@ struct LoRAWeightStateGetter {
...
+    std::optional<LoRAConstantNode> operator() (NodePtr node) const {
+        std::string name = node->get_friendly_name();
+        if (auto params = getter(name)) {
+            // FIXME: Potential name conflict if LoRA is applied multiple times by using this infrastructure independently each time (not a recommended approach).
+            // TODO: Check for name collisions searching for existing variables with the same names.
+            std::string variable_id_name = "lora_constant_" + std::to_string(model->get_sinks().size()) + "_" +name;
+            LoRAConstantNode result;
+            ov::op::util::VariableInfo variable_info;
+
+            // FIXME: No guarantees on ordering of state in InferRequest makes impossible using indices of variables later, forced to use variable_id instead
```
Wovchena: How can it be fixed?

src/cpp/src/lora/adapter.cpp
```diff
@@ -460,11 +552,102 @@ class LoRATransformBase : public ov::pass::MatcherPass {
...
+    size_t applied = 0; // For debug statistics only
+};
 
-private:
+class LoRAReplaceConstantTransform : public ov::pass::MatcherPass {
+public:
+    LoRAReplaceConstantTransform(const LoRAConstantByNodeGetter& getter) {
+        register_matcher(
+            std::make_shared<ov::pass::pattern::Matcher>(ov::pass::pattern::wrap_type<v0::Constant>(), this->get_type_info().name),
+            ([getter, this](ov::pass::pattern::Matcher& m) {
```
Wovchena, line 558:
```suggestion
            ([&getter, this](ov::pass::pattern::Matcher& m) {
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -959,21 +1159,37 @@ bool operator== (const Adapter& a, const Adapter& b) {
...
         for(auto const& adapter : current_config.get_adapters()) {
             auto adapter_impl = get_adapter_impl(adapter);
-            params_getter.weight_getter.push_back(LoRAWeightGetterDefault(&adapter_impl->get_tensors(), config.get_tensor_name_prefix().value_or("")));
+            if (!adapter_impl->get_constant_tensors().empty()) {
+                OPENVINO_ASSERT(const_getter.empty(), "OpenVINO.GenAI does not support several LoRA adapters with constants!");
+                const_getter_impl = std::make_shared<LoRAWeightGetterDefault<LoRAConstantNode, LoRAConstantNode>>(
+                    &adapter_impl->get_constant_tensors(),
+                    config.get_tensor_name_prefix().value_or(""));
+
+                const_getter.push_back([this](const std::string& name) {
```
Wovchena:
```suggestion
                const_getter.push_back([const_getter_impl](const std::string& name) {
```
It seems that you are interested in `const_getter_impl` and not in `this` in this lambda.

`name` is always obtained from `NodePtr::get_friendly_name()`. Move that part into the lambda changing the arg from `string` to `NodePtr`

src/cpp/src/lora/adapter.cpp
```diff
@@ -1136,9 +1385,37 @@ struct AdapterControllerImpl {
...
+                state[const_lora_index].set_state(const_tensor);
+
+            } else if (!const_getter.empty()) {
+                auto opt_lora_const = const_getter[0](const_name);
+
+                if (opt_lora_const && const_getter_impl) {
+                    const_getter_impl->used_tensors.insert(const_name);
+                }
+
+                OPENVINO_ASSERT(!const_getter.empty(), "LoRA constant tensor not found for name: ", const_name);
```
Wovchena: Move to `else`. This is under `if (!const_getter.empty())` currently so it's never triggered

src/cpp/src/lora/adapter.cpp
```diff
@@ -137,6 +140,14 @@ LoRAPartsParser default_lora_patterns () {
     );
 }
 
+// Default LoRA tensor name patterns observed in the existing LoRA weights, captures the prefix that should correspond to a layer name in the base model
```
Wovchena:
```suggestion
// Default LoRA tensor name patterns observed in the existing LoRA weights, captures the prefix that should correspond to a layer name in the base model. Example: https://hf-mirror.com/hfl/llama-3-chinese-8b-lora
```

src/cpp/src/lora/common.hpp
```diff
@@ -15,6 +15,7 @@ namespace ov {
 namespace genai {
 namespace utils {
 
+
```
Wovchena:
```suggestion
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -67,6 +69,7 @@ using ConstantVector = std::vector<std::shared_ptr<v0::Constant>>;
 
 // Holds usual LoRA parameters alpha, A and B of a given type.
 using LoRANode = LoRAParts<std::shared_ptr<ov::Node>>;
+using LoRAConstantNode = std::shared_ptr<ov::Node>;
```
Wovchena:
```suggestion
using NodePtr = std::shared_ptr<ov::Node>;
```
OV has this alias sometimes internally. Stick to it. Plus it is more aligned with what `std::shared_ptr<ov::Node>` actually is

src/cpp/src/lora/adapter.cpp
```diff
@@ -1105,16 +1341,29 @@ struct AdapterControllerImpl {
...
+        std::vector<LoRAConstantGetter> const_getter;
         const auto& adapters = current_config.get_adapters();
         weight_getters.reserve(adapters.size());
-        for(const auto& adapter: adapters) {
-            weight_getters.emplace_back(LoRAWeightGetterDefault(&get_adapter_impl(adapter)->get_tensors(), current_config.get_tensor_name_prefix().value_or("")));
+        for (const auto& adapter : adapters) {
+            auto adapter_impl = get_adapter_impl(adapter);
+            if (!adapter_impl->get_constant_tensors().empty()) {
+                OPENVINO_ASSERT(const_getter.empty(),
+                                "OpenVINO.GenAI does not support several LoRA adapters with constants!");
```
Wovchena, line 1352: What's the reason to disable that?

Change `const_getter`'s type to just `LoRAConstantGetter`. I guess if there's no request, this implementation stays, so no general but unused implementation needed. Same for `std::vector<LoRAConstantGetter> const_getter;` in constructor

src/cpp/src/lora/adapter.cpp
```diff
@@ -342,17 +377,45 @@ void deduce_input_output_dims(NodePtr node, ov::Dimension& input_dim, ov::Dimens
 
 using LoRAVarMap = std::map<std::string, LoRAVarIDs>;
 
+struct BaseStateGetter {
+    std::shared_ptr<ov::Model> model;
+
+    BaseStateGetter(std::shared_ptr<ov::Model> model) : model(model) {}
+
+    NodePtr add_variable(const ov::op::util::VariableInfo& variable_info) const {
```
Wovchena: It seems that the intention is to reuse `add_variable()` in two classes. That forced you to build an inheritance tree. Make `add_variable()` to accept `ov::Model` as the second arg and move the function to anonymous namespace. That should decouple the classes

Wovchena (changes requested):

src/cpp/src/lora/adapter.cpp
likholat: Capturing by reference breaks the behavior: `LoRA Constant replacement is applied for 0 layers`
Probably lambda outlives the `getter` parameter

src/cpp/src/lora/adapter.cpp
likholat: done

src/cpp/src/lora/adapter.cpp
likholat: > name is always obtained from NodePtr::get_friendly_name(). Move that part into the lambda changing the arg from string to NodePtr

Moving this logic into lambda and changing the argument from `std::string` to `NodePtr` would require modifying:
`using LoRAConstantGetter = std::function<std::optional<LoRAConstantNode>(const std::string&)>;`
to
`using LoRAConstantGetter = std::function<std::optional<LoRAConstantNode>(NodePtr)>;`

which will produce incompatibility here
```cpp
auto opt_lora_const = const_getter[0](const_name);
```

Resolving this would require changing
`std::map<std::string, ov::op::util::VariableInfo> constant_variable_ids;` 
to
`std::map<NodePtr node, ov::op::util::VariableInfo> constant_variable_ids;`

I think, this would make the code even more complex:
1. every time we need the name of a constant node from constant_variable_ids we need to call `node->get_friendly_name()` instead of using a simple `std::string`
2. it would also break symmetry with `LoRAVarMap variable_ids` which uses `std::string` as the key

src/cpp/src/lora/adapter.cpp
likholat: Removed this assert: it was a leftover from a previous implementation

src/cpp/src/lora/adapter.cpp
```diff
@@ -373,47 +433,76 @@ struct LoRAWeightStateGetter {
...
+    std::optional<NodePtr> operator() (NodePtr node) const {
+        std::string name = node->get_friendly_name();
+        if (auto params = getter(name)) {
+            // FIXME: Potential name conflict if LoRA is applied multiple times by using this infrastructure independently each time (not a recommended approach).
+            // TODO: Check for name collisions searching for existing variables with the same names.
+            std::string variable_id_name = "lora_constant_" + std::to_string(model->get_sinks().size()) + "_" + name;
+            NodePtr result;
+
+            // No guarantees on ordering of state in InferRequest makes impossible using indices of variables later, forced to use variable_id instead
+            ov::op::util::VariableInfo variable_info = ov::op::util::VariableInfo {
```
Wovchena, line 480:
```suggestion
            ov::op::util::VariableInfo variable_info = ov::op::util::VariableInfo{
```

src/cpp/src/lora/adapter.cpp
```diff
@@ -67,6 +68,7 @@ using ConstantVector = std::vector<std::shared_ptr<v0::Constant>>;
 
 // Holds usual LoRA parameters alpha, A and B of a given type.
 using LoRANode = LoRAParts<std::shared_ptr<ov::Node>>;
+using NodePtr = std::shared_ptr<ov::Node>;
```
Wovchena:
```suggestion
```
It turns out that it's already defined few lines above

src/cpp/src/lora/adapter.cpp
```diff
@@ -373,47 +433,76 @@ struct LoRAWeightStateGetter {
...
-        // Attempt to pre-build initialization expression with empty tensors that should discard LoRA effect by default
-        // FIXME: CPU plugin fails when there is no initialization expression is given and type is not fp32
-        ov::Shape init_shape(shape.rank().get_length());
-        for(size_t i = 0; i < shape.size(); ++i) {
-            init_shape[i] = shape[i].get_min_length();
+struct LoRAStateGetterForConst {
+    std::shared_ptr<ov::Model> model;
+    LoRAConstantGetter getter;
+    std::map<std::string, ov::op::util::VariableInfo>& variable_ids;
+    const std::string if_variable_id = "lora_state_0_replace_orig_constant";
```
Wovchena: `if_variable_id` is only used in `create_if_input()`. Move `if_variable_id` into the function so readers don't need to jump around for it

src/cpp/src/lora/adapter.cpp
Wovchena: I'm still curios if there's a specific reason to prohibit that

likholat: This check prevents overlapping constant tensors between multiple LoRA adapters — I currently don't see a way to apply two different weights to the same constant

src/cpp/src/lora/adapter.cpp
```diff
@@ -137,6 +140,15 @@ LoRAPartsParser default_lora_patterns () {
     );
 }
 
+// Default LoRA tensor name patterns observed in the existing LoRA weights, captures the prefix that should correspond
+// to a layer name in the base model. Example: https://hf-mirror.com/hfl/llama-3-chinese-8b-lora
+std::vector<RegexParser> default_lora_constant_patterns () {
+    return {
+        RegexParser("(.*).lm_head.weight", 0),
+        RegexParser("(.*).embed_tokens.weight", 0),
```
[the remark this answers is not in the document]
Wovchena: This seems to be a valid comment

src/cpp/src/lora/adapter.cpp
likholat: Done

Wovchena (approved):

PR #2398, opened 2025-07-01
Add performance metrics for StructuredOutput generation

pavel-esir:
Ticket: CVS-168960


src/cpp/src/sampling/sampler.hpp
```diff
@@ -123,6 +123,9 @@ class Sampler {
     void create_logit_processor(uint64_t request_id, const GenerationConfig& sampling_parameters, const TokenIds& prompt);
 
     std::map<size_t, int32_t> get_beam_idxs(SequenceGroup::CPtr sequence_group);
+    // pair with map with backend name and corresponding acompiler init time, and vector of compile times for each concrete grammar
```
apaniukov:
```suggestion
    // pair with map with backend name and corresponding compiler init time, and vector of compile times for each concrete grammar
```

src/cpp/include/openvino/genai/perf_metrics.hpp
```diff
@@ -49,6 +54,16 @@ struct OPENVINO_GENAI_EXPORTS MeanStdPair {
     float std;
 };
 
+/**
+* @brief Structure to store list of durations in milliseconds.
+*/
+struct OPENVINO_GENAI_EXPORTS SummaryStats {
+    float mean;
```
Wovchena, line 61: Why not `MicroSeconds`?
Does it make sense to summaries compile time? The time highly depends on what's compiled

pavel-esir: We use float as well for `MeanStdPair`, so that final user don't have to cast manually. 

Here is `SummaryStats` instead of `MeanStdPair`  because need more fine grained values.

It usually it differs highly on (yes you are right) on what is compiled and whether it's compiled for the first time. Example below shows times for the exactly the same grammar (around 1000 json fields), at first it's slow but subsequent ones after much faster because of the caching.

Other values such as tpot, thoughtput, etc. do not change that drastically. After discussion with @apaniukov we decided that it's very handy to see that hight and lows right away, without taking look into raw metrics.

image

Wovchena: What were the arguments why it's handy?
Is it reasonable to report cache retrieval as compilation?

pavel-esir: For the users of xgrammar it's all the same compile call. `m_grammar_compiler->CompileGrammar(grammar)`, caching happens inside the grammar object `std::unique_ptr<xgrammar::GrammarCompiler> m_grammar_compiler`.

In the example above `[5056.2, 10.7, 10.4, 10.5, 10.3]` (converted from microsec to ms for simplicity) if i get only mean and std It would be mean `1000 ms`, std `~2000 ms`. If we wanna know for example how much time it took to compile for the first time this numbers doesn't give us fine grained info.

But when we have min and max we can get the first initial compilation time which is max `5 000 ms`, and subsequent compilation times min `10 ms`.

Wovchena (approved):

.github/workflows/linux.yml
```diff
@@ -903,8 +903,7 @@ jobs:
           PYTHONPATH: "${{ env.BUILD_DIR }}:"
         run: |
           source ${{ env.OV_INSTALL_DIR }}/setupvars.sh
-          python3 -m pytest -v ${{ env.SRC_DIR }}/tests/python_tests/test_llm_pipeline.py
-          # python3 -m pytest -v ${{ env.SRC_DIR }}/tests/python_tests/test_structured_output.py
```
popovaan, line 895: Why test_structured_output.py is not launched?

pavel-esir: Here we chat that when user compiled without structured output (`-DENABLE_XGRAMMAR=OFF`) main parts of the pipeline still work. So this is the test to check that other parts of genai work.

If we run structured output tests they will expectedly fail for a build with `-DENABLE_XGRAMMAR=OFF`, therefore here we skip them by `-k "not test_perf_metrics_with_structured_output"`

src/cpp/src/sampling/structured_output/structured_output_controller.cpp
```diff
@@ -52,12 +54,28 @@ StructuredOutputController::get_logits_transformer(const ov::genai::GenerationCo
...
+    return res;
+}
+
+std::pair<std::map<std::string, float>, std::vector<float>> StructuredOutputController::get_times() const {
+    return {m_init_grammar_compiler_times, m_grammar_compile_times};
+}
+
+void StructuredOutputController::clear_compile_times() {
+    std::lock_guard<std::mutex> lock(m_mutex);
+    m_grammar_compile_times.clear();
```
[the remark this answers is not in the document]
pavel-esir: we don't have to clear `m_init_grammar_compiler_times` we need this map saved.

Wovchena:
There's merge conflict

pavel-esir:
> There's merge conflict

resolved

PR #2407, opened 2025-07-07
Fixed vLLM failure on NPU.

intelgaoxiong:
https://jira.devtools.intel.com/browse/EISW-174949
Fixed issues when running vLLM on NPU:
1.Dynamic shape is not converted to static shape by NPUW. Currently NPUW only used in 
VLMPipelineImpl, not in VLMContinuousBatchingAdapter
2.  In VLMPipelineImpl, vision embedder model is expected to be running on CPU for now, but scheduler_config is not supported.


src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -98,6 +98,14 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
         auto embedder_properties = device_propertes.empty()
             ? properties_copy
             : utils::pop_or_default<ov::AnyMap>(device_propertes, embedder_device, {});
+        // CPU plugin does not support scheduler_config
```
popovaan: scheduler_config is not a plugin property, this config has ContinuousBatching params and should be processed on GenAI side. 
If scheduler_config is passed, it is expected that ContinuousBatching pipeline is used, but this code part is the Stateful pipeline. Here should be no scheduler_config in properties.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -98,6 +98,14 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+        if (embedder_device.find("CPU") != std::string::npos) {
+            const std::string keyToRemove = "scheduler_config";
+            auto it = embedder_properties.find(keyToRemove);
+            if (it != embedder_properties.end()) {
+                embedder_properties.erase(it);
+            }
+        }
```
popovaan: I think this code is not needed. Please remove scheduler config from properties using  `utils::extract_scheduler_config(`) in `VLMPipeline` constructor and pass only plugin properties to `VLMPipelineImpl` constructor.

intelgaoxiong:
@popovaan Could you please take a look once again?
Thanks!

popovaan (approved):

intelgaoxiong:
@Wovchena Could you help to merge the PR?
Thanks!

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -365,30 +366,34 @@ VLMPipeline::VLMPipeline(
...
-    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
 
-    // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
-    if (utils::explicitly_requires_paged_attention(properties)) {
-        auto [plugin_properties, scheduler_config] = utils::extract_scheduler_config(properties, utils::get_latency_oriented_scheduler_config());
-        m_pimpl = std::make_unique<VLMContinuousBatchingAdapter>(models_dir, scheduler_config, device, plugin_properties);
-    } else if (device == "NPU") {
+    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
+    if (device == "NPU") {
+        ov::genai::utils::pop_option(properties, "scheduler_config");
```
Wovchena: If NPU and scheduler_config are set, scheduler_config is silently ignored. This should be an error so the users are aware that scheduler_config isn't used and they should remove it themselves from their config. In case of benchmark_vlm.py it should be the sample responsible for not using scheduler_config in case of NPU. Can you update a sample instead?

intelgaoxiong:
@Wovchena Could you please take a look again?
I push a commit to drop schedule_config in case it's NPU.

samples/python/visual_language_chat/benchmark_vlm.py
```diff
@@ -69,11 +69,14 @@ def main():
...
-    scheduler_config.max_num_batched_tokens = sys.maxsize
+    if device == "NPU":
+        pipe = ov_genai.VLMPipeline(models_path, device)
+    else:
+        scheduler_config = ov_genai.SchedulerConfig()
+        scheduler_config.enable_prefix_caching = False
+        scheduler_config.max_num_batched_tokens = sys.maxsize
 
-    pipe = ov_genai.VLMPipeline(models_path, device, scheduler_config=scheduler_config)
+        pipe = ov_genai.VLMPipeline(models_path, device, scheduler_config=scheduler_config)
```
Wovchena, line 78:
```suggestion
    pipe = ov_genai.VLMPipeline(models_path, device, scheduler_config=scheduler_config)
```

Wovchena (approved):

Wovchena:
build_jenkins

samples/python/visual_language_chat/benchmark_vlm.py
rairatne, line 1: This fix should be carried in benchmark_vlm.cpp too

Wovchena: Will you submit a PR?

rairatne: Sure

rairatne: Here is PR: https://github.com/openvinotoolkit/openvino.genai/pull/2446

PR #2413, opened 2025-07-09
Fix debug build for `XGrammar` on Windows

pavel-esir:
- Resolve this issue https://github.com/openvinotoolkit/openvino.genai/pull/2295#discussion_r2164188160

### DIFF
--- a/src/cpp/CMakeLists.txt
+++ b/src/cpp/CMakeLists.txt
@@ -67,11 +67,6 @@ if(ENABLE_GGUF)
     target_include_directories(gguflib PUBLIC "${gguflib_SOURCE_DIR}")
 endif()
 
-# Check if it's Debug and Windows set ENABLE_XGRAMMAR=OFF forcefully
-if(CMAKE_BUILD_TYPE STREQUAL "Debug" AND WIN32)
-    set(ENABLE_XGRAMMAR OFF CACHE BOOL "Enable XGrammar" FORCE)
-endif()
-
 # Disable XGrammar for Android platform, causes issues in CI
 if(ANDROID)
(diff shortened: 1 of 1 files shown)


src/cpp/CMakeLists.txt
```diff
@@ -88,6 +83,9 @@ if(ENABLE_XGRAMMAR)
       GIT_SUBMODULES_RECURSE ON
     )
 
+    # Set the build type for xgrammar explicitly
+    set(CMAKE_BUILD_TYPE "${CMAKE_BUILD_TYPE}" CACHE STRING "Build type" FORCE)
```
Wovchena: `CMAKE_BUILD_TYPE` is already a cache var.
How does it solve the problem?

pavel-esir: As AI agent suggested this line will `FORCE` flag and the same value will be used when xgrammar configured. I'm not sure if it will indeed solve the issue. Let's see the CI reaction

pavel-esir: It didn't work. Let's see if the new approach works :)

pavel-esir: Now it works!

The reason was that in src\cpp\CMakeLists.txt, line 97
```
# Modify the config.cmake file directly
set(CONFIG_FILE "${xgrammar_SOURCE_DIR}/cmake/config.cmake")
```
this config.cmake is created is was populated by default with `RelWithDebInfo` regardless of `CMAKE_BUILD_TYPE`. And looks like config.cmake has priority over the all other stuff.
```
set(CMAKE_BUILD_TYPE RelWithDebInfo)
```

So i changed manually content of `config.cmake` as we do for python bindings and it worked.

Wovchena (approved):

PR #2416, opened 2025-07-10
Fix get_max_new_tokens()

michalkulakowski:
When prompt_length is bigger than max_length result wraps around and get_max_new_tokens returns huge number. This should fix this issue.


src/cpp/src/generation_config.cpp
```diff
@@ -171,6 +171,9 @@ size_t GenerationConfig::get_max_new_tokens(size_t prompt_length) const {
     if (max_new_tokens != SIZE_MAX) {
         return max_new_tokens;
     } else {
+        if(prompt_length > max_length){
+            return 0;
```
Wovchena: What situation triggers that case? Maybe it should be an exception to let a user know that their prompt is already too big

src/cpp/src/generation_config.cpp
```diff
@@ -171,6 +171,9 @@ size_t GenerationConfig::get_max_new_tokens(size_t prompt_length) const {
     if (max_new_tokens != SIZE_MAX) {
         return max_new_tokens;
     } else {
+        if(prompt_length > max_length){
```
Wovchena:
```suggestion
        if (prompt_length > max_length) {
```

Wovchena (changes requested):

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -238,7 +238,7 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
 
     /// @brief checks that are no conflicting parameters, e.g. do_sample=true and num_beams > 1.
     /// @throws Exception if config is invalid.
-    void validate() const;
+    void validate(size_t prompt_length = 0) const;
```
Wovchena: This is part of public API. Can you change src/cpp/src/continuous_batching/pipeline_impl.cpp instead to additionally assert prompt_len

tests/python_tests/utils/hugging_face.py
```diff
@@ -35,7 +35,9 @@ def generation_config_to_hf(
     # generic parameters
     kwargs['max_length'] = generation_config.max_length
     # has higher priority than 'max_length'
-    kwargs['max_new_tokens'] = generation_config.max_new_tokens
+    SIZE_MAX = 18446744073709551615
+    if generation_config.max_new_tokens != SIZE_MAX:
+        kwargs['max_new_tokens'] = generation_config.max_new_tokens
```
Wovchena, line 40: What was the problem to set 18446744073709551615 for HF?

tests/python_tests/test_sampling.py
```diff
@@ -17,14 +17,16 @@
                          [(dict(max_new_tokens=30), 'table is made of'),
                           (dict(max_new_tokens=30, min_new_tokens=30), '你好！ 你好嗎？'),
                           (dict(max_new_tokens=30, ignore_eos=True), 'Alan Turing was a'),
-                        #   (dict(max_length=40), 'table is made of'),
+                          (dict(max_length=30, ignore_eos=True), 'table is made of'),
+                          (dict(max_length=5, ignore_eos=True), 'table is made of'), # max_length smaller than number of prompt tokens, generation should stop right away
```
Wovchena: I think it's going to throw failing the whole test case

tests/python_tests/utils/hugging_face.py
michalkulakowski: In genai max_new_tokens=SIZE_MAX is treated like its not set but for HF its different and max_new_tokens=SIZE_MAX leds to a test timeout

Wovchena (approved):

tests/python_tests/utils/hugging_face.py
```diff
@@ -35,7 +35,9 @@ def generation_config_to_hf(
     # generic parameters
     kwargs['max_length'] = generation_config.max_length
     # has higher priority than 'max_length'
-    kwargs['max_new_tokens'] = generation_config.max_new_tokens
+    SIZE_MAX = 2**64 - 1
```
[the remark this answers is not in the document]
Wovchena: maxsize is `2**63 - 1`

Wovchena:
https://github.com/openvinotoolkit/openvino.genai/actions/runs/16906111694/job/47897377808?pr=2416 is a valid failure that needs to be fixed

Wovchena:
build_jenkins

PR #2424, opened 2025-07-11
torchcodec for linux to use the latest datasets

Wovchena:


sbalandi (approved):

tools/who_what_benchmark/requirements.txt
```diff
@@ -7,7 +7,8 @@ pandas>=2.0.3
 numpy>=1.23.5
 tqdm>=4.66.1
 diffusers
-datasets==3.6.0
+datasets==4.0.0; sys_platform == "linux"
+datasets==3.6.0; sys_platform != "linux"
```
Wovchena: @sbalandi, should I actually fix the version? Since this is a tool, you may want to relax so it's easier to run for people.

There was upper bound initially that's why I touched this file. Maybe the version requirement can be removed completely?

sbalandi: I would prefer to leave:
datasets<=4.0.0; sys_platform == "linux"
datasets<4.0.0; sys_platform != "linux"

sbalandi: But are you sure, that load_dataset is not fail with `RuntimeError: Dataset scripts are no longer supported, but found ...` for wwb ?

Wovchena: The current state sets the latest datasets version for linux and it passed

Wovchena: So should wwb's datasets version requirement be relaxed or merge as is?

sbalandi: @Wovchena I'm very sorry for the long answer, please, relax requirements

Wovchena: Relaxed

sbalandi (approved):

PR #2426, opened 2025-07-11
Support non string chat templates

yatarkan:
There are models that have non string (nested) chat templates, e.g.:
- CohereLabs/aya-23-8B
- CohereLabs/aya-expanse-8b
- CohereLabs/aya-23-35B
- CohereLabs/c4ai-command-r7b-12-2024
- CohereLabs/c4ai-command-r-v01

Example of non string chat template in `tokinizer_config.json` (CohereLabs/aya-23-8B):
```
"chat_template": [
    {
      "name": "default",
      "template": "{{ ... }}"
    }
  ],
```

This PR enables parsing of such chat templates from json files (`tokenizer_config.json`) and from tokenizer model rt info.


src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -66,10 +66,55 @@ std::optional<std::string> remap_template(const std::string& chat_template) {
...
+    // Handle rt_info chat template format: <chat_template><default value="..." /></chat_template>
+    if (chat_template_value.is<ov::AnyMap>()) {
+        ov::AnyMap chat_template_map = chat_template_value.as<ov::AnyMap>();
+        auto default_iter = chat_template_map.find("default");
+        if (default_iter != chat_template_map.end() && default_iter->second.is<std::string>()) {
+            value = default_iter->second.as<std::string>();
+            return;
+        }
+    }
+    OPENVINO_THROW("Unsupported type for 'chat_template' in tokenizer model: ", chat_template_value.type_info().name());
```
yatarkan: @Wovchena Not sure if we need to throw here, maybe something like `std::cerr << [ WARNING ] ... << std::endl;` would be enough.
And the same question for `parse_chat_template_from_file()`.

Wovchena: If the pipeline is used without chat_template, the exception remains. Because of that warning is better. Convert to `cerr`. Please include info how to silence the warning

yatarkan: Warning message updated

Wovchena (approved):

PR #2428, opened 2025-07-11
Update README.md for llm bench

raymondlo84:
Updating the instruction to ensure people are using the right Python

### DIFF
--- a/tools/llm_bench/README.md
+++ b/tools/llm_bench/README.md
@@ -2,6 +2,8 @@
 
 This script provides a unified approach to estimate performance for Large Language Models (LLMs). It leverages pipelines provided by Optimum-Intel and allows performance estimation for PyTorch and OpenVINO models using nearly identical code and pre-collected models.
 
+### 0. Install Python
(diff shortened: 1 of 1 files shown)


tools/llm_bench/README.md
```diff
@@ -2,6 +2,8 @@
 
 This script provides a unified approach to estimate performance for Large Language Models (LLMs). It leverages pipelines provided by Optimum-Intel and allows performance estimation for PyTorch and OpenVINO models using nearly identical code and pre-collected models.
 
+### 0. Install Python
+We recommend installing Python packages from the [Official Python website](https://www.python.org/downloads). Python versions 3.10-3.12 are recommended.
```
rkazants, line 6: We have other documentation with this missed step. Shall we update it as well?

Wovchena, line 6:
```suggestion
We recommend installing Python packages from the [Official Python website](https://www.python.org/downloads). Python versions 3.10-3.13 are recommended.
```
I think 3.13 is going to be declared as supported in 25.3

Wovchena, line 6: Just curious. What's the reason for the change? There must be someone who used something different

Wovchena (changes requested):

PR #2431, opened 2025-07-14
Enforce failure on PA backend init when explicitly requested in LLMPipeline

WeldonWangwang:
Enforce failure on PA backend init when explicitly requested in LLMPipeline

### DIFF
--- a/src/cpp/src/llm/pipeline.cpp
+++ b/src/cpp/src/llm/pipeline.cpp
@@ -79,6 +79,7 @@ ov::genai::LLMPipeline::LLMPipeline(
     m_device(device) {
     auto start_time = std::chrono::steady_clock::now();
     auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
+    const bool user_specified_backend = user_properties.count("ATTENTION_BACKEND") > 0;
 
     // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
     if (utils::explicitly_requires_paged_attention(user_properties)) {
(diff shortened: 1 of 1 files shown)


src/cpp/src/llm/pipeline.cpp
```diff
@@ -97,6 +98,9 @@ ov::genai::LLMPipeline::LLMPipeline(
             m_pimpl = std::make_unique<ContinuousBatchingAdapter>(models_path, tokenizer, utils::get_latency_oriented_scheduler_config(), device, properties);
 #endif
```
as-suvorov, line 99: This condition doesn't seem to be covered. If architecture is not supported no error will be thrown.

src/cpp/src/llm/pipeline.cpp
```diff
@@ -97,6 +98,9 @@ ov::genai::LLMPipeline::LLMPipeline(
...
         } catch (ov::Exception&) {
+            if (user_specified_backend) {
```
as-suvorov, line 101: I think it make sense to move `user_specified_backend` under catch block as it's used only here.

WeldonWangwang:
Hi @rkazants @as-suvorov @peterchen-intel , seems the functionality of this PR has already been covered by https://github.com/openvinotoolkit/openvino.genai/blob/21693ffbe0f62467435e3eb15cd03b723f365d21/src/cpp/src/llm/pipeline.cpp#L84 . If the construction of the ContinuusBatchingAdapter fails, it should be thrown directly.

There should not be a situation where the construction of ContinuusBatchingAdapter fails but return `m_pimpl==nulliptr`, so will close this PR if there are no issues.

PR #2443, opened 2025-07-16
Increase timeout

Wovchena:
Triggered in https://github.com/openvinotoolkit/openvino.genai/actions/runs/16316173954/job/46082655691


Wovchena:
@as-suvorov, reminder

as-suvorov (approved):

PR #2446, opened 2025-07-17
benchmark_vlm: Fixed vLLM failure on NPU

JeevakaPrabu:
Ticket: EISW-174949

Tested-by: Jeevaka Prabu Badrappan <jeevaka.badrappan@intel.com>

### DIFF
--- a/samples/cpp/visual_language_chat/benchmark_vlm.cpp
+++ b/samples/cpp/visual_language_chat/benchmark_vlm.cpp
@@ -69,19 +69,23 @@ int main(int argc, char* argv[]) try {
 
     std::cout << ov::get_openvino_version() << std::endl;
 
-    ov::genai::VLMPipeline pipe(models_path, device, ov::genai::scheduler_config(scheduler_config));
+    std::unique_ptr<ov::genai::VLMPipeline> pipe;
+    if (device == "NPU")
+        pipe = std::make_unique<ov::genai::VLMPipeline>(models_path, device);
+    else
+        pipe = std::make_unique<ov::genai::VLMPipeline>(models_path, device, ov::genai::scheduler_config(scheduler_config));
 
-    auto input_data = pipe.get_tokenizer().encode(prompt);
+    auto input_data = pipe->get_tokenizer().encode(prompt);
     size_t prompt_token_size = input_data.input_ids.get_shape()[1];
     std::cout << "Number of images:" << images.size() << ", prompt token size:" << prompt_token_size << std::endl;
 
     for (size_t i = 0; i < num_warmup; i++)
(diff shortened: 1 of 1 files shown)


samples/cpp/visual_language_chat/benchmark_vlm.cpp
```diff
@@ -96,6 +100,7 @@ int main(int argc, char* argv[]) try {
     std::cout << "TPOT: " << metrics.get_tpot().mean  << " ± " << metrics.get_tpot().std << " ms/token " << std::endl;
     std::cout << "Throughput: " << metrics.get_throughput().mean  << " ± " << metrics.get_throughput().std << " tokens/s" << std::endl;
 
+    delete pipe;
```
Wovchena: Does `std::optional` work here instead of `VLMPipeline*`? If so, use it. If not, use `std::unique_ptr` so there's no need to care about freeing memory if VLMPipeline throws

samples/cpp/visual_language_chat/benchmark_vlm.cpp
```diff
@@ -69,19 +69,24 @@ int main(int argc, char* argv[]) try {
...
 
-    ov::genai::VLMPipeline pipe(models_path, device, ov::genai::scheduler_config(scheduler_config));
-
-    auto input_data = pipe.get_tokenizer().encode(prompt);
+    std::unique_ptr<ov::genai::VLMPipeline> pipe;
+    if (device == "NPU")
+        pipe = std::make_unique<ov::genai::VLMPipeline>(models_path, device);
+    else
+        pipe = std::make_unique<ov::genai::VLMPipeline>(models_path, device,
+                                                     ov::genai::scheduler_config(scheduler_config));
```
Wovchena:
```suggestion
                                                        ov::genai::scheduler_config(scheduler_config));
```

rairatne:
@Wovchena PR is updated can you please help to review and get it merged?

Wovchena:
build_jenkins

dmatveev:
The real question here is why the scheduler configuration falls to the plugin side at all, how does it work in the CPU/GPU case?

Wovchena:
There's ContinousBatching class between VLMPipeline and OV for CPU and GPU. It intercepts the config

dmatveev:
> There's ContinousBatching class between VLMPipeline and OV for CPU and GPU. It intercepts the config

So we should handle it in the stateful path as well, isn't it?

Wovchena:
`SchedulerConfig` is specific to `ContinousBatching`. It's just doesn't make sense to specify `SchedulerConfig` for stateful path.

PR #2436, opened 2025-07-15
Add text rerank pipeline

as-suvorov:

PR #2457, opened 2025-07-18
Add 2nd input to Tokenizers transformation

pavel-esir:
Description
This PR adds support for a second input (paired input) to the Tokenizers transformation in C++.

Key changes:
- Added 'AddSecondInputPass` which adds second input to the `ov_tokenizer` model.
- Added google tests for the pass.
- Added logic to apply a transformation pass for adding a second input if requested and not already present.
- Updates and extends related tests to cover cases when inputs are added during genai runtime.

Ticket CVS-170415

### DIFF
--- a/src/cpp/src/tokenizer/tokenizer.cpp
+++ b/src/cpp/src/tokenizer/tokenizer.cpp
@@ -377,11 +377,18 @@ public:
         setup_tokenizer(std::make_pair(ov_tokenizer, ov_detokenizer), filtered_properties);
     }
 
-    void setup_tokenizer(const std::pair<std::shared_ptr<ov::Model>, std::shared_ptr<ov::Model>>& models, const ov::AnyMap& properties) {
+    void setup_tokenizer(const std::pair<std::shared_ptr<ov::Model>, std::shared_ptr<ov::Model>>& models, ov::AnyMap properties) {
         auto [ov_tokenizer, ov_detokenizer] = models;
-        auto [filtered_properties, two_input_requested] = utils::extract_paired_input_props(properties);
+        bool two_input_requested = false;
+
+        auto it = properties.find(ov::genai::add_second_input.name());
+        if (it != properties.end()) {
+            two_input_requested = it->second.as<bool>();
+            properties.erase(it);
+        }
+
         // Pass no addtional properties to tokenizer/detokenizer models since it was not used by default
-        filtered_properties = {};
+        properties = {};
         
         is_paired_input = ov_tokenizer && ov_tokenizer->get_parameters().size() == 2;
         
@@ -417,7 +424,7 @@ public:
             manager.register_pass<MakeAddSpecialTokensSatateful>();
             manager.register_pass<MakePaddingSatateful>();
             manager.run_passes(ov_tokenizer);
-            ov::CompiledModel tokenizer = core.compile_model(ov_tokenizer, device, filtered_properties);
+            ov::CompiledModel tokenizer = core.compile_model(ov_tokenizer, device, properties);
             ov::genai::utils::print_compiled_model_properties(tokenizer, "OV Tokenizer");
 
             m_ireq_queue_tokenizer = std::make_unique<CircularBufferQueue<ov::InferRequest>>(
@@ -449,7 +456,7 @@ public:
             ov::pass::Manager manager_detok;
             manager_detok.register_pass<MakeVocabDecoderSatateful>();
             manager_detok.run_passes(ov_detokenizer);
-            ov::CompiledModel detokenizer = core.compile_model(ov_detokenizer, device, filtered_properties);
+            ov::CompiledModel detokenizer = core.compile_model(ov_detokenizer, device, properties);
             ov::genai::utils::print_compiled_model_properties(detokenizer, "OV Detokenizer");
 
             m_ireq_queue_detokenizer = std::make_unique<CircularBufferQueue<ov::InferRequest>>(
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -318,19 +318,6 @@ std::pair<ov::AnyMap, bool> extract_gguf_properties(const ov::AnyMap& external_p
     return {properties, enable_save_ov_model};
 }
 
-std::pair<ov::AnyMap, bool> extract_paired_input_props(const ov::AnyMap& external_properties) {
-    bool add_second_input = false;
-    ov::AnyMap properties = external_properties;
-
-    auto it = properties.find(ov::genai::add_second_input.name());
-    if (it != properties.end()) {
-        add_second_input = it->second.as<bool>();
-        properties.erase(it);
-    }
-
-    return {properties, add_second_input};
-}
-
 void save_openvino_model(const std::shared_ptr<ov::Model>& model, const std::string& save_path, bool compress_to_fp16) {
     try {
         auto serialize_start_time = std::chrono::high_resolution_clock::now();


apaniukov:
https://github.com/openvinotoolkit/openvino.genai/pull/2436

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -300,6 +300,11 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
      * @throws Exception if the detokenizer is not available.
      */
     const std::vector<std::string>& get_vocab_vector() const;
+    
+    /**
+     * @brief Check if the tokenizer supports paired input.
+     */
```
pavel-esir:
```suggestion
     /// @brief Check if the tokenizer supports paired input.
```

tests/python_tests/test_tokenizer.py
```diff
@@ -649,3 +668,6 @@ def test_template_priorities(tmp_path, chat_templates):
     tokenizer = generate_tokenizer(tmp_path, chat_templates)
     assert tokenizer.chat_template == chat_templates.reference
 
+
+# if user didn't request 2d input transformation but feed encode with 2 inputs
+# check if Tokenizer supports 2 inputs
```
pavel-esir:
```suggestion
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -329,6 +335,31 @@ class Tokenizer::TokenizerImpl {
...
         auto [ov_tokenizer, ov_detokenizer] = models;
+        auto [filtered_properties, two_input_requested] = utils::extract_paired_input_props(properties);
+
+        if (ov_tokenizer && ov_tokenizer->get_parameters().size() == 2) {
+            is_paired_input = true;
+        }
+        
+        ov::pass::Manager manager;
+        manager.register_pass<ov::pass::VisualizeTree>("origin.svg");
+        manager.run_passes(ov_tokenizer);
```
pavel-esir:
```suggestion
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -329,6 +335,31 @@ class Tokenizer::TokenizerImpl {
...
+
+        // If model is already converted with 2 inputs, then skip the pass
+        if (ov_tokenizer && two_input_requested && !is_paired_input) {
+            ov::pass::Manager manager;
+            manager.register_pass<ov::pass::VisualizeTree>("before.svg");
```
pavel-esir:
```suggestion
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -329,6 +335,31 @@ class Tokenizer::TokenizerImpl {
...
+            manager.register_pass<ov::genai::AddSecondInputPass>(m_shared_object_ov_tokenizers, m_pass_errors);
+            manager.register_pass<ov::pass::VisualizeTree>("after.svg");
```
pavel-esir:
```suggestion
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -11,13 +11,15 @@
 #include <jinja2cpp/generic_list_iterator.h>
 
 #include "openvino/pass/manager.hpp"
+#include "openvino/pass/visualize_tree.hpp"
```
pavel-esir:
```suggestion
```

pavel-esir:
This PR will not be green until https://github.com/openvinotoolkit/openvino.genai/pull/2465 is merged

src/cpp/src/tokenizer/add_second_input_pass.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * @class AddSecondInputPass
+ * @brief Changes the ov::Model of tokenizer so that it accepts paired input.
+ *
+ * This pass modifies the model to accept two input sequences (paired input).
+ * It concatenates both inputs, processes and tokenizes them together, and then splits them near the end,
+ * before the CombineSegments node. If any constant SpecialToken depends on sequence inputs, they are zeroed.
+ * The truncation operation is also modified to support max_length.
+ */
```
Wovchena, line 16: Move docstring to its class

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -329,6 +334,25 @@ class Tokenizer::TokenizerImpl {
...
+        if (ov_tokenizer && two_input_requested && !is_paired_input) {
+            ov::pass::Manager manager;
+            manager.register_pass<ov::genai::AddSecondInputPass>(m_shared_object_ov_tokenizers, m_pass_errors);
+            manager.run_passes(ov_tokenizer);
+            is_paired_input = true;
+        }
+
+
+        if (two_input_requested && !is_paired_input) {
+            OPENVINO_THROW("Two input requested but AddSecondInputPass failed with " + m_pass_errors.str());
```
Wovchena: Replace with OPENVINO_ASSERT

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -311,6 +314,7 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
 static constexpr ov::Property<bool> add_special_tokens{"add_special_tokens"};
 static constexpr ov::Property<bool> skip_special_tokens{"skip_special_tokens"};
 static constexpr ov::Property<bool> pad_to_max_length{"pad_to_max_length"};
+static constexpr ov::Property<bool> add_second_input{"add_second_input"};
```
Wovchena: It's possible to infer the two input tokenizer with just one input. But it's not possible to infer one input tokenizer with two inputs. Why don't you always apply the pass?

pavel-esir: We decided not to turn in on by default in to secure ourselves if something will go wrong with the pass, or with the resulting model with 2 inputs (e.g. if performance or accuracy is dropped in such cases).

I manually checked on several models (ModernBert, LlamaTokenziers) show no significant perf drops (no more than 1-2%) and accuracy is the same. But we didn't do comprehensive measures.

Since tokenizers mostly is used inside genai pipelines, we know which pipelines require paired inputs and call ctor with required params. For most of the genai users there will be no need to set this arg manually.

tests/python_tests/test_tokenizer.py
```diff
@@ -403,13 +406,29 @@ def test_padding(
...
+# Define model base configs
+base_models_for_paired_input_test = [
+    ("answerdotai/ModernBERT-base", {"padding_side": None}),
+    ("TinyLlama/TinyLlama-1.1B-Chat-v1.0", {"padding_side": None}),
+    ("katuni4ka/tiny-random-llava-next", {"padding_side": "right"}),
+    ("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}),
 ]
 
+def make_model_params():
+    # Parametrize over add_second_input and number_of_inputs
```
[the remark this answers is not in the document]
pavel-esir: We use Parametrize in all other places, also pytest uses Parametrize. So I'll keep it as is

tests/cpp/test_add_second_input_pass.cpp
```diff
@@ -0,0 +1,397 @@
...
+    auto model = make_minimal_model();
+    static const std::string PROCESSED_POST_PROCESSOR_NAME = "processed_post_processor_template";
+    std::vector<int> input_signature = {-1, 42};
+    nlohmann::json post_processor = {{"single", {{"ids", input_signature}}}};
+    model->get_rt_info()[PROCESSED_POST_PROCESSOR_NAME] = post_processor.dump();
+    std::ostringstream pass_errors;
+    bool ok = run_add_second_input_pass(model, pass_errors);
+    ASSERT_FALSE(ok);
+    std::cout << pass_errors.str() << std::endl;
+    ASSERT_NE(pass_errors.str().find("Could not add second input. post_processor does not contain input signature for paired input"), std::string::npos);
```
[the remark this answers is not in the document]
pavel-esir: removed cout

src/cpp/src/tokenizer/add_second_input_pass.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    // Slice for second input
+    std::shared_ptr<ov::Node> begins_2 = std::make_shared<Slice>(begin, second_start, total_size, make_constant({1}));
+    std::shared_ptr<ov::Node> ends_2 = std::make_shared<Slice>(end, second_start, total_size, make_constant({1}));
+
+    // If inputs_2 is empty, we need to zero the second dimension of the broadcasted begins and ends tensors.
+    auto zero_const = make_constant({0});
+    this->equal_node = std::make_shared<Equal>(param_2_shape, zero_const);
+    begins_2 = std::make_shared<Select>(equal_node, zero_const, begins_2);
+    // TODO: indeed for ends it should've been 1 but for thix bug CSV-160624 we set to zero for the moment.
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/tokenizer/add_second_input_pass.cpp
```diff
@@ -0,0 +1,336 @@
...
+        return false;
+    }
+
+    if (post_processor["single"]["ids"].get<std::vector<int>>() != input_signature) {
+        m_pass_errors << "Could not add second input. Input signature from rt_info does not match to the CombineSegments node inputs." << std::endl;
+    }
+
+    auto pair_ids = post_processor["pair"]["ids"].get<std::vector<int>>();
+    if (std::vector<int>(pair_ids.begin(), pair_ids.begin() + input_signature.size()) != input_signature) {
+        m_pass_errors << "Could not add second input. Paried inputs are allowed only when it's widening the single input." << std::endl;
```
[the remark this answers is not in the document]
pavel-esir: done

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -253,22 +259,21 @@ class Tokenizer::TokenizerImpl {
...
+        ov_tokenizer_filesystem_path = std::filesystem::path(std::wstring(ov_tokenizer_path_w));
+#else
+        const char* ov_tokenizer_path = getenv(ScopedVar::ENVIRONMENT_VARIABLE_NAME);
+        ov_tokenizer_filesystem_path = std::filesystem::path(ov_tokenizer_path);
+#endif
+        m_shared_object_ov_tokenizers = load_shared_object(ov_tokenizer_filesystem_path);
+
         std::shared_ptr<ov::Model> ov_tokenizer = nullptr;
         std::shared_ptr<ov::Model> ov_detokenizer = nullptr;
         auto [filtered_properties, enable_save_ov_model] = utils::extract_gguf_properties(properties);
```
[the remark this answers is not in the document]
pavel-esir: it was not introduced by me, will not change that

src/cpp/src/tokenizer/add_second_input_pass.hpp
pavel-esir: done, also added docstring for CTOR

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: done

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Let's review and merge the current pass, and decide whether we want to enable it by default later. Currently I enabled it for `TextRerankPipeline`

pavel-esir: Removed public declaration of this flag. Now it's for our internal use only

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -208,6 +209,11 @@ class Tokenizer::TokenizerImpl {
     // handle to shared_object with openvino tokenizers
     std::shared_ptr<void> m_shared_object_ov_tokenizers = nullptr;
 
+    // buffer to store the error messages from the pass
+    std::ostringstream m_pass_errors;
```
Wovchena: `m_pass_errors` is only used in constructor once. Move it to `setup_tokenizer()`

src/cpp/src/utils.cpp
```diff
@@ -318,6 +318,19 @@ std::pair<ov::AnyMap, bool> extract_gguf_properties(const ov::AnyMap& external_p
     return {properties, enable_save_ov_model};
 }
 
+std::pair<ov::AnyMap, bool> extract_paired_input_props(const ov::AnyMap& external_properties) {
+    bool add_second_input = false;
+    ov::AnyMap properties = external_properties;
```
Wovchena: Why don't you modify it inplace? The current solution spawns a copy which leads to bugs like one fixed in https://github.com/openvinotoolkit/openvino.genai/pull/2534 user_properties vs properties

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -376,6 +381,21 @@ class Tokenizer::TokenizerImpl {
 
     void setup_tokenizer(const std::pair<std::shared_ptr<ov::Model>, std::shared_ptr<ov::Model>>& models, const ov::AnyMap& properties) {
         auto [ov_tokenizer, ov_detokenizer] = models;
+        auto [filtered_properties, two_input_requested] = utils::extract_paired_input_props(properties);
+
+        if (ov_tokenizer && ov_tokenizer->get_parameters().size() == 2) {
+            is_paired_input = true;
+        }
```
apaniukov:
```cpp
is_paired_input = ov_tokenizer && ov_tokenizer->get_parameters().size() == 2;
```

tests/python_tests/utils/hugging_face.py
```diff
@@ -207,10 +207,7 @@ def download_and_convert_embeddings_models(request):
 @pytest.fixture()
 def download_and_convert_rerank_model(request):
     model_id = request.param
-    opt_model, hf_tokenizer, models_path = _download_and_convert_model(model_id, OVModelForSequenceClassification)
-    ov_tokenizer = convert_tokenizer(hf_tokenizer, with_detokenizer=False, number_of_inputs=2)
-    save_model(ov_tokenizer, models_path / "openvino_tokenizer.xml")
-    return opt_model, hf_tokenizer, models_path
+    return _download_and_convert_model(model_id, OVModelForSequenceClassification)
```
apaniukov, line 210: It seems like we don't need a separate fixture for rerank models anymore.

tests/python_tests/test_tokenizer.py
```diff
@@ -424,14 +427,47 @@ def test_padding(
...
+    ["hi" * 20, "buy" * 90],
+]])
+def test_two_inputs_string_list_of_lists_batched(hf_ov_genai_models, input_pair):
+    # Check with batched inputs: list of [str, str] pairs, consistent with HF format.
+    hf_tokenizer, genai_tokenizer = hf_ov_genai_models
+    ov_encoded = genai_tokenizer.encode(input_pair).input_ids.data
+    hf_encoded = hf_tokenizer(input_pair, return_tensors="np", padding=True)["input_ids"]
+    assert np.all(ov_encoded == hf_encoded)
 
+# Original single batch cases
```
apaniukov:
```suggestion
```

tests/python_tests/test_tokenizer.py
```diff
@@ -424,14 +427,47 @@ def test_padding(
...
-    ("katuni4ka/tiny-random-llava-next", {"padding_side": "left", 'number_of_inputs': 2}),
+# Define model base configs
+base_models_for_paired_input_test = [
+    ("answerdotai/ModernBERT-base", {"padding_side": None}),
+    ("TinyLlama/TinyLlama-1.1B-Chat-v1.0", {"padding_side": None}),
+    ("katuni4ka/tiny-random-llava-next", {"padding_side": "right"}),
+    ("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}),
 ]
 
+def make_model_params():
```
apaniukov, line 438: Can be an indirect fixture

src/cpp/src/tokenizer/add_second_input_pass.hpp
```diff
@@ -0,0 +1,66 @@
...
+private:
+    FactoryCreateType node_factory;
+    std::vector<int> input_signature;
+    std::vector<ov::Output<ov::Node>> inputs;
+    nlohmann::json post_processor;
+    ov::ParameterVector new_parameters;
+    std::shared_ptr<ov::Node> equal_node;
+    std::vector<ov::Output<ov::Node>> trunc_values;
+    std::array<ov::Output<ov::Node>, 3> first_input;
+    std::array<ov::Output<ov::Node>, 3> second_input;
```
as-suvorov: I think it make sense to mark members with `m_` prefix for consistency or readability 
```suggestion
    std::array<ov::Output<ov::Node>, 3> m_second_input;
```

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -300,22 +306,21 @@ class Tokenizer::TokenizerImpl {
...
+        const char* ov_tokenizer_path = getenv(ScopedVar::ENVIRONMENT_VARIABLE_NAME);
+        ov_tokenizer_filesystem_path = std::filesystem::path(ov_tokenizer_path);
+#endif
+        m_shared_object_ov_tokenizers = load_shared_object(ov_tokenizer_filesystem_path);
+
         std::shared_ptr<ov::Model> ov_tokenizer = nullptr;
         std::shared_ptr<ov::Model> ov_detokenizer = nullptr;
         auto [filtered_properties, enable_save_ov_model] = utils::extract_gguf_properties(properties);
-        // Pass no addtional properties to tokenizer/detokenizer models since it was not used by default
-        filtered_properties = {};
```
as-suvorov, line 307: @apaniukov FYI

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -376,6 +381,21 @@ class Tokenizer::TokenizerImpl {
...
+        if (ov_tokenizer && ov_tokenizer->get_parameters().size() == 2) {
+            is_paired_input = true;
+        }
+
+        // If model is already converted with 2 inputs, then skip the pass
+        if (ov_tokenizer && two_input_requested && !is_paired_input) {
+            ov::pass::Manager manager;
+            manager.register_pass<ov::genai::AddSecondInputPass>(m_shared_object_ov_tokenizers, m_pass_errors);
+            manager.run_passes(ov_tokenizer);
+            is_paired_input = true;
```
as-suvorov: In case of transformation fail will `run_passes` return `false` or throw an error?

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -302,6 +302,9 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
      * @throws Exception if the detokenizer is not available.
      */
     const std::vector<std::string>& get_vocab_vector() const;
+    
+     /// @brief Check if the tokenizer supports paired input.
+    bool is_paired_input() const;
```
as-suvorov: nitpick
```suggestion
    bool supports_paired_input() const;
```
or
```suggestion
    bool has_paired_input() const;
```

src/cpp/src/utils.cpp
as-suvorov: Is `external_properties` a user object? Or a copy already?
If it's a user object is it expected to modify it?

tests/cpp/test_add_second_input_pass.cpp
```diff
@@ -0,0 +1,396 @@
...
+static bool run_add_second_input_pass(std::shared_ptr<ov::Model> model, std::ostringstream& pass_errors) {
+    ov::pass::Manager manager;
+    manager.register_pass<ov::genai::AddSecondInputPass>(shared_object_ov_tokenizers, pass_errors);
+    try {
+        manager.run_passes(model);
+    } catch (...) {
+        // Some errors may throw, but pass should return false
+        return false;
+    }
+    return pass_errors.str().empty();
```
as-suvorov: It seems in case of normal execution `run_passes` will return `bool` so it's possible to `return manager.run_passes(model);`.
What is the reason to catch the error here? Wouldn't it hide unexpected fails?

as-suvorov (commented):
Could you please also remove manual python transformation from [samples](https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/conftest.py#L185)

tests/cpp/test_add_second_input_pass.cpp
pavel-esir: You are right, try/catch is redundant and masks potential problems. I removed it.

Regarding run_passes return flag it equivalent of checking if pass_errors is empty, because in any case if i exit from the pass leaving the model unchanged i a populate pass_errors with a relevant message, so i left this part unchanged.

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: Good catch! Thanks for noticing that. At 385 we don't know for sure if the transofrmation successfully added the seocnd input. 

We can either check the return pass of run_passes or check if error is empty. I will rely on error message, i am bit afraid of using return flag from run_passes because i'm not if it always works, but for error masse i'm 100% sure.

src/cpp/src/tokenizer/add_second_input_pass.hpp
pavel-esir: agreed, and renamed with `m_*`

tests/python_tests/test_tokenizer.py
pavel-esir: removed

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: done

src/cpp/src/tokenizer/add_second_input_pass.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+    // Slice for second input
+    std::shared_ptr<ov::Node> begins_2 = std::make_shared<Slice>(begin, second_start, total_size, make_constant({1}));
+    std::shared_ptr<ov::Node> ends_2 = std::make_shared<Slice>(end, second_start, total_size, make_constant({1}));
+
+    // If inputs_2 is empty, we need to zero the second dimension of the broadcasted begins and ends tensors.
+    auto zero_const = make_constant({0});
+    this->equal_node = std::make_shared<Equal>(param_2_shape, zero_const);
+    begins_2 = std::make_shared<Select>(equal_node, zero_const, begins_2);
+    // TODO: indeed for ends it should've been 1 but for this bug CSV-160624 we set to zero for the moment.
```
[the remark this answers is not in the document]
pavel-esir: done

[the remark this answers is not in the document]
pavel-esir: done

tests/cpp/test_add_second_input_pass.cpp
as-suvorov: I think it's more 'cleaner' to rely on `run_passes` interface directly, over `pass_errors`. Another reader could think this is desired patter and one should not use `run_passes` return value. But up to you.

src/cpp/src/tokenizer/tokenizer.cpp
as-suvorov: That should be at least documented at 2nd input pass header. Or better fixed as it's a pass interface as I understand. Up to you.

src/cpp/src/utils.cpp
pavel-esir: Its a copy. We cannot modify `external_properties` in place since it's a const. If we make argument non const this can confuse things later, because in that case `setup_tokenizer()` will modify input arguments is something unexpected.

As i understood from #2534 the bug occured because the flag which is removed from the config indeed was needed subseguently. But in our case this flag is needed only in one place and we remove it immediately, because if we lieave in the map compile model which accepts config map will fail because plugins do not understand this name `add_second_input`. So we are obliged to remove it as early as possible. I didn't invert anything new, i used the approach in the same file al ittle bit above with gguf 
https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/tokenizer/tokenizer.cpp#L305

tests/python_tests/test_tokenizer.py
pavel-esir: I tried, but it looks like it's not possible. Indirect fixture receives one set or arguments and returns result for them, but here i feed model_id and return as a result a list of values. Fixture does not allow "spawning" only transformation.

If i make make_model_params a fixture with model_id argument, and parametrize over all this extra args, e.g.
```
@pytest.mark.fixture
def make_model_params(base_models_for_paired_input_test, extra_args):
      {concatenate args and return}
      
@pytest.mark.parametrize("make_model_params", [{}, {"add_second_input": True}, {"number_of_inputs": 2}, ...], indirect=True)
@pytest.mark.parametrize("hf_ov_genai_models", make_model_params???, indirect=True)
def test(
```
i cannot then pass them into `hf_ov_genai_models`. If the whole model id and params `("katuni4ka/tiny-random-llava-next", {"padding_side": "left"}), ("katuni4ka/tiny-random-llava-next", {"padding_side": "left",  "number_of_inputs": 2})` etc. went directly into the model and if i could've called fixture inside it would've been possible, but fixture cannot be called directly inside function.

So the current approach is the one i found working and which keeps `hf_ov_genai_models` fixture and does not do unnecessary extra reconversions of the model

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: `supports_paired_input` sounds better. Renamed

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Returned back public declaration

pavel-esir:
> Could you please also remove manual python transformation from [samples](https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/conftest.py#L185)

done

apaniukov (approved):

tests/python_tests/utils/hugging_face.py
```diff
@@ -199,20 +199,11 @@ def download_and_convert_model(model_id: str, **tokenizer_kwargs):
...
-def download_and_convert_embeddings_models(request):
+def download_and_convert_models(request):
     model_id = request.param
     return _download_and_convert_model(model_id, OVModelForFeatureExtraction)
 
 
-@pytest.fixture()
-def download_and_convert_rerank_model(request):
-    model_id = request.param
-    opt_model, hf_tokenizer, models_path = _download_and_convert_model(model_id, OVModelForSequenceClassification)
```
as-suvorov: Test failing because of this change: https://github.com/openvinotoolkit/openvino.genai/actions/runs/16830630697/job/47678550764?pr=2457#step:7:180
Please revert it

as-suvorov (changes requested):

src/cpp/src/tokenizer/tokenizer.cpp
as-suvorov: `m_pass_errors` is not used anymore, I guess we can remove it
```suggestion
```

as-suvorov (changes requested):

src/cpp/src/utils.cpp
Wovchena: `setup_tokenizer()` should take `properties` as a non const copy

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -376,6 +382,24 @@ class Tokenizer::TokenizerImpl {
...
+
+        // If model is already converted with 2 inputs, then skip the pass
+        if (ov_tokenizer && two_input_requested && !is_paired_input) {
+            ov::pass::Manager manager;
+            manager.register_pass<ov::genai::AddSecondInputPass>(m_shared_object_ov_tokenizers, pass_errors);
+            manager.run_passes(ov_tokenizer);
+        }
+        
+        is_paired_input = pass_errors.str().empty();
+        OPENVINO_ASSERT(!two_input_requested || is_paired_input, "Two input requested but AddSecondInputPass failed with " + pass_errors.str());
```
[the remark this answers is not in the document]
pavel-esir: applied manually

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: done

tests/python_tests/utils/hugging_face.py
pavel-esir: reverted the old division of names

src/cpp/src/tokenizer/add_second_input_pass.cpp
```diff
@@ -0,0 +1,347 @@
...
+    // If inputs_2 is empty, we need to zero the second dimension of the broadcasted begins and ends tensors.
+    auto zero_const = make_constant({0});
+    this->m_equal_node = std::make_shared<Equal>(param_2_shape, zero_const);
+    begins_2 = std::make_shared<Select>(m_equal_node, zero_const, begins_2);
+    // TODO: For correct behavior, 'ends_2' should be set to 1 when the second input is empty.
+    // However, due to bug CSV-160624 (incorrect handling of empty second input in Select and broadcast),
+    // we temporarily set it to zero as a workaround. 
+    // The following line to use 1 instead of 0 is kept for future reference: once bug CSV-160624 is resolved
+    // auto one_const = make_constant({1});
+    ends_2 = std::make_shared<Select>(m_equal_node, zero_const, ends_2);
```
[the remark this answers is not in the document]
pavel-esir: will not apply

tests/cpp/test_add_second_input_pass.cpp
```diff
@@ -0,0 +1,391 @@
...
+#else
+        const char* ov_tokenizer_path = getenv(ScopedVar::ENVIRONMENT_VARIABLE_NAME);
+#endif
+
+auto shared_object_ov_tokenizers = load_shared_object(std::filesystem::path(ov_tokenizer_path));
+
+FactoryCreateType create_func = reinterpret_cast<FactoryCreateType>(get_symbol(shared_object_ov_tokenizers, "create_tokenizer_node"));
+
+
+TEST(AddSecondInputTest, add_second_input_test_1) {
```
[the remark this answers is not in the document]
pavel-esir: left the current approach

tests/python_tests/test_rag.py
```diff
@@ -6,7 +6,7 @@
 import gc
 from pathlib import Path
 from openvino_genai import TextEmbeddingPipeline, TextRerankPipeline
-from utils.hugging_face import download_and_convert_embeddings_models, download_and_convert_rerank_model
+from utils.hugging_face import download_and_convert_models, download_and_convert_models
```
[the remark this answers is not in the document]
pavel-esir: this lines have been changed. not actual anymore

src/cpp/src/utils.cpp
pavel-esir: Done

Wovchena (approved):

PR #2460, opened 2025-07-18
[llm_bench] Remove unused processor for genAI vlm pipeline

sbalandi:
Fix was caused by failing of run phi-3.5-vision-instruct and phi-4-multimodal-instruct with GenAI, but anyway processor config is not used in case of GenAI.

Note:
Converted to IR format versions of these models from hf master will still fails in runs via optimum.
Why runs are fails ?
Converted to IR format versions of phi-3.5-vision-instruct and phi-4-multimodal-instruct models from hf master includes chat_template.json/chat_template.jinja ,  AutoTokenizer take into account chat_template as extra processor config args, but processing_phi3_v.py/processing_phi4mm.py is not allows to do that. 
How these models can be run with optimum ?
1. phi-3.5 - take converted model from OpenVINO repo on hf: https://huggingface.co/OpenVINO/Phi-3.5-vision-instruct-fp16-ov
2. convert model based on special revision, where processing.py files was updated, it is `--revision  refs/pr/78` for phi-3.5 and `--revision  refs/pr/35` for phi-3.5

### DIFF
--- a/tools/llm_bench/llm_bench_utils/ov_utils.py
+++ b/tools/llm_bench/llm_bench_utils/ov_utils.py
@@ -595,8 +595,6 @@ def create_genai_image_text_gen_model(model_path, device, ov_config, memory_moni
     if not (model_path / "openvino_tokenizer.xml").exists() or not (model_path / "openvino_detokenizer.xml").exists():
         convert_ov_tokenizer(model_path)
 
-    processor_config = get_vlm_processor(model_path)
-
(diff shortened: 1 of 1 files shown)


AlexanderKalistratov:
Thanks, it helps to run `phi-3.5-vision` with `llm_bench\benchmark.py`
@Wovchena can we merge this?

Wovchena (approved):
There's going to be a lag with the merge because CI isn't stable now

PR #2470, opened 2025-07-22
Run finish_chat() before start_chat() in VLM pipeline.

popovaan:
Run finish_chat() before start_chat() in VLM pipeline and CB pipeline to prevent accumulating of history images when multiple start_chat() are used. 
Ticket: CVS-170106


Wovchena (approved):

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -279,6 +279,7 @@ std::vector<VLMDecodedResults> ContinuousBatchingPipeline::generate(
 
 
 void ContinuousBatchingPipeline::start_chat(const std::string& system_message) {
+    m_impl->finish_chat();
```
rkazants, line 282: can we test it? as i understand, you just need to check output from the second chat.

popovaan: Yes, done.

PR #2471, opened 2025-07-22
Fixed VLMpipeline sample failure on NPU

JohnLeFeng:
Fixed issues when running VLMpipeline sample on NPU.
When run on NPU, it will provide image tensor even it already processes it to avoid below issue. 

```sh
Check 'rgbs.size() == 1u' failed at C:\Jenkins\workspace\private-ci\ie\build-windows-vs2022\b\repos\openvino.genai\src\cpp\src\visual_language\pipeline.cpp:179:
Currently only batch size equal to 1 is supported for NPU device!
```

### DIFF
--- a/samples/cpp/visual_language_chat/visual_language_chat.cpp
+++ b/samples/cpp/visual_language_chat/visual_language_chat.cpp
@@ -32,7 +32,11 @@ int main(int argc, char* argv[]) try {
 
     std::string prompt;
 
-    pipe.start_chat();
+    if (device != "NPU") {
+        // Start the chat session only for CPU and GPU.
+        // NPU does not support chat mode.
+        pipe.start_chat();
+    }
     std::cout << "question:\n";
 
     std::getline(std::cin, prompt);
--- a/samples/cpp/visual_language_chat/visual_language_chat.cpp
+++ b/samples/cpp/visual_language_chat/visual_language_chat.cpp
@@ -43,9 +43,16 @@ int main(int argc, char* argv[]) try {
     std::cout << "\n----------\n"
         "question:\n";
     while (std::getline(std::cin, prompt)) {
-        pipe.generate(prompt,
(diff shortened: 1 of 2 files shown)


sammysun0711:
@intelgaoxiong @Wovchena @popovaan, could you please kindly review it? BTW, may I know if it is known limitation that VLM on NPU does not support prefix cache, so we need to explicitly pass both prompt and input image for each VLM generation even they are exactly same?

dmatveev:
@AlexanderKalistratov FYI

intelgaoxiong:
@sammysun0711 I know NPU expect batch size = 1, but it's not clear for me how this is solved by passing prompts and images.
@JohnLeFeng Could you explain how the issue happens? Thanks you!

samples/cpp/visual_language_chat/visual_language_chat.cpp
Wovchena, line 1: That must be the general lack of chat support. @sbalandi, is that correct?

JohnLeFeng:
Sorry for missed no chat mode support on NPU.

JohnLeFeng:
@intelgaoxiong Customer runs model with this sample on NPU, and they met several issues in this sample.
Initially, the issue is from `start_chat()` function and I noticed NPU isn't support chat mode, so I request them to remove it when run VLMpipeline on NPU. After it, they report another issue after further chat with LLM. With it, I notice NPU on VLMpipeline needs to provide `image_tensor` in each conversation.

samples/cpp/visual_language_chat/visual_language_chat.cpp
sbalandi: yes, after chat_mode with full history will be applied for VLM  that sample should work on NPU

Wovchena:
Fixed in https://github.com/openvinotoolkit/openvino.genai/pull/2486

PR #2475, opened 2025-07-22
[llm_bench] Add apply_chat_template for LLM

sbalandi:
[CVS-169182](https://jira.devtools.intel.com/browse/CVS-169182)

example of usage:
python ../tools/llm_bench/benchmark.py -m ./TinyLlama-1.1B-Chat-v1.0-int8-ov/ --disable_prompt_permutation  --apply_chat_template

By default chat template is not used as before. For GenAI cases it is better to use --apply_chat_template with disable_prompt_permutation  , otherwise after applying chat template prompt will be tokenized and mixed and chat template will not make sense

### DIFF
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -194,6 +194,9 @@ def get_argprser():
     parser.add_argument("--embedding_normalize", action="store_true", help="Normalize embeddings. Applicable only for text embeddings")
     parser.add_argument("--embedding_max_length", type=int, default=None,
                         help="Max length for text embeddings. Input text will be padded or truncated to specified value")
+    parser.add_argument("--apply_chat_template", action="store_true",
+                        help="Apply chat template for LLM. By default chat template is not applied. It's better to use with --disable_prompt_permutation, \
+                              otherwise the prompt will be modified after applying the chat template, so the structure of chat template will not be kept.")
     return parser.parse_args()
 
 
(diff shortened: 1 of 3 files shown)


Wovchena (approved):

tools/llm_bench/task/text_generation.py
```diff
@@ -33,6 +33,11 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
     if args["output_dir"] is not None and num == 0:
         for bs_index, in_text in enumerate(input_text_list):
             llm_bench_utils.output_file.output_input_text(in_text, args, model_precision, prompt_index, bs_index, proc_id)
+    if args["apply_chat_template"]:
+        input_text_hist = [{'role': 'user', 'content': input_text}]
+        tempalted_input_text = tokenizer.apply_chat_template(input_text_hist, tokenize=False, add_generation_prompt=True)
```
as-suvorov:
```suggestion
        templated_input_text = tokenizer.apply_chat_template(input_text_hist, tokenize=False, add_generation_prompt=True)
```

sbalandi: fixed

tools/llm_bench/benchmark.py
```diff
@@ -194,6 +194,9 @@ def get_argprser():
     parser.add_argument("--embedding_normalize", action="store_true", help="Normalize embeddings. Applicable only for text embeddings")
     parser.add_argument("--embedding_max_length", type=int, default=None,
                         help="Max length for text embeddings. Input text will be padded or truncated to specified value")
+    parser.add_argument("--apply_chat_template", action="store_true",
+                        help="Apply chat template for LLM. By default chat template is not applied. It's better to use with --disable_prompt_permutation, \
```
rkazants, line 198: by default, don't we use chat template stored in IR if it presents?

Wovchena: In WWB but not in llm_bench because variable number of tokens affects performance

PR #2478, opened 2025-07-23
Suppress C4146 warning for xgrammar

olpipi:


src/cpp/CMakeLists.txt
```diff
@@ -106,7 +106,16 @@ if(ENABLE_XGRAMMAR)
         set(XGRAMMAR_ENABLE_CPPTRACE OFF CACHE BOOL "Enable CPPTrace" FORCE)
         set(XGRAMMAR_ENABLE_CUDA OFF CACHE BOOL "Enable CUDA" FORCE)
 
+        # Suppress C4146 warning for MSVC compiler
+        set(ORIGINAL_CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS})
+        set(ORIGINAL_CMAKE_C_FLAGS ${CMAKE_C_FLAGS})
+        if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+            set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4146")
+            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4146")
+        endif()
```
Wovchena: https://github.com/openvinotoolkit/openvino.genai/pull/2100/commits/b7a96c2873116a8af8755ccf6dc3d49c606d9f13 feels to be a cleaner solution. Does it work replacing sdl with this warning?

olpipi: We have to know cmake target name for this approach. I can use `xgrammar`. But if it is changed in future, it will trigger not obvious issues.

Wovchena: `xgrammar` is already used below `target_link_libraries(${TARGET_NAME} PRIVATE xgrammar)`

olpipi: changed as you suggested

PR #2481, opened 2025-07-24
Dont require merge queeue

Wovchena:


.github/workflows/linux.yml
```diff
@@ -316,6 +320,8 @@ jobs:
           if-no-files-found: 'error'
 
   store_artifacts:
+    if: github.event_name != 'merge_group'
+    if: github.event_name != 'merge_group'
```
as-suvorov:
```suggestion
    if: github.event_name != 'merge_group'
```

as-suvorov:
Are all jobs excluded from `merge_group`? If yes, can we exclude `merge_group` from workflow `on` instead?
image

Wovchena:
Some of them are required by branch protection so they must be triggered in queue as well

PR #2486, opened 2025-07-24
Extend using of full chat history mode for stateful pipeline for VLM and LLM with encoded inputs

sbalandi:


src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -23,7 +23,7 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
     // Finish reason of last generation for chat scenario
     ov::genai::GenerationStatus m_chat_generation_finish_status = ov::genai::GenerationStatus::RUNNING;
     // if True, full history will be used as prompt on each chat generation
-    bool m_use_full_chat_history = false;
+    bool m_use_full_chat_history = true;
```
sbalandi, line 26: it is here to make sure that everything works by testing in ci, it will be removed when review done

sbalandi:
@dmatveev @AlexanderKalistratov @AsyaPronina could you please help with review ?

PR #2498, opened 2025-07-29
Fix Generated tokens in sd debug metrics

sbalandi:
[CVS-167754](https://jira.devtools.intel.com/browse/CVS-167754)

removed `Generated tokens` per requests, as we can't keep accurate numbers for last step, when requests for main model have removed
added common `Generated tokens` 
added `generated token by draft model`

### DIFF
--- a/src/cpp/src/speculative_decoding/speculative_decoding_metrics.cpp
+++ b/src/cpp/src/speculative_decoding/speculative_decoding_metrics.cpp
@@ -59,15 +59,15 @@ float SpeculativeDecodingMetrics::get_draft_accepted_tokens_percentage(int64_t r
         size_t total_iteration_cnt = 0;
         for (const auto& accepten_token_cnt : m_draft_accepted_tokens) {
             avg_acceptance_rate += accepten_token_cnt.second;
-            total_iteration_cnt += m_generated_len[request_id];
+            total_iteration_cnt += m_draft_generated_len[request_id];
         }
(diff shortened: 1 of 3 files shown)


src/cpp/src/speculative_decoding/speculative_decoding_metrics.cpp
```diff
@@ -95,17 +95,25 @@ void SpeculativeDecodingMetrics::update_draft_accepted_tokens(int64_t request_id
     }
 }
 
-void SpeculativeDecodingMetrics::set_generated_len(int64_t request_id, size_t generated_len) {
-    m_generated_len.insert({ request_id, generated_len });
+void SpeculativeDecodingMetrics::update_draft_generated_len(int64_t request_id, size_t generated_len) {
+    if (m_draft_generated_len.count(request_id)) {
+        m_draft_generated_len[request_id] += generated_len;
+    } else {
+        m_draft_generated_len.insert({ request_id, generated_len });
```
Wovchena: operator[] default constructs a value if it's missing. So it can be just `m_draft_generated_len[request_id] += generated_len;`.

Some people find this to be weird and avoid using that. If you want to do so, you can use find() returning an iterator so you don't need to search for `request_id` twice

sbalandi: updated

PR #2499, opened 2025-07-29
Added clarification regarding the use of memory consumption mode to help

sbalandi:


tools/llm_bench/benchmark.py
```diff
@@ -83,8 +83,10 @@ def get_argprser():
         default=0,
         required=False,
         type=int,
-        help='if the value is 1, output the maximum memory consumption in warm-up iterations. If the value is 2,'
-        ' output the maximum memory consumption in all iterations.',
+        help='Enables memory usage information collection mode. If the value is 1, output the maximum memory consumption in warm-up iterations.'
+        ' If the value is 2, output the maximum memory consumption in all iterations.\nIt is not recommended to run memory consumption and'
+        ' performance benchmarking in the same time. Effect on performance can be reduced by specifying a longer --memory_consumption_delay,'
+        ' but the impact will still be expected '
```
Wovchena:
```suggestion
        ' performance benchmarking at the same time. Effect on performance can be reduced by specifying a longer --memory_consumption_delay,'
        ' but the impact is still expected '
```

sbalandi: updated

tools/llm_bench/benchmark.py
```diff
@@ -83,8 +83,10 @@ def get_argprser():
         default=0,
         required=False,
         type=int,
-        help='if the value is 1, output the maximum memory consumption in warm-up iterations. If the value is 2,'
-        ' output the maximum memory consumption in all iterations.',
+        help='Enables memory usage information collection mode. If the value is 1, output the maximum memory consumption in warm-up iterations.'
+        ' If the value is 2, output the maximum memory consumption in all iterations.\nIt is not recommended to run memory consumption and'
+        ' performance benchmarking at the same time. Effect on performance can be reduced by specifying a longer --memory_consumption_delay,'
+        ' but the impact is still expected. '
```
Wovchena, line 90:
```suggestion
        ' but the impact is still expected.'
```
And there's a conflict

PR #2507, opened 2025-07-30
remove extra cpu->gpu tensor copies in genai VLM pipeline

liangali:
There are some extra tensor copies in VLM pipeline, which would cause overhead in VLM 2nd+ token generation and impact E2E token rate. This PR is trying to remove these copies by creating tensors as "usm_shared" memory

### DIFF
--- a/src/cpp/src/continuous_batching/model_runner.hpp
+++ b/src/cpp/src/continuous_batching/model_runner.hpp
@@ -48,6 +48,21 @@ class ModelRunner {
     // Output shape: [1, conversation length, hidden_size].
     EmbeddingsModel::Ptr m_embedding;
 
+    // Cached pre-allocated tensors to avoid CPU->GPU copy
+    ov::Tensor m_cached_input_ids;
+    ov::Tensor m_cached_inputs_embeds;
+    ov::Tensor m_cached_position_ids;
+    ov::Tensor m_cached_past_lens;
+    ov::Tensor m_cached_subsequence_begins;
+    ov::Tensor m_cached_block_indices_begins;
+    ov::Tensor m_cached_max_context_len;
+    ov::Tensor m_cached_score_aggregation_window;
+    
+    // Maximum allocated sizes for dynamic resizing
+    size_t m_max_allocated_tokens;
+    size_t m_max_allocated_sequences;
+    size_t m_max_allocated_blocks;
+
 public:
     /**
      * Constructs the ModelRunner.
@@ -81,6 +96,10 @@ public:
           m_is_aggregate_attention_scores(is_aggregate_attention_scores) {
         OPENVINO_ASSERT(m_num_decoder_layers != 0, "num_decoder_layers must be non-zero");
         _reset_cache_rotation_coefficients();
+        _init_cached_tensors();
+        
+        // Uncomment the following line to debug tensor optimization
+        // _debug_tensor_optimization();
     }
 
     /**
@@ -147,18 +166,26 @@ public:
             max_context_len_val = std::max(max_context_len_val, sequence_group->get_context_len());
         }
 
-        ov::Tensor
-            input_ids(ov::element::i64, {total_num_tokens}),
-            inputs_embeds(ov::element::f32, {total_num_tokens, hidden_size}),
-            position_ids(ov::element::i64, {total_num_tokens}),
-            // PA specific parameters
-            past_lens(ov::element::i32, {batch_size_in_sequences}),
-            subsequence_begins(ov::element::i32, {batch_size_in_sequences + 1}),
-            // block_indices are handled in a special fashion below
-            block_indices_begins(ov::element::i32, {batch_size_in_sequences + 1}),
-            max_context_len(ov::element::i32, {});
+        // Use cached pre-allocated tensors instead of creating new ones
+        ov::Tensor input_ids = _get_or_resize_tensor(m_cached_input_ids, "input_ids", 
+                                                     {total_num_tokens}, ov::element::i64);
+        ov::Tensor inputs_embeds = _get_or_resize_tensor(m_cached_inputs_embeds, "inputs_embeds", 
+                                                        {total_num_tokens, hidden_size}, ov::element::f32);
+        ov::Tensor position_ids = _get_or_resize_tensor(m_cached_position_ids, "position_ids", 
+                                                       {total_num_tokens}, ov::element::i64);
+        
+        // PA specific parameters
+        ov::Tensor past_lens = _get_or_resize_tensor(m_cached_past_lens, "past_lens", 
+                                                    {batch_size_in_sequences}, ov::element::i32);
+        ov::Tensor subsequence_begins = _get_or_resize_tensor(m_cached_subsequence_begins, "subsequence_begins", 
+                                                            {batch_size_in_sequences + 1}, ov::element::i32);
+        ov::Tensor block_indices_begins = _get_or_resize_tensor(m_cached_block_indices_begins, "block_indices_begins", 
+                                                              {batch_size_in_sequences + 1}, ov::element::i32);
+        ov::Tensor max_context_len = _get_or_resize_tensor(m_cached_max_context_len, "max_context_len", 
+                                                          {}, ov::element::i32);
 
-        ov::Tensor score_aggregation_window(ov::element::i32, {batch_size_in_sequences});
+        ov::Tensor score_aggregation_window = _get_or_resize_tensor(m_cached_score_aggregation_window, "score_aggregation_window", 
+                                                                  {batch_size_in_sequences}, ov::element::i32);
 
         ov::Tensor generated_ids_embeds;
         float *generated_ids_embeds_data = nullptr;
@@ -298,23 +325,21 @@ public:
             sequence_group->set_output_seq_len(matmul_gathering_is_available ? output_seq_len : num_scheduled_tokens);
         }
 
-        if (sequence_group_type == SequenceGroupType::TOKENS) {
-            m_request.set_tensor("input_ids", input_ids);
-        }
-        else if (sequence_group_type == SequenceGroupType::EMBEDDINGS) {
-            m_request.set_tensor("inputs_embeds", inputs_embeds);
-        }
+        // Note: No need to call set_tensor() for cached tensors as they are already bound to the request
+        // The tensors we're using are pre-allocated and directly obtained from m_request.get_tensor()
+        
+        // For validation, we can optionally verify that the tensors are properly bound:
+        // OPENVINO_ASSERT(input_ids.data() == m_request.get_tensor("input_ids").data(), 
+        //                 "input_ids tensor should be the same as the one in the request");
 
-        // typical LLM parameters
-        m_request.set_tensor("position_ids", position_ids);
-
-        // PA specific parameters
-        m_request.set_tensor("past_lens", past_lens);
-        m_request.set_tensor("subsequence_begins", subsequence_begins);
-
+<<<<<<< Updated upstream
         _set_block_indices(sequence_groups, scheduler_output, total_num_blocks, seq_id_to_skipped_blocks_map);
         m_request.set_tensor("block_indices_begins", block_indices_begins);
         m_request.set_tensor("max_context_len", max_context_len);
+=======
+        _set_block_indices(sequence_groups, scheduler_output, total_num_blocks);
+        // block_indices_begins and max_context_len are also pre-allocated, no need to set them
+>>>>>>> Stashed changes
 
         if (m_is_use_rotation_inputs) {
             m_request.set_tensor("rotation_trig_lut", m_cache_rotation_trig_lut);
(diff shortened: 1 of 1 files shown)


src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -298,38 +325,42 @@ class ModelRunner {
...
-        m_request.set_tensor("past_lens", past_lens);
-        m_request.set_tensor("subsequence_begins", subsequence_begins);
+        // Note: No need to call set_tensor() for cached tensors as they are already bound to the request
+        // The tensors we're using are pre-allocated and directly obtained from m_request.get_tensor()
+        
+        // For validation, we can optionally verify that the tensors are properly bound:
+        // OPENVINO_ASSERT(input_ids.data() == m_request.get_tensor("input_ids").data(), 
+        //                 "input_ids tensor should be the same as the one in the request");
 
+<<<<<<< Updated upstream
```
as-suvorov: Please fix git diffs

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -405,7 +436,69 @@ class ModelRunner {
     }
 
 private:
+<<<<<<< Updated upstream
```
as-suvorov: Same here

as-suvorov (changes requested):

rkazants (commented):
please share CVS ticket. Did you run validation for your changes to ensure that there is no regression?

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -405,6 +427,51 @@ class ModelRunner {
...
+            m_cached_inputs_embeds = m_request.get_tensor("inputs_embeds");
+            m_cached_position_ids = m_request.get_tensor("position_ids");
+            m_cached_past_lens = m_request.get_tensor("past_lens");
+            m_cached_subsequence_begins = m_request.get_tensor("subsequence_begins");
+            m_cached_block_indices_begins = m_request.get_tensor("block_indices_begins");
+            m_cached_max_context_len = m_request.get_tensor("max_context_len");
+        } catch (const ov::Exception&) {
+            // If pre-allocated tensors don't exist or are not accessible,
+            // we'll fall back to the original method
+            // This shouldn't happen in continuous batching models, but it's a safety net
```
Wovchena: This statement about just safety net contradicts with
```
            // If they exist and are USM Host tensors, use them
            // Otherwise, this will throw and we'll handle it in the catch block
```

If there's no `input_ids`, `m_request` throws skipping initialization of the reset of the tensors.

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -298,33 +315,38 @@ class ModelRunner {
...
-
-        // PA specific parameters
-        m_request.set_tensor("past_lens", past_lens);
-        m_request.set_tensor("subsequence_begins", subsequence_begins);
+        // Note: No need to call set_tensor() for cached tensors as they are already bound to the request
+        // The tensors we're using are pre-allocated and directly obtained from m_request.get_tensor()
+        
+        // For validation, we can optionally verify that the tensors are properly bound:
+        // OPENVINO_ASSERT(input_ids.data() == m_request.get_tensor("input_ids").data(), 
+        //                 "input_ids tensor should be the same as the one in the request");
```
Wovchena: You expect exceptions here and there which are performance costly but you hesitate to run an simple assert. I'd recommend to uncomment it.

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -298,33 +315,38 @@ class ModelRunner {
...
+            // Try to use pre-allocated tensor for gather_indices as well
+            try {
+                ov::Tensor gather_indices = m_request.get_tensor("sampled_tokens_indices");
+                gather_indices.set_shape({gather_indices_values.size()});
+                std::memcpy(gather_indices.data(), gather_indices_values.data(), gather_indices_values.size() * sizeof(int64_t));
+            } catch (const ov::Exception&) {
+                // Fallback to creating new tensor if pre-allocated one doesn't exist
+                ov::Tensor gather_indices(ov::element::i64, {gather_indices_values.size()});
+                std::memcpy(gather_indices.data(), gather_indices_values.data(), gather_indices_values.size() * sizeof(int64_t));
+                m_request.set_tensor("sampled_tokens_indices", gather_indices);
```
Wovchena: Why isn't `sampled_tokens_indices` cached?

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: done

jingyua1:
code refined and validation WIP.

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: cached in latest change.

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: good point! aleady fall back to original method in latest change.

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: comment cleaned.

jingyua1:
perf and accuracy test are passed

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -434,6 +452,53 @@ class ModelRunner {
...
+            cached_tensor.get_shape().empty() || 
+            ov::shape_size(cached_tensor.get_shape()) < ov::shape_size(required_shape)) {
+
+            try {
+                // Try to get the tensor from the model and set its shape
+                cached_tensor = m_request.get_tensor(tensor_name);
+                cached_tensor.set_shape(required_shape);
+            } catch (const ov::Exception&) {
+                // If tensor doesn't exist, create a regular tensor
+                cached_tensor = ov::Tensor(element_type, required_shape);
```
Wovchena: In what cases this catch is triggered? I thought there's always a tensor to return from ireq

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -153,20 +164,28 @@ class ModelRunner {
...
-            past_lens(ov::element::i32, {batch_size_in_sequences}),
-            subsequence_begins(ov::element::i32, {batch_size_in_sequences + 1}),
-            // block_indices are handled in a special fashion below
-            block_indices_begins(ov::element::i32, {batch_size_in_sequences + 1}),
-            max_context_len(ov::element::i32, {});
-
-        ov::Tensor score_aggregation_window(ov::element::i32, {batch_size_in_sequences});
+        // Use cached pre-allocated tensors instead of creating new ones
+        ov::Tensor input_ids = _get_or_resize_tensor(m_cached_input_ids, "input_ids", 
+                                                     {total_num_tokens}, ov::element::i64);
```
Wovchena: The alignment doesn't match the alignment below. Can you just use 4 spaces instead of that type of alignment? If something is renamed, your way of alignment needs to be adjusted.

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: those tensors handled in this PR are not common for all models. For example token_type_ids is not used by Qwen2.5-VL-7B-Instruct-int4. When do inference with Qwen2.5-VL-7B-Instruct-int4, m_request.get_tensor("token_type_ids") in _init_cached_tensors() will fail and throw exception, thus the following get_tensor call would be skipped and related cached tensor is null.

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: done.

src/cpp/src/continuous_batching/model_runner.hpp
Wovchena: I think if an ireq doesn't have a name, the corresponding tensor isn't going to be used. It should be OK to leave the tensor default constructed. (State that in code comment to keep it clear)

`_init_cached_tensors()` still has global try catch. If one of the names fail, the following names won't be initialized.

Wovchena:
@popovaan, please, review

popovaan (approved):
Looks good to me.

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: comments added.

Wovchena: Fix `_init_cached_tensors()` instead

jingyua1: done.

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -351,12 +359,19 @@ class ModelRunner {
...
+            } catch (const ov::Exception&) {
+                // Fallback to creating new tensor if pre-allocated one doesn't exist
+                ov::Tensor gather_indices(ov::element::i64, {gather_indices_values.size()});
+                std::memcpy(gather_indices.data(), gather_indices_values.data(), gather_indices_values.size() * sizeof(int64_t));
+                m_request.set_tensor("sampled_tokens_indices", gather_indices);
         }
 
-        if (m_is_aggregate_attention_scores) {
+        if (m_is_aggregate_attention_scores && !m_cached_score_aggregation_window) {
             m_request.set_tensor("score_aggregation_window", score_aggregation_window);
```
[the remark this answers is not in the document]
jingyua1: not necessary

src/cpp/src/continuous_batching/model_runner.hpp
Wovchena: You still allocate tensors under `ov::Exception`

jingyua1: score_aggregation_window is used even there's no such name in ireq

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -319,28 +336,20 @@ class ModelRunner {
             }
             sequence_group->set_output_seq_len(matmul_gathering_is_available ? output_seq_len : num_scheduled_tokens);
         }
-
-        if (sequence_group_type == SequenceGroupType::TOKENS) {
-            m_request.set_tensor("input_ids", input_ids);
```
Wovchena, line 324: I can't find `input_ids` being set. I assume caching with `_get_or_resize_tensor()` handles this implicitly. The problem is that `_get_or_resize_tensor()` may not use the result of `m_request.get_tensor()`. The reader is going to assume it may happen to any tensor because `score_aggregation_window` is listed as an example. Please revert removal of `set_tensor()` calls for all such tensors

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -319,27 +336,18 @@ class ModelRunner {
...
-        }
-        else if (sequence_group_type == SequenceGroupType::EMBEDDINGS) {
-            m_request.set_tensor("inputs_embeds", inputs_embeds);
-            if (have_token_type_ids) {
-                m_request.set_tensor("token_type_ids", token_type_ids);
-            }
+        
+        if (sequence_group_type == SequenceGroupType::EMBEDDINGS && have_token_type_ids && !m_cached_token_type_ids) {
+            m_request.set_tensor("token_type_ids", token_type_ids);
```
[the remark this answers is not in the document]
Wovchena: `ov::Tensor token_type_ids = _get_or_resize_tensor(m_cached_token_type_ids, "token_type_ids", {1, total_num_tokens}, ov::element::i64);` is called unconditionally so you know `m_cached_token_type_ids` is always initialized. Remove ` && !m_cached_token_type_ids` from the condition. It's more preferable to fail because ireq tries to access uninitialized tensor rather then get an accuracy issue because `m_request.set_tensor("token_type_ids", token_type_ids);` wasn't called

src/cpp/src/continuous_batching/model_runner.hpp
jingyua1: yes, the logic to set tensor to ireq is handles implicitly inside `_get_or_resize_tensor()`.
Reverted removal of` set_tensor() `calls to align their behavior with `score_aggregation_window`

src/cpp/src/continuous_batching/model_runner.hpp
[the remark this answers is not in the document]
jingyua1: When a ireq requires token_type_ids  as input, token_type_ids  will always be gotten from ireq and no need to set again.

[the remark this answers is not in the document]
Wovchena: Add this info to code comment please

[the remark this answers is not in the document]
jingyua1: done

PR #542, opened 2024-06-21
Add openvino_genai>=2024.2.0 to requirements.txt

skuros:

PR #2512, opened 2025-07-30
Extend padding

pavel-esir:
Ticket CVS-163374

### DIFF
--- a/src/python/openvino_genai/py_openvino_genai.pyi
+++ b/src/python/openvino_genai/py_openvino_genai.pyi
@@ -2974,22 +2974,55 @@ class Tokenizer:
     def encode(self, prompts: collections.abc.Sequence[str], add_special_tokens: bool = True, pad_to_max_length: bool = False, max_length: typing.SupportsInt | None = None, padding_side: str | None = None) -> TokenizedInputs:
         """
         Encodes a list of prompts into tokenized inputs.
+        Args:
+         'prompts' - list of prompts to encode
+         'add_special_tokens' - whether to add special tokens like BOS, EOS, PAD. Default is True.
+         'pad_to_max_length' - whether to pad the sequence to the maximum length. Default is False.
+         'max_length' - maximum length of the sequence. If None (default), the value will be taken from the IR (where default value from original HF/GGUF model is stored).
+         'padding_side' - side to pad the sequence, can be 'left' or 'right'. If None (default), the value will be taken from the IR (where default value from original HF/GGUF model is stored).
+        Returns:
+         TokenizedInputs object containing input_ids and attention_mask tensors.
         """
     @typing.overload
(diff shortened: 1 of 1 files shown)


pavel-esir:
Untill https://github.com/openvinotoolkit/openvino_tokenizers/pull/542 is merged and submodules updated tokenizer tests will remain red.

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -165,6 +165,7 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     * @param add_special_tokens whether to add special tokens
     * @param max_length optional maximum length to which output will be truncated and/or padded. If not defined, taken from IR.
     * @param pad_to_max_length either pad to max_length, or pad to the longest sequence in the batch. Default is false.
+    * @param padding_side side to pad, either "left" or "right". Default is "right".
```
pavel-esir: There is no default value. If it's not defined padding side will rely on the value from the IR

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -178,6 +179,7 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
     * @param add_special_tokens whether to add special tokens
     * @param max_length optional maximum length to which output will be truncated and/or padded. If not defined, taken from IR.
     * @param pad_to_max_length either pad to max_length, or pad to the longest sequence in the batch. Default is false.
+    * @param padding_side side to pad, either "left" or "right". If not defined value is taken from IR.
```
Wovchena: Define what if there's no value in IR?
Add python docstrings

Wovchena (changes requested):

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: fixed

src/cpp/include/openvino/genai/tokenizer.hpp
pavel-esir: Done. 
- The attribute is always present in the op. Default value into IR is written from the original HF/gguf model. Written that in the docstring.
- Added python docstring. It looks like this now
image

pavel-esir:
> Untill [openvinotoolkit/openvino_tokenizers#542](https://github.com/openvinotoolkit/openvino_tokenizers/pull/542) is merged and submodules updated tokenizer tests will remain red.


Managed to get default value of attribute inside GenAI transformation. So there is no need for #542 PR. The current PR should suffice

src/cpp/include/openvino/genai/tokenizer.hpp
Wovchena: What happens if it's not written? You don't fully control IRs because users are free to modify it

PR #2515, opened 2025-07-30
Allow building GenAI with free-threaded Python

p-wysocki:
Explanation for `Python3_FIND_ABI` modification can be found in https://cmake.org/cmake/help/latest/module/FindPython3.html#hints:

"The Python3_FIND_ABI variable is a 4-tuple specifying, in that order, pydebug (d), pymalloc (m), unicode (u) and gil_disabled (t) flags."

Ticket: CVS-169212


CMakeLists.txt
```diff
@@ -112,6 +112,14 @@ set(CPACK_INCLUDE_TOPLEVEL_DIRECTORY OFF)
 set(CPACK_COMPONENTS_ALL core_genai core_genai_dev core_c_genai core_c_genai_dev cpp_samples_genai licensing_genai openvino_tokenizers openvino_tokenizers_docs)
 if(ENABLE_PYTHON)
     list(APPEND CPACK_COMPONENTS_ALL pygenai_${Python3_VERSION_MAJOR}_${Python3_VERSION_MINOR})
+    if(ENABLE_NO_GIL_PYTHON_API)
+        if(Python3_VERSION VERSION_LESS "3.13")
+            message(FATAL_ERROR "ENABLE_NO_GIL_PYTHON_API requires Python >= 3.13, but found Python ${Python3_VERSION}")
```
Wovchena: Is it actually possible to switch GIL on or off for a given python binary? I though it's not possible to switch between them without switching python itself

p-wysocki: It is not possible, Python has to be built as a non-GIL version. If it's a non-GIL version it's noted with a `t` appendix, ex. `Python 3.14.4t`.

Wovchena: In that case `ENABLE_NO_GIL_PYTHON_API` can only be used to generate an error. The actual choice is already made when python binary is sleeted. Do we need `ENABLE_NO_GIL_PYTHON_API` at all?

p-wysocki: It is required to pass `py::mod_gil_not_used()` to [pybind module definition](https://pybind11.readthedocs.io/en/stable/advanced/misc.html#free-threading-support) and for CMake's `Python3_FIND_ABI`. It's also how it has been done for [OpenVINO](https://github.com/openvinotoolkit/openvino/pull/30978).

Wovchena: Is there a reason it's called `ENABLE_NO_GIL_PYTHON_API` instead of `ENABLE_GIL_PYTHON_API` with default to `ON`? `ENABLE_NO_GIL_PYTHON_API=NO` introduces double negation which is harder to read compared to `ENABLE_GIL_PYTHON_API=ON`

Please, add the option to https://github.com/openvinotoolkit/openvino.genai/blob/08987949421b25843af94378811a8722a58c5bb6/cmake/features.cmake

p-wysocki: The reasons why I called it `ENABLE_NO_GIL_PYTHON_API` are:

- it's an experimental, non-default feature in Python 3.13 and even though it's a double negation it makes sense to me
- compliance with OpenVINO, which has `ENABLE_NO_GIL_PYTHON_API`

One way or another, I implemented `ENABLE_GIL_PYTHON_API=ON` by default. I leave the decision up to you - if you prefer it let's just resolve the conversation.

Wovchena: Compliance with OpenVINO is important. But that just shifts the question to OV's decision. Having different names for the same feature is the worst of all results. Will you update OV after GenAI's `ENABLE_GIL_PYTHON_API`?

1. How does experimentality define the name?
2. There's still default behavior even if the default value for ENABLE_GIL_PYTHON_API isn't defined.

p-wysocki: > How does experimentality define the name?

The way I see it is GIL Disabled is an experimental feature and the flag is `ENABLE_EXPERIMENTAL_FEATURE`.

`ENABLE_NO_GIL_PYTHON_API` makes more sense to me, but that's subjective and I see your point. If you insist I'll keep `ENABLE_GIL_PYTHON_API` here and change it in OpenVINO. Also, @almilosz what do you think?

Wovchena: That makes sense for internal users who add the feature. Other users won't be able to parse out that easily that it should be read as `(ENABLE)_(NO_GIL_PYTHON_API)`

p-wysocki: I aligned the cmake option names in https://github.com/openvinotoolkit/openvino/pull/31546.

p-wysocki:
Could you please review? We plan to enable free-threaded wheel building for the upcoming release, I need this PR to make changes to CI configurations. Thanks!

PR #2518, opened 2025-07-30
transformers 4.51.3

Wovchena:
Failed in https://github.com/openvinotoolkit/openvino.genai/actions/runs/16622670547/job/47033188241

### DIFF
--- a/tests/python_tests/requirements.txt
+++ b/tests/python_tests/requirements.txt
@@ -1,4 +1,5 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
+transformers<=4.51.3
 diffusers==0.34.0
 optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@main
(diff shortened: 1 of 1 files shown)


tests/python_tests/requirements.txt
```diff
@@ -1,4 +1,5 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
+transformers<=4.51.3
```
rkazants, line 2: you can install needed version just only for MacOS x86 using condition

`platform_system == "Darwin" and platform_machine == "x86_64"`

as-suvorov (approved):

tests/python_tests/requirements.txt
as-suvorov: Test fails on Linux

mryzhov (changes requested):
That is incorrect fix, please do not introduce one more test requirement, we need to use tokenizers or genai product requirements instead

PR #2519, opened 2025-07-31
[CI][GHA] Limit transformer 4.51.3 & regenerate cache

sammysun0711:


.github/workflows/linux.yml
```diff
@@ -25,7 +25,7 @@ env:
   SCCACHE_CACHE_SIZE: 30G
   SCCACHE_AZURE_KEY_PREFIX: genai/ubuntu/22_04/x64
   HF_HOME: /mount/caches/huggingface/lin
-  OV_CACHE: /mount/caches/huggingface/.ov_cache/lin
+  OV_CACHE: /mount/caches/huggingface/.ov_cache/lin/187f4d4
```
rkazants, line 28: why changed?

samples/export-requirements.txt
```diff
@@ -9,6 +9,6 @@ transformers_stream_generator==0.0.5  # For Qwen
 diffusers==0.34.0 # For image generation pipelines
 timm==1.0.19  # For exporting InternVL2
 torchvision  # For visual language models
-transformers>=4.43 # For Whisper
+transformers>=4.43,<=4.51.3 # For Whisper
```
rkazants, line 12: why do we need it?

.github/workflows/linux.yml
sammysun0711: This is test purpose, since I found that transformers version used for model export are different as python test, would like to regenerate OV cache to check if which version is more stable for CI test.

samples/export-requirements.txt
sammysun0711: Same here, this is test purpose, since I found that transformers version used for model export are different as python test, would like to regenerate OV cache to check which version is more stable for CI test.

sammysun0711: BTW, transformer version should already limited by optimum-intel: [setup.py](https://github.com/huggingface/optimum-intel/blob/55bc71281c4e6317cfbaee46c6fb600d871273dd/setup.py#L32), but it allow transformers 4.52.* version to e installed.
```
INSTALL_REQUIRE = [
    "torch>=1.11",
    "optimum@git+https://github.com/huggingface/optimum.git",
    "transformers>=4.36,<4.53",
    "datasets>=1.4.0",
    "setuptools",
    "scipy",
    "onnx",
]
```

rkazants: what is bad with 4.52 transformers?

tests/python_tests/requirements.txt
```diff
@@ -1,4 +1,5 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
+transformers<=4.51.3
```
rkazants, line 2: I think we should no limit upper bound for transformers on all platforms.
Moreover, it is not convenient to update it regularly in all requirements files.

samples/export-requirements.txt
sammysun0711: Test failure w/ model export converted with transformer 4.52.4: https://github.com/openvinotoolkit/openvino.genai/actions/runs/16630409967/job/47060373943?pr=2518

tests/python_tests/requirements.txt
mryzhov: That is a good practice to limit  the upper bounds in the requirements, dependabot will updated them automatically, 
But here is the wrong place to do that, requirement came from wwb and should be limited there

sammysun0711: What if it not only need for wwb, but it also affect LLM/VLM test or general precommit test? 

> That is a good practice to limit the upper bounds in the requirements, dependabot will updated them automatically, But I here is the wrong place to do that, requirement came from wwb and should be limited there

sammysun0711: Should I add transformers version limitation in wwb too? https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/requirements.txt

samples/export-requirements.txt
Wovchena: https://github.com/openvinotoolkit/openvino.genai/pull/2465 happened to regenerate cache with 4.52.4 and that broke runtime time that uses 4.51.3?
According to the same PR test results there are other issues with 4.52.4

PR #2521, opened 2025-07-31
Update tokenizers version

mryzhov:
Need to limit transformers version (see https://github.com/openvinotoolkit/openvino_tokenizers/blob/master/pyproject.toml#L46)


sammysun0711:
https://github.com/openvinotoolkit/openvino.genai/actions/runs/16643548227/job/47100821584?pr=2521
MacOS CI test failed due to transformer & pytorch compatibility: https://github.com/huggingface/transformers/issues/38329, but MacOS-13 (x86) only support pytorch 2.2

PR #2524, opened 2025-07-31
Update tokenizers

Wovchena:


tests/python_tests/utils/constants.py
```diff
@@ -45,6 +45,11 @@ def get_ov_cache_dir(temp_dir=TemporaryDirectory()):
             ov_cache = os.path.join(ov_cache, optimum_intel_version)
         except metadata.PackageNotFoundError:
             pass
+        try:
+            optimum_intel_version = metadata.version("transformers")
+            ov_cache = os.path.join(ov_cache, optimum_intel_version)
```
sammysun0711:
```suggestion
            transformers_version = metadata.version("transformers")
            ov_cache = os.path.join(ov_cache, transformers_version)
```

pavel-esir (approved):

tools/llm_bench/llm_bench_utils/hook_forward_whisper.py
```diff
@@ -18,8 +18,7 @@ def get_time_list(self):
             if 'enc_token_time' in data:
                 first_token_latency += data['enc_token_time']
             if 'dec_token_time' in data:
-                first_token_latency += data['dec_token_time'][0]
-                self.tm_list.extend(copy.deepcopy(data['dec_token_time'][1:]))
+                assert not data['dec_token_time']
```
sbalandi: Why we need to have assert here ? 
If there is no particular need for throw an exception, I would replace it with log.,warning

Wovchena: The main reason was to test that `data['dec_token_time']` is indeed empty not just in my env. After that I though someone may want to get an error as well, so I kept it and summoned you to review.

There's no way to fix the error with cmd args. Converted to warning

sbalandi: that is, theoretically there could be a dec_token_time here, but all the cases you tried didn’t contain it?

Wovchena: I only tried the cases covered by tests

sbalandi: ok, I'm sorry, I misunderstood you first. It can be not empty with another transformers, I see that you restricts transformers version here, but llm_bench still have logic for other versions and user can run it with old one, 
so, please leave the old logic and put here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/llm_bench_utils/hook_forward_whisper.py#L132
if self.greedy_hook is not None and (self.greedy_hook.get_time_list() or self.greedy_hook.get_time_infer_list()):

Wovchena: Updated

PR #2534, opened 2025-08-05
Switch to SDPA Qwen2VL and Qwen2.5VL

popovaan:
Temporarily switch to SDPA Qwen2VL and Qwen2.5VL until CVS-167316 is fixed

### DIFF
--- a/src/cpp/src/visual_language/pipeline.cpp
+++ b/src/cpp/src/visual_language/pipeline.cpp
@@ -348,6 +348,17 @@ public:
     }
 };
 
+// TODO: remove it when CVS-167316 is fixed
+bool requires_sdpa(const std::filesystem::path& models_dir) {
+    auto vlm_config = utils::from_config_json_if_exists<VLMConfig>(models_dir, "config.json");
+    if (vlm_config.model_type == VLMModelType::QWEN2_VL) {
+        return true;
+    } else if (vlm_config.model_type == VLMModelType::QWEN2_5_VL) {
+        return true;
+    }
+    return false;
+}
+
 VLMPipeline::VLMPipeline(
     const std::filesystem::path& models_dir,
     const std::string& device,
@@ -362,10 +373,10 @@ VLMPipeline::VLMPipeline(
         m_pimpl = std::make_unique<VLMPipelineImpl>(models_dir, device, properties);
     } else {
         // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
-        if (utils::explicitly_requires_paged_attention(properties)) {
+        if (utils::explicitly_requires_paged_attention(user_properties)) {
             auto [plugin_properties, scheduler_config] = utils::extract_scheduler_config(properties, utils::get_latency_oriented_scheduler_config());
             m_pimpl = std::make_unique<VLMContinuousBatchingAdapter>(models_dir, scheduler_config, device, plugin_properties);
-        } else if (attention_backend == PA_BACKEND) {
+        } else if (attention_backend == PA_BACKEND && !requires_sdpa(models_dir)) {
             // try to call CB adapter one more time, but with safe guard to silent exception
             try {
                 auto [plugin_properties, scheduler_config] = utils::extract_scheduler_config(properties, utils::get_latency_oriented_scheduler_config());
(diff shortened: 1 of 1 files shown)


src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -362,10 +373,10 @@ VLMPipeline::VLMPipeline(
         m_pimpl = std::make_unique<VLMPipelineImpl>(models_dir, device, properties);
     } else {
         // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
-        if (utils::explicitly_requires_paged_attention(properties)) {
+        if (utils::explicitly_requires_paged_attention(user_properties)) {
             auto [plugin_properties, scheduler_config] = utils::extract_scheduler_config(properties, utils::get_latency_oriented_scheduler_config());
             m_pimpl = std::make_unique<VLMContinuousBatchingAdapter>(models_dir, scheduler_config, device, plugin_properties);
-        } else if (attention_backend == PA_BACKEND) {
+        } else if (attention_backend == PA_BACKEND && !requires_sdpa(models_dir)) {
```
Wovchena, line 375: benmark_vlm blindly passes SchedulerConfig which is going to be ignored for qwen. https://github.com/openvinotoolkit/openvino.genai/blob/0a4dbc0a4a412223e9489f91251f3f82391aa4c2/samples/cpp/visual_language_chat/benchmark_vlm.cpp#L76
NPU gives an error (see above). I don't think we want to do the same for specific models handling them in benmark_vlm as well. But maybe we should...

I think the solution is to wrap `std::make_unique<ov::genai::VLMPipeline>(models_path, device, ov::genai::scheduler_config(scheduler_config))` with try except in the sample. Do you have other ideas?

Wovchena: @sbalandi, do you know if the tools pass scheduler_config anywhere by default?

popovaan: Scheduler config won't be ignored, when scheduler config is set `explicitly_requires_paged_attention()` method will return true, so `ContinuousBatching` pipeline will be executed: https://github.com/openvinotoolkit/openvino.genai/blob/0a4dbc0a4a412223e9489f91251f3f82391aa4c2/src/cpp/src/utils.cpp#L594-L597

So benmark_vlm will run CB for Qwen2VL anyway... As Qwen2VL have only accuracy issues, not performance, it should not be a problem, but the results might be confusing, as we measure PA in benmark_vlm and SDPA in WWB. I'm not sure if that's okay.

We can add try-except for Gemma, but for Qwen still CB will be executed.

Wovchena: Yeah, I mixed it with Gemma. Can you add a comment in benchmark_vlm explaining that it's going to trigger PA which is known to have accuracy issue and which is not the default backend for these models

popovaan: Yes, done.

Wovchena (approved):

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -348,6 +348,17 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
 
+// TODO: remove it when CVS-167316 is fixed
+bool requires_sdpa(const std::filesystem::path& models_dir) {
+    auto vlm_config = utils::from_config_json_if_exists<VLMConfig>(models_dir, "config.json");
+    if (vlm_config.model_type == VLMModelType::QWEN2_VL) {
+        return true;
+    } else if (vlm_config.model_type == VLMModelType::QWEN2_5_VL) {
+        return true;
+    }
+    return false;
```
yatarkan: Just a small nit/refactor
```suggestion
    return vlm_config.model_type == VLMModelType::QWEN2_VL || 
           vlm_config.model_type == VLMModelType::QWEN2_5_VL;
```
or
```suggestion
    static const std::unordered_set<VLMModelType> sdpa_required_models = {
        VLMModelType::QWEN2_VL,
        VLMModelType::QWEN2_5_VL
    };
    return sdpa_required_models.contains(vlm_config.model_type);
```

yatarkan (approved):

PR #2539, opened 2025-08-05
[NPU] Fix for whisper fp8 models

eshiryae:
Fix for exposing runtime states as inputs pass for cross attention SDPA, as latest converted models in f8e4m3 have additional FakeConvert for values tensor also.

Ticket: E-175362

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,7 +25,7 @@ env:
   SCCACHE_CACHE_SIZE: 30G
   SCCACHE_AZURE_KEY_PREFIX: genai/ubuntu/22_04/x64
   HF_HOME: /mount/caches/huggingface/lin
-  OV_CACHE: /mount/caches/huggingface/.ov_cache/lin
+  OV_CACHE: /mount/caches/huggingface/.ov_cache/lin/3ace0cbed82ce5d376658b52ca30824b5960dbc8
   GENAI_ARCHIVE_NAME: genai.tar.gz
   GENAI_SAMPLES_NAME: genai_samples.tar.gz
   ARTIFACTS_SHARE: '/mount/build-artifacts'
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -498,6 +498,13 @@ jobs:
           requirements_files: "${{ env.SRC_DIR }}/tests/python_tests/requirements.txt"
           local_wheel_dir: ${{ env.INSTALL_DIR }}/wheels
 
+      # transformers >= 4.52 require torch >= 2.6 and raise an error otherwise:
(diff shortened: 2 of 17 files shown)


src/cpp/src/whisper/pipeline_static.cpp
```diff
@@ -867,16 +877,24 @@ void expose_runtime_states_as_inputs(std::shared_ptr<ov::Model>& model) {
...
-                auto sdpa_node = rv_reader.get_node();
 
                 auto shape = rv_node->get_output_partial_shape(0);
                 auto new_param = std::make_shared<ov::op::v0::Parameter>(rv_node->get_output_element_type(0), shape);
-                
-                auto key_or_value = (sdpa_in.get_index() == 1 || is_fake_cvt) ? "key" : "value";
+
+                // In case there's additional FakeConvert: ReadValue -> FakeConvert -> SDPA
+                auto is_fc_key_tensor = false;
+                if (is_fake_cvt) {
```
smirnov-alexey: Could you make it a function to avoid code duplication?

src/cpp/src/whisper/pipeline_static.cpp
```diff
@@ -761,8 +761,18 @@ void expose_runtime_states_as_outputs(std::shared_ptr<ov::Model>& model) {
...
             if (strstr(reader.get_node()->get_type_name(), "ScaledDotProductAttention") != nullptr || is_fake_cvt) {
                 auto sdpa_in = reader;
+
+                // In case there's additional FakeConvert node: ReadValue -> FakeConvert -> SDPA
+                auto is_fc_key_tensor = false;
+                if (is_fake_cvt) {
+                    auto fc_reader = reader.get_node()->outputs()[0].get_target_inputs();
+                    // FakeConvert node has only 1 consumer
+                    OPENVINO_ASSERT(fc_reader.size() == 1);
+                    is_fc_key_tensor = fc_reader.begin()->get_index() == 1;
```
smirnov-alexey: What does it mean that index is 1?

eshiryae: By index we decide if it's `key` (it's 1st input to SDPA) or `value` (it's 2nd input to SDPA) tensor.
The same approach is used for fp16 models (there ReadValue output goes straight to SDPA, without FakeConvert layer):
[line 775](https://github.com/openvinotoolkit/openvino.genai/pull/2539/files/a481da16dfe08040abdc6a31176eda82bfdbd35e#diff-e9a19fecb7ef8f410831fddd75ee0c086b6cdf3cb1e32a81dfed1f8387a61cfdR775)

dmatveev (commented):
I echo @smirnov-alexey comments here - thanks for the quick turnaround!

src/cpp/src/whisper/pipeline_static.cpp
eshiryae: Done

PR #2540, opened 2025-08-06
eagle2 impl

songbell:


peterchen-intel:
Replaced by https://github.com/openvinotoolkit/openvino.genai/pull/2740 ?

PR #2544, opened 2025-08-06
Adding non-CB pipeline for Speculative Decoding

AsyaPronina:
- Adds Speculative Decoding pipeline, that works without Continious Batching (mainly for purpose to work with NPU)

### DIFF
--- a/src/cpp/src/llm/pipeline.cpp
+++ b/src/cpp/src/llm/pipeline.cpp
@@ -123,9 +123,11 @@ ov::genai::LLMPipeline::LLMPipeline(
     const ov::AnyMap& user_properties) :
     m_device(device) {
     auto start_time = std::chrono::steady_clock::now();
-    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
 
-    if (ov::genai::utils::is_npu_requested(device, properties)) {
+    bool is_npu_requested = ov::genai::utils::is_npu_requested(device, user_properties);
+    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties, is_npu_requested);
+
+    if (is_npu_requested) {
         m_pimpl = StatefulPipeline::create(models_path, tokenizer, device, properties);
     } else if (utils::explicitly_requires_paged_attention(user_properties)) {
         // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
@@ -160,9 +162,10 @@ ov::genai::LLMPipeline::LLMPipeline(
     m_device(device) {
     auto start_time = std::chrono::steady_clock::now();
 
-    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
+    bool is_npu_requested = ov::genai::utils::is_npu_requested(device, user_properties);
+    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties, is_npu_requested);
 
-    if (ov::genai::utils::is_npu_requested(device, properties)) {
+    if (is_npu_requested) {
         m_pimpl = StatefulPipeline::create(models_path, device, properties);
     } else if (utils::explicitly_requires_paged_attention(user_properties)) {
         // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
@@ -200,9 +203,10 @@ ov::genai::LLMPipeline::LLMPipeline(
     m_device(device) {
     auto start_time = std::chrono::steady_clock::now();
 
-    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
+    bool is_npu_requested = ov::genai::utils::is_npu_requested(device, user_properties);
+    auto [properties, attention_backend] = utils::extract_attention_backend(user_properties, is_npu_requested);
 
-    if (ov::genai::utils::is_npu_requested(device, properties)) {
+    if (is_npu_requested) {
         m_pimpl = StatefulPipeline::create(
             utils::singleton_core().read_model(model_str, weights_tensor),
             tokenizer,
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -631,7 +631,7 @@ SchedulerConfig get_latency_oriented_scheduler_config() {
     return default_config;
 }
 
-bool explicitly_requires_paged_attention(const ov::AnyMap& properties) {
+bool explicitly_requires_paged_attention(const ov::AnyMap& properties, bool is_npu_requested) {
     auto attention_backend_it = properties.find("ATTENTION_BACKEND");
 
     if (properties.find(ov::genai::scheduler_config.name()) != properties.end() ||
@@ -643,7 +643,7 @@ bool explicitly_requires_paged_attention(const ov::AnyMap& properties) {
         }
     }
 
-    if (properties.find(utils::DRAFT_MODEL_ARG_NAME) != properties.end()) {
+    if (properties.find(utils::DRAFT_MODEL_ARG_NAME) != properties.end() && !is_npu_requested) {
         if (is_paged_attention_available()) {
             return true;
         } else {
@@ -662,7 +662,8 @@ bool explicitly_requires_paged_attention(const ov::AnyMap& properties) {
     return false;
 }
 
-std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& external_properties) {
+std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& external_properties,
+                                                             bool is_npu_requested) {
     std::string attention_backend = PA_BACKEND;
     ov::AnyMap properties = external_properties;
 
@@ -674,7 +675,7 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
         properties.erase(it);
     }
 
-    if (explicitly_requires_paged_attention(external_properties)) {
+    if (explicitly_requires_paged_attention(external_properties, is_npu_requested)) {
         OPENVINO_ASSERT(attention_backend == PA_BACKEND,
             "User properties are conflicting: some of them requires PagedAttention backend, while 'ATTENTION_BACKEND' is set to 'SDPA'");
     }
--- a/src/cpp/src/utils.hpp
+++ b/src/cpp/src/utils.hpp
@@ -278,9 +278,9 @@ std::pair<ov::AnyMap, SchedulerConfig> extract_scheduler_config(const ov::AnyMap
 
 SchedulerConfig get_latency_oriented_scheduler_config();
 
-bool explicitly_requires_paged_attention(const ov::AnyMap& properties);
+bool explicitly_requires_paged_attention(const ov::AnyMap& properties, bool is_npu_requested = false);
 
-std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& external_properties);
+std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& external_properties, bool is_npu_requested = false);
 
 void save_openvino_model(const std::shared_ptr<ov::Model>& model, const std::string& save_path, bool compress_to_fp16);
 


src/cpp/src/speculative_decoding/speculative_decoding_npu.cpp
```diff
@@ -0,0 +1,662 @@
...
+        auto mismatched_candidates = candidates.size() - accepted_tokens_number;
+
+        // Phase 4: Update inference wrappers based on found matches and mismatches
+        // This is the case when main model accepted all candidates from draft model
+        // we need to collect kv cache for draft last generated token by infering it.n
+        if (mismatched_candidates == 0) {
+            draft_prefix_token = candidate;
+        } else {
+            m_draft_request->remove_last_generated_tokens(mismatched_candidates);
+            m_draft_request->trimm_kv_cache(mismatched_candidates - 1);
```
AsyaPronina: Use `position_ids` for NPU

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
```diff
@@ -0,0 +1,126 @@
...
+
+    ov::Tensor get_logits();
+
+    std::size_t get_num_processed_tokens() const;
+
+    ov::genai::GenerationHandle create_generation_handle();
+
+    void remove_last_generated_tokens(const std::size_t tokens_to_remove); 
+
+    void trimm_kv_cache(const std::size_t tokens_to_remove);
```
dmatveev: it must be `trim_` I believe

AsyaPronina: Fixed, thanks!

songbell:
I have a common question: is it able to scale to CPU/GPU as well, if we want to run speculative on non CB mode?

AsyaPronina:
> I have a common question: is it able to scale to CPU/GPU as well, if we want to run speculative on non CB mode?

Hello! It works on CPU now. But GenAI will choose this non-CB Speculative Decoding path only if we mentioned NPU in main or draft or both models. Do we need to change it to allow to run on other devices?

songbell:
> > I have a common question: is it able to scale to CPU/GPU as well, if we want to run speculative on non CB mode?
> 
> Hello! It works on CPU now. But GenAI will choose this non-CB Speculative Decoding path only if we mentioned NPU in main or draft or both models. Do we need to change it to allow to run on other devices?

I think it depends on whether we have scenarios that non CB path can have perf out-perf CB path? BTW, does it plan to  support eagle speculative decoding as well?

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+    int64_t infer_next(int64_t out_token, bool skip_perf_stat = false);
+
+    std::vector<int64_t> infer_next_return_all(const std::vector<int64_t> tokens);
+
+    ov::Tensor get_logits();
+
+    std::size_t get_num_processed_tokens() const;
+
+    ov::genai::GenerationHandle create_generation_handle();
```
AsyaPronina: Should be removed.

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+    void remove_last_generated_tokens(const std::size_t tokens_to_remove);
```
AsyaPronina: Should be removed.

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+    std::size_t get_num_processed_tokens() const;
+
+    ov::genai::GenerationHandle create_generation_handle();
+
+    void remove_last_generated_tokens(const std::size_t tokens_to_remove); 
+
+    void trim_kv_cache(const std::size_t tokens_to_remove);
+
+    ov::genai::EncodedResults finalize();
```
AsyaPronina: Should be removed.

src/cpp/src/llm/pipeline_stateful_npu.cpp
```diff
@@ -0,0 +1,100 @@
...
+}
+
+void StatefulLLMPipelineNPU::start_chat(const std::string& system_message) {
+    m_pimpl->start_chat(system_message);
+}
+
+// FIXME: Do we need it?
+// void StatefulLLMPipelineNPU::reset_kv_state() {
+//     m_pimpl->reset_kv_state();
+// }
```
smirnov-alexey: Please don't forget to address it

AsyaPronina: Thanks a lot!

src/cpp/src/llm/pipeline_stateful_npu.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include "openvino/runtime/core.hpp"
+#include "openvino/core/parallel.hpp"
+#include "openvino/genai/text_streamer.hpp"
+
+namespace {
+    ov::genai::ModelDesc
+    extract_draft_model_from_config(ov::AnyMap& config) {
+        ov::genai::ModelDesc draft_model;
+        if (config.find(ov::genai::utils::DRAFT_MODEL_ARG_NAME) != config.end()) {
```
smirnov-alexey: Maybe some throw otherwise?

src/cpp/src/utils.cpp
```diff
@@ -205,6 +204,27 @@ ProcessorConfig from_any_map(
...
+}
+
+bool is_npu_requested(const std::string& device, const ov::AnyMap& properties) {
+    if (device == "NPU") {
+        return true;
+    }
+
+    auto draft_model_descr = get_draft_model_from_config(properties);
+    if (draft_model_descr.model != nullptr) {
+        return draft_model_descr.device == "NPU";
```
smirnov-alexey, line 232: Just to double-check: it works correctly in all configurations? Draft, main on CPU/NPU

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+namespace ov {
+namespace genai {
+
+class LLMInferWrapper {
+public:
+    LLMInferWrapper(const ov::genai::ModelDesc& model_desc);
+
+    std::string device() const;
+
```
smirnov-alexey, line 20: Could you please remove some of the empty lines between the methods?

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
```diff
@@ -0,0 +1,123 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+
+#include "speculative_decoding_metrics.hpp"
+#include "llm/pipeline_base.hpp"
+#include "sampling/sampler.hpp"
+#include "utils.hpp"
+#include <openvino/genai/perf_metrics.hpp>
+#include <openvino/genai/speculative_decoding/perf_metrics.hpp>
```
smirnov-alexey: `""` instead of `<>`

src/cpp/src/speculative_decoding/speculative_decoding_npu.cpp
```diff
@@ -0,0 +1,729 @@
...
+ov::genai::GenerationConfig LLMInferWrapper::get_generation_config() const {
+    return m_generation_config;
+}
+
+void LLMInferWrapper::set_generation_config(ov::genai::GenerationConfig config) {
+    m_generation_config = config;
+}
+
+int64_t LLMInferWrapper::get_kvcache_capacity() const {
+    if (m_device == "NPU") {
```
smirnov-alexey, line 94: Why is it based on the device?

src/cpp/src/llm/pipeline_stateful_npu.cpp
AsyaPronina: No, no, this function returns draft model if it exists. If not, then pipeline will be switched just to `StatefulLLMPipeline`

src/cpp/src/utils.cpp
AsyaPronina: Yes, if either of models on `NPU`, then `is_npu_requested(...)` will return `true`.

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
AsyaPronina: I add them on purpose, because class looks a bit messy without them. Lets sync offline

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
AsyaPronina: Why?

src/cpp/src/llm/pipeline.cpp
```diff
@@ -9,10 +9,10 @@
 #include "openvino/genai/llm_pipeline.hpp"
 #include "openvino/genai/perf_metrics.hpp"
 
-#include "llm/pipeline_static.hpp"
 #include "llm/pipeline_stateful.hpp"
 #include "llm/pipeline_continuous_batching_adapter.hpp"
 #include "speculative_decoding/speculative_decoding_impl.hpp"
+#include "llm/pipeline_stateful_npu.hpp"
```
dmatveev: I believe we avoided `_npu` in the files here. That's the reason why static shape pipeline was called "static shape" but not "npu" one. Also, why do we need one more, again?

src/cpp/src/llm/pipeline_stateful_npu.cpp
dmatveev, line 1: We didn't have this before, why do we need this now? What makes it NPU specific?

src/cpp/src/speculative_decoding/speculative_decoding_impl.hpp
```diff
@@ -8,33 +8,10 @@
 #include "speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp"
 #include "speculative_decoding/speculative_decoding_metrics.hpp"
 #include "openvino/genai/speculative_decoding/perf_metrics.hpp"
+#include "utils.hpp"
```
dmatveev, line 11: are you sure it is the right location?

src/cpp/src/speculative_decoding/speculative_decoding_npu.cpp
```diff
@@ -0,0 +1,729 @@
...
+    return m_request.reset_state();
+}
+
+void LLMInferWrapper::release_memory() {
+    m_request.get_compiled_model().release_memory();
+}
+
+void LLMInferWrapper::set_already_allocated_input_for_1_token() {
+    m_request.set_tensor("input_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1},  reinterpret_cast<void*>(&m_new_input_token)));
+    m_request.set_tensor("position_ids", ov::Tensor(ov::element::i64, ov::Shape{1,1}, reinterpret_cast<void*>(&m_new_position_id)));
```
dmatveev, line 321: why do you cast to void?

dmatveev (commented):
It shouldn't be that npu-ish

src/cpp/src/llm/pipeline.cpp
AsyaPronina: Because there become a lot of specific logic in the GenAI for NPU : separate static pipeline that can be requested, Speculative Decoding pipeline, that doesn't use Continuous Batching.

For example, `llm\pipeline.cpp`, that handled such NPU-specific logic previously: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/llm/pipeline.cpp#L87-L9, doesn't explicitly seeks for draft model in properties to call CB Speculative Decoding. It just calls Continuous Batching Pipeline, that itself seeks for a draft model to dispatch which CB pipeline implementation to call to in the end: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching/pipeline.cpp#L76.

So, the separate class is created to handle all specific `if`-s for dispatching of NPU pipelines inside itself to not load the `LLMPipeline` class with additional `if`-s.

AsyaPronina:
AR APr: Add class diagram and sequence diagram
Aim is to reduce NPU-specific logic

src/cpp/src/speculative_decoding/speculative_decoding_npu.hpp
smirnov-alexey: We use `<>` for std headers and `""` for repo files, it's the case for all other files, so I'd keep in consistent

src/cpp/src/speculative_decoding/speculative_decoding_impl.cpp
```diff
@@ -222,6 +222,7 @@ void ContinuousBatchingPipeline::SpeculativeDecodingImpl::step() {
         auto main_model_gen_duration = main_timer.get_duration_microsec();
         auto m_main_pipeline_metrics = m_main_pipeline->get_metrics();
         main_raw_perf_counters.m_durations.push_back(MicroSeconds(main_model_gen_duration));
+        // TODO: Ask about +=
         main_raw_perf_counters.m_inference_durations[0] = MicroSeconds(m_main_pipeline_metrics.inference_duration);
```
AsyaPronina: Hello Dear @sbalandi!
Could you please say, do you think the `=` is a typo and `+=` should be used here?

sbalandi: Hello @AsyaPronina ! yes, it looks like typo
also here https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp#L323

AsyaPronina: Thanks, will fix!

src/cpp/src/speculative_decoding/speculative_decoding_impl.hpp
AsyaPronina: Yes, it is done specifically to relocate `ModelDesc` from this file to `utils.hpp`, as I am using the same `ModelDesc` in Speculative Decoding implementation

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
dmatveev, line 1: Please make sure to review it with the GenAI team

src/cpp/src/speculative_decoding/speculative_decoding_impl.cpp
dmatveev, line 1: This too

src/cpp/src/utils.cpp
```diff
@@ -205,6 +205,36 @@ ProcessorConfig from_any_map(
...
 
+ov::genai::ModelDesc get_draft_model_from_config(const ov::AnyMap& config) {
+    ov::genai::ModelDesc draft_model;
+    if (config.find(utils::DRAFT_MODEL_ARG_NAME) != config.end()) {
+        draft_model = config.at(utils::DRAFT_MODEL_ARG_NAME).as<ov::genai::ModelDesc>();
+    }
+    return draft_model;
+}
+
+ov::genai::ModelDesc extract_draft_model_from_config(ov::AnyMap& config) {
```
dmatveev, line 216: should it be `pop_` instead? Like we've had with `pop_`options?

tools/llm_bench/llm_bench_utils/model_utils.py
dmatveev, line 1: Please review it with the GenAI team

src/cpp/src/continuous_batching/timer.hpp
dmatveev, line 1: Please review it with the GenAI team

src/cpp/src/llm/pipeline.cpp
```diff
@@ -60,6 +60,55 @@ std::pair<std::string, Any> draft_model(
     return { utils::DRAFT_MODEL_ARG_NAME, Any::make<ModelDesc>(model, tokenizer, device, plugin_config, scheduler_config, generation_config) };
 }
 
+// NOTE: Should be used only when NPU device is requested
+// either for main model or for draft model if last exists.
+class StatefulNPUPipelineCreator {
```
dmatveev:
```suggestion
class StatefulPipeline {
```

src/cpp/src/llm/pipeline.cpp
```diff
@@ -80,14 +129,12 @@ ov::genai::LLMPipeline::LLMPipeline(
     auto start_time = std::chrono::steady_clock::now();
     auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
 
-    // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
-    if (utils::explicitly_requires_paged_attention(user_properties)) {
+    if (ov::genai::utils::is_npu_requested(device, properties)) {
+        m_pimpl = StatefulNPUPipelineCreator::create(models_path, tokenizer, device, properties);
```
dmatveev:
```suggestion
    if (ov::genai::utils::is_npu_requested(device, properties)) {
        m_pimpl = StatefulPipeline::create(models_path, tokenizer, device, properties);
```

src/cpp/src/llm/pipeline.cpp
```diff
@@ -60,6 +60,55 @@ std::pair<std::string, Any> draft_model(
     return { utils::DRAFT_MODEL_ARG_NAME, Any::make<ModelDesc>(model, tokenizer, device, plugin_config, scheduler_config, generation_config) };
 }
 
+// NOTE: Should be used only when NPU device is requested
+// either for main model or for draft model if last exists.
```
dmatveev: I'd remove this limitation and let is also create the classic `StatefulLLMPipeline`.

AsyaPronina: Thanks! Discussed with GenAI team, will rework.

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -202,9 +202,10 @@ def analyze_args(args):
     if args.cb_config:
         cb_config = get_config(args.cb_config)
     model_args["cb_config"] = cb_config
-    if args.draft_model and (args.device == "NPU" or model_args['config']['ATTENTION_BACKEND'] != PA_ATTENTION_BACKEND):
-        log.warning("Speculative Decoding is supported only with Page Attention Backend and not supported for NPU device")
-        args.draft_model = None
+    if args.draft_model:
+        if (args.draft_device != "NPU" and args.device != "NPU" and model_args['config']['ATTENTION_BACKEND'] != PA_ATTENTION_BACKEND):
+            log.warning("Speculative Decoding is supported only with Page Attention Backend for non-NPU devices")
+            args.draft_model = None
```
AsyaPronina, line 215: Update for Non-PA models

src/cpp/src/utils.cpp
asyadeveloper: I just re-used the same name that it has previously, I just relocated that function from `src/cpp/src/continuous_batching/pipeline.cpp`

samples/python/text_generation/speculative_decoding_lm.py
Wovchena, line 1: The accompanying Accuracy_Performance.xlsx doesn't contain perf data comparison with greedy. The point of speculative decoding is to outperform the corresponding single model sampling implementation. If you are going to collect that data it's worth paying attention to text generated by greedy. It should match speculative decoding exactly (if the device is the same), you can cover that in tests as well.

Does random sampling work?

I won't be able review in time, so I'd rely on @as-suvorov, @popovaan, @sbalandi

src/cpp/src/llm/pipeline.cpp
```diff
@@ -80,14 +121,12 @@ ov::genai::LLMPipeline::LLMPipeline(
     auto start_time = std::chrono::steady_clock::now();
     auto [properties, attention_backend] = utils::extract_attention_backend(user_properties);
 
-    // If CB is invoked explicitly, create CB adapter as is and re-throw in case if internal issues
-    if (utils::explicitly_requires_paged_attention(user_properties)) {
+    if (ov::genai::utils::is_npu_requested(device, properties)) {
```
AlexanderKalistratov: So we basically have 3 constructors with different parameters.
The logic is looking complex enough not to repeat it 3 times.

We could try to use constructor delegation, move the logic into separate function with most generic parameters or (least preferrable) use template function with parameters pack and forwarding.

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,779 @@
...
+
+void update_perf_metrics(ov::genai::RawPerfMetrics& raw_perf_counters, const float duration_microsec,
+                         const ov::genai::TimePoint new_token_time, const std::size_t num_generated_tokens) {
+    raw_perf_counters.m_token_infer_durations.emplace_back(duration_microsec);
+    raw_perf_counters.m_inference_durations[0] += ov::genai::MicroSeconds(duration_microsec);
+    raw_perf_counters.m_new_token_times.emplace_back(new_token_time);
+    raw_perf_counters.m_batch_sizes.emplace_back(num_generated_tokens);
+}
+
+void update_perf_metrics(ov::genai::RawPerfMetrics& raw_perf_counters,
```
AlexanderKalistratov: Two `update_perf_metrics` functions updates different metrics and seems to supposed to be applied to different object.

I would recommend not to use function overloading and give these two function different names clearly stating functions intention

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,779 @@
...
+    m_draft_request->reset_state();
+    m_main_request->reset_state();
+};
+
+StatefulSpeculativeLLMPipeline::~StatefulSpeculativeLLMPipeline() {
+    m_main_request->release_memory();
+    m_draft_request->release_memory();
+}
+
+void StatefulSpeculativeLLMPipeline::update_candidate_strategy(const std::size_t matches_num) {
```
sbalandi, line 807: I just didn't fully get the matching between num_assistant_tokens and accepted_tokens_number in CB. If I understand correctly, the accept_tokens_number strategy doesn't work in statefull, num_assistant_tokens works, but it's not strictly enforced, and if num_assistant_tokens is not set, it will default to 5 ? I think it would be good if setting of the value was consistent with the common generation_config, and if num_assistant_tokens is not set, pipeline fails and other nuances should be reflected in generation_config.

src/cpp/src/speculative_decoding/speculative_decoding_npu.cpp
AsyaPronina: It has constructor of that type:
```cpp

    /**
     * @brief Constructs Tensor using element type and shape. Wraps allocated host memory.
     * @note Does not perform memory allocation internally
     * @param type Tensor element type
     * @param shape Tensor shape
     * @param host_ptr Pointer to pre-allocated host memory with initialized objects
     * @param strides Optional strides parameters in bytes. Strides are supposed to be computed automatically based
     * on shape and element size
     */
    Tensor(const element::Type& type, const Shape& shape, void* host_ptr, const Strides& strides = {});
```

samples/python/text_generation/speculative_decoding_lm.py
AsyaPronina: I will share results of validation against stateful pipepline with greedy decoding. Thanks!

Random sampling doesn't work for now.

src/cpp/src/llm/pipeline.cpp
AsyaPronina: Discussed internally, will create a follow up refactoring ticket

src/cpp/src/continuous_batching/timer.hpp
```diff
@@ -14,6 +14,8 @@ class ManualTimer {
 public:
     ManualTimer(const std::string& title) :
         m_total(0.),
+        m_start(std::chrono::steady_clock::duration::zero()),
+        m_end(std::chrono::steady_clock::duration::zero()),
```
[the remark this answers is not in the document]
AsyaPronina: Fixed, thanks!

src/cpp/src/continuous_batching/timer.hpp
```diff
@@ -42,6 +44,14 @@ class ManualTimer {
         return m_total;
     }
 
+    void clear() {
+        m_total = 0.0;
+        m_start = std::chrono::steady_clock::time_point(
+            std::chrono::steady_clock::duration::zero());
+        m_end = std::chrono::steady_clock::time_point(
+            std::chrono::steady_clock::duration::zero());
```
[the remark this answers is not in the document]
AsyaPronina: Fixed, thanks!

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,779 @@
...
+    // Dynamically adjust number of generated candidates based on number of matches,
+    // we want to balance the benefits of getting candidates tokens correct with the
+    // cost of forecasting incorrect candidates tokens.
+    if (matches_num == m_candidates_num) {
+        m_candidates_num = std::min(m_candidates_num + 2, m_max_candidates_num);
+    } else {
+        m_candidates_num = std::max(int64_t(m_candidates_num) - 1, int64_t(1));
```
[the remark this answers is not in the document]
AsyaPronina: Fixed, thanks!

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
AsyaPronina: Fixed, thanks!

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
AsyaPronina: Thanks a lot for the question, let me contact you on that matter!

samples/python/text_generation/speculative_decoding_lm.py
Wovchena: Add a test verifying that optimum-intel greedily generated text matches exactly new speculative decoding. That should cover the alignment with optimum and the case I described earlier.

There are tests validating perf metrics for LLMs: https://github.com/openvinotoolkit/openvino.genai/blob/384084eddc5ce2e52a8f643cf6e330620bf2a34c/tests/python_tests/test_llm_pipeline.py#L680 and https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_continuous_batching.py Add your case.

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
AsyaPronina: Discussed - fixed, thanks a lot!

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -274,6 +274,8 @@ operator|(const StructuredOutputConfig::CompoundGrammar& lhs,
  * Assisting generation parameters:
  * @param assistant_confidence_threshold the lower token probability of candidate to be validated by main model in case of dynamic strategy candidates number update.
  * @param num_assistant_tokens the defined candidates number to be generated by draft model/prompt lookup in case of static strategy candidates number update.
+ *        NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial value and adjusts it
```
sbalandi: please, add that assistant_confidence_threshold is not supported for stateful

AsyaPronina:
Final alignment on `GenerationConfig.num_assistant_tokens` utilization between ContinuousBatching and Stateful Speculative Decodings:

> ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set it will be defaulted to 5 for both backends.

Wovchena:
If `num_assistant_tokens` is not set, it defaults to 5 for both backends.

zero conditional

sbalandi (approved):

samples/python/text_generation/speculative_decoding_lm.py
```diff
@@ -29,10 +31,15 @@ def main():
...
-    # Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded
-    # add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration
-    config.num_assistant_tokens = 5
-    # add parameter to enable speculative decoding to generate candidates by draft_model while candidate probability is higher than `assistant_confidence_threshold`
+    # Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded.
+    # Add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration.
+    # NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial
+    # value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both
+    # backends.
+    # config.num_assistant_tokens = 5
```
sbalandi: please, leave setting uncomments, it could be set to 4, for cpp too

AsyaPronina: Fixed, thanks!

src/cpp/src/utils.hpp
```diff
@@ -12,9 +12,10 @@
 #include "openvino/runtime/core.hpp"
 
 #include "openvino/genai/generation_handle.hpp"
```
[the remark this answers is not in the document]
AsyaPronina: Thank you, I think I will leave as is

src/cpp/src/utils.hpp
```diff
@@ -12,9 +12,10 @@
...
+#include "openvino/genai/scheduler_config.hpp"
+#include "openvino/genai/generation_config.hpp"
 #include "visual_language/processor_config.hpp"
 
-#include "openvino/genai/generation_handle.hpp"
 #include "openvino/genai/streamer_base.hpp"
```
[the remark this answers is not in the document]
AsyaPronina: Thank you, I think I will leave as is

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+}
+
+void StatefulSpeculativeLLMPipeline::update_candidate_strategy(const std::size_t matches_num) {
+    // Dynamically adjust number of generated candidates based on number of matches,
+    // we want to balance the benefits of getting candidates tokens correct with the
+    // cost of forecasting incorrect candidates tokens.
+    if (matches_num == m_candidates_num) {
+        m_candidates_num = std::min(m_candidates_num + 2, m_max_candidates_num);
+    } else {
+        m_candidates_num = static_cast<std::size_t>(std::max(static_cast<int64_t>(m_candidates_num) - 1, int64_t(1)));
```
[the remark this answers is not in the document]
AsyaPronina: If `m_candidates_num` is unsigned, then `m_candidates_num - 1` could lead to overflow and be chosen from set of `{overflowed_number, 1}` what is not intended here

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+            // We don't need last candidate in KVCache of draft model, as
+            // it fails validation.
+            draft_prefix_token = -1;
+        }
+        update_candidate_strategy(accepted_tokens_number);
+
+        auto& main_perf_generated_tokens = m_main_request->raw_perf_metrics.m_batch_sizes.back();
+        main_perf_generated_tokens -= mismatched_candidates;
+        m_sd_metrics.update_draft_generated_len(0 /* request_id */, candidates_to_generate);
+        m_sd_metrics.update_acceptance_rate(0 /* request_id */, (accepted_tokens_number /  candidates_to_generate) * 100);
```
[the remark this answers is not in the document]
AsyaPronina: Great catch! Fixed, thanks!

samples/python/text_generation/speculative_decoding_lm.py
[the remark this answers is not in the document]
AsyaPronina: Fixed after Sofia's comment

samples/python/text_generation/speculative_decoding_lm.py
AsyaPronina: Done, thanks!!

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+    }
+    if (draft_model_desc_copy.properties.empty() && (draft_model_desc_copy.device == main_model_desc.device)) {
+        draft_model_desc_copy.properties = main_model_desc.properties;
+    }
+    m_draft_request = std::make_unique<LLMInferWrapper>(draft_model_desc_copy);
+
+    // Specifying number candidates to generate
+    ensure_num_assistant_tokens_is_set(m_generation_config);
+    m_candidates_num = m_generation_config.num_assistant_tokens;
+    m_max_candidates_num = m_candidates_num * 2;
```
AsyaPronina, line 395: Add comment why we do this

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -14,6 +14,11 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::Contin
     bool is_validation_mode_enabled) {
     m_tokenizer = tokenizer;
     m_generation_config = generation_config;
+    if (m_generation_config.assistant_confidence_threshold == 0.f) {
+        if (m_generation_config.num_assistant_tokens == 0) {
+            m_generation_config.num_assistant_tokens = 5;
```
AsyaPronina: Make as constant

src/cpp/src/speculative_decoding/speculative_decoding_impl.cpp
```diff
@@ -262,9 +262,15 @@ ContinuousBatchingPipeline::SpeculativeDecodingImpl::generate(const std::vector<
     std::vector<GenerationHandle> main_generations;
     for (size_t request_id = 0; request_id < input_ids.size(); ++request_id) {
         OPENVINO_ASSERT(1 == input_ids[request_id].get_shape().at(0), "Use multiple tensors to pass a batch.");
-        main_generations.push_back(m_main_pipeline->add_request(request_id, input_ids[request_id], sampling_params[request_id]));
+        auto main_sampling_params = sampling_params[request_id];
+        if (main_sampling_params.assistant_confidence_threshold == 0.f) {
+            if (main_sampling_params.num_assistant_tokens == 0) {
+                main_sampling_params.num_assistant_tokens = 5;
```
AsyaPronina: make as constant

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+        raw_perf_metrics.m_durations.back() +=
+            ov::genai::MicroSeconds(infer_next_timer.get_duration_microsec());
+        raw_perf_metrics.m_inference_durations[0] += 
+            ov::genai::MicroSeconds(ov::genai::PerfMetrics::get_microsec(infer_end - infer_start));
+    }
+
+    return last_token;
+}
+
+std::vector<int64_t> LLMInferWrapper::infer_next_return_all(const std::vector<int64_t> tokens) {
```
AsyaPronina: add reference

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+    infer_next_return_all_timer.start();
+
+    size_t tokens_size = tokens.size();
+    auto input_ids = m_request.get_tensor("input_ids");
+    ov::Tensor new_input_ids(input_ids.get_element_type(), ov::Shape{BATCH_SIZE, tokens_size});
+    std::copy_n(tokens.begin(), tokens_size, new_input_ids.data<int64_t>());
+    m_request.set_tensor("input_ids", new_input_ids);
+
+    auto attention_mask = m_request.get_tensor("attention_mask");
+    ov::Tensor new_attention_mask(attention_mask.get_element_type(), ov::Shape{BATCH_SIZE, m_num_processed_tokens + tokens_size});
```
AsyaPronina, line 236: Create a foolow-up for performance imrpovement and removing of allocation

AsyaPronina: Check why not just reshape : old issue?

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+    OPENVINO_ASSERT(m_num_processed_tokens > 0, "trim_kv_cache() can be called only after infer_first()!");
+
+    OPENVINO_ASSERT(tokens_to_remove < m_num_processed_tokens);
+    // For NPU "trim" is done by position ids on NPUW side.
+    if (m_device != "NPU") {
+        // Trim kv_cache values on tokens_to_remove
+        ov::genai::utils::KVCacheState to_trim_state;
+        to_trim_state.num_tokens_to_trim = tokens_to_remove;
+        to_trim_state.seq_length_axis =  m_kv_pos.seq_len;
+        to_trim_state.reset_mem_state = false;
```
AsyaPronina, line 294: Check if it is needed for LoRA states?

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+    }
+    // If stop_token_ids were not provided, take value from default m_generation_config
+    if (config.stop_token_ids.empty())
+        config.stop_token_ids = m_generation_config.stop_token_ids;
+    // If eos_token_id was not provided, take value from default m_generation_config
+    if (config.eos_token_id == -1)
+        config.set_eos_token_id(m_generation_config.eos_token_id);
+    config.validate();
+
+    OPENVINO_ASSERT(config.is_greedy_decoding() || config.is_multinomial(),
```
AsyaPronina: Remove multinomial

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+    OPENVINO_ASSERT(config.num_return_sequences == 1u,
+        "Currently only \"num_return_sequences\" equal to 1 is supported!");
+
+    m_main_request->set_generation_config(config);
+
+    // Config draft model to not stop on EOS and remove stop strings:
+    ov::genai::GenerationConfig draft_config = m_draft_request->get_generation_config();
+    draft_config.ignore_eos = true;
+    draft_config.stop_strings = {};
+    draft_config.max_new_tokens = config.get_max_new_tokens();
```
AsyaPronina, line 541: to remove? Will it work?

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+    // Last generated token by draft model needs to be prepended before next run if it is accepted by the main model!
+    // So it will get into the context too.
+    int64_t draft_prefix_token = -1;
+    while (m_main_request->can_infer() && (streaming_status == ov::genai::StreamingStatus::RUNNING)) {
+        iteration_timer.start();
+
+        // Phase 1: Generation of candidates with the draft model.
+        candidates_timer.start();
+
+        std::vector<int64_t> candidates;
```
AsyaPronina, line 612: Comment logic of capacities in more details

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,796 @@
...
+
+    auto streaming_status = stream_generated_tokens(streamer_ptr, std::vector<int64_t> {out_token});
+    results.tokens[0].push_back(out_token);
+
+    // Creating timers for performance metrics calculation:
+    ManualTimer iteration_timer("Speculative decode: infer iteration");
+    ManualTimer candidates_timer("Draft model: candidates generation");
+    ManualTimer main_timer("Main model");
+
+    /* Speculative decoding works the following way. The draft model predicts the next K
```
AsyaPronina: Need to fix

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
AsyaPronina: Seems like no: [src/cpp/src/speculative_decoding/speculative_decoding_stateful.hpp](https://github.com/openvinotoolkit/openvino/pull/31433/files#diff-c5ef1bc36a918c8519581ec720cb4488c9a3736dd6c88f371a63bea898801c06R28)

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
AsyaPronina: Removed

AsyaPronina: Returned back as needed to be setup when `ignore_eos` is set to `true`. Currently returned as is, as this parameter doesn't influence on pipeline execution.

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -14,6 +14,11 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::Contin
     bool is_validation_mode_enabled) {
     m_tokenizer = tokenizer;
     m_generation_config = generation_config;
+    if (m_generation_config.assistant_confidence_threshold == 0.f) {
```
as-suvorov, line 17: It's redundant as generation config has assert for this case: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/generation_config.cpp#L342

It also seems to change default behavior where we explicitly ask user to provide `num_assistant_tokens` or `assistant_confidence_threshold`.

Let's remove this checks and also `default_num_assistant_tokens` to keep current behavior. What's your thoughts?

samples/python/text_generation/speculative_decoding_lm.py
```diff
@@ -29,10 +31,15 @@ def main():
...
-    # Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded
-    # add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration
-    config.num_assistant_tokens = 5
-    # add parameter to enable speculative decoding to generate candidates by draft_model while candidate probability is higher than `assistant_confidence_threshold`
+    # Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded.
+    # Add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration.
+    # NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial
+    # value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both
+    # backends.
+    config.num_assistant_tokens = 4
```
as-suvorov, line 38:
```suggestion
    config.num_assistant_tokens = 5
```

samples/cpp/text_generation/speculative_decoding_lm.cpp
```diff
@@ -13,10 +13,15 @@ int main(int argc, char* argv[]) try {
...
-    // Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded
-    // add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration
-    config.num_assistant_tokens = 5;
-    // add parameter to enable speculative decoding to generate candidates by draft_model while candidate probability is higher than `assistant_confidence_threshold`
+    // Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded.
+    // Add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration.
+    // NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial
+    // value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both
+    // backends.
+    config.num_assistant_tokens = 4;
```
as-suvorov, line 21:
```suggestion
    config.num_assistant_tokens = 5;
```

tests/python_tests/test_stateful_speculative_decoding.py
```diff
@@ -0,0 +1,205 @@
...
+    #        It seems like temporary directory from model downloading stage isn't removed after test
+    #        launch for SmolLM2-360M model, that is why it is not used now.
+    MODEL_UNDER_TEST = {
+        "name": "HuggingFaceTB/SmolLM2-135M",
+        "convert_args": ['--trust-remote-code']
+    }
+    prompt = "Alan Turing was a"
+
+    # Download and convert model:
+    main_opt_model, main_hf_tokenizer, main_model_path = download_and_convert_model(MODEL_UNDER_TEST["name"])
```
as-suvorov: Please parametrize model name

tests/python_tests/test_stateful_speculative_decoding.py
```diff
@@ -0,0 +1,205 @@
...
+    ('NPUW:CPU', 'NPUW:CPU')
+]
+@pytest.mark.parametrize("main_device,draft_device", test_cases)
+@pytest.mark.precommit
+def test_string_inputs(main_device, draft_device):
+    # FIXME: For now SmolLM2-135M is used as a main and a draft model in the test.
+    #        However, it is more desirable to use SmolLM2-360M as a main one to simulate the real case
+    #        for speculative decoding.
+    #        It seems like temporary directory from model downloading stage isn't removed after test
+    #        launch for SmolLM2-360M model, that is why it is not used now.
```
as-suvorov: There is no temporary directory for the models in the CI. Models are downloaded to OV_CACHE path which is shared between workflows.
You can use SmolLM2-360M here.

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
AsyaPronina: Hi, this was done specifically to align behaviour between CB and Stateful Speculative Decode per agreement with Sofia (@sbalandi) and Vladimir (@Wovchena), because for Speculative Decoding it seems that user can just provide a draft model and setting `num_assistant_tokens` might be extra step. However, samples still reflect this step

samples/python/text_generation/speculative_decoding_lm.py
AsyaPronina: It was done specially per agreement with Sofia (@sbalandi ), because `5` is our default behaviour now

tests/python_tests/test_stateful_speculative_decoding.py
AsyaPronina: Thank you!!

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
as-suvorov: Yeah, I've missed that chain, thanks, agree

tests/python_tests/test_stateful_speculative_decoding.py
AsyaPronina: Fixed, thanks!

tests/python_tests/test_stateful_speculative_decoding.py
AsyaPronina: Fixed, thanks!

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
AsyaPronina: Issue is reproduced here:
```bash
RuntimeError: Exception from src\core\src\runtime\tensor.cpp:85:
Check 'shape_size(new_shape) <= ov::shape_size(m_capacity)' failed at src\inference\src\dev\make_tensor.cpp:96:       
Could set new shape: [1,5]
```

AsyaPronina:
Here is a commit to restrict StatefulSpeculativeLLMPipeline to work only if one of the models is requested to be executed on NPU currently: https://github.com/openvinotoolkit/openvino.genai/pull/2544/commits/efcbb66e8b14fba9567a06ae9f1fb15aac06bfbd

Please add thumbs-up to approve it or comment if you don't agree: @dmatveev @as-suvorov  @sbalandi

samples/cpp/text_generation/speculative_decoding_lm.cpp
```diff
@@ -25,6 +30,8 @@ int main(int argc, char* argv[]) try {
 
     // User can run main and draft model on different devices.
     // Please, set device for main model in `LLMPipeline` constructor and in in `ov::genai::draft_model` for draft.
+    // CPU, GPU and NPU can be used. Please be aware that GPU is performant only with Continuous Batching pipeline, so it is not recommented
+    // to use it in conjuction with NPU.
```
dmatveev: I wouldn't put it this way here. Let's not make statements on other IPs with our changes.
I'd rather put it like for NPU, NPU+NPU configuration is preferrable - and that's it.

samples/python/text_generation/speculative_decoding_lm.py
```diff
@@ -19,8 +19,10 @@ def main():
     args = parser.parse_args()
 
     # User can run main and draft model on different devices.
-    # Please, set device for main model in `openvino_genai.LLMPipeline` constructor and in openvino_genai.draft_model` for draft.
-    main_device = 'CPU'  # GPU can be used as well
+    # Please, set device for main model in `openvino_genai.LLMPipeline` constructor and in `openvino_genai.draft_model` for draft.
+    # CPU, GPU and NPU can be used. Please be aware that GPU is performant only with Continuous Batching pipeline, so it is not
+    # recommented to use it in conjuction with NPU.
```
dmatveev: Lets avoid these statements, see my past comment.

src/cpp/src/llm/pipeline.cpp
```diff
@@ -60,6 +60,51 @@ std::pair<std::string, Any> draft_model(
...
+    const ov::AnyMap& properties,
+    const ov::genai::GenerationConfig& generation_config) {
+
+    auto properties_without_draft_model = properties;
+    auto draft_model_descr = ov::genai::utils::extract_draft_model_from_config(properties_without_draft_model);
+    if (draft_model_descr.model != nullptr) {
+        // FIXME: Add support for StatefulSpeculativeLLMPipeline for non-NPU devices for both models.
+        OPENVINO_ASSERT(device == "NPU" || draft_model_descr.device == "NPU",
+            "Stateful Speculative Decoding is expected to be launched when NPU is requsted as "
+            "execution device for one or both models.");
```
dmatveev, line 98:
```suggestion
            "Stateful Speculative Decoding pipeline should only run on NPU");
```

src/cpp/src/llm/pipeline.cpp
```diff
@@ -60,6 +60,51 @@ std::pair<std::string, Any> draft_model(
...
+    const ov::genai::Tokenizer& tokenizer,
+    const std::string& device,
+    const ov::AnyMap& properties,
+    const ov::genai::GenerationConfig& generation_config) {
+
+    auto properties_without_draft_model = properties;
+    auto draft_model_descr = ov::genai::utils::extract_draft_model_from_config(properties_without_draft_model);
+    if (draft_model_descr.model != nullptr) {
+        // FIXME: Add support for StatefulSpeculativeLLMPipeline for non-NPU devices for both models.
+        OPENVINO_ASSERT(device == "NPU" || draft_model_descr.device == "NPU",
```
dmatveev, line 96: don't we have
```
(ov::genai::utils::is_npu_requested(device, properties)
```
can this work here as well?

src/cpp/src/utils.cpp
dmatveev: okay so it is not ours naming convention here

dmatveev (commented):
Note - reviewed integration-related changes only. Generally I don't have issues with limiting it to NPU+NPU case only

src/cpp/src/llm/pipeline.cpp
AsyaPronina: It is not a public message, we shouldn't step into this assert, it is for internal testing, that we haven't allowed `StatefulSpeculativeLLMPipeline` to work with other than `NPU` devices to early (with removing of another assert in `explicitly_requires_paged_attention()` code)

src/cpp/src/llm/pipeline.cpp
AsyaPronina: Hello! It will work 100%, however I would like to leave the check as is as it is an internal sanity check that we haven't enabled `StatefulSpeculativeLLMPipeline` too early

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -0,0 +1,818 @@
...
+    // If stop_token_ids were not provided, take value from default m_generation_config
+    if (config.stop_token_ids.empty())
+        config.stop_token_ids = m_generation_config.stop_token_ids;
+    // If eos_token_id was not provided, take value from default m_generation_config
+    if (config.eos_token_id == -1)
+        config.set_eos_token_id(m_generation_config.eos_token_id);
+    config.validate();
+
+    OPENVINO_ASSERT(config.is_greedy_decoding(),
+        "Currently only greedy decoding are supported");
```
AsyaPronina: is

PR #2545, opened 2025-08-06
[OV JS] Implement PerfMetrics for the LLMPipeline

almilosz:
### Details:
- Add class `DecodedResults` in _llmPipeline.ts_ 
- Expose class `PerfMetrics` in js perf_metrics.hpp and perf_metrics.cpp
- Update _samples/js/text_generation/beam_search_causal_lm.js_ to return the same output without warning. 
- Add tests for `generationConfig['return_decoded_results']` and perfMetrics similar to python api [test](https://github.com/openvinotoolkit/openvino.genai/blob/d9c1e5d48eb35ceb56a28a9ca042abf4b556a169/tests/python_tests/test_llm_pipeline.py#L680)
- Till now `llmPipeline.generate()` always returned `string`. I added `generationConfig['return_decoded_results']` option to get `DecodedResults` and used `console.warn` to give warning that since 2026.0 llmPipeline.generate() will return `DecodedResults` instead string. This way JS api will be aligned with C++ API. 
In PythonAPI string is return only when input is string and [num_return_sequences](https://github.com/openvinotoolkit/openvino.genai/blob/0a4dbc0a4a412223e9489f91251f3f82391aa4c2/src/python/py_llm_pipeline.cpp#L87) is 1, which in my opinion can be confusing.
 **TO_DISCUSS**: Is there other way to support generate returning `DecodedResults` without breaking backward compatibility? If no, when `generationConfig['return_decoded_results']` should be deprecated?


src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -8,6 +8,40 @@ export type Options = {
   'max_new_tokens'?: number
 };
 
+class DecodedResults {
+  constructor(texts: string[] = [],
+    scores: number[] = [],
+    perfMetrics: any = {}) {
```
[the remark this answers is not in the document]
almilosz: Added

src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -8,6 +8,40 @@ export type Options = {
...
+      result += `${this.scores[i].toFixed(6)}: ${this.texts[i]}\n`;
+    }
+    result += `${this.scores[this.scores.length - 1]
+      .toFixed(6)}: ${this.texts[this.texts.length - 1]}`;
+
+    return result;
+  }
+  texts: string[];
+  scores: number[];
+  perfMetrics: any;
```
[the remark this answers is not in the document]
almilosz: Added

src/js/include/helper.hpp
```diff
@@ -58,4 +65,44 @@ Napi::Value cpp_to_js<ov::genai::EmbeddingResults, Napi::Value>(
...
+    const auto& durations = instance.*member;
+    res.reserve(durations.size());
+    std::transform(durations.begin(), durations.end(), std::back_inserter(res),
+                   [](const auto& duration) { return duration.count(); });
+    return res;
+}
+
+// function from src/python/py_perf_metrics.cpp
+template <typename T, typename U>
+std::vector<double> timestamp_to_ms(const T& instance, U T::*member) {
```
Wovchena: Move to a separate header that would be included by py_perf_metrics.cpp and helper.hpp. You can make any dir to be include dir in https://github.com/openvinotoolkit/openvino.genai/blob/99fb07961bd733d82370b99beb29833cb7c8110f/src/js/CMakeLists.txt#L49C1-L49C27

src/js/tests/module.test.js
```diff
@@ -159,6 +161,119 @@ describe('generation parameters validation', () => {
...
+
+  it('DecodedResults.perfMetrics', async () => {
+    const config = {
+      'max_new_tokens': 20,
+      'return_decoded_results': true,
+    };
+    const prompt = 'The Sky is blue because';
+    const start = hrtime.bigint();
+    pipeline = await LLMPipeline(MODEL_PATH, 'CPU');
+    await pipeline.startChat();
```
yatarkan, line 204: Why do we need to re-create LLMPipeline here? It should be available from `before()`

almilosz: I recreated python api tests where `loadTime` was supposed to be smaller than creating pipeline and call to generate() 
https://github.com/openvinotoolkit/openvino.genai/blob/d9c1e5d48eb35ceb56a28a9ca042abf4b556a169/tests/python_tests/test_llm_pipeline.py#L682-L685

almilosz:
Other features/refactors that are planned for next releases: 
- Add PerfMetrics.get_grammar_compiler_init_times() and PerfMetrics.get_grammar_compile_time()
- Expose ExtendedPerfMetrics

Wovchena (approved):

PR #2563, opened 2025-08-08
[GGUF] Fix Q4_1 accuracy

wine99:
Q4_1 model's output was irrational before this fix.


sammysun0711:
LGTM, thanks for your contribution, please add a test case for Q4_1 in python_test, e.g. https://github.com/openvinotoolkit/openvino.genai/blob/f55880da109a4706778c782a097bc28b1e206817/tests/python_tests/data/models.py#L33-L37

It will be used in GGUF Reader test in pre commit: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_gguf_reader.py

TianmengChen (approved):

src/cpp/src/gguf_utils/gguf_quants.cpp
```diff
@@ -15,8 +15,8 @@ using namespace std;
 void unpack_32_4(uint8_t* data, uint8_t* dst) {
     std::fill_n(dst, 16, 0);
     for (int j = 0; j < 16; ++j) {
-        uint8_t x = (data[j + 2] & 0x0F);  // j+2 to skip scale bytes.
-        uint8_t y = (data[j + 2] >> 4);
+        uint8_t x = (data[j] & 0x0F);
+        uint8_t y = (data[j] >> 4);
```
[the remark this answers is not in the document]
wine99: This is intended. The function expects `data` to point to exactly where the data is and leave the caller to skip scales and optionally bias.

tests/python_tests/data/models.py
```diff
@@ -35,6 +35,11 @@ def get_gguf_model_list():
             "gguf_filename": "qwen2.5-0.5b-instruct-q4_0.gguf",
             "dynamic_quantization_group_size": None,
         },
+        {
+            "gguf_model_id": "unsloth/Llama-3.2-1B-Instruct-GGUF",
+            "gguf_filename": "Llama-3.2-1B-Instruct-Q4_1.gguf",
+            "dynamic_quantization_group_size": None,
```
Wovchena: Is there smaller alternative? If not, generate a random model similar to `sammysun0711/tiny-random-deepseek-distill-qwen-gguf`

TianmengChen: I think this add a Q4_1 model for test case, cause this pr fix Q4_1 accuracy.

Wovchena: The request is to add a smaller Q4_1 model

TianmengChen: @wine99 Hi, can you help to find smaller Q4_1 model or generate a dummy model?

yangsu2022: To create a dummy model:
• Copy all json files from source model to new folder
• Modify the num_hidden_layers =1 in config.json: model size will be smaller

Reference code:

from transformers import LlamaForCausalLM, AutoConfig, AutoTokenizer, LlamaTokenizerFast
import torch
from pathlib import Path

model_dir = Path("tiny-random-llama3.2")
config = AutoConfig.from_pretrained(model_dir / "config.json")
model = LlamaForCausalLM(config)

tokenizer = AutoTokenizer.from_pretrained(model_dir, trust_remote_code=True)

output_dir = model_dir

model.save_pretrained(output_dir)
tokenizer.save_pretrained(output_dir)
print("Create fake model with tokenizer successful!")

wine99: I changed the model to be the Q4_1 version of SmolLM2-135M which already exists in the tests.

Wovchena (approved):

src/cpp/src/gguf_utils/gguf_quants.cpp
```diff
@@ -58,8 +58,8 @@ void extract_q4_1_data(const gguf_tensor& tensor,
     auto biases = biases_arr.data<ov::element_type_traits<ov::element::f16>::value_type>();
     ov::parallel_for(scales_arr.get_size(), [&](size_t i) {
         scales[i] = ov::float16::from_bits(*((uint16_t*)(data + i * bytes_per_block)));
-        biases[i] = ov::float16::from_bits(*((uint16_t*)(data + i * bytes_per_block + 1)));
-        unpack_32_4(data + i * bytes_per_block, weights + i * 16);
+        biases[i] = ov::float16::from_bits(*((uint16_t*)(data + i * bytes_per_block + 2)));
+        unpack_32_4(data + i * bytes_per_block + 4, weights + i * 16);
```
TianmengChen, line 62: Should there be unpack_32_4(data + i * bytes_per_block + 4, weights + i * **32**);

TianmengChen:
@yangsu2022 , could you please take a look at Q4_1 changes, LGTM but it have different result between gguf-ov and gguf-hf.
https://github.com/openvinotoolkit/openvino.genai/actions/runs/16901879893/job/47884852467#step:7:469

TianmengChen:
@Wovchena Hi, we think that the new test case has this accuracy problem is because the way we calculate zero_point, you can see code [here](https://github.com/openvinotoolkit/openvino.genai/blob/5c36f2f088d0fce12e5e3f703cabe9c3d356e474/src/cpp/src/gguf_utils/building_blocks.cpp#L672). For original gguf, they directly use w * scale + bias, they don't need to calculate (q - zp) * scale, but we need to keep consistency with the original OV model structure. This may lead to accuracy problem.

So do we need to delete this test case or give a hardcore answer. Thx

Wovchena:
Why do we need to keep consistency with the original OV model structure? it seems that if the model is GGUF, the baseline should be GGUF based

wine99:
I tried modifying the `make_int4_weights` to use the logic `w * scale + bias`, the accuracy is good (verified in another project) but performance is very bad as it did not trigger compressed kernel (so it was actually doing convert + computation in fp32 in runtime)

Wovchena:
It's pointless to perform a fast computation if it's known to be incorrect. I assume OV was supposed to trigger the kernel. In that case you should report the problem to OV

yangsu2022:
Could we temporarily mark the failing pytest as xfail for now? 
This PR's target is to fix the Q4_1 irrational outputs. We could file a ticket for this accuracy alignment.

Wovchena:
Sure. Please add the ticket number to xfail's reason

tests/python_tests/data/models.py
```diff
@@ -35,6 +35,15 @@ def get_gguf_model_list():
...
             "dynamic_quantization_group_size": None,
         },
+        pytest.param(
+            {
+                "hf_model_id": "HuggingFaceTB/SmolLM2-135M",
+                "gguf_model_id": "QuantFactory/SmolLM2-135M-GGUF",
+                "gguf_filename": "SmolLM2-135M.Q4_1.gguf",
+                "dynamic_quantization_group_size": None,
+            },
+            marks=pytest.mark.xfail(reason="CVS-172345"),
```
Wovchena:
```suggestion
            marks=pytest.mark.xfail(reason="Prediction mismatch. Ticket 172345", raises=AssertionError),
```

wine99:
@Wovchena Could you help trigger CI again

Wovchena:
build_jenkins

Wovchena:
build_jenkins

PR #2568, opened 2025-08-08
Add thread control and verbose performance stats to chat_sample.py

juliensimon:
Hi OpenVINO team,

This makes it easier to get a quick sense of performance for chat applications, similar to llama-cli.

- Add -t/--num_threads argument for CPU thread control (CPU only)
- Add -v/--verbose flag to display performance statistics
- Implement accurate token counting using perf_metrics
- Add TTFT, TPOT, and throughput measurements
- Display warning when thread count is specified for non-CPU devices

### DIFF
--- a/samples/python/text_generation/chat_sample.py
+++ b/samples/python/text_generation/chat_sample.py
@@ -3,36 +3,99 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import argparse
+import time
+
 import openvino_genai
 
 
 def streamer(subword):
-    print(subword, end='', flush=True)
+    print(subword, end="", flush=True)
     # Return flag corresponds whether generation should be stopped.
     return openvino_genai.StreamingStatus.RUNNING
 
+
 def main():
     parser = argparse.ArgumentParser()
-    parser.add_argument('model_dir', help='Path to the model directory')
-    parser.add_argument('device', nargs='?', default='CPU', help='Device to run the model on (default: CPU)')
+    parser.add_argument("model_dir", help="Path to the model directory")
+    parser.add_argument(
+        "device",
+        nargs="?",
+        default="CPU",
+        help="Device to run the model on (default: CPU)",
+    )
+    parser.add_argument(
+        "-mt",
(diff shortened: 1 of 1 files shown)


samples/python/text_generation/chat_sample.py
```diff
@@ -3,36 +3,99 @@
...
+            throughput = perf_metrics.get_throughput().mean  # Tokens per second
+
+            print(f"\n----------")
+            print(f"Performance Stats:")
+            print(f"  Generation time: {generation_time:.2f} seconds")
+            print(f"  Tokens generated: {tokens_generated}")
+            print(f"  Time to first token (TTFT): {ttft:.2f} ms")
+            print(f"  Time per output token (TPOT): {tpot:.2f} ms")
+            print(f"  Throughput: {throughput:.2f} tokens/s")
+            print(f"----------")
```
Wovchena, line 92: There's already https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/benchmark_genai.py which covers most of these features.

Why do you need INFERENCE_NUM_THREADS?

juliensimon: Hi,

These proposed changes aim to make the example more useful for quick model tests, similar to `llama-cli` in llama.cpp.

The ability to set the number of threads is an easy way to try different configurations and gain a visual sense of how quickly the model generates tokens.

Likewise, the optional stats give you a quick sense of performance, without having to run full benchmarks.

Wovchena: This is already done in https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/benchmark_genai.py

juliensimon: I'm perfectly aware of that, and again, that's not my point. My point is to have a quick interactive way to test models and get perf info. If that's not interesting, no problem, close the PR. Thanks.

Wovchena: What aspect is less interactive in benchmark_genai.py?

PR #2569, opened 2025-08-08
Add thread control and device detection to benchmark_genai.py

juliensimon:
Hi OpenVINO team,

This makes it easier to batch benchmarking jobs with different configurations, and get detailed information in the logs.

- Add -t/--num_threads argument for CPU thread control (CPU only)
- Add platform-agnostic device detection using OpenVINO properties
- Display device information and thread count at startup
- Show warning when thread count is specified for non-CPU devices

### DIFF
--- a/samples/python/text_generation/benchmark_genai.py
+++ b/samples/python/text_generation/benchmark_genai.py
@@ -1,43 +1,116 @@
 # Copyright (C) 2023-2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-import sys
 import argparse
+import platform
+import sys
+
+import openvino as ov
 import openvino_genai as ov_genai
 from openvino import get_version
 
+
+def get_device_info(device_name):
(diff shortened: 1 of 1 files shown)


samples/python/text_generation/benchmark_genai.py
```diff
@@ -1,73 +1,167 @@
...
+        help="Maximal number of new tokens",
+    )
     parser.add_argument("-d", "--device", type=str, default="CPU", help="Device")
-    
+    parser.add_argument(
+        "-t",
+        "--num_threads",
+        type=int,
+        default=None,
+        help="Number of threads to use for inference (CPU only, ignored for GPU/NPU)",
```
Wovchena, line 74: You can trigger properties logging by setting `OPENVINO_LOG_LEVEL` env var: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/docs/DEBUG_LOG.md

As for `--num_threads`, there are dozens of other properties that could be exploited. I'd like avoid making it a tool

PR #2575, opened 2025-08-08
Skip whisper tests for transformers 4.53.0

as-suvorov:
There is an issue with whisper model export. Starting with transformers version 4.53.0 ([commit](https://github.com/huggingface/transformers/commit/badc71b9f604ca910bb87a43979c795eaf6e7d64)) optimum-cli exports bad stateless whisper model. Such model has nan values in output tensor. The issue only applied for stateless model (--disable-stateful), default exported model (stateful) works as expected.
This PR skips stateless model tests for CPU and NPU.


as-suvorov:
optimum-cli 1.25.2 supports transformers 4.53.0: https://github.com/openvinotoolkit/openvino.genai/actions/runs/16935024954/job/47990502315?pr=2611#step:7:190

PR #2576, opened 2025-08-08
Expose StructuredOutputConfig validation method

mzegla:
Currently, as we pass JSON schema to structured output config as the source of our grammar, it gets validated as grammar gets created and if we pass bad schema there is an error. 

Grammar creation happens in Sampler/LogitProcessor which means we learn about bad schema when request is already in the pipeline and if the schema is bad, the pipeline is broken. 

This PR exposes additional `validate` method on StructuredOutputConfig level that attempts to create grammar, so it's possible to check if grammar is not malformed before the request is scheduled for processing.

### DIFF
--- a/src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
+++ b/src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
@@ -75,8 +75,8 @@ xgrammar::Grammar XGrammarStructuredOutput::create_grammar(const std::optional<S
         grammar = xgrammar::Grammar::FromStructuralTag(
             xgrammar_structural_tags, structured_output_config.value().structural_tags_config.value().triggers
         );
-    } else if (structured_output_config.compound_grammar.has_value()) {
-        grammar = parse_compound_grammar(*structured_output_config.compound_grammar);
+    } else if (structured_output_config.value().compound_grammar.has_value()) {
+        grammar = parse_compound_grammar(structured_output_config.value().compound_grammar.value());
     } else {
         OPENVINO_THROW("No grammar definition provided for structured output generation.");
     }
--- a/src/cpp/src/tokenizer/tokenizer_impl.cpp
+++ b/src/cpp/src/tokenizer/tokenizer_impl.cpp
@@ -231,6 +231,29 @@ Tokenizer::TokenizerImpl::TokenizerImpl(const std::pair<std::shared_ptr<ov::Mode
     setup_tokenizer(models, properties);
 }
 
+void filter_properties(ov::AnyMap& properties) {
+    // Properties allowed for tokenizer/detokenizer on CPU
+    std::set<std::string> allowed_argnames = {
+        ov::hint::performance_mode.name(),
+        ov::hint::num_requests.name(),
+        ov::hint::enable_cpu_pinning.name(),
+        ov::hint::execution_mode.name(),
+        ov::hint::compiled_blob.name(),
+        ov::hint::enable_hyper_threading.name(),
+        ov::hint::enable_cpu_reservation.name(),
+        ov::enable_profiling.name(),
+    };
+
+    for (auto prop_it = properties.begin(); prop_it != properties.end();) {
+        auto it = allowed_argnames.find(prop_it->first);
+        if (it == allowed_argnames.end()) {
+            prop_it = properties.erase(prop_it);
+        } else {
+            ++prop_it;
+        }
+    }
+}
+
 void Tokenizer::TokenizerImpl::setup_tokenizer(const std::filesystem::path& models_path, const ov::AnyMap& properties) {
     ScopedVar env_manager(tokenizers_relative_to_genai());
     auto core = get_core_singleton();
@@ -320,8 +343,8 @@ void Tokenizer::TokenizerImpl::setup_tokenizer(const std::pair<std::shared_ptr<o
         properties.erase(it);
     }
 
-    // Pass no addtional properties to tokenizer/detokenizer models since it was not used by default
-    properties = {};
+    // Filter properties by leaving only params from the allowlist
+    filter_properties(properties);
     
     is_paired_input = ov_tokenizer && ov_tokenizer->get_parameters().size() == 2;
     


src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -156,9 +156,9 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
  * 
  * @param apply_chat_template whether or not to apply chat_template for non-chat scenarios
  */
-
+class StructuredOutputController;
 class OPENVINO_GENAI_EXPORTS GenerationConfig {
-
+    std::shared_ptr<ov::genai::StructuredOutputController> structured_output_controller = nullptr;
```
apaniukov: We need a mapping from tokenizer to structured_output_controller to support SO with multiple LLM pipelines in one application.

src/cpp/src/sampling/sampler.cpp
```diff
@@ -1020,7 +1020,10 @@ void Sampler::create_logit_processor(uint64_t request_id, const GenerationConfig
     if (!m_structured_output_controller) {
         // We don't have vocab size (actually logits size) and also we don't have access to the logits.
         // vocab size will be taken from the tokenizer during LogitProcessor initialization.
-        m_structured_output_controller = std::make_shared<StructuredOutputController>(m_tokenizer);
+        const auto& structured_output_controller = sampling_params.get_structured_output_controller();
+        if (structured_output_controller) {
+            m_structured_output_controller = structured_output_controller;
```
apaniukov: We also don't need this shared pointer in Sampler anymore.

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: I think it is better to move this from GenerationConfig to SOConfig, as well as the validate method.

src/cpp/src/sampling/structured_output/structured_output_controller.cpp
```diff
@@ -63,10 +59,16 @@ StructuredOutputController::get_logits_transformer(const ov::genai::GenerationCo
...
     const auto end = std::chrono::steady_clock::now();
     m_grammar_compile_times.emplace_back(std::chrono::duration_cast<std::chrono::microseconds>(end - start).count());
-    return res;
+}
+
+std::shared_ptr<LogitTransformers::ILogitTransformer> StructuredOutputController::get_logits_transformer() const {
+    if (!m_logits_transformer) {
+        OPENVINO_THROW("Logit transformer is not initialized. Call initialize_logits_transformer() first.");
+    }
+    return m_logits_transformer;
```
apaniukov: Initialising and storing the logit transformer in the single SOController might be bad for a parallel scenario, so I suggest avoiding it.

src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
```diff
@@ -24,35 +24,31 @@ XGrammarStructuredOutput::XGrammarStructuredOutput(const Tokenizer& tokenizer, s
...
             auto structural_tag = xgrammar::StructuralTagItem{tag.begin, tag.schema, tag.end};
             xgrammar_structural_tags.push_back(std::move(structural_tag));
         }
         grammar = xgrammar::Grammar::FromStructuralTag(
-            xgrammar_structural_tags, structured_output_config.structural_tags_config->triggers
+            xgrammar_structural_tags, structured_output_config.structural_tags_config.value().triggers
         );
     } else {
         OPENVINO_THROW("No grammar definition provided for structured output generation.");
     }
```
apaniukov: This might be sufficient for grammar validation, and can be extracted into a separate method.

src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -934,11 +943,17 @@ Vocab Tokenizer::get_vocab() const {
...
 
 const std::vector<std::string>& Tokenizer::get_vocab_vector() const {
+    OPENVINO_ASSERT(m_pimpl != nullptr, "Tokenizer is not initialized. Please check if the tokenizer model was provided and loaded correctly.");
     OPENVINO_ASSERT(!m_pimpl->m_vocab.empty(), "Tokenizer vocab is empty. Please check if the detokenizer model was provided and loaded correctly.");
     return m_pimpl->m_vocab;
 }
 
+std::shared_ptr<StructuredOutputController> Tokenizer::get_structured_output_controller() const {
+    return m_structured_output_controller;
+}
```
apaniukov, line 953: I think we need to init the controller lazily, the first time `get_structured_output_controller` is called.

apaniukov (approved):

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -303,11 +313,17 @@ class OPENVINO_GENAI_EXPORTS Tokenizer {
      */
     const std::vector<std::string>& get_vocab_vector() const;
 
+    std::shared_ptr<StructuredOutputController> get_structured_output_controller();
```
Wovchena: Make it private as well as StructuredOutputController. You should be able to use `friend` for that. Ideally you should be able to call `m_pimpl->get_structured_output_controller()` so `get_structured_output_controller()` doesn't even apear in the public header.

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -25,6 +25,16 @@ struct TokenizedInputs {
     std::optional<ov::Tensor> token_type_ids;
 };
 
+struct TokenizerProperties {
```
Wovchena: `TokenizerProperties` isn't used publicly. Move to private headers. Copilot's note about const ref is valid. `Tokenizer` is just a `shared_ptr` what was the problem using it directly?

src/cpp/src/generation_config.cpp
```diff
@@ -368,6 +368,10 @@ void StructuredOutputConfig::validate() const {
     );
 }
 
+void StructuredOutputConfig::validate(Tokenizer& tokenizer) const {
+    tokenizer.get_structured_output_controller()->validate_grammar(*this);
```
Wovchena:
```suggestion
    tokenizer.get_structured_output_controller()->validate_grammar(*this);
    validate();
```

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: That would introduce circular dependency. This PR ties `StructuredOutputController` to `Tokenizer` and up to this point `StructuredOutputController` required `Tokenizer` to initialize. When I used `Tokenizer` directly I had some issues with certain members being null pointers. Introducing such container seemed safer as we have more linear initialization and it worked instantly which was important factor regarding code freeze time pressure.

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: Used friend as suggested. SO controller is required in GenerationConfig for validation and Sampler for applying logit processor, so we still need some kind of intereface at least for those two cases.

Wovchena: Make this interface private.
I see `get_structured_output_controller()` isn't moved to `TokenizerImpl`. What was the problem?

src/cpp/include/openvino/genai/tokenizer.hpp
Wovchena: > up to this point `StructuredOutputController` required `Tokenizer` to initialize

But there's no circular dependency anymore, isn't it? `StructuredOutputController` is constructed from `Tokenizer` and `Tokenizer` doesn't require `StructuredOutputController` to be constructed. So `TokenizerProperties` is no longer needed

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: TokenizerImpl is completely private. If `get_structured_output_controller()` is moved from Tokenizer to TokenizerImpl it's not going to be visible in Sampler and GenerationConfig where it's needed. There is no standalone TokenizerImpl instance anywhere else and to make it available to other classes I would still need to have a method on Tokenizer level.

Wovchena: `TokenizerImpl` should be moved to a private header. That would expose `TokenizerImpl`'s methods

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: You're right. Since we have lazy initialization it can now be done. Just pushed the commit with such change.

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: Okay, assuming we have `get_structured_output_controller` in `TokenizerImpl` and it has it's own header that exposes that method. Do we want to work on TokenizerImpl objects? 
Because use case for early validation looks like that now:
```
auto tokenizer = pipeline.get_tokenizer();
structrured_output_config.validate(tokenizer);
```
where validate calls tokenizer::get_structured_output_controller method. Would we pass `TokenizerImpl` to the validate method then? I thought `TokenizerImpl` should be hidden.

Wovchena: `validate()` should still accept `Tokenizer`. `StructuredOutputConfig` is able unpack to the tokenizer because they are fiends. So `StructuredOutputConfig` can now read `Tokenizer::m_pimpl`. After `StructuredOutputConfig`'s cpp includes private `TokenizerImpl` header, it will be able to call `tokenizer.m_pimpl->get_structured_output_controller()`

Wovchena:
FYI, Linux is broken: https://github.com/openvinotoolkit/openvino.genai/pull/2609

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: Extracted TokenizerImpl to separate header and used it in required places. `Tokenizer` does not expose method to get SO controller now.

src/cpp/include/openvino/genai/tokenizer.hpp
```diff
@@ -25,6 +25,9 @@ struct TokenizedInputs {
     std::optional<ov::Tensor> token_type_ids;
 };
 
+class TokenizerImpl;
+class StructuredOutputController;
```
Wovchena: Move `class TokenizerImpl;` to `private` where it was.

`class StructuredOutputController;` is now constructed from `TokenizerImpl` (great idea!). `StructuredOutputConfig` shouldn't be a `friend` anymore

src/cpp/src/sampling/sampler.cpp
```diff
@@ -1017,13 +1019,11 @@ LogitProcessor& Sampler::get_logit_processor(uint64_t request_id) {
 
 
 void Sampler::create_logit_processor(uint64_t request_id, const GenerationConfig& sampling_params, const TokenIds& prompt) {
-    if (!m_structured_output_controller) {
-        // We don't have vocab size (actually logits size) and also we don't have access to the logits.
-        // vocab size will be taken from the tokenizer during LogitProcessor initialization.
-        m_structured_output_controller = std::make_shared<StructuredOutputController>(m_tokenizer);
+    std::shared_ptr<StructuredOutputController> structured_output_controller = nullptr;
+    if (m_tokenizer.m_pimpl != nullptr) {
+        structured_output_controller = m_tokenizer.m_pimpl->get_structured_output_controller();
```
Wovchena, line 1024: I think there's no way to construct an instance of `Tokenizer` with null `m_pimpl`. If you want to verify that, it should be an error to avoid misleading that it's a valid situation if `m_pimpl` is null

src/cpp/src/tokenizer/tokenizer_impl.hpp
```diff
@@ -0,0 +1,113 @@
...
+#include "gguf_utils/gguf_tokenizer.hpp"
+#include "tokenizer/chat_template_fallback_map.hpp"
+#include "tokenizer/make_tokenizer_stateful.hpp"
+#include "tokenizer/tokenizers_path.hpp"
+#include "circular_buffer_queue.hpp"
+#include "json_utils.hpp"
+#include "utils.hpp"
+
+
+namespace {
```
Wovchena: Move to cpp

src/cpp/include/openvino/genai/tokenizer.hpp
mzegla: If I move `TokenizerImpl` back to private scope inside `Tokenizer`, then I have to expand friends of `Tokenizer` class as `TokenizerImpl` full type is needed in `StructuredOutputController`, `XGrammarStructuredOutput`. Additionally it's used in static method outside of any class - `static bool registerXGrammarBackend()`. Seems like it would be harder to maintain.

src/cpp/src/sampling/sampler.cpp
mzegla: Not in a "real life" case. But in tests we create default Sampler with Tokenizer created via default constructor and it doesn't have m_pimpl:
https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/cpp/sampler.cpp#L71

src/cpp/src/tokenizer/tokenizer_impl.hpp
mzegla: moved definitions to cpp
declarations are kept in the header as those functions are used by `Tokenizer`

src/cpp/src/sampling/sampler.hpp
```diff
@@ -22,6 +22,7 @@
 #include "sequence_group.hpp"
 #include "threadpool.hpp"
 #include "sampling/structured_output/structured_output_controller.hpp"
+#include "tokenizer/tokenizer_impl.hpp"
```
Wovchena:
```suggestion
```
Is it needed?

src/cpp/src/sampling/structured_output/structured_output_controller.hpp
```diff
@@ -9,10 +9,11 @@
 #include <vector>
 #include <unordered_map>
 #include <functional>
+#include <set>
```
Wovchena:
```suggestion
```

src/cpp/src/sampling/sampler.hpp
mzegla: removed

src/cpp/src/sampling/structured_output/structured_output_controller.hpp
mzegla: removed

Wovchena (approved):

PR #2594, opened 2025-08-11
Add tests and samples dependencies gha component

as-suvorov:


.github/components.yml
```diff
@@ -4,6 +4,28 @@ GHA:
 tokenizers:
   revalidate: 'all'
 
+tests_dependencies:
+  revalidate:
```
as-suvorov, line 8: @Wovchena Maybe it would be safer to `revalidate: all`. It will trigger samples tests but we would not need to manually add new components here.

Wovchena: The components in this list are large. The only change that requires extending the list is new use case introduction. But new use case would require updating other lists as well

Wovchena (approved):

PR #2595, opened 2025-08-11
[NPUW] Disable chunking on NPU for VLM pipeline

AlexanderKalistratov:
Disabling chunking on NPU for VLMs as is breaks the model.

### DIFF
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -31,12 +31,6 @@ const std::string SDPA_BACKEND = "SDPA";
 
 namespace {
 
-void update_config(ov::AnyMap& config, const std::pair<std::string, ov::Any>& pair) {
-    if (config.count(pair.first) == 0) {
-        config.insert(pair);
-    }
-}
-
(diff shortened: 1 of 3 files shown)


src/cpp/src/utils.cpp
```diff
@@ -647,6 +643,12 @@ void release_core_plugin(const std::string& device) {
     }
 }
 
+void update_config(ov::AnyMap& config, const std::pair<std::string, ov::Any>& pair) {
+    if (config.count(pair.first) == 0) {
+        config.insert(pair);
+    }
+}
```
dmatveev: Doesn't stl have something for this?

AlexanderKalistratov: You are right. Checking for `.count() == 0` is not needed as it is checked inside `insert` method.
So, probably remove `update_config` at all and replace it with `insert`?

AlexanderKalistratov: Just used `insert` to disable chunking without touching the rest.

dmatveev (approved):

PR #2603, opened 2025-08-12
[wwb] Add tests for mac/win to ci

sbalandi:
+ add type of prompt long/short to gt_data, to run data with target_model without specifying prompt type and fix typo


Wovchena:
The fix for windows is in https://github.com/openvinotoolkit/openvino.genai/pull/2604

PR #2604, opened 2025-08-13
preserve properties from allowlist in Tokenizer

pavel-esir:


src/cpp/src/tokenizer/tokenizer.cpp
```diff
@@ -369,9 +392,9 @@ class Tokenizer::TokenizerImpl {
             two_input_requested = it->second.as<bool>();
             properties.erase(it);
         }
-
-        // Pass no addtional properties to tokenizer/detokenizer models since it was not used by default
-        properties = {};
+        
+        // Filter properties by leaving only params from the allowlist
+        filter_properties(properties);
```
Wovchena, line 397: This overload doesn't have a path so `enable_save_ov_model` should result in failure. I think you shouldn't filter `properties`

Wovchena: You shouldn't filter*

pavel-esir: If we don't filter, and allow all properties to pass to 'CPU' then it fails because cpu plugin 
```
2025-08-12T13:35:28.4948146Z E       Exception from src/inference/src/dev/plugin.cpp:53:
2025-08-12T13:35:28.4948893Z E       Exception from src/plugins/intel_cpu/src/config.cpp:457:
2025-08-12T13:35:28.4949480Z E       NotFound: Unsupported property NPUW_DEVICES by CPU plugin.
...
2025-08-12T13:35:28.4924214Z E           Exception from src/plugins/intel_cpu/src/config.cpp:457:
2025-08-12T13:35:28.4924878Z E           NotFound: Unsupported property prompt_lookup by CPU plugin.
...
```
and so on. We should leave only properties which `core.compile_model(ov_[de]tokenizer, properties)` can handle.

I also didn't quite understand bout saved model. At this point regardless of whether i filter or we don't have `'enable_save_ov_model'` in properties. Could you please elaborate what did you mean?

Wovchena: If there's no `enable_save_ov_model`, what property leads to the exception?

pavel-esir: At least 3 arguments above `NPUW_DEVICES`, `prompt_lookup`, `draft_model`, probably some more

Wovchena: I guess that should do for now. But this is not a full list of valid properties. It will need to be converted to a black list instead of white list in the next PR because it's easier to resolve a problem reported as an error rather then just observe perf degradation.

Wovchena (approved):

src/cpp/src/tokenizer/tokenizer.cpp
pavel-esir: Agree with you

Wovchena:
Ticket 172236

PR #2611, opened 2025-08-13
Upgrade optimum-intel and transformers

Wovchena:


samples/export-requirements.txt
```diff
@@ -1,7 +1,7 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
 openvino-tokenizers[transformers]~=2025.3.0.0.dev
-optimum-intel[nncf]==1.25.1
+optimum-intel[nncf] @ git+https://github.com/huggingface/optimum-intel.git@v1.25.2
```
as-suvorov:
```suggestion
optimum-intel[nncf]==v1.25.2
```

tests/python_tests/samples/test_text2speech.py
```diff
@@ -34,6 +35,10 @@ def teardown_class(self):
     @pytest.mark.precommit
     @pytest.mark.parametrize("convert_model", ["tiny-random-SpeechT5ForTextToSpeech"], indirect=True)
     @pytest.mark.parametrize("input_prompt", ["Hello everyone"])
+    @pytest.mark.xfail(
```
as-suvorov: If it's not ci issue, then speech generation wont work with deps update. Should this be a blocker?

Wovchena: That's what I asked you to take a look at in addition to WWB :) You delegated WWB but text2speeh is still on you

as-suvorov: Then it's blocker for the moment

Wovchena: We will need to move to the new versions in any case. This problem can indeed block the release. But merging that now enables merge queue by making pre-commit green and ensures the following PRs are validated with the versions targeted for the release

as-suvorov: Sure, agree

as-suvorov (approved):

tests/python_tests/samples/test_text2speech.py
```diff
@@ -34,6 +35,10 @@ def teardown_class(self):
...
+        reason="Missing config.json",
+        raises=subprocess.CalledProcessError,
```
as-suvorov: Fails on win only. Checking win locally

as-suvorov: Sample works locally on Win, it's CI issue.
Let's xfail or skip for win only. Or wait for tomorrow:)

Wovchena: Today isn't the first day it fails

Wovchena: xpassed. So the problem was in cached model

PR #2617, opened 2025-08-14
Fix CLEANUP_CACHE

Wovchena:


tests/python_tests/README.md
```diff
@@ -17,17 +17,12 @@ pip install -r tests/python_tests/requirements.txt
 python -m pytest tests/python_tests/ -m precommit
 ```
 
-During the test downloaded HuggingFace (HF) models will be saved into the current directory. If you wish to place them somewhere else you can specify `GENAI_MODELS_PATH_PREFIX` environment variable, e.g.
```
as-suvorov, line 20: We could keep the message and change `GENAI_MODELS_PATH_PREFIX` -> `OV_CACHE`. Might be helpful, I'm using this locally.

Wovchena: OV_CACHE is described below

PR #2619, opened 2025-08-14
Add cmake target creation helpers

olpipi:
CVS-146355


cmake/add_target_helpers.cmake
```diff
@@ -0,0 +1,354 @@
...
+        openvino::important_plugin
+   OBJECT_FILES
+        object libraries
+   DEFINES
+        DEF1 DEF2
+   LINK_FLAGS
+        flag1 flag2
+)
+#]]
+function(genai_add_sample_target)
```
Wovchena: `cmake/add_target_helpers.cmake` isn't packaged. It won't be possible to use samples outside of the repo.
There're probably similar functions for OV. If this is the case, maybe it could be reused

olpipi: I almost addressed this issue. There is still variables GENAI_SAMPLES_RUNTIME_OUTPUT_DIRECTORY that need to be removed.

olpipi: now samples can be build independently

cmake/global_properties.cmake
```diff
@@ -0,0 +1,33 @@
+# Define output directories
+if (CMAKE_ARCHIVE_OUTPUT_DIRECTORY)
+    set (GENAI_ARCHIVE_OUTPUT_DIRECTORY "$<1:${CMAKE_RUNTIME_OUTPUT_DIRECTORY}>")
+    set (GENAI_SAMPLES_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/samples/")
+    set (GENAI_TOOLS_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}/tools/")
```
Wovchena, line 5: Great idea with `load_openvino()`. Is this file needed given that there's already `load_openvino()`? GenAI could keep using CMAKE vars as read only directly

samples/CMakeLists.txt
```diff
@@ -2,6 +2,125 @@
...
+    set(multiValueArgs
+        )
+    cmake_parse_arguments(ARG "${options}" "${oneValueRequiredArgs}" "${multiValueArgs}" ${ARGN} )
+
+    _add_sample(${ARGV})
+
+    target_link_libraries(${ARG_NAME} PRIVATE openvino::genai)
+endfunction()
+
+function(genai_add_c_sample_target)
```
Wovchena: That file isn't packaged either I believe. Why not `ov_add_sample()` as I suggested earlier?

olpipi: reused ov_add_sample

cmake/global_properties.cmake
olpipi: I tried to not set CMAKE_ARCHIVE_OUTPUT_DIRECTORY explicitly. But 3rdparty dependency targets in this case are built in wrong dir

cmake/add_target_helpers.cmake
```diff
@@ -0,0 +1,214 @@
...
+        openvino::important_plugin
+   OBJECT_FILES
+        object libraries
+   DEFINES
+        DEF1 DEF2
+   LINK_FLAGS
+        flag1 flag2
+)
+#]]
+function(genai_add_target)
```
Wovchena: Is it used?

cmake/global_properties.cmake
Wovchena: How is that possible that third-party targets are affected while genai itself is fine?

olpipi: genai targets are affected also but they have explicit property of OUTPUT_DIRECTORY.
And as I can see, isolating of load_openvino helps only partly. So we have to set CMAKE_..._OUTPUT_DIRECTORY to build everything into required directory.

cmake/add_target_helpers.cmake
olpipi: not really. I wanted to reuse it in a separate PR for other targets

src/cpp/CMakeLists.txt
```diff
@@ -191,9 +191,6 @@ endif()
 set_target_properties(${TARGET_NAME} PROPERTIES
     EXPORT_NAME genai
     INTERPROCEDURAL_OPTIMIZATION_RELEASE ${ENABLE_LTO}
-    ARCHIVE_OUTPUT_DIRECTORY "$<1:${CMAKE_BINARY_DIR}/openvino_genai/>"
-    LIBRARY_OUTPUT_DIRECTORY "$<1:${CMAKE_BINARY_DIR}/openvino_genai/>"
-    RUNTIME_OUTPUT_DIRECTORY "$<1:${CMAKE_BINARY_DIR}/openvino_genai/>"
```
Wovchena, line 196: Why is it removed? Should it use new vars instead?

samples/CMakeLists.txt
```diff
@@ -1,6 +1,103 @@
...
+    else()
+        set (GENAI_SAMPLES_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/samples/")
+    endif()
+endif()
+
+if (NOT GENAI_SAMPLES_RUNTIME_OUTPUT_DIRECTORY)
+    if (CMAKE_RUNTIME_OUTPUT_DIRECTORY)
+        set (GENAI_SAMPLES_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/samples/")
+    else()
+        set (GENAI_SAMPLES_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/samples/")
```
Wovchena: These vars are already set in `cmake/global_properties.cmake`. Why set them again here?

cmake/add_target_helpers.cmake
Wovchena: Remove it for now. You should be able to recover it from git history of your branch when needed. That should also motivate you not to forget about the separate PR for other targets. What will that new PR do?

olpipi: removed

src/cpp/CMakeLists.txt
olpipi: yeap. It uses global variables instead

samples/CMakeLists.txt
olpipi: removed.

src/cpp/CMakeLists.txt
olpipi: But as you both with copilot recommended, return them back

CMakeLists.txt
```diff
@@ -39,32 +39,37 @@ if(NOT DEFINED Python3_FIND_VIRTUALENV)
...
+    find_package(OpenVINODeveloperPackage ${OV_COMPATIBILITY_VERSION} QUIET
+                COMPONENTS Runtime Threading
+                PATHS "${OpenVINO_DIR}")
+    if(NOT OpenVINODeveloperPackage_FOUND)
+        find_package(OpenVINO ${OV_COMPATIBILITY_VERSION} REQUIRED
+                    COMPONENTS Runtime Threading
+                    PATHS "${OpenVINO_DIR_PY}")
+    endif()
+endfunction()
+load_openvino()
```
mryzhov, line 69: Is there any other places where this function is used?

src/python/CMakeLists.txt
```diff
@@ -36,13 +36,17 @@ endif()
 target_link_libraries(${TARGET_NAME} PRIVATE openvino::genai)
 target_include_directories(${TARGET_NAME} PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/.." "${OpenVINOGenAI_SOURCE_DIR}/src/cpp/src") # for tokenizer/tokenizers_path.hpp
 set_target_properties(${TARGET_NAME} PROPERTIES
-    ARCHIVE_OUTPUT_DIRECTORY "$<1:${CMAKE_BINARY_DIR}/openvino_genai/>"
-    LIBRARY_OUTPUT_DIRECTORY "$<1:${CMAKE_BINARY_DIR}/openvino_genai/>"
+    ARCHIVE_OUTPUT_DIRECTORY "$<1:${GENAI_ARCHIVE_OUTPUT_DIRECTORY}>"
```
mryzhov, line 39: Why we can't use PROJECT_<>_DIR (BINARY, ARCHIVE etc) here instead?

olpipi:
build_jenkins

Issue #2526, opened 2025-08-01
Slow model loading time comparing to optimum OV model


Quallyjiang:
In our application we have a need to switch models due to resource constraint. When we compare model loading time in genai pipeline vs in optimum openvino model, we found the difference is quite big.

To Reproduce
Update testing script for model files:
```python
import gc
import itertools
from pathlib import Path
import time
from optimum.intel import OVModelForCausalLM
from transformers import AutoTokenizer
from openvino_genai import LLMPipeline 


class TimerContext:
    def __init__(self, name: str):
        self.name = name
        self.start_time = None

    def __enter__(self):
        self.start_time = time.time()
        return self

    def __exit__(self, exc_type, exc_value, traceback):
        elapsed_time = time.time() - self.start_time
        print(f"{self.name} took {elapsed_time:.2f} seconds")


def test_load_model(loader_func, model_a: Path, model_b: Path, iterations=10):
    print(f"**** Starting {loader_func.__name__} test {'***'*20}")
    with TimerContext(f"Intialize by loading model {model_b.name:40}"):
        model, tok = loader_func(model_b)
    for _, model_path in itertools.product(range(iterations), (model_a, model_b)):
        with TimerContext(f"Deleting existing model"):
            del model
            del tok
            gc.collect()
        with TimerContext(f"Loading model {model_path.name:40}"):
            model, tok = loader_func(model_path)
        if isinstance(model, LLMPipeline):
            result = model.generate(["The Sun is yellow because"], max_new_tokens=2)
            print(f"Load time: {result.perf_metrics.get_load_time():.2f} ms")
    del model, tok
    gc.collect()


def ov_optimum_load_model(model_path: Path):
    model = OVModelForCausalLM.from_pretrained(
        model_path, compile=True, compile_only=True, device="GPU", ov_config={"INFERENCE_PRECISION_HINT": "f16"}
    )
    tok = AutoTokenizer.from_pretrained(model_path)
    return model, tok


def ov_genai_load_model(model_path: Path):
    model = LLMPipeline(
        model_path,
        device="GPU",
        CACHE_DIR= str(model_path / ".ov_genai_cache"), 
        ATTENTION_BACKEND="PA",
    )
    return model, None


if __name__ == "__main__":
    model_a = Path(r"local_models\Qwen2.5-7B-Instruct-int4-ov")
    model_b = Path(r"local_models\Phi-3-mini-4k-instruct-int4-ov")

    import os

[...]

PR #2627, opened 2025-08-18
create signature property

Quallyjiang:
There's a [new PR in openvino for model_transform_signature](https://github.com/openvinotoolkit/openvino/pull/31761), which can be used to accelerate model hash computation and thus model loading time. #2526 
This PR will create this property based on transformation done in ContinuousBatchingPipeline and supply to compile_model call.


Wovchena:
Solved in https://github.com/openvinotoolkit/openvino.genai/pull/2720

PR #2630, opened 2025-08-19
Wangwang/optimize qwen2vl encoder

WeldonWangwang:


src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -7,15 +7,104 @@
...
 // Chat template hardcodes char sequence instead of referring to tag values, so NATIVE_TAG is hardcoded as well.
-std::string NATIVE_TAG = "<|vision_start|><|image_pad|><|vision_end|>";
+// std::string NATIVE_TAG = "<|vision_start|><|image_pad|><|vision_end|>";
+const std::string NATIVE_TAG = "<|vision_start|><|image_pad|><|vision_end|>";
+
+std::shared_ptr<ov::Model> patch_preprocess_into_model(std::shared_ptr<ov::Model> model_org) {
+    auto raw_images = std::make_shared<ov::op::v0::Parameter>(ov::element::u8, ov::PartialShape{-1, -1, -1, -1});
+    auto resize_target_shape = std::make_shared<ov::op::v0::Parameter>(ov::element::i64, ov::PartialShape{2});
+    auto image_mean = std::make_shared<ov::op::v0::Parameter>(ov::element::f32, ov::PartialShape{1, -1, 1, 1});
+    auto image_scale = std::make_shared<ov::op::v0::Parameter>(ov::element::f32, ov::PartialShape{1, -1, 1, 1});
```
xipingyan: mean and std are constant, maybe you can remove the both Parameter, replace them with const node.
The 2 nodes should be fused in compile model when we change them to const.

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -331,53 +450,57 @@ EncodedImage VisionEncoderQwen2VL::encode(const ov::Tensor& image, const ov::Any
...
+    for(auto& v : a_image_mean) v *= 255.0f;
+    for(auto& v : a_image_scale) v = 1.0f / (v*255.0f);
+
+    ov::Tensor image_mean(ov::element::f32, ov::Shape{1,a_image_mean.size(),1,1}, a_image_mean.data());
+    ov::Tensor image_scale(ov::element::f32, ov::Shape{1,a_image_scale.size(),1,1}, a_image_scale.data());
+
+    encoder.set_tensor("raw_images", raw_images);
+    encoder.set_tensor("resize_target_shape", target_shape);
+    encoder.set_tensor("image_mean", image_mean);
+    encoder.set_tensor("image_scale", image_scale);
```
xipingyan: remove after changing mean, std to const.

Wovchena:
> Evaluate pipeline: 24it [06:26, 16.10s/it]
INFO:main:Metrics for model: /mnt/llm_irs/WW33_llm_optimum_cb26ee3e/qwen2.5-vl-3b-instruct/pytorch/dldt/
INFO:main: similarity
0 0.941586

Is it CPU or GPU? CPU is expected to be ~0.97 depending on a platform. See the report in ticket 171393

WeldonWangwang:
> > Evaluate pipeline: 24it [06:26, 16.10s/it]
> > INFO:main:Metrics for model: /mnt/llm_irs/WW33_llm_optimum_cb26ee3e/qwen2.5-vl-3b-instruct/pytorch/dldt/
> > INFO:main: similarity
> > 0 0.941586
> 
> Is it CPU or GPU? CPU is expected to be ~0.97 depending on a platform. See the report in ticket 171393

It's GPU

Wovchena:
Should be fine in that case. Please run CPU as well

WeldonWangwang:
> 171393

Okay, thanks.

src/cpp/src/visual_language/qwen2vl/classes.cpp
WeldonWangwang: Converted image_mean and image_scale to constant.

WeldonWangwang:
Hi @Wovchena , update the local wwb test with CPU in CVS-170991, please review.

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -7,15 +7,93 @@
...
 #include "visual_language/vl_sdpa_transformations.hpp"
 
 namespace ov::genai {
 
 namespace {
 
 // Chat template hardcodes char sequence instead of referring to tag values, so NATIVE_TAG is hardcoded as well.
-std::string NATIVE_TAG = "<|vision_start|><|image_pad|><|vision_end|>";
-
+const std::string NATIVE_TAG = "<|vision_start|><|image_pad|><|vision_end|>";
```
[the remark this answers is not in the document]
WeldonWangwang: The normalize API accessing NATIVE_TAG requires const std::string& type.

WeldonWangwang:
Update test data in [CVS-170991](https://jira.devtools.intel.com/browse/CVS-170991) based on commit: https://github.com/openvinotoolkit/openvino.genai/pull/2630/commits/dcdba9c91147a71811be4f29434363475f2e6743

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -331,53 +539,47 @@ EncodedImage VisionEncoderQwen2VL::encode(const ov::Tensor& image, const ov::Any
...
+    ov::Tensor reshape_shape8d(ov::element::i64, ov::Shape{8}, a_temp_shape8d);
+    ov::Tensor reshape_shape4d(ov::element::i64, ov::Shape{4}, a_temp_shape4d);
+    ov::Tensor reshape_shape2d(ov::element::i64, ov::Shape{2}, last_output_shape);
+
+    encoder.set_tensor("input_images", input_images);
+    encoder.set_tensor("resize_shape", target_shape);
+    encoder.set_tensor("tile_shape", tile_shape);
+    encoder.set_tensor("reshape_shape8d", reshape_shape8d);
+    encoder.set_tensor("reshape_shape4d", reshape_shape4d);
+    encoder.set_tensor("reshape_shape2d", reshape_shape2d);
```
Wovchena, line 657: Please add a hidden option `IMAGE_PRERPOCESS` with `OV` or `С++` values `similar to `ATTENTION_BACKEND`. That option should activate the old implementation.

Context: there's an idea to fully replicate image preprocessing from the original framework. That would allow investigating accuracy issues caused by implementation differences in future.

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -314,6 +470,58 @@ ov::Tensor merge_text_and_image_embeddings(
     
 } // namespace qwen2vl_utils
 
+std::unique_ptr<CircularBufferQueue<ov::InferRequest>> VisionEncoderQwen2VL::create_ireq(ov::CompiledModel& compiled_model) {
```
Wovchena: Good idea moving a part of a constructor to a separate function. Move more parts of the constructors here. Also please make it function since it doesn't access any class members

src/cpp/src/visual_language/qwen2vl/classes.cpp
Wovchena, line 1: @yatarkan, please, review

src/cpp/src/visual_language/qwen2vl/classes.cpp
WeldonWangwang: Moved more parts of the constructors here. Thanks!

src/cpp/src/visual_language/qwen2vl/classes.cpp
WeldonWangwang: OK, will add it.

zhaixuejun1993: @Wovchena Can we use an env variable for this? Some reasons: 
1) this property unlike the ATTENTION_BACKEND, which is special only for QW2VL Vision Encoder; 
2) the ATTENTION_BACKEND is used on high level, but the IMAGE_PREPROCESS used in low level, which means there is a long way to pass this property.
3) if pass this property “IMAGE_PREPROCESS” from top level, need to do many changes on the exist interface.
BTW I think keep the original is only for developer verify & debug, so I think we can use this simple way. How do u think?

Wovchena: The reason for having it as a property is to be able to pass it from Jenkins via `--ov_config` in who what benchmark. Other models should implement `IMAGE_PREPROCESS` as well over time.

But I think it's OK to implement it as an env var for now. It can be replaced with property later after it's proven that the feature is needed.

src/cpp/src/visual_language/qwen2vl/classes.hpp
```diff
@@ -14,9 +14,13 @@ namespace ov::genai {
...
 class VisionEncoderQwen2VL : public VisionEncoder {
 public:
-    using VisionEncoder::VisionEncoder;
+    explicit VisionEncoderQwen2VL(const std::filesystem::path& model_dir, const std::string& device, const ov::AnyMap properties);
+    explicit VisionEncoderQwen2VL(const ModelsMap& models_map, const std::filesystem::path& config_dir_path, const std::string& device, const ov::AnyMap properties);
 
     EncodedImage encode(const ov::Tensor& image, const ov::AnyMap& config_map) override;
+    EncodedImage encode_with_imagepreprocess_cpp(const ov::Tensor& image, const ov::AnyMap& config_map);
+    EncodedImage encode_with_imagepreprocess_ov(const ov::Tensor& image, const ov::AnyMap& config_map);
+    bool is_use_ov_image_preprocess = true; // default use ov image preprocess, control by env IMAGE_PREPROCESS=CPP to use cpp image preprocess
```
Wovchena:
```suggestion
    bool use_ov_image_preprocess = true; // default use ov image preprocess, control by env IMAGE_PREPROCESS=CPP to use cpp image preprocess
```
`use` and `is` are both verbs

Wovchena (approved):

yatarkan (approved):

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -314,15 +470,71 @@ ov::Tensor merge_text_and_image_embeddings(
...
+}
+
+VisionEncoderQwen2VL::VisionEncoderQwen2VL(const std::filesystem::path& model_dir,
+                                           const std::string& device,
+                                           const ov::AnyMap properties)
+    : VisionEncoder(model_dir, device, properties) {
+    const char* env = std::getenv("IMAGE_PREPROCESS");
+    if (env && std::string(env) == "CPP") {
+        is_use_ov_image_preprocess = false;
+    }
```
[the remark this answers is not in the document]
Wovchena: That also allows initializing `use_ov_image_preprocess` with member initializer list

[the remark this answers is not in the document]
Wovchena: @WeldonWangwang, address it please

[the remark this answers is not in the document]
WeldonWangwang: Hi @Wovchena Updated it, thanks!

PR #2631, opened 2025-08-19
[JS] Add prettier and align eslint 

almilosz:
- Add prettier. Now you can run `npm run format` to format code.
- Align with https://github.com/openvinotoolkit/openvino/pull/31240 except
  - file extension, eslint.config.js is eslint.config.cjs
  - allow some properties in snake_case e.g. "max_new_tokens". 
  - use typescript-eslint config recommended and add "@typescript-eslint/no-explicit-any": "off"
- Remove lint check on windows cause it errors out on line endings. Alternative is to set `"endOfLine": "auto"` but it may introduce inconsistencies. Another alternative is to enforce LF everywhere using `.gitattributes`.
- Bump modules marked by `npm audit` 

[CVS-170355](https://jira.devtools.intel.com/browse/CVS-170355)


.github/workflows/windows.yml
```diff
@@ -916,10 +916,6 @@ jobs:
           Copy-Item -Recurse ${{ env.INSTALL_DIR }}/openvino_node_npm_package/bin node_modules/openvino-node/bin
           npm install --verbose
 
-      - name: Check lint
-        working-directory: ${{ env.SRC_DIR }}/src/js
-        run: npm run lint
-
```
yatarkan, line 922: Why lint step is removed?
Linting and formatting with prettier are different processes.
I believe we should still check linting in CI.

src/js/BUILD.md
yatarkan, line 1: Do we really need to format markdown files with prettier?

.github/workflows/windows.yml
almilosz: I added `const prettierConfig = require("eslint-plugin-prettier/recommended");` to eslint configuration. It disables eslint stylistic rules and also shows formatting errors from prettier configuration.
Currently we check linting both on windows and linux. I added rule "endOfLine": "lf". Then lint check on windows errored out with "error Delete ␍ prettier/prettier". Linux check passed. Alternative to removing check on windows is to set "endOfLine": "auto" but it may introduce inconsistencies. Another alternative is to enforce LF everywhere using .gitattributes.
Even if we remove linting check on windows, we still check it on linux

src/js/BUILD.md
almilosz: We do not need to, but it helps with consistency. Do you want to set `"proseWrap": "preserve"`?

.github/workflows/windows.yml
almilosz: https://github.com/openvinotoolkit/openvino.genai/blob/630c9736f0b4ac27d56dde670f9833baeba50593/.github/workflows/linux.yml#L851-L853

Retribution98 (approved):
Looks good for me

PR #2632, opened 2025-08-19
Make image 2 image pipeline can run text 2 image task

JohnLeFeng:
Customer is requesting image 2 image pipeline to support text 2 image task to avoid load and unload model during change task.

Ticket 172202


JohnLeFeng:
Use #2701 instead.

PR #2655, opened 2025-08-21
[CMAKE] Solve pybind targets conflict

mryzhov:
When GenAI is built via the OpenVINO CMake option OPENVINO_EXTRA_MODULES, it throws errors:

[cmake] add_library cannot create imported target "pybind11::pybind11" because [cmake] another target with the same name already exists.

### DIFF
--- a/src/python/CMakeLists.txt
+++ b/src/python/CMakeLists.txt
@@ -12,10 +12,14 @@ if (NOT TARGET pybind11::pybind11)
         URL_HASH SHA256=453b1a3e2b266c3ae9da872411cadb6d693ac18063bd73226d96cfb7015a200c
     )
     FetchContent_GetProperties(pybind11)
-    FetchContent_Populate(pybind11)
+    
     # search for FindPython3.cmake instead of legacy modules
     set(PYBIND11_FINDPYTHON ON)
-    add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
+    
+    if(NOT pybind11_POPULATED)
+        FetchContent_Populate(pybind11)
+        add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
+    endif()
 else()
     message(STATUS "Reusing outer project's pybind11 target")
 endif()


src/python/CMakeLists.txt
```diff
@@ -4,20 +4,23 @@
...
+    )
+    FetchContent_GetProperties(pybind11)
     FetchContent_Populate(pybind11)
     add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
+else()
+    message(STATUS "Reusing outer project's pybind11 target")
 endif()
 
+# search for FindPython3.cmake instead of legacy modules
+set(PYBIND11_FINDPYTHON ON)
```
Wovchena: Should `set(PYBIND11_FINDPYTHON ON)` go before `add_subdirectory()`?

Wovchena (commented):
> When GenAI is built via the OpenVINO CMake option OPENVINO_EXTRA_MODULES, it throws errors:

OV had a test for that. Is it disabled?

src/python/CMakeLists.txt
mryzhov: you are right, fixed

mryzhov:
```suggestion
```

src/python/CMakeLists.txt
```diff
@@ -4,20 +4,23 @@
...
+# OpenVINO uses pybind11, and they can conflict.
+if (NOT TARGET pybind11::pybind11)
+    FetchContent_Declare(
+        pybind11
+        URL https://github.com/pybind/pybind11/archive/refs/tags/v3.0.0.tar.gz
+        URL_HASH SHA256=453b1a3e2b266c3ae9da872411cadb6d693ac18063bd73226d96cfb7015a200c
+    )
+    FetchContent_GetProperties(pybind11)
     FetchContent_Populate(pybind11)
     add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
```
mryzhov:
```suggestion
    FetchContent_Populate(pybind11)
    # search for FindPython3.cmake instead of legacy modules
    set(PYBIND11_FINDPYTHON ON)
    add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
```

src/python/CMakeLists.txt
```diff
@@ -4,20 +4,23 @@
...
+    FetchContent_GetProperties(pybind11)
     FetchContent_Populate(pybind11)
+    # search for FindPython3.cmake instead of legacy modules
+    set(PYBIND11_FINDPYTHON ON)
     add_subdirectory(${pybind11_SOURCE_DIR} ${pybind11_BINARY_DIR})
+else()
+    message(STATUS "Reusing outer project's pybind11 target")
 endif()
 
+
```
mryzhov:
```suggestion
```

mryzhov:
> > When GenAI is built via the OpenVINO CMake option OPENVINO_EXTRA_MODULES, it throws errors:
> 
> OV had a test for that. Is it disabled?
No, we don't build OV with genai in CI (

Wovchena:
I was asking about this https://github.com/openvinotoolkit/openvino/blob/8da895bee769c8e81b213cbcf91e94214b9ee6f8/.github/workflows/android_x64.yml#L156

Wovchena:
But it has `-DENABLE_PYTHON=OFF`. Can it be enabled?

mryzhov:
> But it has `-DENABLE_PYTHON=OFF`. Can it be enabled?

let;s try - https://github.com/openvinotoolkit/openvino/pull/31851

PR #2663, opened 2025-08-22
[GHA] Disable Cacheopt tests on mac

mryzhov:
Those tests takes 2 hours and don't use the hf cache, mac is deprecating soon.
So let's reduce our costs


akladiev (approved):

Wovchena:
@l-bat, FYI

PR #2671, opened 2025-08-26
WWB empty_adapters mode

likholat:


tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -194,6 +194,12 @@ def parse_args():
         help="LLMPipeline specific parameter that defines the use of a long context prompt",
     )
 
+    parser.add_argument(
+        "--empty_adapters",
+        action="store_true",
+        help="Inference without adapters",
```
sbalandi: smth like `Inference with empty adapters. Applicable for GenAI text-to-image task.`
If I understand correctly, the user can run it as is and it will be a mode without adapters.

tools/who_what_benchmark/whowhatbench/text2image_evaluator.py
```diff
@@ -115,7 +117,7 @@ def worst_examples(self, top_k: int = 5, metric="similarity"):
         return res
 
     def _generate_data(self, model, gen_image_fn=None, image_dir="reference"):
-        def default_gen_image_fn(model, prompt, num_inference_steps, generator=None):
+        def default_gen_image_fn(model, prompt, num_inference_steps, *, generator=None, **kwargs):
```
sbalandi: `*` - why is it needed ?
I think you can left just **kwargs and specify {'empty_adapters': self.empty_adapters} on str 167 or add empty_adapters=False here

tools/who_what_benchmark/whowhatbench/wwb.py
likholat: Updated. It will be available for text generation as well

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -194,6 +194,9 @@ def load_text2image_model(
         logger.info("Using OpenVINO GenAI API")
         model = load_text2image_genai_pipeline(model_id, device, ov_config, **kwargs)
     elif use_hf:
+        if kwargs['empty_adapters']:
```
as-suvorov:
```suggestion
        if 'empty_adapters' in kwargs:
```

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -194,6 +194,9 @@ def load_text2image_model(
         logger.info("Using OpenVINO GenAI API")
         model = load_text2image_genai_pipeline(model_id, device, ov_config, **kwargs)
     elif use_hf:
+        if 'empty_adapters' in kwargs:
```
sbalandi: please, move it to check_args() to fix ci https://github.com/openvinotoolkit/openvino.genai/blob/b9a5efdca9b317e88b1be3e4c3e4b56813019f2b/tools/who_what_benchmark/whowhatbench/wwb.py#L200

 ```
if args.hf and args.empty_adapters:
    raise ValueError
```

PR #2679, opened 2025-08-28
Add docs entry about building GenAI with free threaded Python

p-wysocki:


almilosz (approved):
Looks good to me. Is there a similar note/PR for `openvino.genai`?

as-suvorov (approved):

PR #2690, opened 2025-09-01
[Docs] Convert whisper as stateless in the quantization example

nikita-savelyevv:
There are some accuracy issues with quantized stateful whisper models. Updating documentation until it is not fixed. See CVS-172705.


nikita-savelyevv:
I see a warning
```text
[WARN] Whisper decoder models with past is deprecated. Support will be removed in 2026.0.0 release.
```
I hope the issue will be resolved until 2026.0 release. 

cc @as-suvorov

as-suvorov (approved):

PR #2693, opened 2025-09-02
[llm bench] Add possibility to setup cache eviction config for LLM

sbalandi:


tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -221,7 +251,13 @@ def create_genai_text_gen_model(model_path, device, ov_config, memory_monitor, *
     if kwargs.get("mem_consumption"):
         memory_monitor.start()
     start = time.perf_counter()
-    llm_pipe = openvino_genai.LLMPipeline(model_path, device.upper(), **ov_config)
+    del ov_config['CACHE_DIR']
```
as-suvorov: Why 'CACHE_DIR' removed?

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -221,7 +251,13 @@ def create_genai_text_gen_model(model_path, device, ov_config, memory_monitor, *
...
+    if cb_pipeline_required(kwargs):
+        ov_config.pop("ATTENTION_BACKEND", None)
+        log.info("Pipeline will be initialized with ContinuousBatchingPipeline")
+        llm_pipe = openvino_genai.ContinuousBatchingPipeline(model_path, device=device.upper(), **ov_config)
+    else:
+        llm_pipe = openvino_genai.LLMPipeline(model_path, device.upper(), **ov_config)
```
as-suvorov: Can we set `cache_eviction_config`/`sparse_attention_config` for `LLMPipeline`?

tools/llm_bench/task/text_generation.py
```diff
@@ -207,6 +207,56 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
         bench_hook.clear_time_infer_list()
 
 
+def genai_generate(streaming, model, tokens_len, gen_config, empty_lora, input_data, batch_size):
```
as-suvorov, line 210: nitpick: types are nice to have

tools/llm_bench/llm_bench_utils/ov_utils.py
sbalandi: by mistake, was restored, thank you !

tools/llm_bench/llm_bench_utils/ov_utils.py
sbalandi: yes, we can set these options, but we can't call get_metrcis and get avg_cache_usage, max_cache_usage from CB pipeline metrics

tools/llm_bench/task/text_generation.py
sbalandi: agree, but I would prefer to perform changes in another pr and alight all functions with that

vshampor (approved):

as-suvorov (approved):

PR #2697, opened 2025-09-02
Tune automatic memory allocation

popovaan:
1. Clear KV-cache at the end of generate().
2. Update KV-cache growth formula to align with GPU.

Ticket: CVS-162894


src/cpp/src/continuous_batching/block_manager.hpp
```diff
@@ -1138,6 +1151,17 @@ class BlockManager {
             }
         }
     }
+
+    void clear() {
+        // KV-cache should not be cleared in prefix caching is enabled
```
as-suvorov:
```suggestion
        // KV-cache should not be cleared if prefix caching is enabled
```

as-suvorov (approved):

PR #2719, opened 2025-09-10
Fixture-based VLM models reusing

sgonorov:
Part of [CVS-173285](https://jira.devtools.intel.com/browse/CVS-173285)

Fixes and performance optimizations for VLM tests

### DIFF
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -3,31 +3,35 @@
 
 """
 Fixture hierarchy
-cat_image ─────────┬─── cat_image_336x336
-    |              │
-cat_tensor         └─── cat_image_32x32
+synthetic_video ──────── synthetic_video_32x32_tensor
+cat_image ───────────┬── cat_tensor
+    │                ├── cat_image_384x384
+    │                ├── cat_image_336x336
+    │                └── cat_image_32x32
     │
-    ├──── iteration_images
-    │          │
-    │          ├─── cat_tensor
-    │          ├─── car_tensor
-    │          └─── handwritten_tensor
+    ├── iteration_images
+    │       ├── cat_tensor
+    │       ├── car_tensor
+    │       └── handwritten_tensor
     │
-    ├──── image_sequence
+    ├── image_sequence
+    │       └── cat_tensor
     │
-    └──── conversation_requests
-                │
-                ├─── cat_tensor
-                ├─── car_tensor
-                └─── handwritten_tensor
+    └── conversation_requests
+            ├── cat_tensor
+            ├── car_tensor
+            └── handwritten_tensor
 car_tensor
 handwritten_tensor
-model_and_tag
+ov_pipe_model
+ov_continious_batching_pipe
 """
 
+import collections
+from dataclasses import dataclass
+from typing import Callable, Generator
 import openvino_tokenizers
 import openvino
-import gc
 import PIL
 import pytest
 import platform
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_vlm_pipeline.py
```diff
@@ -96,32 +131,69 @@ def get_ov_model(model_id):
...
     return model_dir
 
 
-prompts = [
-    "What is on the image?",
-    "What is special about this image?",
-]
+@pytest.fixture(scope="module")
+def vlm_pipeline_provider() -> Generator[Callable[[str, str], VLMPipeline], None, None]:
+    cache: dict[tuple[str, str], VLMPipeline] = {}
```
Wovchena: cache needed?

tests/python_tests/test_vlm_pipeline.py
Wovchena, line 1: `master` reports 336 collected tests for `py -m pytest tests/python_tests/test_vlm_pipeline.py` while this branch reports 334. Please check if anything is lost

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -131,8 +203,7 @@ def implementation():
 
 @pytest.fixture(scope="module")
 def cat_image(pytestconfig):
-    cat_url = "https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11"
-    return from_cache_or_download(pytestconfig, cat_url, "cat.jpg")
+    return from_cache_or_download(pytestconfig, TEST_IMAGE_URLS['cat'], "cat.jpg")
```
Wovchena, line 324: What's the benefit of moving URLs to `TEST_IMAGE_URLS`? The argument not to do so is that the reader doesn't need to jump to `TEST_IMAGE_URLS` to figure out what `cat.jpg` corresponds to

sgonorov: It's better to keep this kind of constants (especially ones like URLs) centralized - more consistency, easier to manage and extend. Also fixture code is more readable that way.

tests/python_tests/test_vlm_pipeline.py
sgonorov: Yes, in the last version there's a code like `pipeline.finish_chat() `for state resetting between tests inside a module.

Wovchena:
By the way, you ca use ccache to speed up compilation on your machine

tests/python_tests/test_vlm_pipeline.py
Wovchena: `module` scoped fixture should be equivalent to caching because it lives as log as the module. My expectation is that if `cache` is removed, the return values would still be cached by pytest

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -215,23 +283,22 @@ def streamer(word: str) -> bool:
 @pytest.mark.parametrize("config", configs)
 def test_vlm_continuous_batching_generate_vs_add_request(config, cat_tensor):
     scheduler_config = SchedulerConfig()
-    models_path = get_ov_model(model_ids[0])
+    models_path = _get_ov_model(MODEL_IDS[0])
```
Wovchena: Reminder to try to remove `get_default_llm_properties()` to enable `ov_pipe` fixture in tests like that with `@pytest.mark.parametrize("ov_backend", ["PA"], indirect=True)`

tests/python_tests/test_vlm_pipeline.py
Wovchena: Reminder

sgonorov: Yes, they will be cached, but we won't be able to reset their state upon requesting the model second time for example (using `pipeline.finish_chat()`).

tests/python_tests/test_vlm_pipeline.py
sgonorov: Fixed where i could 👍

tests/python_tests/test_vlm_pipeline.py
Wovchena, line 1: Just a reminder, there's Fixture hierarchy on top. It should be updated once the implementation is finalized. Copilot should be able to generate it.

FYI, a new test will be added: https://github.com/openvinotoolkit/openvino.genai/pull/2738

sgonorov: Fixed 👍

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -98,33 +188,93 @@ def get_ov_model(model_id):
...
 
-prompts = [
-    "What is on the image?",
-    "What is special about this image?",
-]
+# On macOS, transformers<4.52 is required, but this causes gemma3 to fail
+GEMMA3_MACOS_XFAIL_REASON = "gemma3 not supported on macOS with older transformers"
+
+
+@pytest.fixture(scope="session")
```
apaniukov:
```suggestion
@pytest.fixture(scope="module")
```
Fixtures defined in a module should be contained within the module

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -133,131 +283,101 @@ def implementation():
...
     for images in [], [cat_tensor], [cat_tensor, handwritten_tensor, car_tensor]:
         result_from_streamer = []
         res = ov_pipe.generate(
-            prompts[0],
+            PROMPTS[0],
             images=images,
             generation_config=generation_config,
             streamer=streamer,
         )
         assert res.texts[0] == "".join(result_from_streamer)
```
apaniukov: Looks like "three <del>raccoons</del> tests stacked on top of each other in a trench coat pretending to be <del>a person</del> one test"
It is better to create a fixture that will yield `[], [cat_tensor], [cat_tensor, handwritten_tensor, car_tensor]` one by one.
`PROMPTS[0]` should be a parameter as well.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -717,255 +831,279 @@ def generate(vlm, requests):
...
+
+    retry(workaround_inconsistent_inference)
+
+
+@pytest.mark.precommit
+@model_and_tag_parametrize()
+def test_model_tags_older(ov_pipe_model: VlmModelInfo, car_tensor: openvino.Tensor):
+    ov_pipe = ov_pipe_model.pipeline
+    
+    generation_config = _setup_generation_config(ov_pipe, set_eos_token=False)
```
apaniukov, line 1292: Should we reduce the max_new_tokens parameter in such tests?

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -717,255 +831,279 @@ def generate(vlm, requests):
...
+    def workaround_inconsistent_inference():
+        automatic_tags = ov_pipe.generate(prompt, images=[cat_tensor], do_sample=False)
+        reference_tags = ov_pipe.generate(
+            templated_prompt, images=[cat_tensor], apply_chat_template=False, do_sample=False
+        )
+        assert automatic_tags.texts == reference_tags.texts
+        assert automatic_tags.scores == reference_tags.scores
 
-        retry(workaround_inconsistent_inference)
+    retry(workaround_inconsistent_inference)
```
apaniukov, line 1161:
```suggestion
    def workaround_inconsistent_inference():
        __tracebackhide__ = True
        automatic_tags = ov_pipe.generate(prompt, images=[cat_tensor], do_sample=False)
        reference_tags = ov_pipe.generate(
            templated_prompt, images=[cat_tensor], apply_chat_template=False, do_sample=False
        )
        assert automatic_tags.texts == reference_tags.texts
        assert automatic_tags.scores == reference_tags.scores

    retry(workaround_inconsistent_inference)
```
`__tracebackhide__ = True` also should be added to retry.
https://docs.pytest.org/en/6.2.x/example/simple.html#writing-well-integrated-assertion-helpers

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -717,255 +831,279 @@ def generate(vlm, requests):
...
+@pytest.mark.precommit
+@model_and_tag_parametrize(TAG_INSERTED_BY_TEMPLATE)
+def test_model_tags_representation(ov_pipe_model: VlmModelInfo, cat_tensor: openvino.Tensor):
+    ov_pipe = ov_pipe_model.pipeline
+    model_id = ov_pipe_model.model_id
+    
+    generation_config = _setup_generation_config(ov_pipe, set_eos_token=False)
+    generation_config = ov_pipe.get_generation_config()
+    generation_config.max_new_tokens = 30
+    ov_pipe.set_generation_config(generation_config)
```
apaniukov, line 1118:
```suggestion
    generation_config = _setup_generation_config(ov_pipe, set_eos_token=False, max_new_tokens=30)
    ov_pipe.set_generation_config(generation_config)
```
Also, why do we return the generation config from `_setup_generation_config` when we have a pipe there? `_set_generation_config` seems better and will reduce boilerplate. Or pass this config to `.generate` like in other tests.

tests/python_tests/test_vlm_pipeline.py
sgonorov: Frankly speaking i've tried not to change logic and parameters of the tests without a need. Maybe less values will bring some speed benefits, but not sure it won't affect test quality.

tests/python_tests/test_vlm_pipeline.py
sgonorov: They are still contained within the module, lifecycle is as same as module just in case. Sure, I'll change it for the clarity👍 It's a leftover from the previous experiments, thanks for noticing.

tests/python_tests/test_vlm_pipeline.py
sgonorov: Yeah, I'm trying to keep the original logic, but looks like we can change it here 👍

tests/python_tests/test_vlm_pipeline.py
sgonorov: Fixed

tests/python_tests/test_vlm_pipeline.py
sgonorov: A little bit out of scope, but added them to all assertion functions which need it. inside the file

tests/python_tests/test_vlm_pipeline.py
apaniukov: We check that `ov_pipe.generate("", images=[car_tensor])` works and `ov_pipe.generate("<ov_genai_image_0>", images=[car_tensor])` raises the error. It seems like the test should work fine with fewer output tokens.

apaniukov: If that is not the case, leave the config unchanged; otherwise, set max new tokens to `2`.

sgonorov: Yeah, checked and it works for this test 👍

as-suvorov:
@apaniukov Your comments seems to be resolved. Could you please review once more?

PR #2720, opened 2025-09-11
Use model path property for caching

praasz:
### Description
- Use new property for cache feature to bind `ov::Model` (after local transformations) with model path, to speed up cache ID calculation for blob.

### Blocked By
- #openvinotoolkit/openvino#31924


Wovchena:
Is this approach still needed considering my reply? https://github.com/openvinotoolkit/openvino/pull/31761#issuecomment-3269543072

mlukasze:
depends on https://github.com/openvinotoolkit/openvino.genai/pull/2743

Wovchena:
Increased timeouts further: https://github.com/openvinotoolkit/openvino.genai/pull/2756

PR #2723, opened 2025-09-11
WWB Text Generation with LoRA

likholat:
Found issues:
[173476](https://jira.devtools.intel.com/browse/CVS-173476)
[173478](https://jira.devtools.intel.com/browse/CVS-173478)

### DIFF
--- a/src/js/tests/module.test.js
+++ b/src/js/tests/module.test.js
@@ -266,20 +266,22 @@ describe("LLMPipeline.generate()", () => {
 
     // assert that calculating statistics manually from the raw counters
     // we get the same restults as from PerfMetrics
-    assert.strictEqual(
-      (perfMetrics.rawMetrics.generateDurations / 1000).toFixed(2),
-      generateDuration.mean.toFixed(2),
-    );
+    //
+    // Disabled due to potential floating-point differences. (CVS-175568)
+    // assert.strictEqual(
+    //   (perfMetrics.rawMetrics.generateDurations / 1000).toFixed(3),
+    //   generateDuration.mean.toFixed(3),
+    // );
 
-    assert.strictEqual(
-      (perfMetrics.rawMetrics.tokenizationDurations / 1000).toFixed(2),
-      tokenizationDuration.mean.toFixed(2),
-    );
+    // assert.strictEqual(
+    //   (perfMetrics.rawMetrics.tokenizationDurations / 1000).toFixed(3),
+    //   tokenizationDuration.mean.toFixed(3),
+    // );
 
-    assert.strictEqual(
-      (perfMetrics.rawMetrics.detokenizationDurations / 1000).toFixed(2),
-      detokenizationDuration.mean.toFixed(2),
-    );
+    // assert.strictEqual(
+    //   (perfMetrics.rawMetrics.detokenizationDurations / 1000).toFixed(3),
+    //   detokenizationDuration.mean.toFixed(3),
+    // );
 
     assert.ok(perfMetrics.rawMetrics.timesToFirstToken.length > 0);
     assert.ok(perfMetrics.rawMetrics.newTokenTimes.length > 0);
--- a/src/js/tests/structuredOutput.test.js
+++ b/src/js/tests/structuredOutput.test.js
@@ -231,10 +231,9 @@ table::= "users" | "orders" | "products"`),
   });
 
   it("generate with TriggeredTags in structural_tags_config", async (testContext) => {
-    if (os.platform() === "darwin" || os.platform() === "win32") {
-      testContext.skip("Skipped due to inconsistent LLM outputs. CVS-175278");
-      return;
-    }
+    testContext.skip("Skipped due to inconsistent LLM outputs. CVS-175278");
+    return;
+    // eslint-disable-next-line no-unreachable
     const tools = [
       {
         name: "get_weather",


tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -119,6 +125,26 @@ def load_text_hf_pipeline(model_id, device):
                 model_id, trust_remote_code=True, device_map=device.lower(), **model_kwargs
             )
 
+    if 'adapters' in kwargs and kwargs['adapters'] is not None:
```
sbalandi:
```suggestion
    if kwargs.get("adapters") is not None:
```

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -119,6 +125,26 @@ def load_text_hf_pipeline(model_id, device):
...
+        adapter_names = ["adapter_0"]
+        model = PeftModel.from_pretrained(model, adapters[0], adapter_name=adapter_names[0])
+
+        for idx, adapter in enumerate(adapters[1:], start=1):
+            model.load_adapter(adapter, adapter_name=f"adapter_{idx}")
+            adapter_names.append(f"adapter_{idx}")
+
+        if alphas is not None:
+            assert len(alphas) == len(adapter_names), "`alphas` must be the same length as `adapters`"
+            model.add_weighted_adapter([f"adapter_{idx}" for idx in range(len(kwargs['adapters']))], alphas, "merged_lora")
```
sbalandi: Isn't this [f"adapter_{idx}" for idx in range(len(kwargs['adapters']))] the same as adapter_names ?

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -63,15 +63,21 @@ def load_text_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwargs):
...
+            ov_adapter = openvino_genai.Adapter(adapter)
+            adapter_config.add(ov_adapter, alpha)
+
     is_continuous_batching = kwargs.get("cb_config", None) is not None
 
     if is_continuous_batching:
         logger.info("Using OpenVINO GenAI Continuous Batching API")
         scheduler_config = get_scheduler_config_genai(kwargs["cb_config"])
-        pipeline = openvino_genai.LLMPipeline(model_dir, device=device, scheduler_config=scheduler_config, **ov_config)
+        pipeline = openvino_genai.LLMPipeline(model_dir, device=device, adapters=adapter_config, scheduler_config=scheduler_config, **ov_config)
```
sbalandi: Are behavior with: 
```
adapter_config = openvino_genai.AdapterConfig()
openvino_genai.LLMPipeline(model_dir, device=device, adapters=adapter_config, **ov_config) 
```
and 
`openvino_genai.LLMPipeline(model_dir, device=device, **ov_config) `
the same ?

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -63,15 +63,21 @@ def load_text_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwargs):
             "Failed to import openvino_genai package. Please install it.")
         exit(-1)
 
+    adapter_config = openvino_genai.AdapterConfig()
+    if "adapters" in kwargs and kwargs["adapters"] is not None:
+        for adapter, alpha in zip(kwargs['adapters'], kwargs['alphas']):
```
sbalandi, line 75: If the alpha value does not match the number of adapters, then here it will be processed and adapters and alphas will be combined, but for hf it will fail. Why ? Is it impossible for hf to combine adapters with and without alphas?

tools/who_what_benchmark/whowhatbench/model_loaders.py
likholat: Yes, it is, I'll also add a separate check for the validation team on this

tools/who_what_benchmark/whowhatbench/model_loaders.py
likholat: I’d just add an assert here too: if the number of adapters and alphas don’t match, it’s unclear which alpha applies to which adapter, most likely the user forgot to provide a weight

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -63,15 +63,22 @@ def load_text_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwargs):
             "Failed to import openvino_genai package. Please install it.")
         exit(-1)
 
+    adapter_config = openvino_genai.AdapterConfig()
+    if "adapters" in kwargs and kwargs["adapters"] is not None:
```
sbalandi:
```suggestion
      if kwargs.get("adapters") is not None:
```

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -63,15 +63,22 @@ def load_text_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwargs):
...
+        assert len(kwargs['alphas']) == len(kwargs["adapters"]), "`alphas` must be the same length as `adapters`"
```
sbalandi: Can adapter be set without alpha for genAI ? if yes, this case needs to be handled as well
```suggestion
        if  kwargs.get('alphas') is not None:
            assert len(kwargs['alphas']) == len(kwargs["adapters"]), "`alphas` must be the same length as `adapters`"
            for adapter, alpha in zip(kwargs['adapters'], kwargs['alphas']):
                ....
```
or if adapters can't be set without alphas
```suggestion
        assert len(kwargs['alphas'], []) == len(kwargs["adapters"]), "`alphas` must be the same length as `adapters`"
```

likholat: Adapter can be set without alpha for all the cases (genai, optimum, hf).
I don’t really understand the purpose of the suggested change - it already works correctly when alpha is not provided

sbalandi: if alphas is not provided len(kwargs['alphas']) will be mean len(None) and it should rise TypeError. This doesn't happen now?

sbalandi: you are right, I have found [the place](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/wwb.py#L594) , alphas are converted to list and it never will be None

Then, this check `assert len(kwargs['alphas']) == len(kwargs["adapters"]), "alphas must be the same length as adapters" ` can be moved to https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/wwb.py#L592 and it doesn't need to keep https://github.com/openvinotoolkit/openvino.genai/blob/68c20f8f6e7901c0e1cfcbb1487625ce6e2f731d/tools/who_what_benchmark/whowhatbench/model_loaders.py#L145 since this code will never be executed

likholat: this check is already done in `wwb.py` `def check_args(args):` , deleted this assert

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -63,15 +63,22 @@ def load_text_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwargs):
             "Failed to import openvino_genai package. Please install it.")
         exit(-1)
 
+    adapter_config = openvino_genai.AdapterConfig()
+    if kwargs.get("adapters") is not None:
+        assert len(kwargs['alphas']) == len(kwargs["adapters"]), "`alphas` must be the same length as `adapters`"
+        for adapter, alpha in zip(kwargs['adapters'], kwargs['alphas']):
```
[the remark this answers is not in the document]
likholat: Adapters can be set without alphas

sbalandi (approved):

PR #2727, opened 2025-09-12
chang gpu_block_size to 256

ceciliapeng2011:
Tickets:


ceciliapeng2011:
close, as it is replaced by https://github.com/openvinotoolkit/openvino.genai/pull/2764

PR #2729, opened 2025-09-12
Allow additional_params for tokenizer decode in TextStreamer

dkalinowski:
This is needed in OVMS to be able to parse GPT-OSS tool/reasoning output in streaming manner.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -513,7 +513,7 @@ jobs:
           - name: 'Cacheopt E2E'
             cmd: 'tests/python_tests/test_kv_cache_eviction.py'
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).continuous_batching.test }}
-            timeout: 240
+            timeout: 360
           - name: 'LLM & VLM'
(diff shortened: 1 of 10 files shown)


src/cpp/include/openvino/genai/text_streamer.hpp
```diff
@@ -25,13 +25,14 @@ class OPENVINO_GENAI_EXPORTS TextStreamer : public StreamerBase {
 
     void end() override;
 
-    TextStreamer(const Tokenizer& tokenizer, std::function<CallbackTypeVariant(std::string)> callback);
+    TextStreamer(const Tokenizer& tokenizer, std::function<CallbackTypeVariant(std::string)> callback, bool skip_special_tokens = true);
```
as-suvorov: @apaniukov @pavel-esir Is it possible to set `skip_special_tokens` default for the tokenizer in the client code?
If not, than it seems better to pass `ov::AnyMap& tokenization_properties` or `Properties&&... properties` for extensibility.

as-suvorov (approved):

pavel-esir (approved):

src/python/openvino_genai/py_openvino_genai.pyi
Wovchena, line 1: You need to regenerate .pyi once more

PR #2730, opened 2025-09-12
[VLMPipeline] Run embed models on GPU when run LM on NPU

JohnLeFeng:
In default, when run VLM pipeline on NPU, it will run embed models on CPU, but CPU's computation resource is low, and it also need to handle other tasks, therefore, this change runs embed model on GPU to have better performance.

### DIFF
--- a/src/cpp/src/visual_language/pipeline.cpp
+++ b/src/cpp/src/visual_language/pipeline.cpp
@@ -71,13 +71,13 @@ public:
         //     ov::device::properties("NPU", ...),
         //     ov::device::properties("CPU", ...)
         // }
-        auto device_propertes = utils::pop_or_default<ov::AnyMap>(
+        auto device_properties = utils::pop_or_default<ov::AnyMap>(
             properties_copy, ov::device::properties.name(), { }
         );
         // Otherwise, the same properties are used for all models and devices
-        auto lm_properties = device_propertes.empty()
+        auto lm_properties = device_properties.empty()
             ? properties_copy
-            : utils::pop_or_default<ov::AnyMap>(device_propertes, device, {});
+            : utils::pop_or_default<ov::AnyMap>(device_properties, device, {});
 
         ov::CompiledModel compiled_language_model;
         auto embedder_device = device;
@@ -95,9 +95,9 @@ public:
         m_language = compiled_language_model.create_infer_request();
         m_language.get_tensor("attention_mask").set_shape({1, 0});
 
-        auto embedder_properties = device_propertes.empty()
+        auto embedder_properties = device_properties.empty()
             ? properties_copy
-            : utils::pop_or_default<ov::AnyMap>(device_propertes, embedder_device, {});
+            : utils::pop_or_default<ov::AnyMap>(device_properties, embedder_device, {});
 
         m_inputs_embedder = std::make_shared<InputsEmbedder>(models_dir, embedder_device, embedder_properties);
         m_tokenizer = m_inputs_embedder->get_tokenizer();


src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -71,18 +71,18 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+        auto lm_properties = device_properties.empty()
             ? properties_copy
-            : utils::pop_or_default<ov::AnyMap>(device_propertes, device, {});
+            : utils::pop_or_default<ov::AnyMap>(device_properties, device, {});
 
         ov::CompiledModel compiled_language_model;
         auto embedder_device = device;
         if (m_is_npu) {
-            embedder_device = "CPU";
+            embedder_device = "GPU";
```
intelgaoxiong, line 85: I think it's better to use an option to toggle CPU or GPU?

intelgaoxiong:
@AlexanderKalistratov Could you please take a look?
Thanks!

src/cpp/src/visual_language/pipeline.cpp
as-suvorov, line 85: Is GPU a safe default? What if drivers were not installed?

Wovchena: GenAI should allow per model user configuration to solve that problem

JohnLeFeng: @Wovchena, may I know how to config it? In NPU `VLMPipline`, it is using CPU as default, if GenAI has API to make user can handle this, I think that is better. 
If GenAI does not have this API, I am also glad to contribute.

Wovchena: GenAI does not have this API. @as-suvorov, will you guide how to implement it?

as-suvorov: @JohnLeFeng As a starting point can refer to image generation pipelines. Here is the sample: https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/image_generation/heterogeneous_stable_diffusion.cpp

And also to static pipeline members where initialized models can be passed: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/image_generation/text2image_pipeline.hpp#L95

Wovchena: These examples aren't really applicable to VLM because no building blocks are exposed

as-suvorov: I understand. There are references of building blocks for image generation above. The same can be implemented for VLMs.

Wovchena: It's bad idea to ask to implement someone outside of the team to implement that big feature requiring architecture review.
I was thinking about `ov::Property`

as-suvorov: Your intention about `ov::Property` wasn't clear.
Regarding implemeneation approaches. Currently we have image generation pipeline where per model devices specified by `.compile('model1_device', 'model2_device', ..., 'properties')` or with dedicated models builder `stable_diffusion(compiled_model1, compiled_model2, ...)`.
So you propose a different approach for VLMs with something like `ov::Property{{'language_model_device', 'CPU'}, {'vision_embeddings_device', 'GPU'}, ...}`?

Wovchena: Yes

JohnLeFeng: thanks for recommendation, I will update it once it ready

PR #2733, opened 2025-09-15
[VLM] Add nanoLLaVA

popovaan:
Ticket: CVS-155518

WWB results:

```
nanoLLaVA
genai PA    0.937279
genai SDPA  0.937279
optimum     0.997287

nanoLLaVA 1.5
genai PA    0.913653
genai SDPA  0.913653
optimum     0.995103
```


yatarkan (commented):
Please add tests with tiny-random-nanollava to test_vlm_pipeline.py

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include "visual_language/nanollava/classes.hpp"
+
+#include "visual_language/clip.hpp"
+
+#include "utils.hpp"
+
+namespace ov::genai {
+
+const std::string NATIVE_TAG = "<image>\n";
```
yatarkan, line 10: Please add nanollava image tag to docstrings and docs:
1. ./src/cpp/include/openvino/genai/visual_language/pipeline.hpp
2. ./src/python/py_vlm_pipeline.cpp
3. ./src/python/openvino_genai/py_openvino_genai.pyi
4. ./site/docs/use-cases/image-processing/_sections/_usage_options/index.mdx

src/cpp/src/visual_language/vlm_config.hpp
```diff
@@ -12,6 +12,7 @@ namespace ov::genai {
 enum class VLMModelType {
     MINICPM,
     LLAVA,
+    NANOLLAVA,
```
yatarkan, line 15: Please add nanollava to supported models list in docs:
`.site/docs/supported-models/_components/vlm-models-table/models.ts`

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Done.

src/cpp/src/visual_language/vlm_config.hpp
popovaan: Done.

src/cpp/src/utils.cpp
```diff
@@ -704,6 +704,35 @@ ov::Tensor merge_text_and_image_embeddings_llava(const ov::Tensor& input_ids, ov
...
+    size_t image_idx = 0;
+    while (text_token_idx < text_tokens_size) {
+        if (input_ids_data[text_token_idx] == image_tok) {
+            const auto im_embed = image_embeds[image_idx];
+            image_idx++;
+            std::memcpy(inputs_embeds.data() + res_embeds_position * sizeof(float), im_embed.data(), im_embed.get_byte_size());
+            res_embeds_position += ov::shape_size(im_embed.get_shape());
+        }
+        else {
+            std::memcpy(inputs_embeds.data() + res_embeds_position * sizeof(float), text_embeds_data + text_token_idx * hidden_size, hidden_size * sizeof(float));
```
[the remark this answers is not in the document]
popovaan: It's wrong, `inputs_embeds.data()` returns bytes pointer, for this reason position is multiplied by `sizeof(float)`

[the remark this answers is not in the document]
popovaan: It's wrong, `inputs_embeds.data()` returns bytes pointer, for this reason position is multiplied by `sizeof(float)`

site/docs/supported-models/_components/vlm-models-table/models.ts
```diff
@@ -44,6 +44,26 @@ export const VLM_MODELS: VLMModelType[] = [
...
+  {
+    architecture: 'nanoLLaVA',
+    models: [
+      {
+        name: 'nanoLLaVA-1.5',
+        loraSupport: false,
+        links: ['https://huggingface.co/qnguyen3/nanoLLaVA-1.5'],
+      },
+    ],
+  },
```
yatarkan, line 61:
```suggestion
  {
    architecture: 'nanoLLaVA',
    models: [
      {
        name: 'nanoLLaVA',
        loraSupport: false,
        links: ['https://huggingface.co/qnguyen3/nanoLLaVA'],
      },
      {
        name: 'nanoLLaVA-1.5',
        loraSupport: false,
        links: ['https://huggingface.co/qnguyen3/nanoLLaVA-1.5'],
      },
    ],
  },
```

src/cpp/src/utils.cpp
```diff
@@ -704,6 +704,35 @@ ov::Tensor merge_text_and_image_embeddings_llava(const ov::Tensor& input_ids, ov
     return inputs_embeds;
 }
 
+ov::Tensor merge_text_and_image_embeddings_nanollava(const ov::Tensor& input_ids, ov::Tensor& text_embeds, const std::vector<ov::Tensor>& image_embeds, int64_t image_tok) {
```
Wovchena: `merge_text_and_image_embeddings_llava()` is in utils because it's shared between models. `merge_text_and_image_embeddings_nanollava()` is unique to nanolava. Move it to classes.cpp

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,178 @@
...
+    embeds.reserve(single_images.size());
+    for (const ov::Tensor& image : single_images) {
+        embeds.emplace_back(m_vision_encoder->encode(image, vision_config));
+    }
+    return embeds;
+}
+
+std::pair<std::string, std::vector<size_t>> InputsEmbedderNanoLLaVA::normalize_prompt(const std::string& prompt, size_t base_id, const std::vector<EncodedImage>& images) const {
+    auto [unified_prompt, images_sequence] = normalize(prompt, NATIVE_TAG, NATIVE_TAG, base_id, images.size());
+    return {std::move(unified_prompt), std::move(images_sequence)};
```
Wovchena:
```suggestion
    return normalize(prompt, NATIVE_TAG, NATIVE_TAG, base_id, images.size());
```

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,178 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "visual_language/nanollava/classes.hpp"
+#include "visual_language/clip.hpp"
+#include "utils.hpp"
+
+namespace ov::genai {
+
+const std::string NATIVE_TAG = "<image>\n";
+const int64_t IMAGE_TOKEN_ID = -200;
```
Wovchena:
```suggestion
const int64_t IMAGE_PLACEHOLDER = -200;
```
It was confusing to see `tokenize_without_image_tag()` extracting `<image>` just to insert `IMAGE_TOKEN_ID`. Without noticing its value is negative it seemed like reverting. Keep `IMAGE_TOKEN_ID` if it's the name used in python implementation.

Would it be possible to search for `<image>` corresponding token id? I guess it was implemented the way you did it originally, just curious

src/cpp/src/visual_language/nanollava/classes.cpp
Wovchena, line 1: > Please add tests with tiny-random-nanollava to test_vlm_pipeline.py

That's a good request. I'd just comment that nanollava seem to be already small enough so you could probably use it directly

Wovchena (changes requested):

src/cpp/src/visual_language/nanollava/classes.cpp
Wovchena, line 1: ` OPENVINO_LOG_LEVEL=9 gdb -ex 'set args ./i/nanoLLaVA-1.5 ./i/images' /home/vzlobin/z/v/bin/intel64/Debug/visual_language_chat -ex r` and question `How many images are here?` fails with `free(): corrupted unsorted chunks`
The backtrace points to `#11 0x00007ffff7549247 in ov::genai::ContinuousBatchingPipeline::IContinuousBatchingPipeline::generate(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&, std::vector<std::vector<ov::Tensor, std::allocator<ov::Tensor> >, std::allocator<std::vector<ov::Tensor, std::allocator<ov::Tensor> > > > const&, std::vector<ov::genai::GenerationConfig, std::allocator<ov::genai::GenerationConfig> > const&, std::variant<std::function<bool (std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)>, std::function<ov::genai::StreamingStatus (std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)>, std::shared_ptr<ov::genai::StreamerBase>, std::monostate> const&) (this=0x555556f5bcf0, prompts=std::vector of length 1, capacity 1 = {...}, rgbs_vector=std::vector of length 1, capacity 1 = {...}, 
    sampling_params=std::vector of length 1, capacity 1 = {...}, streamer=std::variant [no contained value])
    at /home/vzlobin/z/g/src/cpp/src/continuous_batching/pipeline_base.cpp:193` which isn't helpful.

There are 3 images in the folder

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: I tried it, but tokenization result is different for following cases:
`encode("text1<image>text2")` != `encode("text1")+im_token+encode("text2")`, even if image token id from tokenizer is used, tokens after image token id are different for these two cases. So we need to tokenize text without image tag, in hugging face and optimum it is done the same way.
Regarding `-200` value, I think we can probably use image token id from tokenizer, as `-200` is removed from final embeddings tensor anyway and replaced by output from vision encoder, but `-200` is used in python, so I used here the same value.

Wovchena: The value is fine. The name of the var was confusing

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Fixed.

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Agree, renamed.

src/cpp/src/utils.cpp
popovaan: Done.

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Done.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -68,8 +68,21 @@ def get_ov_model(model_id):
...
+        lambda: OVModelForVisualCausalLM.from_pretrained(
+            model_id,
+            compile=False,
+            device="CPU",
+            export=True,
+            load_in_8bit=False,
+            trust_remote_code=True,
+            ov_config=get_default_llm_properties(),
+        )
+    )
```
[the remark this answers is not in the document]
popovaan: In lines below this code is deleted, so there's no duplications here.

popovaan:
> Will you add WWB test here?

Let's add it in a separate PR along with the fix to WWB to support nanollava there.

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,289 @@
...
+            new_chat_tokens_data[0] = IMAGE_PLACEHOLDER;
+            new_chat_tokens_data ++;
+        }
+    }
+    return new_chat_tokens;
+}
+
+ov::Tensor InputsEmbedderNanoLLaVA::apply_chat_template_tokenize(const std::string& prompt, ov::genai::VLMPerfMetrics& metrics) {
+    if (m_is_chat_conversation) {
+        std::string prompt_to_encode = prompt;
```
yatarkan: Seems to be unused

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,289 @@
...
+        auto start_tokenizer_time = std::chrono::steady_clock::now();
+        ov::Tensor new_chat_tokens = tokenize_without_image_tag(prompt);
```
yatarkan: `IInputsEmbedder::apply_chat_template_tokenize(...)` method for encoding also uses `add_special_tokens` property with different values for chat mode (`false`), chat template application (`false`) and pure string encoding (`true`).
I understand that `tokenize_without_image_tag(...)` is encoding prompt by chunks, but can we face here a discrepancy with special tokens? Or this particular tokenizer does not add them?

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+
+ov::Tensor InputsEmbedderNanoLLaVA::tokenize_without_image_tag(const std::string& prompt) {
+    std::string prompt_substr;
+    size_t prev_pos = 0;
+    std::vector<ov::Tensor> encoded_substrings;
+    size_t encoded_size = 0;
+    size_t n_images = 0;
+    const std::string image_tag = "<image>";
```
yatarkan, line 186: Can we use `NATIVE_TAG` here? And should we split by string with newline?

src/cpp/src/visual_language/nanollava/classes.cpp
```diff
@@ -0,0 +1,289 @@
...
+    if (m_image_features_size > 0)
+        new_chat_tokens = insert_image_placeholders(new_chat_tokens, m_image_features_size);
+    auto new_input_ids = IInputsEmbedder::update_history(new_chat_tokens);
+    m_prev_hist_length = m_kv_cache_state.get_state().size();
+    m_kv_cache_state.add_inputs(new_input_ids);
+
+    if (m_image_features_size > 0)
+        new_input_ids = drop_image_placeholders(new_input_ids, m_image_features_size);
+    return new_input_ids;
+}
```
yatarkan: Is it possible to insert image placeholders (`m_image_features_size` times) in `apply_chat_template_tokenize(...)` or in `tokenize_without_image_tag(...)`?
If so, we will not need to change default `get_encoded_input_ids(...)` method.
And if we need to remove image placeholders later then, it can be done in `InputsEmbedderNanoLLaVA::get_inputs_embeds()`, after calling `get_encoded_input_ids()` before passing `input_ids` to text embedding model.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -956,21 +996,38 @@ def test_vlm_pipeline_match_optimum_preresized(request, model_id, image_name, ba
...
+    if model.config.model_type == "llava-qwen2":
+        processor = get_nanollava_processor()
+        tokenizer = transformers.AutoTokenizer.from_pretrained(model_id, trust_remote_code=True)
 
-    # Gemma3 input_ids has two bos tokens when running with optimum: one in chat template + "add_bos_token" is set to True in tokenizer_config.json
-    if model.config.model_type == "gemma3":
-        processor.tokenizer.add_bos_token = False
+        from optimum.intel.openvino.modeling_visual_language import MODEL_TYPE_TO_CLS_MAPPING
+        preprocess_inputs = MODEL_TYPE_TO_CLS_MAPPING[model.config.model_type].preprocess_inputs
+        inputs = preprocess_inputs(prompt, resized_image, processor, tokenizer, config=model.config)
```
yatarkan, line 1011: Probably we can use `preprocess_inputs` not only for nanoLLaVa but for all other models.
But I believe this is out of scope of this PR

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: I tried to split by '`<image>\n`', but inference result started to mismatch with pytorch.
Optimum uses "`<image>`" for splitting as well: https://github.com/huggingface/optimum-intel/blob/04db016571d1a19c14918553365ee4c05c8b4697/optimum/intel/openvino/modeling_visual_language.py#L2317

tests/python_tests/test_vlm_pipeline.py
popovaan: Yes, that's what WWB does, so it should be applicable for all models.

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Yes, removed.

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Agree, it's a bug. Added `add_special_tokens` flag to `tokenize_without_image_tag()` and passed correct values for chat template/non chat template cases.

src/cpp/src/visual_language/nanollava/classes.cpp
popovaan: Yes, done.

PR #2743, opened 2025-09-18
Increase timeouts

Wovchena:
Failed in https://github.com/openvinotoolkit/openvino.genai/actions/runs/17818164864/job/50662923279?pr=2720 and https://github.com/openvinotoolkit/openvino.genai/actions/runs/17818164863/job/50657024432?pr=2720


.github/workflows/linux.yml
```diff
@@ -513,15 +513,15 @@ jobs:
           - name: 'Cacheopt E2E'
             cmd: 'tests/python_tests/test_kv_cache_eviction.py'
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).continuous_batching.test }}
-            timeout: 240
+            timeout: 360
```
mryzhov, line 516: Folks, we have to change the tests instead of increasing the timeouts, 360 minutes is unacceptable time for one job!

PR #2744, opened 2025-09-18
Using pytest cache instead of ov cache env variable

sgonorov:
Part of [CVS-173285](https://jira.devtools.intel.com/browse/CVS-173285)

Adds support for pytest cache plugin, also some fixture refactorings in order to support it better

### DIFF
--- a/tests/python_tests/test_image_generation.py
+++ b/tests/python_tests/test_image_generation.py
@@ -9,7 +9,6 @@ import numpy as np
 import openvino as ov
 import openvino_genai as ov_genai
 
-from utils.constants import get_ov_cache_converted_models_dir
 from utils.atomic_download import AtomicDownloadManager
 from utils.network import retry_request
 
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -515,7 +514,7 @@ jobs:
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).continuous_batching.test }}
             timeout: 360
           - name: 'LLM & VLM'
-            cmd: 'tests/python_tests/test_llm_pipeline.py tests/python_tests/test_llm_pipeline_static.py tests/python_tests/test_vlm_pipeline.py tests/python_tests/test_structured_output.py --override-ini cache_dir=/mount/caches/pytest/'
+            cmd: 'tests/python_tests/test_llm_pipeline.py tests/python_tests/test_llm_pipeline_static.py tests/python_tests/test_vlm_pipeline.py tests/python_tests/test_structured_output.py -o cache_dir=/mount/caches/pytest/lin'
```
Wovchena: `-o` is an alias for `--override-ini`. `--override-ini` is more verbose. Why replace with `-o`?

sgonorov: It's for consistency - all other commands use `-o`

Wovchena: I can't fine any. Can you like a specific place?

PR #2757, opened 2025-09-22
[Embeddings] Add last token pooling

as-suvorov:
Required to enable qwen3 embedding models: https://huggingface.co/Qwen/Qwen3-Embedding-0.6B
Ticket: 173649

### DIFF
--- a/tests/python_tests/requirements.txt
+++ b/tests/python_tests/requirements.txt
@@ -4,7 +4,7 @@ optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@04db016571d
 numpy==1.26.4; platform_system == "Darwin" and platform_machine == "x86_64"
 safetensors==0.6.2; platform_system == "Darwin" and platform_machine == "x86_64"
 pytest==8.4.2
-transformers==4.52.4
+transformers==4.53.3
 hf_transfer==0.1.9
 gguf==0.17.1
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,13 @@
+## Description
+<!-- Please include a summary of the change. Also include relevant motivation and context. -->
+
+<!--- Jira ticket number (e.g., 123). Delete if there's no ticket. Don't include full link or project name. -->
+Ticket: 
+
+<!-- Remove if not applicable -->
+Fixes #(issue)
+
+## Checklist:
+- [ ] Tests have been updated or added to cover the new code <!--- If the change isn't maintenance related, update the tests at https://github.com/openvinotoolkit/openvino.genai/tree/master/tests or explain in the description why the tests don't need an update. -->
+- [ ] This patch fully addresses the ticket. <!--- If follow-up pull requests are needed, specify in description. -->
+- [ ] I have made corresponding changes to the documentation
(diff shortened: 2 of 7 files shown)


tests/python_tests/requirements.txt
```diff
@@ -1,6 +1,6 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 diffusers==0.34.0
-optimum-intel==1.25.2
+optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@04db016571d1a19c14918553365ee4c05c8b4697
```
as-suvorov, line 3: optimum-intel PR https://github.com/huggingface/optimum-intel/pull/1415 required.

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -102,6 +118,8 @@ std::shared_ptr<Model> apply_postprocessing(std::shared_ptr<Model> model, const
             return get_cls_pooling_op(node);
         } else if (config.pooling_type == TextEmbeddingPipeline::PoolingType::MEAN) {
             return get_mean_pooling_op(model, node);
+        } else if (config.pooling_type == TextEmbeddingPipeline::PoolingType::LAST_TOKEN) {
+            return get_last_token_pooling_op(node);
```
apaniukov: Should we check if the left padding is applied here, and either:
1) Raise an exception if not or
2) Pass the model and apply the transformation with the attention mask
?

as-suvorov: Added support to default right padding for LAST_TOKEN pooling

tests/python_tests/test_rag.py
```diff
@@ -241,6 +275,22 @@ def test_embedding_constructors(download_and_convert_embeddings_models):
...
+@pytest.mark.parametrize("download_and_convert_embeddings_models", ["Qwen/Qwen3-Embedding-0.6B"], indirect=True)
+@pytest.mark.parametrize(
+    "config",
+    [
+        TextEmbeddingPipeline.Config(normalize=False, pooling_type=TextEmbeddingPipeline.PoolingType.LAST_TOKEN, padding_side="left"),
+        TextEmbeddingPipeline.Config(normalize=False, pooling_type=TextEmbeddingPipeline.PoolingType.LAST_TOKEN),
+    ],
+)
+@pytest.mark.precommit
+def test_qwen3_embedding(download_and_convert_embeddings_models, dataset_documents, config):
```
apaniukov, line 287: Should we have a separate parametrize for the padding side?

apaniukov (approved):

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -6,6 +6,7 @@
 #include <fstream>
 #include <nlohmann/json.hpp>
 
+#include "debug_utils.hpp"
```
yatarkan: Why do we need debug utils here?

as-suvorov:
```suggestion
```

as-suvorov: We don't, removed

as-suvorov:
> Should we also update [docs](https://openvinotoolkit.github.io/openvino.genai/docs/use-cases/text-embedding/#pooling-strategies)?

Good point, thanks. We also need to include fixed shapes properties. I created a ticket: 174069

Wovchena (approved):

PR #2751, opened 2025-09-22
Added Model Caching example in Whisper README.md

luke-lin-vmc:

PR #2759, opened 2025-09-23
Enable model caching for Whisper pipeline on GPU and NPU

luke-lin-vmc:
Whisper sample code to enable model caching on GPU and NPU

This is https://github.com/openvinotoolkit/openvino.genai/pull/2751 follow up

Sample Code Reference: 
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/visual_language_chat/encrypted_model_vlm.py#L87
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/text_generation/encrypted_model_causal_lm.cpp#L52

OPTIMIZE_SIZE and encryption are not included. The main performance concern for Whisper is pipeline speed. Since Whisper is much smaller than LLMs, size optimization offers only very little savings while potentially adding latency. Similarly, model encryption can also introduce additional latency.

### DIFF
--- a/samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
+++ b/samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
@@ -4,6 +4,12 @@
 #include "audio_utils.hpp"
 #include "openvino/genai/whisper_pipeline.hpp"
 
+auto get_config_for_cache() {
+    ov::AnyMap config;
+    config.insert({ov::cache_dir("whisper_cache")});
+    return config;
+}
+
 int main(int argc, char* argv[]) try {
     if (argc < 3 || argc > 4) {
         throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> \"<WAV_FILE_PATH>\" <DEVICE>");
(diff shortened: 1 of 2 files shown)


samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
```diff
@@ -13,7 +19,14 @@ int main(int argc, char* argv[]) try {
     std::string wav_file_path = argv[2];
     std::string device = (argc == 4) ? argv[3] : "CPU";  // Default to CPU if no device is provided
 
-    ov::genai::WhisperPipeline pipeline(models_path, device);
+    ov::AnyMap ov_config;
+    if (device == "NPU" || device.find("GPU") != std::string::npos) {  // need to handle cases like "GPU", "GPU.0" and "GPU.1"
+        // Cache compiled models on disk for GPU and NPU to save time on the
+        // next run. It's not beneficial for CPU.
```
as-suvorov, line 25: Why it's not beneficial for CPU?

luke-lin-vmc: 1. This comment is simply copied from the [reference sample code](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/visual_language_chat/encrypted_model_vlm.py#L118).
2. AFAIK CPU plugin's "compile" step is mostly graph rewrites and primitive selection. It’s typically milliseconds–a few hundred ms, not seconds–minutes like on GPU/NPU.
3. Most importantly, enable model caching on CPU causes Whisper pipeline crashed. This looks like a bug which needs further investigation. So currently model caching is enabled only on GPU and NPU to avoid the issue.

Wovchena:
build_jenkins

PR #2763, opened 2025-09-23
[llm_bench] Fix OpenVINO config not being passed for speech-to-text and Whisper models

aobolensk:


tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -543,12 +543,14 @@ def create_speech_2_txt_model(model_path, device, memory_monitor, **kwargs):
                 log.info("Selected OpenVINO GenAI for benchmarking")
                 return create_genai_speech_2_txt_model(model_path, device, memory_monitor, **kwargs)
         log.info("Selected Optimum Intel for benchmarking")
+        ov_config = kwargs['config']
```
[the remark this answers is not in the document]
aobolensk: Not sure that this is relevant, since `ov_config = kwargs['config']` is used for other models as well

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -580,12 +580,14 @@ def create_speech_2_txt_model(model_path, device, memory_data_collector, **kwarg
                 log.info("Selected OpenVINO GenAI for benchmarking")
                 return create_genai_speech_2_txt_model(model_path, device, memory_data_collector, **kwargs)
```
sbalandi, line 582: Should config also be added to the creation of [ov_genai.WhisperPipeline](https://github.com/openvinotoolkit/openvino.genai/blob/891a5803097709047211849de488e53e6146964e/tools/llm_bench/llm_bench_utils/ov_utils.py#L548C18-L548C42) ​​?

aobolensk: Yes, it should. Added

PR #2764, opened 2025-09-24
update gpu block size based on xattn

rnwang04:
- block size for XAttention
- work with
-- https://github.com/openvinotoolkit/openvino/pull/32064
-- https://github.com/openvinotoolkit/openvino/pull/32551

Tickets: CVS-173845

Related PRs
- https://github.com/openvinotoolkit/openvino.genai/pull/2924
- https://github.com/openvinotoolkit/openvino.genai/pull/2927

### DIFF
--- a/tools/llm_bench/llm_bench_utils/ov_utils.py
+++ b/tools/llm_bench/llm_bench_utils/ov_utils.py
@@ -190,12 +190,10 @@ def get_scheduler_config_genai(config_data, config_name="CB config"):
             sparse_attention_kwargs = user_config.pop('sparse_attention_config')
             if "mode" in sparse_attention_kwargs.keys():
                 sparse_attention_kwargs["mode"] = getattr(openvino_genai.SparseAttentionMode, sparse_attention_kwargs["mode"])
-            if user_config.pop('use_sparse_attention', True):
-                scheduler_config.use_sparse_attention = True
-                scheduler_config.sparse_attention_config = openvino_genai.SparseAttentionConfig(**sparse_attention_kwargs)
-                log.info("Sparse Attention mode ON")
-            else:
-                raise RuntimeError("==Failure==: sparse_attention_config value can't be used with use_sparse_attention=False")
+
+            scheduler_config.use_sparse_attention = True
+            scheduler_config.sparse_attention_config = openvino_genai.SparseAttentionConfig(**sparse_attention_kwargs)
+            log.info("Sparse Attention mode ON")
 
         for param, value in user_config.items():
             setattr(scheduler_config, param, value)
--- a/tools/llm_bench/task/text_generation.py
+++ b/tools/llm_bench/task/text_generation.py
@@ -294,7 +294,7 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
             "Enabled input prompt permutations. It means that generated results may vary on different steps. "
             "If it is not expected, please specify --disable_prompt_permutation in your benchmarking command to disable this behavior"
         )
-        from openvino_genai import TokenizedInputs
+        from openvino_genai import TokenizedInputs, GenerationConfig
         import openvino as ov
 
         input_ids = input_data.input_ids.data
@@ -315,7 +315,6 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
         if args['infer_count'] is not None:
             out_str += 'all max_output_token_size: {} * {}'.format(args['infer_count'], args['batch_size'])
         log.info(out_str)
-    from openvino_genai import GenerationConfig
     gen_config = model.get_generation_config() if hasattr(model, 'get_generation_config') else GenerationConfig()
     gen_config.max_new_tokens = max_gen_tokens
     # llama-3-8b-instruct's generation_config.json has 4096 max_length.
--- a/tools/who_what_benchmark/whowhatbench/model_loaders.py
+++ b/tools/who_what_benchmark/whowhatbench/model_loaders.py
@@ -42,36 +42,14 @@ class GenAIModelWrapper:
             return getattr(self.model, attr)
 
 
-def configure_sparse_attention(scheduler_params, scheduler_config):
-    """
-    Configures sparse attention settings based on scheduler parameters.
-    """
-    import openvino_genai
-    sparse_attention_kwargs = scheduler_params.pop('sparse_attention_config', None)
-
-    if sparse_attention_kwargs:
-        # Convert mode string to enum if present
-        mode = sparse_attention_kwargs.get("mode")
-        if mode:
-            sparse_attention_kwargs["mode"] = getattr(openvino_genai.SparseAttentionMode, mode)
-
-        # Check if sparse attention is enabled
-        if scheduler_params.pop('use_sparse_attention', True):
-            scheduler_config.use_sparse_attention = True
-            scheduler_config.sparse_attention_config = openvino_genai.SparseAttentionConfig(**sparse_attention_kwargs)
-            logger.info("Sparse Attention mode ON")
-        else:
-            raise RuntimeError("==Failure==: sparse_attention_config value can't be used with use_sparse_attention=False")
-
-
 def get_scheduler_config_genai(cb_config):
     import openvino_genai
+
     default_cb_config = {"cache_size": 1}
     scheduler_config = openvino_genai.SchedulerConfig()
     scheduler_params = cb_config or default_cb_config
     if scheduler_params:
         logger.info(f"Scheduler parameters for:\n{scheduler_params}")
-        configure_sparse_attention(scheduler_params, scheduler_config)
         for param, value in scheduler_params.items():
             if param == "cache_eviction_config":
                 value = openvino_genai.CacheEvictionConfig(aggregation_mode=openvino_genai.AggregationMode.NORM_SUM, **value)


src/cpp/src/continuous_batching/cache_manager.hpp
```diff
@@ -44,8 +44,14 @@ class CacheManager {
         OPENVINO_ASSERT(all_gpu_device || execution_devices.size() == 1,
                         "Continuous batching: execution device is expected to be single CPU / single GPU / multi GPUs");
         m_device = execution_devices[0];
+        // decide gpu_block_size based on xattn enable status
+        const char* env_p = std::getenv("OV_GPU_XATTN_BLOCK_SIZE");
+        size_t gpu_block_size = 256;
+        if (env_p != nullptr && std::stoi(env_p) == 1) {
+            gpu_block_size = 16;
```
ceciliapeng2011: This ticket https://jira.devtools.intel.com/browse/CVS-161326 makes genai get kv cache actual precision from plugin after compiling. Can you please refer it to make genai get kv_block_size from plugin too?

rnwang04: Thanks for the comment! Have updated related code, would you mind taking a look again ?

tools/llm_bench/task/text_generation.py
```diff
@@ -315,6 +315,7 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
         if args['infer_count'] is not None:
             out_str += 'all max_output_token_size: {} * {}'.format(args['infer_count'], args['batch_size'])
         log.info(out_str)
+    from openvino_genai import GenerationConfig
```
Wovchena: There's `from openvino_genai import TokenizedInputs, GenerationConfig` on top. Please remove `GenerationConfig`

rnwang04: If remove this line, I will meet below error when I benchmark with `--disable_prompt_permutation`:
```python
  File "/home/arda/ruonan/openvino.genai/tools/llm_bench/task/text_generation.py", line 319, in run_text_generation_genai
    gen_config = model.get_generation_config() if hasattr(model, 'get_generation_config') else GenerationConfig()
                                                                                               ^^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'GenerationConfig' where it is not associated with a value
```

Above `GenerationConfig` only works for `enable_prompt_permutations` case:
image

So I perfer to keep this line here and change line 297 from `from openvino_genai import TokenizedInputs, GenerationConfig` to `from openvino_genai import TokenizedInputs`.

Not sure do you agree with it ? 😊

Wovchena: Agree

rnwang04: Thanks ! Have changed to this way.

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -190,10 +190,14 @@ def get_scheduler_config_genai(config_data, config_name="CB config"):
...
                 sparse_attention_kwargs["mode"] = getattr(openvino_genai.SparseAttentionMode, sparse_attention_kwargs["mode"])
-
-            scheduler_config.use_sparse_attention = True
-            scheduler_config.sparse_attention_config = openvino_genai.SparseAttentionConfig(**sparse_attention_kwargs)
-            log.info("Sparse Attention mode ON")
+            if 'use_sparse_attention' in user_config.keys():
+                use_sparse_attention = user_config.pop('use_sparse_attention')
+            else:
+                use_sparse_attention = None
+            if use_sparse_attention != False:
```
Wovchena:
```suggestion
            if user_config.pop('use_sparse_attention', True):
```
Seems to be eqvivalent

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -190,10 +190,14 @@ def get_scheduler_config_genai(config_data, config_name="CB config"):
...
+                scheduler_config.use_sparse_attention = True
+                scheduler_config.sparse_attention_config = openvino_genai.SparseAttentionConfig(**sparse_attention_kwargs)
+                log.info("Sparse Attention mode ON")
```
Wovchena:
```suggestion
                log.info("Sparse Attention mode ON")
            else:
                raise RuntimeError("==Failure ==: sparse_attention_config value can't be used with use_sparse_attention=False")
```
@sbalandi is on holidays, so let's make it conservative avoiding implicit `use_sparse_attention`

tools/llm_bench/llm_bench_utils/ov_utils.py
rnwang04: Thanks for your suggestion! Have changed to this way.

tools/llm_bench/llm_bench_utils/ov_utils.py
rnwang04: Thanks for your suggestion! Have changed to this way.

vshampor:
The test at https://github.com/openvinotoolkit/openvino.genai/blob/2650995f72798a5ef70173ad3905c9ddd7e30b8e/tests/python_tests/test_kv_cache_eviction.py#L105 must be extended with the XAttention case, and/or additional tests must be added to demonstrate just what behaviour you had in mind for the case where the user just tries to switch from `SparseAttentionMode::TRISHAPE` to `SparseAttentionMode::XATTENTION` without changing his expectations about the block size of 16 that TRISHAPE used to work perfectly with.

Also, I guess I'll have to ask again - why is it that only enabling xattention changes the GPU block size implicitly to 128? If it's such a performant block size anyway, why do you not change the GPU block size to 128 everywhere and avoid implicit block size changes of which the user is unaware?

vshampor (dismissed):
See my previous comment

ceciliapeng2011:
> The test at
> 
> https://github.com/openvinotoolkit/openvino.genai/blob/2650995f72798a5ef70173ad3905c9ddd7e30b8e/tests/python_tests/test_kv_cache_eviction.py#L105
> 
> must be extended with the XAttention case,
 
Could do it. But how about a separate PR after CPU xattn is merged? Hope this won't a blocker to GPU xattn.



>and/or additional tests must be added to demonstrate just what behaviour you had in mind for the case where the user just tries to switch from `SparseAttentionMode::TRISHAPE` to `SparseAttentionMode::XATTENTION` without changing his expectations about the block size of 16 that TRISHAPE used to work perfectly with.

If I understand correctly, "SparseAttentionMode" is set during compiling time when users create a LLMPipeline. The switch between TRISHAPE and XATTENTION will trigger plugin to re-compile the model. Any problem here?



> Also, I guess I'll have to ask again - why is it that only enabling xattention changes the GPU block size implicitly to 128? If it's such a performant block size anyway, why do you not change the GPU block size to 128 everywhere and avoid implicit block size changes of which the user is unaware?

That's a good question. I think here you are mentioning GPU KV block size. It is 256. There are two sets of KV block size: 256 for xattention + xe2/xe3+ new GPU architectures, and 16 for legacy GPUs or without xattention. Sergey once suggested to change the block size to 256 everywhere. We have a phase-by-phase plan to do that. Considering execution priority and effort, this work will be happening after full functionality of xattention is productized. We've talked about this in the email thread.

peterchen-intel:
> > The test at
> > https://github.com/openvinotoolkit/openvino.genai/blob/2650995f72798a5ef70173ad3905c9ddd7e30b8e/tests/python_tests/test_kv_cache_eviction.py#L105
> > 
> > must be extended with the XAttention case,
> 
> Could do it. But how about a separate PR after CPU xattn is merged? Hope this won't a blocker to GPU xattn.
> 

@vshampor  I think we have to go in this way, or else the test case will always fail. Created a ticket to track CVS-175120

src/cpp/src/continuous_batching/cache_manager.hpp
```diff
@@ -75,6 +74,19 @@ class CacheManager {
...
+                has_xattention = true;
+            }
+            if (utils::env_setup_for_print_debug_info()) {
+                if (has_xattention)
+                    std::cout << "[XAttention]: ENABLED on GPU device." << std::endl;
+                else
+                    std::cout << "[XAttention]: DISABLED on GPU device." << std::endl;
+            }
+        }
+        m_block_size = all_gpu_device ? ( has_xattention ? gpu_block_size_xattn : gpu_block_size ) : cpu_block_size;
```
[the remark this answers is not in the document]
ceciliapeng2011: won't apply

peterchen-intel:
> > The test at
> > https://github.com/openvinotoolkit/openvino.genai/blob/2650995f72798a5ef70173ad3905c9ddd7e30b8e/tests/python_tests/test_kv_cache_eviction.py#L105
> > 
> > must be extended with the XAttention case,
> 
> Could do it. But how about a separate PR after CPU xattn is merged? Hope this won't a blocker to GPU xattn.
> 
> > and/or additional tests must be added to demonstrate just what behaviour you had in mind for the case where the user just tries to switch from `SparseAttentionMode::TRISHAPE` to `SparseAttentionMode::XATTENTION` without changing his expectations about the block size of 16 that TRISHAPE used to work perfectly with.
> 
> If I understand correctly, "SparseAttentionMode" is set during compiling time when users create a LLMPipeline. The switch between TRISHAPE and XATTENTION will trigger plugin to re-compile the model. Any problem here?
> 
> > Also, I guess I'll have to ask again - why is it that only enabling xattention changes the GPU block size implicitly to 128? If it's such a performant block size anyway, why do you not change the GPU block size to 128 everywhere and avoid implicit block size changes of which the user is unaware?
> 
> That's a good question. I think here you are mentioning GPU KV block size. It is 256. There are two sets of KV block size: 256 for xattention + xe2/xe3+ new GPU architectures, and 16 for legacy GPUs or without xattention. Sergey once suggested to change the block size to 256 everywhere. We have a phase-by-phase plan to do that. Considering execution priority and effort, this work will be happening after full functionality of xattention is productized. We've talked about this in the email thread.

@vshampor  @ceciliapeng2011  Let's continue the discussion in CVS-175590 (GPU KV block size) and CVS-175120 (test XATTN with CPU)

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -42,14 +42,36 @@ def __getattr__(self, attr):
             return getattr(self.model, attr)
 
 
-def get_scheduler_config_genai(cb_config):
+def configure_sparse_attention(scheduler_params, scheduler_config):
```
peterchen-intel: https://github.com/openvinotoolkit/openvino.genai/pull/2895  is working on the same purpose.

ceciliapeng2011 (approved):

peterchen-intel:
As talked in chat, merge this one and the follow up CVS-175590 (GPU KV block size) and CVS-175120 (test XATTN with CPU)

Wovchena:
Was this chat with Vasily?

PR #2765, opened 2025-09-24
Add .github/pull_request_template.md

Wovchena:
This should remind the contributors to add tests

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,5 +1,13 @@
-Ticket <!--- Jira ticket number (e.g., 123). Delete if there's no ticket. Don't include full link or project name. -->
+## Description
+<!-- Please include a summary of the change. Also include relevant motivation and context. -->
 
-This patch fully addresses the ticket. <!--- If follow-up pull requests are needed, specify here. -->
+<!--- Jira ticket number (e.g., 123). Delete if there's no ticket. Don't include full link or project name. -->
+Ticket: 
 
-<!--- If the change isn't maintenance related, update the tests at https://github.com/openvinotoolkit/openvino.genai/tree/master/tests or explain in the description why the tests don't need an update. -->
+<!-- Remove if not applicable -->
+Fixes #(issue)
+
+## Checklist:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,5 @@
+Ticket <!--- Jira ticket number (e.g., 123). Delete if there's no ticket. Don't include full link or project name. -->
+
+This patch fully addresses the ticket. <!--- If follow-up pull requests are needed, specify here. -->
+
+<!--- If the change isn't maintenance related, update the tests at https://github.com/openvinotoolkit/openvino.genai/tree/master/tests or explain in the description why the tests don't need an update. -->
```
as-suvorov:
```suggestion
## Description
<!-- Please include a summary of the change. Also include relevant motivation and context. -->

<!--- Jira ticket number (e.g., 123). Delete if there's no ticket. Don't include full link or project name. -->
Ticket: 

<!-- Remove if not applicable -->
Fixes #(issue)

## Checklist:
- [ ] Tests have been updated or added to cover the new code <!--- If the change isn't maintenance related, update the tests at https://github.com/openvinotoolkit/openvino.genai/tree/master/tests or explain in the description why the tests don't need an update. -->
- [ ] This patch fully addresses the ticket. <!--- If follow-up pull requests are needed, specify in description. -->
- [ ] I have made corresponding changes to the documentation
```

as-suvorov: Maybe we can also include Who can review section, for example: https://github.com/huggingface/transformers/blob/main/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L30

Wovchena: Transformers limit the reviewers to wait for the tests. We don't have that limitation so there's nothing to specify.

Adding `Description` heading to PR's description is tautology. And Copilot summarizes the changes these days.

My concern with the checklist is that the templates end up being ignored. For example https://github.com/openvinotoolkit/openvino/pull/30383 So I moved into the comments as much as possible. `Ticket` stayed because it's OV's template so I wasn't sure if I should have hidden it. `This patch fully addresses the ticket.` is the default PR's state.

I think we can do it your way and see how it goes

MaximProshin (approved):

MaximProshin:
LGTM

PR #2768, opened 2025-09-24
Remove hardcoded fields and pass special variables to `apply_chat_template()`

yatarkan:
## Description
This PR:
- Extends chat history message from `map<string, string>` to `ov::AnyMap` to support nested messages, while previous message types should still work. 
- Extends `apply_chat_template()` method with tools (`vector<ov::AnyMap>`) and extra context (`ov::AnyMap`) arguments to provide them to chat template.
- Adds python tests for nested chat history messages and additional chat template params.

This is first part of better chat history management story within agentic scenario.
The following PRs will include:
- Introducing JSON wrapper for chat history messages, tools and extra context instead of `AnyMap`.
- Extending and exposing chat history, providing chat history to `generate()` method.
- Aligning API between `generate()` and `apply_chat_template()` methods through config.
- Updating Python and JS bindings to support new API.

Ticket: CVS-170884, CVS-170931

## Checklist:
- [x] This patch fully addresses the ticket
- [x] I have made corresponding changes to the documentation - Not applicable for this PR


tests/python_tests/test_tokenizer.py
```diff
@@ -152,6 +152,82 @@ def test_apply_chat_template(model_tmp_path, chat_config: tuple[str, dict], ov_h
...
+    add_generation_prompt = True
+
+    extra_context = { "enable_thinking": False }
+    
+    hf_full_history_str = hf_tokenizer.apply_chat_template(
+        conversation, add_generation_prompt=add_generation_prompt, tokenize=False, tools=tools, **extra_context
+    )
+
+    ov_full_history_str = genai_tokenzier.apply_chat_template(
+        conversation, add_generation_prompt=add_generation_prompt, tools=tools, extra_context=extra_context
```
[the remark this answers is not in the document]
yatarkan: Defined outside the test function

tests/python_tests/test_tokenizer.py
```diff
@@ -152,6 +152,82 @@ def test_apply_chat_template(model_tmp_path, chat_config: tuple[str, dict], ov_h
...
+    # In GenAI order of dict keys is not preserved (sorted alphabetically, due to conversion to AnyMap)
+    tools = [json.loads(json.dumps(tool, sort_keys=True)) for tool in tools]
+
+    add_generation_prompt = True
+
+    extra_context = { "enable_thinking": False }
+    
+    hf_full_history_str = hf_tokenizer.apply_chat_template(
+        conversation, add_generation_prompt=add_generation_prompt, tokenize=False, tools=tools, **extra_context
+    )
```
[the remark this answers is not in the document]
yatarkan: Defined outside the test function

src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -9,7 +9,15 @@ export type Options = {
 };
 
 interface Tokenizer {
-  /** Embeds input prompts with special tags for a chat scenario. */
+  /** Applies a chat template to format chat history into a prompt string. */
+  // TODO Update bindings for new ChatHistory type, tools and extra context
```
as-suvorov, line 13: @Retribution98 FYI

tests/python_tests/test_tokenizer.py
```diff
@@ -152,6 +151,82 @@ def test_apply_chat_template(model_tmp_path, chat_config: tuple[str, dict], ov_h
...
+            "parameters": {
+                "type": "object",
+                "properties": {
+                    "arg": { "type": "string" }
+                },
+                "required": ["arg"]
+            }
+        }
+    }]
+    # In GenAI order of dict keys is not preserved (sorted alphabetically, due to conversion to AnyMap)
```
apaniukov, line 201: That might be a problem when we compare the full pipeline results. Maybe we can pass a string representation to preserve the original keys' order?
I would argue that we need to add a second version of this test (via parametrize?) - one that does not sort the keys and mark it as failing for now.

tests/python_tests/test_tokenizer.py
```diff
@@ -168,8 +243,8 @@ def test_non_string_chat_template(hf_ov_genai_models):
     ov_full_history_str = genai_tokenzier.apply_chat_template(conversation, add_generation_prompt=False)
 
     if ov_full_history_str != hf_full_history_str:
-        print(f"hf reference: {hf_full_history_str}")
-        print(f"ov_genai out: {ov_full_history_str}")
+        print(f"HF reference:\n", hf_full_history_str)
+        print(f"GenAI output:\n", ov_full_history_str)
     assert ov_full_history_str == hf_full_history_str
```
apaniukov:
```suggestion
assert ov_full_history_str == hf_full_history_str, f"HF reference:\n{hf_full_history_str}\nGenAI output:\n{ov_full_history_str}"
```

src/js/lib/pipelines/llmPipeline.ts
yatarkan: @Retribution98 JS bindings should be updated later when API and all data structures/classes are finalized in the next PRs.

apaniukov (approved):

tests/python_tests/test_tokenizer.py
yatarkan: Fixed

tests/python_tests/test_tokenizer.py
yatarkan: Discussed, will remove key sorting after migrating to Json Wrapper

src/cpp/src/tokenizer/tokenizer_impl.cpp
```diff
@@ -726,34 +726,49 @@ std::vector<std::string> Tokenizer::TokenizerImpl::decode(const std::vector<std:
...
+    if (!tools_json.empty()) {
+        minja_inputs.tools = tools_json;
+    }
     minja_inputs.add_generation_prompt = add_generation_prompt;
     minja_inputs.extra_context = nlohmann::ordered_json::object();
     minja_inputs.extra_context["bos_token"] = m_bos_token;
     minja_inputs.extra_context["eos_token"] = m_eos_token;
     minja_inputs.extra_context["pad_token"] = m_pad_token;
 
+    if (!extra_context.empty()) {
```
Wovchena, line 767: Should we distinguish explicit empty and completely missing value?
Same for `apply_chat_template()` API

yatarkan: As for now I suppose we should not distinguish them. No matter if it is a missing value or explicitly empty map - extra variables are not provided to chat template

Wovchena: Are these if `(!*.empty())` needed in that case? Removing them would convey the message, that we don't care distinguishing the cases

PR #2771, opened 2025-09-25
[llm_bench] Add possibility to run tool with models in gguf format

sbalandi:
## Description
To run with GenAI:
`python tools/llm_bench/benchmark.py -m ./models/SmolLM2-135M-GGUF/SmolLM2-135M.F16.gguf -n 2 -d CPU`

To run with transformers:
`python tools/llm_bench/benchmark.py -m ./models/SmolLM2-135M-GGUF/SmolLM2-135M.F16.gguf -n 2 -d CPU -f pt`

Ticket: [CVS-171473](https://jira.devtools.intel.com/browse/CVS-171473)

## Checklist:
- [+] Tests have been updated or added to cover the new code 
- [+] This patch fully addresses the ticket. 
- [-] I have made corresponding changes to the documentation


tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -296,7 +294,7 @@ def get_time_list(self):
             return self.token_generation_time
     streamer = TokenStreamer(llm_pipe.get_tokenizer()) if use_streamer_metrics else None
 
-    return llm_pipe, tokenizer, end - start, streamer, True
+    return llm_pipe, None, end - start, streamer, True
```
as-suvorov, line 289: If tokenizer is always `None`, can we remove it from the return tuple?

sbalandi: tokenizer is not None for optimum/pt cases, for example here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/llm_bench_utils/ov_utils.py#L171 and we have the same signature for all cases

as-suvorov (approved):

PR #2798, opened 2025-10-06
feat: Add GGUF file support for GenAI and HF text pipelines

sswierze:
- Handle gguf_file parameter in GenAI pipeline with path resolution
- Pass gguf_file to HF AutoModelForCausalLM via model_kwargs
- Skip quantization logic when GGUF file is specified

## Description

Ticket: 

Fixes #(issue)

## Checklist:


tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -203,11 +203,20 @@ def parse_args():
...
+        help="Path to GGUF model file for tokenizer loading. "
+        "If the base/target model is a local path, gguf-file should be just the filename (e.g., 'model.gguf'). "
+        "If the base/target model is a HuggingFace model ID, gguf-file should be an absolute path.",
+    )
 
     return parser.parse_args()
 
 
 def check_args(args):
+    logger.info('TEST'*100)
```
sbalandi: remove debug print

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -203,11 +203,20 @@ def parse_args():
         action="store_true",
         help="Inference with empty adapters. Applicable for GenAI only.",
     )
+    parser.add_argument(
+        "--gguf-file",
+        type=str,
+        default=None,
+        help="Path to GGUF model file for tokenizer loading. "
+        "If the base/target model is a local path, gguf-file should be just the filename (e.g., 'model.gguf'). "
+        "If the base/target model is a HuggingFace model ID, gguf-file should be an absolute path.",
```
sbalandi: why absolute path for hf model ?

tools/who_what_benchmark/whowhatbench/model_loaders.py
sbalandi, line 1: also need to add changes for optimum case https://github.com/openvinotoolkit/openvino.genai/blob/aefb60e888e428b3a42c55e52e10f8a459c57859/tools/who_what_benchmark/whowhatbench/model_loaders.py#L156

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: changed to relative

tools/who_what_benchmark/whowhatbench/model_loaders.py
sswierze: N/A in this case atm

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: done

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -65,15 +65,20 @@ def load_text_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwargs):
             "Failed to import openvino_genai package. Please install it.")
         exit(-1)
 
+    pipeline_path = model_dir
+    if kwargs.get('gguf_file'):
+        import os
```
as-suvorov: Any specific reason for import not in the beginning of the file? If not, please move to the beginning

sswierze: Done, 
I used to do this in case when part of imported code is used only sometimes. For os it is not a case..

sbalandi:
build_jenkins

PR #2801, opened 2025-10-07
[WWB] Fix to support nanoLLaVA.

popovaan:
## Description
Fixed WWB to support nanoLLaVA.

Ticket: CVS-163185

## Checklist:

### DIFF
--- a/tools/who_what_benchmark/tests/test_cli_vlm.py
+++ b/tools/who_what_benchmark/tests/test_cli_vlm.py
@@ -2,7 +2,7 @@ import subprocess  # nosec B404
 import pytest
 import logging
 import sys
-from test_cli_image import run_wwb
+from test_cli_image import run_wwb, get_similarity
 
 
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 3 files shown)


.github/workflows/linux.yml
```diff
@@ -535,7 +535,7 @@ jobs:
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).RAG.test }}
             timeout: 30
           - name: 'WWB tests'
-            cmd: '/tools/who_what_benchmark/tests'
+            cmd: '/tools/who_what_benchmark/tests -m "not nanollava"'
```
sbalandi, line 538: nitpick: -m "common scope"/"general tests"

sbalandi (approved):

.github/workflows/linux.yml
popovaan: Okay, done.

tools/who_what_benchmark/tests/test_cli_image.py
```diff
@@ -55,6 +55,7 @@ def get_similarity(output: str) -> float:
     return float(matches[-1])
 
 
+@pytest.mark.common_scope
```
as-suvorov: I'm a bit confused with `common_scope` mark. How we define what is common what's not?

popovaan: Common scope includes models that work with default version of transformers. nanollava is in a separate scope, because it works only with the older version.

sbalandi: it looks like my mistake, I have thought 'not nanollava' it's mark, not condition and ask to add something more general

popovaan: Maybe rename it to `@pytest.mark.default_transformers` to make it less confusing?

as-suvorov: > Maybe rename it to `@pytest.mark.default_transformers` to make it less confusing?

@sbalandi , @popovaan don't you mind to revert this `common_scope` mark?
`not nanollava` seems more simple, doesn't require to add `@pytest.mark.default_transformers` for each new test.
What do you think?

sbalandi: I'm ok with `not nanollava` as condition, @popovaan sorry for inconvenience

sbalandi: the only thing is that if new tests with dependencies on some versions will appear, then the list of not ... will grow

sbalandi (approved):

PR #2804, opened 2025-10-08
WIP-1: Create asr pipeline and paraformerASR constructor.

10vliu13:
## Description
This PR intends to integrate ASR pipeline which include paraformer ASR.
WIP-1: Create ASRPipeline and paraformerASR constructor.

### DIFF
--- a/src/cpp/include/openvino/genai/speech_recognition/asr_pipeline.hpp
+++ b/src/cpp/include/openvino/genai/speech_recognition/asr_pipeline.hpp
@@ -1,0 +1,31 @@
+// Copyright (C) 2023-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "utils.hpp"
+
+namespace ov {
+namespace genai {
+
+// forward declaration
+class ASRPipelineImplBase;
+
+class OPENVINO_GENAI_EXPORTS ASRPipeline {
+public:
+    explicit ASRPipeline(const std::filesystem::path& model_dir,
+                         const std::string& device,
+                         const ov::AnyMap& properties = {});
+
+    static ASRPipeline paraformerASR(const std::filesystem::path& model_dir,
+                                     const std::string& device,
+                                     const ov::AnyMap& properties = {});
+
+private:
+    std::shared_ptr<ASRPipelineImplBase> m_impl;
+
+    explicit ASRPipeline(const std::shared_ptr<ASRPipelineImplBase>& impl);
+};
+
+} // namespace genai
+} // namespace ov
--- a/src/cpp/src/speech_recognition/paraformer_impl.cpp
+++ b/src/cpp/src/speech_recognition/paraformer_impl.cpp
@@ -1,0 +1,21 @@
+// Copyright (C) 2023-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "paraformer_impl.hpp"
+
+namespace ov {
+namespace genai {
+
+ParaformerImpl::ParaformerImpl(const std::filesystem::path& model_dir,
+                               const std::string& device,
+                               const ov::AnyMap& properties)
+: ASRPipelineImplBase(model_dir, device, properties)
+{
+
+}
+
+ParaformerImpl::~ParaformerImpl()
+{}
+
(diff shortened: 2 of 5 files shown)


rkazants (changes requested):
Please create/provide JIRA ticket to access if we need arch review for these changes.
In Jira ticket, we expect description of reason of these changes.

src/cpp/include/openvino/genai/speech_recognition/asr_pipeline.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#pragma once
+
+#include "utils.hpp"
+
+namespace ov {
+namespace genai {
+
+// forward declaration
+class ASRPipelineImplBase;
```
rkazants, line 12: what is about Whisper pipeline? Will it work with ASRPipeline? I expect -yes. Please add tests for this.

src/cpp/src/speech_recognition/paraformer_impl.cpp
```diff
@@ -0,0 +1,21 @@
+// Copyright (C) 2023-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "paraformer_impl.hpp"
+
+namespace ov {
+namespace genai {
+
+ParaformerImpl::ParaformerImpl(const std::filesystem::path& model_dir,
```
rkazants, line 9: how do you get IRs for paraformer models? Do exporting and inference work in optimum-intel? If no, then we must add it for the full consistent user experience.

src/cpp/include/openvino/genai/speech_recognition/asr_pipeline.hpp
as-suvorov: Agree, `ASRPipeline` should work with Whisper model. Once we add `ASRPipeline` I expect whisper pipeline to be deprecated.

10vliu13:
> Please create/provide JIRA ticket to access if we need arch review for these changes. In Jira ticket, we expect description of reason of these changes.

Do you have any existing JIRA link i can refers to? I can create new one based on this. Thanks

src/cpp/include/openvino/genai/speech_recognition/asr_pipeline.hpp
10vliu13: Yes, I think Whisper pipeline will work with ASRPipeline. Let me also work on this to integrate with Whisper. @rkazants also mentioned tests for whisper. could you also show me the current tests for Whisper pipeline? I can refer to the current implementation. Thanks

src/cpp/src/speech_recognition/paraformer_impl.cpp
10vliu13: The peraformer IRs upstream to huggingface. Here is the link.
https://huggingface.co/apinge/paraformer-zh-int8-ov

rkazants: No, that is not how it should work. All supported models must pass optimum-cli tool conversion.
This is our deployment pipeline. If not optimum-intel support, model is not enabled and no claim in our documentation.
Create jira ticket for the corresponding task.

rkazants:
> > Please create/provide JIRA ticket to access if we need arch review for these changes. In Jira ticket, we expect description of reason of these changes.
> 
> Do you have any existing JIRA link i can refers to? I can create new one based on this. Thanks

Please consult with @EfimovIlia, @moslex about this JIRA tickets creation.

src/cpp/include/openvino/genai/speech_recognition/asr_pipeline.hpp
as-suvorov: whisper tests are here:
https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_whisper_pipeline.py
https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_whisper_pipeline_static.py

src/cpp/src/speech_recognition/paraformer_impl.cpp
Wovchena: Just to make it explicit. The ticket was created: CVS-174743

PR #2817, opened 2025-10-10
[StructuredOutput] Update XGrammar

apaniukov:
## Description
Update XGrammar to 0.1.25 version.
- New version reduces grammar compile time
- Support new structural tags that work similarly to compound grammar
- Deprecate compound grammar and old structural tags
- Update samples with new structural tags API
- Fix python benchmark sample - the actual number of tokens was different to the reported number of tokens
- Update python formatting with ruff.

OVMS StructuralTags compilation benchmark results:
0.1.18:   2613.6 ms
0.1.19:  10242.3 ms
0.1.20:  15852.5 ms
0.1.21:  17536.1 ms
0.1.22: 205489.0 ms
0.1.23:    703.5 ms
0.1.24:    692.8 ms 
0.1.25      86.4 ms

Ticket: CVS-173972

## Checklist:

### DIFF
--- a/samples/python/text_generation/benchmark_genai.py
+++ b/samples/python/text_generation/benchmark_genai.py
@@ -1,8 +1,9 @@
 # Copyright (C) 2023-2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
-import sys
 import argparse
+import sys
+
 import openvino_genai as ov_genai
 from openvino import get_version
 
@@ -20,19 +21,19 @@ def main():
     args = parser.parse_args()
 
     if args.prompt is not None and args.prompt_file is not None:
-        raise RuntimeError(f'Prompt and prompt file should not exist together!')
+        raise RuntimeError("Prompt and prompt file should not exist together!")
     else:
         if args.prompt_file is not None:
-            with open(args.prompt_file, 'r', encoding='utf-8') as f:
+            with open(args.prompt_file, "r", encoding="utf-8") as f:
                 prompt = [f.read()]
         else:
-            prompt = ['The Sky is blue because'] if args.prompt is None else [args.prompt]
+            prompt = ["The Sky is blue because"] if args.prompt is None else [args.prompt]
     if len(prompt) == 0:
-        raise RuntimeError(f'Prompt is empty!')
+        raise RuntimeError("Prompt is empty!")
 
-    print(f'openvino runtime version: {get_version()}, genai version: {ov_genai.__version__}')
+    print(f"openvino runtime version: {get_version()}, genai version: {ov_genai.__version__}")
 
-    # Perf metrics is stored in DecodedResults. 
+    # Perf metrics is stored in DecodedResults.
     # In order to get DecodedResults instead of a string input should be a list.
     models_path = args.model
     device = args.device
@@ -63,15 +64,22 @@ def main():
     for _ in range(num_iter - 1):
         res = pipe.generate(prompt, config)
         perf_metrics += res.perf_metrics
-    
+
     print(f"Output token size: {res.perf_metrics.get_num_generated_tokens()}")
     print(f"Load time: {perf_metrics.get_load_time():.2f} ms")
-    print(f"Generate time: {perf_metrics.get_generate_duration().mean:.2f} ± {perf_metrics.get_generate_duration().std:.2f} ms")
-    print(f"Tokenization time: {perf_metrics.get_tokenization_duration().mean:.2f} ± {perf_metrics.get_tokenization_duration().std:.2f} ms")
-    print(f"Detokenization time: {perf_metrics.get_detokenization_duration().mean:.2f} ± {perf_metrics.get_detokenization_duration().std:.2f} ms")
+    print(
+        f"Generate time: {perf_metrics.get_generate_duration().mean:.2f} ± {perf_metrics.get_generate_duration().std:.2f} ms"
+    )
+    print(
+        f"Tokenization time: {perf_metrics.get_tokenization_duration().mean:.2f} ± {perf_metrics.get_tokenization_duration().std:.2f} ms"
+    )
+    print(
+        f"Detokenization time: {perf_metrics.get_detokenization_duration().mean:.2f} ± {perf_metrics.get_detokenization_duration().std:.2f} ms"
+    )
     print(f"TTFT: {perf_metrics.get_ttft().mean:.2f} ± {perf_metrics.get_ttft().std:.2f} ms")
     print(f"TPOT: {perf_metrics.get_tpot().mean:.2f} ± {perf_metrics.get_tpot().std:.2f} ms")
     print(f"Throughput : {perf_metrics.get_throughput().mean:.2f} ± {perf_metrics.get_throughput().std:.2f} tokens/s")
+
 
 if __name__ == "__main__":
     main()
--- a/samples/python/text_generation/benchmark_genai.py
+++ b/samples/python/text_generation/benchmark_genai.py
@@ -6,6 +6,7 @@ import argparse
 import openvino_genai as ov_genai
 from openvino import get_version
 
+
 def main():
     parser = argparse.ArgumentParser(description="Help command")
     parser.add_argument("-m", "--model", type=str, required=True, help="Path to model and tokenizers base directory")
@@ -15,7 +16,7 @@ def main():
     parser.add_argument("-n", "--num_iter", type=int, default=2, help="Number of iterations")
     parser.add_argument("-mt", "--max_new_tokens", type=int, default=20, help="Maximal number of new tokens")
     parser.add_argument("-d", "--device", type=str, default="CPU", help="Device")
-    
+
     args = parser.parse_args()
 
     if args.prompt is not None and args.prompt_file is not None:
@@ -37,9 +38,10 @@ def main():
     device = args.device
     num_warmup = args.num_warmup
     num_iter = args.num_iter
-    
+
     config = ov_genai.GenerationConfig()
     config.max_new_tokens = args.max_new_tokens
+    config.apply_chat_template = False
 
     if device == "NPU":
         pipe = ov_genai.LLMPipeline(models_path, device)
@@ -55,7 +57,7 @@ def main():
 
     for _ in range(num_warmup):
         pipe.generate(prompt, config)
-    
+
     res = pipe.generate(prompt, config)
     perf_metrics = res.perf_metrics
     for _ in range(num_iter - 1):
--- a/src/cpp/include/openvino/genai/generation_config.hpp
+++ b/src/cpp/include/openvino/genai/generation_config.hpp
@@ -403,8 +403,7 @@ public:
      * @brief Tag defines a begin/end wrapper with constrained inner content.
      *
      * The generator will output `begin`, then the `content` (a StructuralTag),
-     * and finally `end`. Useful for embedding structured fragments inside free
-     * text using explicit markers.
+     * and finally `end`.
      *
      * Example: Tag("<think>", AnyText(), "</think>") represents thinking portion of the model output.
      */
--- a/src/cpp/include/openvino/genai/generation_config.hpp
+++ b/src/cpp/include/openvino/genai/generation_config.hpp
@@ -261,7 +261,7 @@ public:
                              std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::TagsWithSeparator>>) {
             return g ? g->to_string() : std::string("null");
         } else {
-            return std::string("unsupported_structural_tag");
+            OPENVINO_THROW("Unsupported structural tag, cannot convert to string:" + std::string(typeid(g).name()));
         }
     }
 
@@ -283,7 +283,7 @@ public:
                              std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::TagsWithSeparator>>) {
             return g ? g->to_json() : std::string("null");
         } else {
-            return std::string("unsupported_structural_tag");
+            OPENVINO_THROW("Unsupported structural tag, cannot convert to json:" + std::string(typeid(g).name()));
         }
     }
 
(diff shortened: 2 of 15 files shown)


samples/python/text_generation/benchmark_genai.py
```diff
@@ -15,31 +17,32 @@ def main():
...
     # In order to get DecodedResults instead of a string input should be a list.
     models_path = args.model
     device = args.device
     num_warmup = args.num_warmup
     num_iter = args.num_iter
-    
+
     config = ov_genai.GenerationConfig()
     config.max_new_tokens = args.max_new_tokens
+    config.apply_chat_template = False
```
Wovchena, line 45:
```suggestion
```
Why?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -91,7 +101,32 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
...
+                case '\\': stream << "\\\\"; break;
+                case '\b': stream << "\\b"; break;
+                case '\f': stream << "\\f"; break;
+                case '\n': stream << "\\n"; break;
+                case '\r': stream << "\\r"; break;
+                case '\t': stream << "\\t"; break;
+                default:
+                    if (static_cast<unsigned char>(character) < 0x20) {
+                        stream << "\\u" << std::hex << std::uppercase << std::setfill('0') << std::setw(4)
+                                << static_cast<int>(static_cast<unsigned char>(character)) << std::dec << std::nouppercase;
```
Wovchena: What is `unsigned char` cast for?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -126,88 +167,323 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
...
+        }
+        std::string to_json() const {
+            return std::string("{\"type\": \"const_string\", \"value\": ") + format_for_json(value) + "}";
+        }
+        bool operator==(const ConstString& other) const {
+            return value == other.value;
+        }
+    };
+
+    struct AnyText {
```
Wovchena, line 216: Is it possible to include `AnyText` into a sample? It seems like it's equivalent to not applying any schema, but why `AnyText` is needed in that case

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -126,88 +167,323 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
...
+                             std::is_same_v<T, ov::genai::StructuredOutputConfig::ConstString> ||
+                             std::is_same_v<T, ov::genai::StructuredOutputConfig::AnyText> ||
+                             std::is_same_v<T, ov::genai::StructuredOutputConfig::QwenXMLParametersFormat>) {
+            return g.to_json();
+        } else if constexpr (std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::Concat>> ||
+                             std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::Union>> ||
+                             std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::Tag>> ||
+                             std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::TriggeredTags>> ||
+                             std::is_same_v<T, std::shared_ptr<ov::genai::StructuredOutputConfig::TagsWithSeparator>>) {
+            return g ? g->to_json() : std::string("null");
```
Wovchena, line 311: Can json really be `null`, should it be `{}`?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -126,88 +167,323 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
...
+        }
+        std::string to_json() const {
+            return "{\"type\": \"any_text\"}";
+        }
+        bool operator==(const AnyText& other) const {
+            return true;
+        }
+    };
+
+    struct QwenXMLParametersFormat {
```
Wovchena, line 234: Seem to be a niche thing. Can you add a description a a link to learn more to the docstrings

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -126,88 +167,323 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
...
+        AnyText,
+        QwenXMLParametersFormat,
         std::shared_ptr<Concat>,
-        std::shared_ptr<Union>
+        std::shared_ptr<Union>,
+        std::shared_ptr<Tag>,
+        std::shared_ptr<TriggeredTags>,
+        std::shared_ptr<TagsWithSeparator>
     >;
+    using CompoundGrammar = StructuralTag;
```
Wovchena, line 271: Is there a way to deprecate `CompoundGrammar`?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -126,88 +167,323 @@ class OPENVINO_GENAI_EXPORTS StructuredOutputConfig {
...
+                    oss << ", ";
+                }
+            }
+            oss << "], at_least_one=" << (at_least_one ? "True" : "False") <<
+                   ", stop_after_first=" << (stop_after_first ? "True" : "False") << ")";
+            return oss.str();
+        };
+    };
+
+    struct TagsWithSeparator {
```
Wovchena, line 504: Would be nice to have a text describing big picture and relations between the structs. Sure it's possible to infer that info, but it requires more effort

tests/python_tests/test_structured_output.py
```diff
@@ -158,3 +171,92 @@ def test_structural_tags(ov_pipe, prompt_and_structural_tag):
     match = re.search(rf"{structural_tag.begin}(.*?){structural_tag.end}", res_str)
     assert match, f"Output `{res_str}` does not contain structural tag {structural_tag.begin}...{structural_tag.end}"
     RESTAPIResponse.model_validate_json(match.group(1))
+
+
+@pytest.mark.precommit
+@pytest.mark.parametrize(
+    "ov_pipe", [model_id for model_id in structured_id_models if "random" not in model_id], indirect=True
```
Wovchena: That's a tricky way to test just `TinyLlama/TinyLlama-1.1B-Chat-v1.0`. Make it explicit.

Why not `katuni4ka/tiny-random-phi3`?

tests/python_tests/test_structured_output.py
```diff
@@ -158,3 +171,92 @@ def test_structural_tags(ov_pipe, prompt_and_structural_tag):
...
+            """{
+        "type": "structural_tag",
+        "format": {
+            "type": "const_string",
+            "value": "abc"
+        }
+    }""",
+            lambda x: x == "abc",
+        ),
+        ("regex", SOC.Regex("a*"), lambda x: re.match(r"^a*$", x) is not None),
```
Wovchena: Seems like `regex` is the value for `prompt` arg. Looks more like a test id to me. Am I misreading it?

apaniukov: Yes, I am putting the "test id" as a prompt in cases where a particular prompt is not needed.

samples/python/text_generation/benchmark_genai.py
apaniukov: Benchmark reports the number of input tokens to a user like this:
```python
input_data = pipe.get_tokenizer().encode(prompt)
prompt_token_size = input_data.input_ids.get_shape()[1]
```
Since the default is to apply the chat template implicitly, the real number of input tokens might (and usually is) different.

```bash
python benchmark_genai.py -m path/to/llama-3.2-3b-instruct-ov -n 5 -p "hi"
openvino runtime version: 2025.4.0-20060-f8950915195, genai version: 2025.4.0.0-2551-b978edd8873-update-xgrammar
...
Prompt token size: 2  # from benchmark output
Prompt token size from generate: 36  # from res.perf_metrics.get_num_input_tokens()
...
```

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: `AnyText` is useful as part of the bigger schema. For example, we can enforce thinking like this:
`Tag(begin="<think>", content=AnyText(), end="</think>") + AnyText()`

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: The `compound_grammar` field in `StructuredOutputConfig` is deprecated and will print a warning.

tests/python_tests/test_structured_output.py
Wovchena: Use `pytest.param()`. It allows specifying `id` inplace. If you don't care about a prompt for some of the params, use `""`.
If you replace `Generate a date with EBNF` with `""` will the schema still enforce the date?

Wovchena: If so `prompt` could be removed

tests/python_tests/test_structured_output.py
apaniukov: The `TriggeredTags` is not suitable for a random model - I expect that the model repeats the word `"function"`, so the structured output is triggered. The filtering allows us to have one source of model ids.
I can extract the `TriggeredTags` test, so we can use random models here too.

Wovchena: Ok. Can it be `structured_id_models[0]`?

apaniukov: `if "random" not in model_id` conveys the intent, while `structured_id_models[0]` obscures it. There is also an implicit assumption that the first model is not random, which may lead to unexpected test failures if `structured_id_models` is changed in the future.

src/cpp/include/openvino/genai/generation_config.hpp
Wovchena: Will you extend one of the samples? Seems to be harmless

tests/python_tests/test_structured_output.py
Wovchena: `if "random" not in model_id` may trigger multiple models if `structured_id_models` is extended. The putrpose of these tests is to verify the tags and not models. I guess there's no way to resolve that?

tests/python_tests/test_structured_output.py
```diff
@@ -158,3 +171,111 @@ def test_structural_tags(ov_pipe, prompt_and_structural_tag):
...
+                    SOC.Tag(begin="function", content=SOC.ConstString("B"), end="</function>"),
+                ],
+                at_least_one=True,
+                stop_after_first=True,
+            ),
+            lambda x: re.match(r"(function(A|B)</function>)", x) is not None,
+            id="TriggeredTags",
+        ),
+        pytest.param(
+            "TagsWithSeparator",
```
Wovchena: This and `QwenXMLParametersFormat` remained as prompts. Why?

I guess "prompt" or "dummy prompt" would be more readable than "". I didn't come up with that earlier. I leave it up to you

apaniukov: Fixed.

tests/python_tests/test_structured_output.py
apaniukov: Done

tests/python_tests/test_structured_output.py
apaniukov: Well...
```python
@pytest.mark.parametrize(
    "ov_pipe", 
    [
        next(
            (model_id for model_id in structured_id_models if "random" not in model_id), 
            "TinyLlama/TinyLlama-1.1B-Chat-v1.0"
        )
    ], 
    indirect=True,
)
```
Is it a good solution? I don't know.

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: `null` is a valid JSON object.

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: After I add the documentation.

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: To avoid errors when `char` is `signed char`.
Added comments to the code.

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: In the future, yes. Not in this PR.

src/cpp/include/openvino/genai/generation_config.hpp
apaniukov: I will address it in the documentation.

src/python/py_generation_config.cpp
```diff
@@ -47,6 +47,75 @@ auto structured_output_config_docstring = R"(
         The compound grammar a Union or Concat of several grammars, where each grammar can be a JSON schema, regex, EBNF, Union or Concat.
 )";
 
+// Docstrings for structural tag building blocks
```
apaniukov:
```suggestion
```

tests/python_tests/test_structured_output.py
Wovchena: That's the point) There seem to be no readable solution.

I'd propose the easiest to implement: `TinyLlama/TinyLlama-1.1B-Chat-v1.0` with a comment describing the intent. "One source of model ids" seem to be the least needed among the named benefits

tests/python_tests/test_structured_output.py
```diff
@@ -158,3 +171,111 @@ def test_structural_tags(ov_pipe, prompt_and_structural_tag):
     match = re.search(rf"{structural_tag.begin}(.*?){structural_tag.end}", res_str)
     assert match, f"Output `{res_str}` does not contain structural tag {structural_tag.begin}...{structural_tag.end}"
     RESTAPIResponse.model_validate_json(match.group(1))
+
+
+@pytest.mark.precommit
+@pytest.mark.parametrize(
+    "ov_pipe", [model_id for model_id in structured_id_models if "random" not in model_id], indirect=True
+)
```
apaniukov:
```suggestion
# use only non-random model for stable output in TriggeredTags test
@pytest.mark.parametrize("ov_pipe", ["TinyLlama/TinyLlama-1.1B-Chat-v1.0"], indirect=True)
```

Wovchena (approved):
Should we wait for @dtrawins?

apaniukov:
> Should we wait for @dtrawins?

The PR is backward compatible, so no need for an additional review.

PR #2825, opened 2025-10-13
Fix missing result_dir parameter in evaluator constructors

sswierze:
- Fix _generate_data call in RerankingEvaluator to include result_dir parameter
- Fix _generate_data call in EmbeddingsEvaluator to include result_dir parameter
- Both evaluators now properly pass reference directory path when generating ground truth data
- Ensures consistent behavior between constructor and score method data generation

## Description

Ticket: 

Fixes #(issue)

## Checklist:


Wovchena:
build_jenkins

PR #2834, opened 2025-10-14
Coverity issues fix

sgonorov:
## Description
Fixes all possible high and very high level Coverity issues
Ticket: 
[CVS-174757](https://jira.devtools.intel.com/browse/CVS-174757)

### DIFF
--- a/src/cpp/src/utils.hpp
+++ b/src/cpp/src/utils.hpp
@@ -26,9 +26,9 @@ extern const std::string SDPA_BACKEND;
 
 struct ModelDesc {
     std::string device;
-    ov::genai::SchedulerConfig scheduler_config{};
+    ov::genai::SchedulerConfig scheduler_config;
     ov::AnyMap properties;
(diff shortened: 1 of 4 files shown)


src/cpp/src/utils.hpp
```diff
@@ -26,9 +26,9 @@ extern const std::string SDPA_BACKEND;
 
 struct ModelDesc {
     std::string device;
-    ov::genai::SchedulerConfig scheduler_config;
+    ov::genai::SchedulerConfig scheduler_config{};
     ov::AnyMap properties;
-    ov::genai::GenerationConfig generation_config;
+    ov::genai::GenerationConfig generation_config{};
```
Wovchena: That doesn't look right. it's just a member declaration, there's no any instance to create with {}

sgonorov: It works but redundant. I've removed these cases 👍

Wovchena (approved):

PR #2851, opened 2025-10-16
Encourage requirements txt usage

Wovchena:
So we do not have to update these places when bumping the versions

### DIFF
--- a/site/docs/guides/model-preparation/convert-to-openvino.mdx
+++ b/site/docs/guides/model-preparation/convert-to-openvino.mdx
@@ -19,11 +19,12 @@ For downloading pre-converted models, see [Download Pre-Converted OpenVINO Model
 
     :::info Note
     Some models may require additional dependencies.
-    To convert models with `optimum-cli` and to run the examples, install the dependencies from [`./samples/requirements.txt`](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/requirements.txt):
-    ```bash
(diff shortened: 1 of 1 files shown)


site/docs/guides/model-preparation/convert-to-openvino.mdx
```diff
@@ -16,9 +16,6 @@ For downloading pre-converted models, see [Download Pre-Converted OpenVINO Model
 ## Converting Models from Hugging Face
 
 1. Install `optimum-intel` package to download, convert and optimize models:
-    ```bash
-    pip install optimum-intel@git+https://github.com/huggingface/optimum-intel.git
-    ```
```
yatarkan, line 21: In note below we are providing install command with `./samples/requirements.txt`, but for modelscope path - `./samples/export-requirements.txt`. I think we should be aligned here.

I would suggest here to have a note first (without install command, just explaining why we are using requirement file, not the pypi package) and install command below the note:
```
    :::info Note
    Some models may require additional dependencies.
    To convert models with `optimum-cli` and to run the examples, install the dependencies from [`./samples/export-requirements.txt`](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/export-requirements.txt):
    :::

    ```bash
    pip install --requirement ./samples/export-requirements.txt
    ```
```

Wovchena: Updated

PR #2854, opened 2025-10-16
Increase wwb timeout

Wovchena:
Failed in https://github.com/openvinotoolkit/openvino.genai/actions/runs/18562758875/job/52920690867?pr=2261


Wovchena:
@as-suvorov, reminder

PR #2876, opened 2025-10-20
[JS] Implement StructuredOutputConfig

Retribution98:
## Description
- StructuredOutputConfig was implemented for JS API
- add parameter `system message` to startChat
- added 3 samples and tests

Ticket: 175108

## Checklist:

### DIFF
--- a/src/js/tests/structuredOutput.test.js
+++ b/src/js/tests/structuredOutput.test.js
@@ -231,8 +231,8 @@ table::= "users" | "orders" | "products"`),
   });
 
   it("generate with TriggeredTags in structural_tags_config", async (testContext) => {
-    if (os.platform() === "darwin") {
-      testContext.skip("Skipped for macOS due to inconsistent LLM outputs. CVS-175278");
+    if (os.platform() === "darwin" || os.platform() === "win32") {
+      testContext.skip("Skipped due to inconsistent LLM outputs. CVS-175278");
       return;
     }
     const tools = [


samples/js/text_generation/README.md
```diff
@@ -92,6 +92,94 @@ Recommended models: Qwen/Qwen2.5-3B-Instruct, Qwen/Qwen2.5-7B-Instruct
   node react_sample.js model_dir
   ```
 
+### 6. Structured Output Sample (`structured_output_sample`)
```
almilosz: Please run this sample in https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_structured_output_sample.py

Retribution98: I tested it, and Python and JavaScript returned similar but not identical results. This occurs because they use different libraries and generate different prompts.
I will fix it and add this test in a separate PR. (CVS 175278)

MaximProshin:
@apaniukov , FYI if you also want to review it

src/js/src/helper.cpp
```diff
@@ -79,24 +62,45 @@ ov::Any js_to_cpp<ov::Any>(const Napi::Env& env, const Napi::Value& value) {
...
 }
 
+template <>
+std::string js_to_cpp<std::string>(const Napi::Env& env, const Napi::Value& value) {
+    if (value.IsString()) {
+        return value.As<Napi::String>().Utf8Value();
+    } else {
+        OPENVINO_THROW("Passed argument must be of type String.");
+    }
+}
```
```suggestion
    OPENVINO_ASSERT(value.IsString(), "Passed argument must be of type String.");
    return value.As<Napi::String>().Utf8Value();
}
```

samples/js/text_generation/README.md
```diff
@@ -92,6 +92,94 @@ Recommended models: Qwen/Qwen2.5-3B-Instruct, Qwen/Qwen2.5-7B-Instruct
...
+  - Combine grammars with Concat and Union operations
+  - Multi-turn chat with grammar switching
+  - Structured tool calling using zod schemas
+- **Run Command:**
+  ```bash
+  node compound_grammar_generation.js model_dir
+  ```
+- **Notes:**
+  This sample is ideal for scenarios requiring strict control over LLM outputs, such as building agents that interact with APIs or require validated structured responses. It showcases how to combine regex triggers and JSON schema enforcement for robust output generation.
+  The sample is verified with `microsoft/Phi-4-mini-instruct` model. Other models may not produce the expected results or might require different system prompt.
```
almilosz, line 199:
```suggestion
  The sample is verified with `Qwen2.5-1.5B-Instruct-int4-ov` model. Other models may not produce the expected results or might require different system prompt.
```
I tested with this one

samples/js/text_generation/README.md
```diff
@@ -92,6 +92,94 @@ Recommended models: Qwen/Qwen2.5-3B-Instruct, Qwen/Qwen2.5-7B-Instruct
...
+  </function>
+  ```
+  The sample includes schemas for each tool, and the model is prompted to use them for tool calling. There are two model calls - with and without structural tags. 
+  You can compare the results to see how the model generates structured outputs when using structural tags.
+  If there is no prompt provided, the sample will use the default prompt: `"What is the weather in London today and in Paris yesterday, and how many pounds can I get for 100 euros?"`
+
+- **Main Feature:** Structured tool call generation with LLM using schema enforcement with structural tags.
+- **Run Command:**
+  ```bash
+  node structural_tags_generation.js model_dir [--prompt "Your prompt here"]
```
almilosz:
```suggestion
  node structural_tags_generation.js model_dir [prompt]
```
`--prompt` looks like parsed arg from `yargs` or sth similar

samples/js/text_generation/README.md
almilosz: Initially you can add those sample and just ran them on CI without comparing results 1:1.
In `test_structural_tag_generation.py` the sample is just ran:
https://github.com/openvinotoolkit/openvino.genai/blob/37864216d33cde26d1f2ca6741ce29471247e38a/tests/python_tests/samples/test_structural_tag_generation.py#L17

samples/js/text_generation/README.md
Retribution98: It works for me. I will investigate why it happens randomly.

PR #2879, opened 2025-10-21
Complete GenAI Logger

yangwang201911:
Complete the logging mechanism with a thread-safe singleton Logger that honors OPENVINO_LOG_LEVEL before printing.
Added macros-based functions. (GENAI_DEBUG_LOG, GENAI_INFO_LOG, etc.)

Tickets: CVS-176017

### DIFF
--- a/src/cpp/src/logger.cpp
+++ b/src/cpp/src/logger.cpp
@@ -4,9 +4,11 @@
 #include "logger.hpp"
 
 #include <cctype>
+#include <chrono>
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <ctime>
 #include <iostream>
 #include <memory>
 #include <stdexcept>
@@ -45,16 +47,10 @@ Logger::Logger() {
 }
 
 void Logger::do_log(ov::log::Level level, const char* file, int line, const std::string& msg) {
-    if (!should_log(level)) {
-        return;
-    }
     write_message(level, file, line, msg);
 }
 
 void Logger::log_format(ov::log::Level level, const char* file, int line, const char* format, ...) {
-    if (!should_log(level)) {
-        return;
-    }
     va_list args;
     va_start(args, format);
     try {
@@ -84,7 +80,11 @@ void Logger::write_message(ov::log::Level level, const char* file, int line, con
     format_prefix(out, level, file, line);
     out << msg;
     if (msg.empty() || msg.back() != '\n') {
-        out << '\n';
+        if (level == ov::log::Level::ERR) {
+            out << std::endl;
+        } else {
+            out << '\n';
+        }
     }
 }
 
@@ -107,7 +107,7 @@ std::string Logger::format_from_variadic(const char* format, va_list args) const
     if (required == 0) {
         return {};
     }
-    std::vector<char> buffer(static_cast<size_t>(required) + 1u, '\0');
+    std::vector<char> buffer(static_cast<size_t>(required) + 1u);
     int result = std::vsnprintf(buffer.data(), buffer.size(), format, args);
     if (result < 0) {
         throw std::runtime_error("Logger format error: failed to format message");
@@ -126,7 +126,7 @@ std::string_view Logger::get_filename(std::string_view file_path) const {
 std::ostream& Logger::format_prefix(std::ostream& out, ov::log::Level level, const char* file, int line) const {
     switch (level) {
     case ov::log::Level::DEBUG:
-        out << "[DEBUG][" << get_filename(file) << ":" << line << "] ";
+        out << "[DEBUG] ";
         break;
     case ov::log::Level::INFO:
         out << "[INFO] ";
@@ -141,6 +141,23 @@ std::ostream& Logger::format_prefix(std::ostream& out, ov::log::Level level, con
         out << "[LOG] ";
         break;
     }
+
+    // Add timestamp and file info only for DEBUG level
+    if (level == ov::log::Level::DEBUG) {
+        {
+            static std::mutex m;
+            time_t tt = std::chrono::system_clock::to_time_t(std::chrono::system_clock::now());
+            std::lock_guard<std::mutex> lock(m);
+            auto tm = gmtime(&tt);
+            if (tm) {
+                char buffer[256];
+                strftime(buffer, sizeof(buffer), "%Y-%m-%dT%H:%M:%Sz", tm);
+                out << buffer << " ";
+            }
+        }
+        out << get_filename(file) << ":" << line << "\t";
+    }
+
     return out;
 }
 
--- a/src/cpp/src/logger.hpp
+++ b/src/cpp/src/logger.hpp
@@ -54,6 +54,9 @@ namespace detail {
 
 inline void log_message(ov::log::Level level, const char* file, int line, const std::string& msg) {
     auto& logger = Logger::get_instance();
+    if (!logger.should_log(level)) {
+        return;
+    }
     logger.do_log(level, file, line, msg);
 }
 
@@ -76,9 +79,9 @@ inline void log_message(ov::log::Level level, const char* file, int line, const 
 
 }  // namespace detail
 
-#define OPENVINO_DEBUG(...) ::ov::genai::detail::log_message(ov::log::Level::DEBUG, __FILE__, __LINE__, __VA_ARGS__)
-#define OPENVINO_INFO(...)  ::ov::genai::detail::log_message(ov::log::Level::INFO, __FILE__, __LINE__, __VA_ARGS__)
-#define OPENVINO_WARN(...)  ::ov::genai::detail::log_message(ov::log::Level::WARNING, __FILE__, __LINE__, __VA_ARGS__)
-#define OPENVINO_ERR(...)   ::ov::genai::detail::log_message(ov::log::Level::ERR, __FILE__, __LINE__, __VA_ARGS__)
+#define GENAI_DEBUG(...) ::ov::genai::detail::log_message(ov::log::Level::DEBUG, __FILE__, __LINE__, __VA_ARGS__)
+#define GENAI_INFO(...)  ::ov::genai::detail::log_message(ov::log::Level::INFO, __FILE__, __LINE__, __VA_ARGS__)
+#define GENAI_WARN(...)  ::ov::genai::detail::log_message(ov::log::Level::WARNING, __FILE__, __LINE__, __VA_ARGS__)
+#define GENAI_ERR(...)   ::ov::genai::detail::log_message(ov::log::Level::ERR, __FILE__, __LINE__, __VA_ARGS__)
 
 }  // namespace ov::genai
--- a/src/cpp/src/rag/text_embedding_pipeline.cpp
+++ b/src/cpp/src/rag/text_embedding_pipeline.cpp
@@ -300,7 +300,7 @@ private:
                         << ")."
                         << " Some models may fail with such configuration."
                         << " Remove max_position_embeddings from config.json to silence this warning.";
-                OPENVINO_WARN(message.str());
+                GENAI_WARN(message.str());
             }
 
             if (m_config.pad_to_max_length.has_value() && *m_config.pad_to_max_length) {
--- a/src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
+++ b/src/cpp/src/sampling/structured_output/xgrammar_backend.cpp
@@ -72,9 +72,9 @@ xgrammar::Grammar XGrammarStructuredOutput::create_grammar(const std::optional<S
             using ConfigType = std::decay_t<decltype(config)>;
             if constexpr (std::is_same_v<ConfigType, StructuralTagsConfig>) {
                 // Old format: StructuralTagsConfig
-                OPENVINO_WARN("The use of \"structural_tags_config\" with StructuralTagsConfig instance is "
-                              "deprecated and will be removed in future releases. "
-                              "Use TriggeredTags instead.");
+                GENAI_WARN("The use of \"structural_tags_config\" with StructuralTagsConfig instance is "
+                           "deprecated and will be removed in future releases. "
+                           "Use TriggeredTags instead.");
 
                 std::ostringstream oss;
                 oss << "{\"type\": \"structural_tag\", \"format\": " << config.to_json() << "}";
@@ -99,8 +99,8 @@ xgrammar::Grammar XGrammarStructuredOutput::create_grammar(const std::optional<S
             }
         }, structured_output_config.value().structural_tags_config.value());
     } else if (structured_output_config.value().compound_grammar.has_value()) {
-        OPENVINO_WARN("The use of \"compound_grammar\" is deprecated and will be removed in future releases.\n"
-                      "Pass the same input to \"structural_tags_config\" instead.");
+        GENAI_WARN("The use of \"compound_grammar\" is deprecated and will be removed in future releases.\n"
+                   "Pass the same input to \"structural_tags_config\" instead.");
         return parse_structural_tag(structured_output_config.value().compound_grammar.value());
     }
 
--- a/src/cpp/src/whisper/models/with_past_decoder.cpp
+++ b/src/cpp/src/whisper/models/with_past_decoder.cpp
@@ -82,10 +82,10 @@ namespace ov::genai {
 WhisperWithPastDecoder::WhisperWithPastDecoder(const std::filesystem::path& models_path,
                                                const std::string& device,
                                                const ov::AnyMap& properties) {
-    OPENVINO_WARN("Whisper decoder models with past is deprecated. Support will be removed in 2026.0.0 release.\n"
-                  "To obtain stateful decoder model use latest `optimum-intel` package:\n"
-                  "pip install optimum-intel@git+https://github.com/huggingface/optimum-intel.git@main\n"
-                  "optimum-cli export openvino --trust-remote-code --model openai/whisper-tiny whisper-tiny");
+    GENAI_WARN("Whisper decoder models with past is deprecated. Support will be removed in 2026.0.0 release.\n"
+               "To obtain stateful decoder model use latest `optimum-intel` package:\n"
+               "pip install optimum-intel@git+https://github.com/huggingface/optimum-intel.git@main\n"
+               "optimum-cli export openvino --trust-remote-code --model openai/whisper-tiny whisper-tiny");
 
     ov::Core core = utils::singleton_core();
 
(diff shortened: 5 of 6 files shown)


src/cpp/src/logger.hpp
```diff
@@ -5,13 +5,89 @@
...
+        if (std::string::npos == index) {
+            return filePath;
+        }
+        return filePath.substr(index + 1);
+    }
+    std::string format_prefix(ov::log::Level level, const char* file, int line) const {
+        std::string level_str;
+        switch (level) {
+        case ov::log::Level::DEBUG:
+            level_str = "[DEBUG][" + get_filename(file) + ":" + std::to_string(line) + "] ";
```
as-suvorov:
```suggestion
            level_str = "[DEBUG][" + get_filename(file) + ":" + std::to_string(line) + "]";
```

src/cpp/src/logger.hpp
```diff
@@ -5,13 +5,89 @@
...
+    static Logger* get_instance() {
+        static Logger instance;
+        return &instance;
+    }
+    void do_log(ov::log::Level level, const char* file, int line, const std::string& msg) {
+        std::lock_guard<std::mutex> lock(log_mutex);
+        if (level > log_level) {
+            return;
+        }
+        std::cout << format_prefix(level, file, line) << msg << std::endl;
```
as-suvorov:
```suggestion
        std::cout << format_prefix(level, file, line) << " " << msg << std::endl;
```

src/cpp/src/logger.hpp
yangwang201911: Updated

src/cpp/src/logger.hpp
yangwang201911: Updated

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+        const auto current_level = log_level.load(std::memory_order_relaxed);
+        return current_level != ov::log::Level::NO && level != ov::log::Level::NO && level <= current_level;
+    }
+
+private:
+    std::atomic<ov::log::Level> log_level{ov::log::Level::NO};
+    std::mutex log_mutex;
+    void write_message(ov::log::Level level, const char* file, int line, const std::string& msg) {
+        std::lock_guard<std::mutex> lock(log_mutex);
+        std::cout << format_prefix(level, file, line) << msg;
```
sgonorov: Maybe errors and warnings should go to stderr instead of simple stdout?

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
 public:
-    static void warn(const std::string& message) {
-        std::cout << "[WARN] " << message << '\n';
-    };
+    Logger() {
+        log_level.store(get_cur_log_level(), std::memory_order_relaxed);
+    }
+    Logger(const Logger&) = delete;
+    Logger& operator=(const Logger&) = delete;
+    static Logger* get_instance() {
```
sgonorov: It's better making it thread-safe just in case

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+namespace detail {
+
+inline void log_message(ov::log::Level level, const char* file, int line, const std::string& msg) {
+    Logger* logger = Logger::get_instance();
+    logger->do_log(level, file, line, msg);
+}
+
+inline void log_message(ov::log::Level level, const char* file, int line, const char* msg) {
+    Logger* logger = Logger::get_instance();
+    if (msg) {
```
sgonorov: If log level is disabled, we still validate format string here

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+        logger->validate_format_string(msg, 0);
+    }
+    logger->do_log(level, file, line, msg ? std::string(msg) : std::string());
+}
+
+template <typename... Args, typename = std::enable_if_t<(sizeof...(Args) > 0)>>
+inline void log_message(ov::log::Level level, const char* file, int line, const char* format, Args&&... args) {
+    Logger* logger = Logger::get_instance();
+    logger->validate_format_string(format, sizeof...(Args));
```
sgonorov: If log level is disabled, we still validate format string here

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+        return ov::log::Level::NO;
+    }
+}
+
+inline ov::log::Level get_cur_log_level() {
+    static ov::log::Level cur_log_level = get_openvino_env_log_level();
+    return cur_log_level;
+}
 
 class Logger {
```
sgonorov, line 21: Better move as much logic/implementations as possible to `cpp` file, otherwise it will be compiled each time, bloating overall compilation time.

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+#include <stdexcept>
 #include <string>
+#include <type_traits>
+#include <utility>
+#include <vector>
+
+#include "openvino/openvino.hpp"
 
 namespace ov::genai {
+inline ov::log::Level get_openvino_env_log_level() {
```
sgonorov: Does it have to be inline?

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+        if (std::string::npos == index) {
+            return filePath;
+        }
+        return filePath.substr(index + 1);
+    }
+    std::string format_prefix(ov::log::Level level, const char* file, int line) const {
+        std::string level_str;
+        switch (level) {
+        case ov::log::Level::DEBUG:
+            level_str = "[DEBUG][" + get_filename(file) + ":" + std::to_string(line) + "] ";
```
sgonorov: Too much allocations. Pass `std::ostream& output` and do something like:
```cpp
output << "[DEBUG][" << get_filename(file) << ":" << std::to_string(line) << "] ";
```

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
 
 namespace ov::genai {
+inline ov::log::Level get_openvino_env_log_level() {
+    const char* env = std::getenv("OPENVINO_LOG_LEVEL");
+    if (!env)
+        return ov::log::Level::NO;
+    try {
+        std::string env_str(env);
+        size_t idx = 0;
+        int env_var_value = std::stoi(env_str, &idx);
```
sgonorov:
```suggestion
        const auto env_var_value = std::stoi(env_str, &idx);
```

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+            return ov::log::Level::NO;
+        }
+        return static_cast<ov::log::Level>(env_var_value);
+    } catch (...) {
+        return ov::log::Level::NO;
+    }
+}
+
+inline ov::log::Level get_cur_log_level() {
+    static ov::log::Level cur_log_level = get_openvino_env_log_level();
```
sgonorov: We already have a logger singleton, i don't think we need this static also.

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,260 @@
...
+            throw std::runtime_error("Failed to format log message");
+        }
+        if (required == 0) {
+            return {};
+        }
+        std::vector<char> buffer(static_cast<size_t>(required) + 1u, '\0');
+        std::vsnprintf(buffer.data(), buffer.size(), format, args);
+        return std::string(buffer.data(), static_cast<size_t>(required));
+    }
+    std::string get_filename(const std::string& filePath) const {
```
sgonorov: We can possibly use `std::string_view` here - this will help us avoid copying string on every call.

src/cpp/src/logger.hpp
yangwang201911: Updated. I moved errors log to stderr for keeping simple.

src/cpp/src/logger.hpp
yangwang201911: Updated with `static std::shared_ptr`+`std::call_once` here.

src/cpp/src/logger.hpp
yangwang201911: updated with checking `should_log() ` first here.

src/cpp/src/logger.hpp
yangwang201911: same as above.

src/cpp/src/logger.hpp
yangwang201911: updated.

src/cpp/src/logger.hpp
yangwang201911: Removed `inline`.

src/cpp/src/logger.hpp
yangwang201911: updated.

src/cpp/src/logger.hpp
yangwang201911: `static `gives us one-time initialization here to avoid calling `get_openvino_env_log_level()` every time.

src/cpp/src/logger.hpp
yangwang201911: updated.

src/cpp/src/logger.hpp
yangwang201911: updated.

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,97 @@
...
-    };
+    Logger();
+    Logger(const Logger&) = delete;
+    Logger& operator=(const Logger&) = delete;
+    static std::shared_ptr<Logger>& get_instance() {
+        static std::shared_ptr<Logger> instance;
+        std::call_once(init_flag, [&]() {
+            auto* obj_ptr = new Logger();
+            OPENVINO_ASSERT(obj_ptr != nullptr);
+            instance.reset(obj_ptr);
```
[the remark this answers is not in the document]
yangwang201911: updated.

Wovchena:
@as-suvorov, please review

src/cpp/src/logger.cpp
```diff
@@ -0,0 +1,221 @@
...
+        return ov::log::Level::NO;
+    }
+}
+
+ov::log::Level get_cur_log_level() {
+    static ov::log::Level cur_log_level = get_openvino_env_log_level();
+    return cur_log_level;
+}
+
+std::once_flag Logger::init_flag;
```
as-suvorov: Is it required here?

src/cpp/src/logger.cpp
```diff
@@ -0,0 +1,221 @@
...
+        va_end(args);
+        throw std::runtime_error(std::string{"Logger format error: "} + ex.what());
+    } catch (...) {
+        va_end(args);
+        throw std::runtime_error("Logger format error: unknown exception");
+    }
+    va_end(args);
+}
+
+void Logger::validate_format_string(const char* format, size_t arg_count) const {
```
as-suvorov: Are there formatting solutions with embedded validation?

yangwang201911: The intention of validate_format_string here is just to guard the subset of printf-style specifiers we want to support. It scans the format string and rejects anything outside our allowed list, so we fail fast before calling vsnprintf. It’s not a full formatting implementation—just a lightweight filter to keep unsupported placeholders out.

src/cpp/src/logger.cpp
yangwang201911: Removed.

src/cpp/src/logger.cpp
as-suvorov: Sure, I understand. The validate code looks relatively hard to support. It also seems to be duplication of what happen inside formatting implementation itself. It great that tests are implemented, so in general it's ok. But I still wonder if we can have formatting solution which doesn't require manual format validation. Can we get error form `vsnprintf` if format string is not ok?

yangwang201911: Agreed. Let's rely on vsnprintf for format validation. The compiler's format attribute will catch errors at compile time anyway.

src/cpp/src/logger.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return current_level != ov::log::Level::NO && level != ov::log::Level::NO && level <= current_level;
+}
+
+void Logger::write_message(ov::log::Level level, const char* file, int line, const std::string& msg) {
+    std::lock_guard<std::mutex> lock(log_mutex);
+    std::ostream& out = (level == ov::log::Level::ERR) ? std::cerr : std::cout;
+    format_prefix(out, level, file, line);
+    out << msg;
+    if (msg.empty() || msg.back() != '\n') {
+        out << std::endl;
```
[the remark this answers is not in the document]
as-suvorov: Looks worth to apply

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,89 @@
...
 namespace ov::genai {
+ov::log::Level get_openvino_env_log_level();
+ov::log::Level get_cur_log_level();
 
 class Logger {
 public:
-    static void warn(const std::string& message) {
-        std::cout << "[WARN] " << message << '\n';
-    };
+    Logger();
```
Wovchena: There's already `LogHelper` in `openvino/src/common/util/include/openvino/util/log.hpp`. Would it make sense to reuse it?

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,89 @@
...
+}  // namespace detail
+
+#define GenAILogger ov::genai::Logger::get_instance()
+
+#define GenAILogPrint(level, ...) ::ov::genai::detail::log_message(level, __FILE__, __LINE__, __VA_ARGS__)
+
+#define GENAI_DEBUG_LOG(...)   GenAILogPrint(ov::log::Level::DEBUG, __VA_ARGS__)
+#define GENAI_INFO_LOG(...)    GenAILogPrint(ov::log::Level::INFO, __VA_ARGS__)
+#define GENAI_WARNING_LOG(...) GenAILogPrint(ov::log::Level::WARNING, __VA_ARGS__)
+#define GENAI_ERROR_LOG(...)   GenAILogPrint(ov::log::Level::ERR, __VA_ARGS__)
```
Wovchena: Align macro names with OV, for example `OPENVINO_ERR`

src/cpp/src/logger.cpp
```diff
@@ -0,0 +1,147 @@
...
+    try {
+        std::string env_str(env);
+        size_t idx = 0;
+        const auto env_var_value = std::stoi(env_str, &idx);
+        if (idx != env_str.size()) {
+            return ov::log::Level::NO;
+        }
+        return static_cast<ov::log::Level>(env_var_value);
+    } catch (...) {
+        return ov::log::Level::NO;
```
Wovchena: I expected 0 to be the default which is `ERR`. I don't know OV's default. Can you check it and align?

src/cpp/src/logger.hpp
```diff
@@ -2,16 +2,89 @@
...
-        std::cout << "[WARN] " << message << '\n';
-    };
+    Logger();
+    Logger(const Logger&) = delete;
+    Logger& operator=(const Logger&) = delete;
+    static std::shared_ptr<Logger>& get_instance() {
+        static std::shared_ptr<Logger> instance;
+        std::call_once(init_flag, [&]() {
+            instance = std::make_shared<Logger>();
+        });
```
Wovchena: What's wrong with traditional Meyers singleton?

Given `Logger` is always used as static, is the class introduction even needed? It's pieces could just be static local variables inside functions. Not sure if it's a better solution, just asking

Wovchena (changes requested):

src/cpp/src/logger.hpp
yangwang201911: OpenVINO's LogHelper only works in `DEBUG` builds (requires `ENABLE_OPENVINO_DEBUG`). I‘m not sure if `GenAI `should follow the same approach or if we need logging in `RELEASE `builds for production diagnostics. Additionally, `GenAI`'s LogHelper supports runtime log level adjustment via env 'OPENVINO_LOG_LEVEL', while `OpenVINO`'s LogHelper is compile-time controlled through macros. What's your preference for the overall architecture?

src/cpp/src/logger.hpp
yangwang201911: updated.

src/cpp/src/logger.hpp
yangwang201911: You're right! I was over-thinking thread safety. C++11 already guarantees thread-safe static local initialization, so the traditional Meyers singleton works well. Simplified as suggested. 

I chose the class approach mainly for state cohesion (keeping log_level and log_mutex together).

src/cpp/src/logger.cpp
yangwang201911: OpenVINO's LogHelper doesn't have a default log level concept - it's purely compile-time controlled via `ENABLE_OPENVINO_DEBUG `(all logs on or all logs off). Let's use `ERR `as the default log level.

src/cpp/src/logger.hpp
Wovchena: GenAI should keep the logging in Release builds. My point was that `LogHelper` already have implementation to print the file name and line number. You can use it as a member of your class or construct it on the fly like OV does.

src/cpp/src/logger.hpp
Wovchena: The example was how it's represented in OV. Defining the same things in GenAI introduces name clash. Sure compiler's processor will figure that out but it reduces readability. Can they be `GENAI_*`?

Sorry for vague formulation earlier

src/cpp/src/logger.hpp
yangwang201911: Got it. When I tried to reuse OV's LogHelper in GenAI, I noticed that `ov::util::LogHelper` requires linking against the `openvino::util` static library from OpenVINO's development package. Is it appropriate for GenAI to depend on the OV dev package, or should we keep the current implementation to avoid this dependency?

src/cpp/src/logger.cpp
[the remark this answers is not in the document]
yangwang201911: updated.

src/cpp/src/logger.hpp
yangwang201911: Sure. Updated to `GENAI_*`.

src/cpp/src/logger.hpp
Wovchena: It's not appropriate. Nice catch! I though everything that has /include/ in OV is fully public

PR #2881, opened 2025-10-21
Fix unclassified issues

sgonorov:
Fixes [CVS-174891](https://jira.devtools.intel.com/browse/CVS-174891)


src/cpp/src/continuous_batching/pipeline_impl.cpp
```diff
@@ -249,24 +252,25 @@ GenerationHandle
 ContinuousBatchingPipeline::ContinuousBatchingImpl::add_request(
     uint64_t request_id,
     const ov::Tensor& input_ids,
-    ov::genai::GenerationConfig sampling_params,
+    const ov::genai::GenerationConfig& sampling_params,
     std::optional<ov::Tensor> token_type_ids) {
+    auto sampling_params_copy = sampling_params;
```
Wovchena, line 257: What was wrong here? Function call makes a copy avoiding extra var contrary to what you have now

sgonorov: It's because the base interface has changed. Some of the implementations doesn't need a copy. So it's better to pass a reference and decide if it's needed or not later.

PR #2882, opened 2025-10-21
Fix spelling errors & Grammar issues

as-suvorov:
## Description
Copilot PRs combination: https://github.com/openvinotoolkit/openvino.genai/pull/2870 https://github.com/openvinotoolkit/openvino.genai/pull/2860

## Checklist:

### DIFF
--- a/samples/cpp/whisper_speech_recognition/README.md
+++ b/samples/cpp/whisper_speech_recognition/README.md
@@ -59,7 +59,7 @@ result = pipeline.generate(raw_speech);
 //  Il s'agit d'une entité très complexe qui consiste...
 ```
 
-If the source audio language is know in advance, it can be specified as an argument to `generate` method:
+If the source audio language is known in advance, it can be specified as an argument to `generate` method:
 
 ```c++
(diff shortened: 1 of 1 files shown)


src/cpp/include/openvino/genai/whisper_pipeline.hpp
```diff
@@ -48,7 +48,7 @@ class OPENVINO_DEPRECATED(
 
     /// @brief write is called every time new vector of tokens is decoded, in case of assisting or prompt lookup
     /// decoding
-    /// @return StreamingStatus flag to indicate whether generation should be countinue to run or stopped or cancelled
+    /// @return StreamingStatus flag to indicate whether generation should continue to run or stopped or cancelled
```
sbalandi: flag to indicate whether generation should continue to run or stop or cancel
or
flag to indicate whether generation should be continued to run or stopped or cancelled

as-suvorov: fixed

PR #2887, opened 2025-10-22
[llm_bench] Fix help message

isanghao:
## Description
- Wrong help message is fixed

### DIFF
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -197,7 +197,7 @@ def get_argprser():
                         help='The task to setup the pipeline type')
     parser.add_argument(
         '--strength', type=float, default=None,
-        help='Applicable for Image to imaage/Inpainting pipelines. Indicates extent to transform the reference `image`. Must be between 0 and 1.')
+        help='Applicable for Image to image/Inpainting pipelines. Indicates extent to transform the reference `image`. Must be between 0 and 1.')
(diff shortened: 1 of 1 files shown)


tools/llm_bench/benchmark.py
```diff
@@ -197,8 +197,8 @@ def get_argprser():
                         help='The task to setup the pipeline type')
     parser.add_argument(
         '--strength', type=float, default=None,
-        help='Applicable for Image to imaage/Inpainting pipelines. Indicates extent to transform the reference `image`. Must be between 0 and 1.')
-    parser.add_argument("--disable_prompt_permutation", action="store_true", help="Disable modification prompt from run to run for avoid prefix caching")
+        help='Applicable for Image to image/Inpainting pipelines. Indicates extent to transform the reference `image`. Must be between 0 and 1.')
+    parser.add_argument("--disable_prompt_permutation", action="store_true", help="Disable modification prompt from run to run to allow prefix caching")
```
sbalandi, line 201: to avoid

as-suvorov: Why to avoid?

sbalandi: we need to prefix caching is not happened

sbalandi: you are right, allow here

sbalandi (approved):

PR #2890, opened 2025-10-22
[wwb] Fix reranker tests

sbalandi:
## Description

- split wwb reranker tests
- skip rerankier genAI tests on MacOS, task created 

## Checklist:

### DIFF
--- a/tools/who_what_benchmark/tests/test_cli_reranking.py
+++ b/tools/who_what_benchmark/tests/test_cli_reranking.py
@@ -1,32 +1,73 @@
 import subprocess  # nosec B404
+import sys
 import pytest
+import shutil
 import logging
+import tempfile
 from test_cli_image import run_wwb
+from pathlib import Path
 
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
+tmp_dir = tempfile.mkdtemp()
+
+
+OV_RERANK_MODELS = {
+    ("cross-encoder/ms-marco-TinyBERT-L2-v2", "text-classification"),
+    ("Qwen/Qwen3-Reranker-0.6B", "text-generation"),
+}
+
+
+def setup_module():
+    for model_info in OV_RERANK_MODELS:
+        model_id = model_info[0]
+        task = model_info[1]
+        MODEL_PATH = Path(tmp_dir, model_id.replace("/", "_"))
+        subprocess.run(["optimum-cli", "export", "openvino", "--model", model_id, MODEL_PATH, "--task", task, "--trust-remote-code"],
+                       capture_output=True,
+                       text=True)
+
+
+def teardown_module():
+    logger.info("Remove models")
+    shutil.rmtree(tmp_dir)
+
+
+@pytest.mark.parametrize(("model_info"), OV_RERANK_MODELS)
+def test_reranking_genai(model_info, tmp_path):
+    if sys.platform == 'darwin':
+        pytest.xfail("Ticket 175534")
+
+    GT_FILE = Path(tmp_dir) / "gt.csv"
+    model_id = model_info[0]
+    MODEL_PATH = Path(tmp_dir) / model_id.replace("/", "_")
+
+    # test GenAI
+    run_wwb([
+        "--base-model",
+        MODEL_PATH,
+        "--num-samples",
+        "1",
+        "--gt-data",
+        GT_FILE,
+        "--device",
+        "CPU",
+        "--model-type",
+        "text-reranking",
+        "--genai"
+    ])
+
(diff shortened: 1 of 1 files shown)


tools/who_what_benchmark/tests/test_cli_reranking.py
```diff
@@ -1,64 +1,94 @@
...
+
+def teardown_module():
+    logger.info("Remove models")
+    shutil.rmtree(tmp_dir)
+
+
+@pytest.mark.parametrize(("model_info"), OV_RERANK_MODELS)
+def test_reranking_genai(model_info, tmp_path):
+    if sys.platform == 'darwin':
+        pytest.xfail("Ticket 175534")
```
as-suvorov, line 40: Would test still hang with xfail? Should we skip it instead?

tools/who_what_benchmark/tests/test_cli_reranking.py
```diff
@@ -1,64 +1,94 @@
...
         "CPU",
         "--model-type",
-        model_type,
-        "--hf",
+        "text-reranking",
+        "--genai"
     ])
 
-    # test Optimum
+    assert Path(tmp_dir, "reference").exists()
```
as-suvorov, line 61: Should we assert similarity for optimum and genai vs hf as well?

tools/who_what_benchmark/tests/test_cli_reranking.py
```diff
@@ -1,64 +1,94 @@
...
-    # Collect reference with HF model
+
+def setup_module():
+    for model_info in OV_RERANK_MODELS:
+        model_id = model_info[0]
+        task = model_info[1]
+        MODEL_PATH = Path(tmp_dir, model_id.replace("/", "_"))
+        subprocess.run(["optimum-cli", "export", "openvino", "--model", model_id, MODEL_PATH, "--task", task, "--trust-remote-code"],
+                       capture_output=True,
+                       text=True)
```
Wovchena, line 29: If someone runs it locally with the intention to run a single test case, all of `OV_RERANK_MODELS` will be exported anyway. You should prefer fixtures. See samples tests. You can also reuse `OV_CACHE` env var and skip exporting if the model is already there.

But I don't think this should be addressed now. Seem to be a separate optional task.

tools/who_what_benchmark/tests/test_cli_reranking.py
Wovchena, line 1: I can't infer from the changes. What was the problem?

tools/who_what_benchmark/tests/test_cli_reranking.py
Wovchena: That questing would be valid for a `mark`. The call here aborts the test.

tools/who_what_benchmark/tests/test_cli_reranking.py
sbalandi: I don't know yet, it's just localization and skip exactly haning tests

tools/who_what_benchmark/tests/test_cli_reranking.py
sbalandi: will add in separate PR

tools/who_what_benchmark/tests/test_cli_reranking.py
Wovchena: There's no any specific test that hangs according to the logs. The logs contain the status of the last executed test and the logs don't contain the name of the next test. So I though it hangs at set up or tear down step. But there was no `setup_module()` for test_cli_reranking.py until this PR

Wovchena: Example: https://github.com/openvinotoolkit/openvino.genai/actions/runs/18756205375/job/53511854025

sbalandi: As far as I could find out, the hang occurs on run of wwb with GenAI ​

sbalandi: not on setup_module , as now it is separate step and works

tools/who_what_benchmark/tests/test_cli_reranking.py
```diff
@@ -1,64 +1,94 @@
...
+
+
+def teardown_module():
+    logger.info("Remove models")
+    shutil.rmtree(tmp_dir)
+
+
+@pytest.mark.parametrize(("model_info"), OV_RERANK_MODELS)
+def test_reranking_genai(model_info, tmp_path):
+    if sys.platform == 'darwin':
```
as-suvorov, line 39: Only genai pipeline hangs on mac?

sbalandi: yes

PR #2893, opened 2025-10-23
[GHA] Add missing jobs to workflow overal status

as-suvorov:
## Description
NodeJS test didn't fail overal workflow status: https://github.com/openvinotoolkit/openvino.genai/actions/runs/18723441840/job/53404679076?pr=2881
This PR includes all workflow jobs to overal_status.needs.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -25,7 +25,7 @@ env:
   SCCACHE_CACHE_SIZE: 30G
   SCCACHE_AZURE_KEY_PREFIX: genai/ubuntu/22_04/x64
   HF_HOME: /mount/caches/huggingface/lin
-  OV_CACHE: /mount/caches/huggingface/.ov_cache/lin
+  OV_CACHE: /mount/caches/huggingface/.ov_cache/lin/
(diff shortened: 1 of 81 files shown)


.github/workflows/linux.yml
```diff
@@ -935,7 +935,7 @@ jobs:
 
   Overall_Status:
     name: ci/gha_overall_status_linux
-    needs: [smart_ci, openvino_download, genai_build_cmake, genai_build_wheel, genai_build_samples, genai_build_nodejs, genai_tests_wheel, genai_samples_tests, genai_tools_tests, genai_nodejs_tests]
+    needs: [smart_ci, openvino_download, genai_build_cmake, genai_build_wheel, store_artifacts, genai_build_samples, genai_build_nodejs, genai_tests_wheel, genai_samples_tests, genai_tools_tests, genai_nodejs_tests, genai_xgrammar_off_tests]
```
Wovchena: `store_artifacts` runs unconditionally. So it's OK if it fails and subsequent jobs aren't needed

as-suvorov: removed

PR #2896, opened 2025-10-23
[JS] Support ChatHistory in .generate()

Retribution98:
## Description
 Support ChatHistory as a input in .generate()

Ticket: 175448

## Checklist:

### DIFF
--- a/src/js/tests/module.test.js
+++ b/src/js/tests/module.test.js
@@ -366,6 +366,7 @@ describe("stream()", () => {
 
 describe("LLMPipeline with chat history", () => {
   let pipeline = null;
+  // We need to keep previous messages between tests to avoid error for SDPA backend (macOS in CI)
   const chatHistory = new ChatHistory();
 
   before(async () => {
@@ -383,6 +384,7 @@ describe("LLMPipeline with chat history", () => {
       return_decoded_results: true,
     };
     const reply = await pipeline.generate(chatHistory, config);
+    // We need to keep previous messages between tests to avoid error for SDPA backend (macOS in CI)
     chatHistory.push({ role: "assistant", content: reply.toString() });
     assert.ok(Array.isArray(reply.texts));
     assert.equal(reply.texts.length, 1);
@@ -412,6 +414,7 @@ describe("LLMPipeline with chat history", () => {
       chunks.push(chunk);
     }
     assert.ok(chunks.length > 0);
+    // We need to keep previous messages between tests to avoid error for SDPA backend (macOS in CI)
     chatHistory.push({ role: "assistant", content: chunks.join("") });
   });
 });


src/js/src/helper.cpp
```diff
@@ -123,13 +123,26 @@ std::vector<std::string> js_to_cpp<std::vector<std::string>>(const Napi::Env& en
...
-        OPENVINO_THROW("Passed argument must be a string or an array of strings");
+GenerateInputs js_to_cpp<GenerateInputs>(const Napi::Env& env, const Napi::Value& value) {
+    try {
+        if (value.IsString()) {
+            return value.As<Napi::String>().Utf8Value();
+        } else if (value.IsArray()) {
+            auto array = value.As<Napi::Array>();
+            if (array.Length() == 0) {
+                OPENVINO_THROW("Passed array must not be empty.");
+            }
```
```suggestion
            OPENVINO_ASSERT(array.Length() != 0, "Passed array must not be empty.");
```
one if-else less

src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -251,15 +254,12 @@ export class LLMPipeline {
...
     generationConfig: GenerationConfig = {},
     callback: (chunk: string) => void | undefined,
   ) {
-    if (
-      typeof prompt !== "string" &&
-      !(Array.isArray(prompt) && prompt.every((item) => typeof item === "string"))
-    )
-      throw new Error("Prompt must be a string or string[]");
+    if (typeof inputs !== "string" && !Array.isArray(inputs))
+      throw new Error("Inputs must be a string, string[], or ChatHistory");
```
almilosz: This should be updated in tests

Retribution98: Thanks,  I check and change this message in the tests

src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -199,10 +201,11 @@ export class LLMPipeline {
     return result;
   }
 
-  stream(prompt: string, generationConfig: GenerationConfig = {}) {
+  stream(inputs: string | ChatHistory, generationConfig: GenerationConfig = {}) {
```
yatarkan, line 201:
```suggestion
  stream(inputs: string | string[] | ChatHistory, generationConfig: GenerationConfig = {}) {
```

src/js/lib/pipelines/llmPipeline.ts
```diff
@@ -8,11 +8,13 @@ export type Options = {
   max_new_tokens?: number;
 };
 
+type ChatHistory = Record<string, any>[];
```
yatarkan: In C++ and Python ChatHistory is a separate class (publicly exposed) that stores/manages messages, tools and extra_context. It has getters/setters for the members and a couple of other methods for convenience.
It would be good to unify such behavior in JS bindings as well, but it can be done as a next step in a separate PR.

Retribution98: Thank you for your comment. Yes, I will add it in the separate PR

yatarkan: BTW, `samples/js/chat_sample.js` should be updated as well to showcase chat history class instead of `start_chat()`/`finish_chat()` methods (after existing chat history PRs are ready and merged).

src/js/lib/pipelines/llmPipeline.ts
Retribution98: The stream doesn't support a list of inputs because it is not designed for parallel execution.

src/js/tests/module.test.js
```diff
@@ -110,10 +110,10 @@ describe("generation parameters validation", () => {
   });
 
   it("should throw an error if temperature is not a number", async () => {
```
yatarkan:
```suggestion
  it("should throw an error if no arguments passed to generate", async () => {
```

src/js/tests/module.test.js
```diff
@@ -199,6 +199,34 @@ describe("LLMPipeline.generate()", () => {
     assert.strictEqual(replyStr, reply.toString());
   });
 
+  it("generate(chatHistory, config)", async () => {
```
yatarkan: It makes sense to move test for generate() with chat history to separate suite (`describe`) - chat history cannot be mixed with other input types (string, string[], tokenized inputs), please see failed test https://github.com/openvinotoolkit/openvino.genai/actions/runs/18878957102/job/53877723377?pr=2896#step:7:1896

src/js/tests/module.test.js
```diff
@@ -356,4 +356,62 @@ describe("stream()", () => {
...
+    ]);
+    const config = {
+      max_new_tokens: 10,
+      return_decoded_results: true,
+    };
+    const reply = await pipeline.generate(chatHistory, config);
+    chatHistory.push({ role: "assistant", content: reply.toString() });
+    assert.ok(Array.isArray(reply.texts));
+    assert.equal(reply.texts.length, 1);
+    assert.ok(typeof reply.texts[0] === "string");
```
yatarkan, line 391: Does it make sense to also check if `reply.toString()` equals `reply.texts[0]`?

Retribution98: No, it doesn't.
If `reply.texts` has length 1 and `reply.texts[0]` is string, then `reply.texts[0] === reply.toString()`.
This may be unclear, but in this test case, it doesn't matter if these values are equal.
We could use reply.texts[0] to store the results; otherwise, if the result doesn't have any texts, storing the message will crash, which will affect other test cases.

I added a comment explaining why chatHistory is used.

PR #2897, opened 2025-10-23
Fix for another batch of coverity issues

sgonorov:
Fixes more issues from [CVS-174757](https://jira.devtools.intel.com/browse/CVS-174757)

### DIFF
--- a/src/cpp/include/openvino/genai/perf_metrics.hpp
+++ b/src/cpp/include/openvino/genai/perf_metrics.hpp
@@ -164,6 +164,8 @@ struct OPENVINO_GENAI_EXPORTS PerfMetrics {
      */
     virtual void evaluate_statistics(std::optional<TimePoint> start_time = std::nullopt);
 
+    virtual ~PerfMetrics() = default;
+
     /**
      * @brief convert duration to microseconds
      *


samples/c/whisper_speech_recognition/whisper_speech_recognition.c
```diff
@@ -29,13 +32,19 @@ int main(int argc, char* argv[]) {
...
+    }
+
     float file_sample_rate;
     if (load_wav_file(wav_file_path, &audio_data, &audio_length, &file_sample_rate) != 0) {
         exit_code = EXIT_FAILURE;
         goto err;
     }
 
-    if (file_sample_rate != 16000.0f) {
+    if (fabsf(file_sample_rate - 16000.0f) > SAMPLE_RATE_TOLERANCE) {
```
Wovchena, line 47: Why shouldn't the audio be resampled if it's something close to 16000.0f but not exact?

src/cpp/src/image_generation/models/sd3_transformer_2d_model.cpp
```diff
@@ -48,6 +49,7 @@ SD3Transformer2DModel::SD3Transformer2DModel(const std::string& model,
                                              const size_t vae_scale_factor) :
     m_config(config), m_vae_scale_factor(vae_scale_factor) {
     m_model = utils::singleton_core().read_model(model, weights);
+    OPENVINO_ASSERT(m_model, "Failed to read model from weights");
```
Wovchena: I believe `m_model` can't be null because `read_model()` will throw if the it failed to read model from weights.

This check is absent in other places so I guess earlier it was parked as false positive trigger

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
```diff
@@ -87,7 +87,7 @@ ov::genai::GenerationConfig LLMInferWrapper::get_generation_config() const {
 }
 
 void LLMInferWrapper::set_generation_config(ov::genai::GenerationConfig config) {
-    m_generation_config = config;
+    m_generation_config = std::move(config);
```
Wovchena, line 90: Function signature should have `&&` in that case

samples/c/whisper_speech_recognition/whisper_speech_recognition.c
sgonorov: It's just a safety measure. It's better never directly compare floats, it can be stored as 15999.999 or 16000.001. In this case we're getting an unnecessary resampling.

Wovchena: That's what I'm asking. Why is the resampling unnecessary for 15999.999?

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
sgonorov: I think it's not very handy. Using `&&` will make it less flexible and won't provide real benefits in my opinion.

src/cpp/src/image_generation/models/sd3_transformer_2d_model.cpp
sgonorov: You're right, fixed it 👍

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
Wovchena: It's even less handy when the config is unexpectedly moved away. The function signature should let callers know

samples/c/whisper_speech_recognition/whisper_speech_recognition.c
sgonorov: If we assume we have some audio with sample rate of 15999.9999hz, I think the best resampling can give is to provide some bad resampling artifacts without improving an overall quality. But the main problem here is that an audio can simply be 16khz, it's just the way the float stores values.

src/cpp/src/speculative_decoding/speculative_decoding_stateful.cpp
sgonorov: It's not moved away - it's always being copied when passed to the function on line 89. I've just removed the second copy and replaced it with moving. So it won't be unexpectedly moved away on `set_generation_config` call.

Wovchena:
@sgonorov, reminder

PR #2911, opened 2025-10-25
[llm_bench] Limit the MAX Input Text In log

peterchen-intel:
## Description
When the Input prompt is long, for example 8K or even 128K, the output of Input Text in the log makes the log hard to scroll up or down. Limit the MAX Input Text In log, and end the Text with " ... [Truncated to 1024]" if the Text was truncated.

## Checklist:

### DIFF
--- a/tools/llm_bench/llm_bench_utils/metrics_print.py
+++ b/tools/llm_bench/llm_bench_utils/metrics_print.py
@@ -6,6 +6,7 @@ import sys
 
 MAX_INPUT_TXT_IN_LOG = 1024
 
+
 def print_metrics(iter_num, iter_data, tms=None, tms_infer=None, warm_up=False,
                   stable_diffusion=None, tokenization_time=None, batch_size=1,
                   prompt_idx=-1, whisper=None, text_emb=None, latency_unit=None,


tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -230,7 +231,8 @@ def run_text_2_speech_benchmark(model_path, framework, device, args, num_iters,
         for idx, input_text in enumerate(text_list):
             p_idx = prompt_idx_list[idx]
             for num in range(num_iters + 1):
-                if num == 0:
+                if args["output_dir"] is None and num == 0:
```
rkazants: please clarify - if we save generated speech to output_dir, right? If yes, I expect that input text will be showed despite output_dir is set. Now it is more convinient.

sbalandi:
Hi @peterchen-intel , if I understand correctly this commit is caused by situations where the prompt is very large, am i right ?
I agree with @rkazants . So maybe to add something like --silence mode and will control print by them ? Or print only n characters, for example 500 and , if prompt is bigger, then add `...` at the end and warning with information ?

tools/llm_bench/llm_bench_utils/metrics_print.py
```diff
@@ -2,7 +2,9 @@
 # Copyright (C) 2023-2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 import logging as log
+import sys
 
+MAX_INPUT_TXT_IN_LOG = 1024
 
 def print_metrics(iter_num, iter_data, tms=None, tms_infer=None, warm_up=False,
```
peterchen-intel, line 10:
```suggestion

def print_metrics(iter_num, iter_data, tms=None, tms_infer=None, warm_up=False,
```

tools/llm_bench/task/text_to_speech_generation.py
peterchen-intel: Take the suggestion to limit the chars number of Input Text in log.

PR #2914, opened 2025-10-26
Restore 4 missing VLM test cases

sgonorov:
Fixes [CVS-175578](https://jira.devtools.intel.com/browse/CVS-175578)

### DIFF
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -1393,27 +1393,30 @@ def test_model_tags_missing_native(ov_pipe_model: VlmModelInfo):
 
 @pytest.mark.precommit
 @pytest.mark.parametrize(
-    "ov_pipe_model",
+    "ov_pipe_model,has_image,has_video",
     [
-        pytest.param(("katuni4ka/tiny-random-qwen2vl","SDPA")),
-        pytest.param(("katuni4ka/tiny-random-qwen2vl", "PA")),
-        pytest.param(("katuni4ka/tiny-random-qwen2.5-vl", "SDPA")),
-        pytest.param(("katuni4ka/tiny-random-qwen2.5-vl", "PA"), marks=pytest.mark.xfail(reason="CVS-167316")),
+        pytest.param(("katuni4ka/tiny-random-qwen2vl","SDPA"), True, False, id="qwen2vl/SDPA/image"),
+        pytest.param(("katuni4ka/tiny-random-qwen2vl", "PA"), True, False, id="qwen2vl/PA/image"),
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1393,27 +1393,30 @@ def test_model_tags_missing_native(ov_pipe_model: VlmModelInfo):
...
-        pytest.param(("katuni4ka/tiny-random-llava-next-video", "PA")),
+        pytest.param(("katuni4ka/tiny-random-gemma3", "PA"), True, False, id="gemma3/PA/image", marks=pytest.mark.xfail(reason="CVS-171180")),
+        pytest.param(("qnguyen3/nanoLLaVA", "SDPA"), True, False, id="nanoLLaVA/SDPA/image"),
+        pytest.param(("qnguyen3/nanoLLaVA", "PA"), True, False, id="nanoLLaVA/PA/image"),
+        pytest.param(("katuni4ka/tiny-random-llava-next-video", "SDPA"), True, False, id="llava-next-video/SDPA/image"),
+        pytest.param(("katuni4ka/tiny-random-llava-next-video", "PA"), True, False, id="llava-next-video/PA/image"),
+        pytest.param(("katuni4ka/tiny-random-llava-next-video", "SDPA"), False, True, id="llava-next-video/SDPA/video"),
+        pytest.param(("katuni4ka/tiny-random-llava-next-video", "PA"), False, True, id="llava-next-video/PA/video"),
+        pytest.param(("katuni4ka/tiny-random-llava-next-video", "SDPA"), True, True, id="llava-next-video/SDPA/image+video"),
+        pytest.param(("katuni4ka/tiny-random-llava-next-video", "PA"), True, True, id="llava-next-video/PA/image+video"),
```
Wovchena, line 1415: Text only case is missing. Is this intentional?

Wovchena (changes requested):

tests/python_tests/test_vlm_pipeline.py
sgonorov: Yes, it wasn't there, in the initial cases. So I've restored only ones which were previously removed.

Wovchena:
I'd like to wait for @yatarkan since it's he who brought the problem

tests/python_tests/test_vlm_pipeline.py
yatarkan, line 1415: Does it make sense to move attention backend (SDPA, PA) to a separate param to avoid duplication?
Or it is better to do in a separate PR?

sgonorov: Model fixture is currently parametrized by 2 parameters, model name and a backend, This way we can create most of the models for testing and reuse them. So it's better stick to this approach. Making backend a separate parameter will make things much more complex in my opinion.

PR #2924, opened 2025-10-28
Update llm bench

rnwang04:
## Description

- This PR is split from https://github.com/openvinotoolkit/openvino.genai/pull/2764 to make the review easier.

- fix GenerationConfig when disable prompt_permutations
- raise error when user set sparse_attention_config and use_sparse_attention=False


peterchen-intel:
Merge after https://github.com/openvinotoolkit/openvino.genai/pull/2764

PR #2929, opened 2025-10-28
Reprodue '(': illegal token on right side of '::'

Wovchena:
CVS-175618


Wovchena:
Fixed in https://github.com/openvinotoolkit/openvino.genai/pull/2958

PR #2941, opened 2025-10-29
[RAG] Add text rerank models task

as-suvorov:
## Description
For text rerank models `--task` needs to be provided during conversion with `optimum-cli`

image

Ticket: 173665

## Checklist:


site/docs/supported-models/_components/text-rerank-models-table/index.tsx
```diff
@@ -3,14 +3,17 @@ import { TEXT_RERANK_MODELS } from './models';
 import { BaseModelsTable, LinksCell } from '../base-models-table';
 
 export default function TextRerankModelsTable(): React.JSX.Element {
-  const headers = ['Architecture', 'Example HuggingFace Models'];
+  const headers = ['Architecture', 'optimum-cli task', 'Example HuggingFace Models'];
```
yatarkan: If it will render as a preformatted (inline code) block
```suggestion
  const headers = ['Architecture', '`optimum-cli` task', 'Example HuggingFace Models'];
```

site/docs/supported-models/_components/text-rerank-models-table/index.tsx
```diff
@@ -3,14 +3,17 @@ import { TEXT_RERANK_MODELS } from './models';
...
-  const rows = TEXT_RERANK_MODELS.map(({ architecture, models }) => (
+  const rows = TEXT_RERANK_MODELS.map(({ architecture, optimumIntelTask, models }) => (
     <>
       <tr key={architecture}>
         <td rowSpan={models.length}>
           <code>{architecture}</code>
         </td>
+        <td rowSpan={models.length}>
+          <code>{optimumIntelTask}</code>
+        </td>
```
yatarkan, line 16: I was thinking about a separate note section with a content like:

> Info: To export rerank models with `optimum-cli` provide `--task` parameter from the corresponding table column below.

Default task for majority of rerank models (text-classification) can be also put to the note, with exception/link note for Qwen3ForCausalLM. However, if models tend to have more unique tasks it might become duplicated info, so keeping separate task column is reasonable.

as-suvorov: I was thinking like we already have a kind of 'note' exception for qwen3 models. I wasn't sure if we need to introduce two exceptions one for generic case and second existing for qwen3 models. I thought a more generic solution is a better choice. I don't have strong opinion here, please tell me if you anyway want to go with 2 notes.

site/docs/supported-models/_components/text-rerank-models-table/models.ts
```diff
@@ -45,6 +49,7 @@ export const TEXT_RERANK_MODELS: TextRerankModelType[] = [
   },
   {
     architecture: 'Qwen3ForCausalLM',
+    optimumIntelTask: 'text-generation-with-past',
```
rkazants, line 52: that is strange. How do you use `hidden states` or `output_tokens`? You need scores for reranking.

as-suvorov: Here is example usage: https://huggingface.co/Qwen/Qwen3-Reranker-0.6B#transformers-usage
We have same prostporcessing: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/rag/text_rerank_pipeline.cpp#L90

rkazants: Thanks, I got it. From my perspective we need to adjust it on optimum-intel side. That means we need to have classification task for it where we should prepend this post-processing on optimum-intel side so that IR will contain all required ops and no need to prepend it on GenAI side. That's for the future, but will be nice to have JIRA ticket for such enhancement. How do you think?

as-suvorov: We discussed with @apaniukov further optimization of qwen3 reranker postprocessing, here is the python reference: https://huggingface.co/Qwen/Qwen3-Reranker-0.6B/discussions/3. Model card with backed in postporcessing: https://huggingface.co/tomaarsen/Qwen3-Reranker-0.6B-seq-cls.

Having such postprocessing on optimum-intel side would basically convert original qwen3-rerank model into: https://huggingface.co/tomaarsen/Qwen3-Reranker-0.6B-seq-cls. Which is ok.

I have only once concern regarding move postprocessing to optimum-cli side. It's passing yes/no_token_id. When postporcessing on genai side we can ask user to provide yes/no token ids or string representation. This feature is not implemented yet on GenAI side. But I guess it would not be possible to implement if we have postprocessing on optimum-intel side.
To summarize if need to pass yes/no token ids as a parameter we should leave post-processing on GenAI side. If no, it's better to move it to optimum-intel.
@apaniukov What's your thoughts on this?

apaniukov: The post-processing applied to the model depends on the tokenizer's padding side. We can sync this in GenAI's embedding pipeline, but the optimum class does not have a direct link to the tokenizer.

apaniukov: The post-processing is applied on optimum side if `--library=sentence-transformers` key is used during export. We need to support this scenario on GenAI side.

apaniukov (approved):

yatarkan (approved):

PR #2950, opened 2025-10-31
[Docs] Add Structured Output Docs

apaniukov:
## Description
Add SO page to the docs

Ticket: CVS-173810

## Checklist:

### DIFF
--- a/site/docs/guides/structured-output.mdx
+++ b/site/docs/guides/structured-output.mdx
@@ -127,4 +127,4 @@ You can find complete sample programs that demonstrate structured output generat
 - [Trigger structured output during regular generation run](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/structural_tags_generation.py)
 - [Test-driven examples](https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_structured_output.py) (comprehensive examples covering JSON, regex, EBNF and structural tags)
 
-OpenVINO GenAI JavaScript bindings also support the Structured Output feature. Check [JS samples directory](https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/js/text_generation) for usage examples.
+OpenVINO GenAI JavaScript bindings also support the Structured Output feature. Check [JS samples](/docs/samples/js/text_generation#7-structured-output-sample-structured_output_sample) for usage examples.


yangsu2022:
Shall we add some figures like this？
image
from https://arxiv.org/pdf/2411.15100

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
+---
+sidebar_position: 7
+---
+
+# Structured Output
+
+Structured output generation lets a model produce machine‑readable responses that follow a defined format — for example JSON, a regular expression, or an EBNF grammar. OpenVINO GenAI provides a flexible structured output enforcement so the model can generate valid, well‑formed output.
+
+## What OpenVINO GenAI provides
```
yatarkan: Let's use title case
```suggestion
## What OpenVINO GenAI Provides
```

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+- Base structured output formats: JSON, regex, and grammar (EBNF) based structured generation.
+- Structural tags:
+  - Compound grammars: combine several grammars to handle more complex tasks or enforce different function calling styles.
+  - Tags: blend regular free‑form generation with structural tags during single generate call.
+
+## Structured output configuration
```
yatarkan: Let's use title case
```suggestion
## Structured Output Configuration
```

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
...
+- `json_schema` — a JSON Schema that the generator will try to satisfy; useful when you need typed fields and nested structures.
+- `regex` — a regular expression that the output must match.
+- `grammar` — an EBNF grammar that describes structure of the output.
+- `structural_tags_config` and `compound_grammar` — advanced options to combine multiple grammars.
+
+:::info
+You should set only one primary structured constraint (for example `json_schema`, `regex`, or `grammar`) on a `StructuredOutputConfig` instance. The library validates the configuration and will raise an error if conflicting constraints are provided (for example both `json_schema` and `regex`).
+:::
+
+Deprecation note: the `StructuralTagsConfig` class and the `compound_grammar` field are deprecated. Use the `structural_tags_config` field on `StructuredOutputConfig` instead; it provides the same functionality and is the recommended API going forward.
```
yatarkan: I would use warning note to increase visibility
```suggestion
:::warning Deprecation Note
The `StructuralTagsConfig` class and the `compound_grammar` field are deprecated. Use the `structural_tags_config` field on `StructuredOutputConfig` instead; it provides the same functionality and is the recommended API going forward.
:::
```

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+## Samples and further reading
+
+You can find complete sample programs that demonstrate structured output generation here:
+
+- Python sample: `samples/python/text_generation/structured_output_generation.py`
+- C++ sample: `samples/cpp/text_generation/structured_output_generation.cpp`
+- Combining multiple grammars for structured output: `samples/python/text_generation/compound_grammar_generation.py`
+- Combine regular sampling and structured output: `samples/python/text_generation/structural_tags_generation.py`
+- Test-driven examples: `tests/python_tests/test_structured_output.py` (comprehensive examples covering JSON, regex, EBNF and structural tags)
```
yatarkan: Would be better to provide GitHub links for samples and tests.

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
...
+    auto result = pipe.generate(prompt, gen_config);
+
+    std::cout << "Model output: " << result.texts[0] << std::endl;
+    return 0;
+}
+```
+    </TabItemCpp>
+</LanguageTabs>
+
+## Samples and further reading
```
yatarkan: Let's use title case
```suggestion
## Samples and Further Reading
```

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+You can find complete sample programs that demonstrate structured output generation here:
+
+- Python sample: `samples/python/text_generation/structured_output_generation.py`
+- C++ sample: `samples/cpp/text_generation/structured_output_generation.cpp`
+- Combining multiple grammars for structured output: `samples/python/text_generation/compound_grammar_generation.py`
+- Combine regular sampling and structured output: `samples/python/text_generation/structural_tags_generation.py`
+- Test-driven examples: `tests/python_tests/test_structured_output.py` (comprehensive examples covering JSON, regex, EBNF and structural tags)
+
+## Tips and best practices
```
yatarkan: Let's use title case
```suggestion
## Tips and Best Practices
```

yatarkan: Does it make sense to move "Tips and Best Practices" and "Samples and Further Reading" headers to level 3 so they will go under more general level 2 header "Using Structured Output"?

yatarkan (changes requested):

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
...
+
+### Examples
+
+<LanguageTabs>
+    <TabItemPython>
+```python
+import json
+import openvino_genai as ov_genai
+
+
```
rkazants, line 62:
```suggestion
```

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,116 @@
+---
+sidebar_position: 7
+---
+
+# Structured Output
+
+Structured output generation lets a model produce machine‑readable responses that follow a defined format — for example JSON, a regular expression, or an EBNF grammar. OpenVINO GenAI provides a flexible structured output enforcement so the model can generate valid, well‑formed output.
```
rkazants: May be it makes sense to share examples of non-structured and structured outputs for the same model and input prompt and show the main benefit of structured output. Will be quite useful for readers non-familiar with this feature.

Also, I think this feature may degrade generation performance. I think we need to make a note about it. What ratio?

site/docs/guides/structured-output.mdx
apaniukov: This is PEP-8 compliant formatting (two lines after imports). The example is short, so I would prefer to leave it as is.

site/docs/guides/structured-output.mdx
apaniukov: Done

site/docs/guides/structured-output.mdx
apaniukov: Done

site/docs/guides/structured-output.mdx
apaniukov: Applied

site/docs/guides/structured-output.mdx
apaniukov: Done

site/docs/guides/structured-output.mdx
apaniukov: > May be it makes sense to share examples of non-structured and structured outputs for the same model and input prompt and show the main benefit of structured output. 

The structural_tags_generation sample actually does that.

> Also, I think this feature may degrade generation performance. I think we need to make a note about it. What ratio?

Good point, added that to the docs page. The actual ratio is dependent on the model, task and input data, so there is no particular number that we could give.

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,124 @@
+---
+sidebar_position: 7
+---
+
+# Structured Output
+
+Structured output generation lets a model produce machine-readable responses that follow a defined format - for example JSON, a regular expression, or an EBNF grammar. OpenVINO GenAI provides a flexible structured output enforcement so the model can generate valid, well‑formed output.
```
Wovchena:
```suggestion
Structured output generation lets to force a model produce machine-readable responses that follow a defined format - for example JSON, a regular expression, or an EBNF grammar. OpenVINO GenAI provides a flexible structured output enforcement so the model can generate valid, well‑formed output.
```
Models are capable of produce machine-readable responses so they don't need additional letting

apaniukov: Add "forced" and delete "machine-readable" (not all enforcements are for machine-readable output)

apaniukov:
> Shall we add some figures like this？ image from https://arxiv.org/pdf/2411.15100

Thanks for the idea, added this example with the link to the original paper.

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,126 @@
...
+</LanguageTabs>
+
+### Samples and Further Reading
+
+You can find complete sample programs that demonstrate structured output generation here:
+
+- [Python sample](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/structured_output_generation.py)
+- [C++ sample](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/text_generation/structured_output_generation.cpp)
+- [Combining multiple grammars for structured output](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/compound_grammar_generation.py)
+- [Trigger structured output during regular generation run](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/structural_tags_generation.py)
```
yatarkan, line 127: There are also JS samples for structured output.
Probably we don't need to provide links to each sample here, maybe it is better to point user to samples page in docs (for C++, Python, JS), e.g. with relative link `/docs/samples/js/text_generation#7-structured-output-sample-structured_output_sample`

site/docs/guides/structured-output.mdx
```diff
@@ -0,0 +1,126 @@
...
+sidebar_position: 7
+---
+
+# Structured Output
+
+Structured output generation forces a model to produce responses that follow a defined format - for example JSON, a regular expression, or an EBNF grammar. OpenVINO GenAI provides a flexible structured output enforcement so the model can generate well‑formed output.
+
+## How It Works
+
+[![Example from XGrammar paper](/img/structured_output_work_example.png)](https://arxiv.org/pdf/2411.15100)
```
yatarkan: It might be not so clear that image is a link to the paper. I would suggest to keep image and add more explicit text in the end of this section/paragraph, e.g.:

> Refer to [XGrammar paper](https://arxiv.org/pdf/2411.15100) for more details.

site/docs/guides/structured-output.mdx
apaniukov: Rearanged this part.

site/docs/guides/structured-output.mdx
apaniukov: Done

site/docs/guides/structured-output.mdx
apaniukov: Done

PR #2940, opened 2025-10-29
Modify sample to demonstrate parsing

pavel-esir:
## Description
Modify compound grammar sample to demonstrate parsing abilities.

Ticket:  CVS-175569

## Checklist:
- [x] Tests have been updated or added to cover the new code (existing test to compound grammar covers this sample)
- [x] This patch addresses parsing part of the ticket. 
- [x] I have made corresponding changes to the documentation

PR #2951, opened 2025-10-31
[JS] Fix structured output test

Retribution98:
## Description
- Add comparison JavaScript samples with Python.

Ticket: 175278

## Checklist:


samples/js/text_generation/compound_grammar_generation.js
```diff
@@ -125,6 +125,7 @@ async function main() {
 
     process.stdout.write("Assistant: ");
     await pipe.generate(modelInput2, generationConfig, streamer);
+    console.log();
```
yatarkan, line 155: Redundant?

Retribution98: No, it adds \n after output from the streamer.

Retribution98:
Compound grammar generation sample was changed in PR #2940. This requires the implementation of Parser for JS.
This PR is blocked

apaniukov:
@Retribution98 you can add an external JS output parsing as a quick fix and then work on parser bindings.

Retribution98:
> @Retribution98 you can add an external JS output parsing as a quick fix and then work on parser bindings.

@apaniukov Yes, I made it

samples/js/text_generation/compound_grammar_generation.js
```diff
@@ -97,34 +120,38 @@ async function main() {
     const userText1 = "Do dolphins have fingers?";
     console.log("User: ", userText1);
     chatHistory.push({ role: "user", content: userText1 });
-    const modelInput = tokenizer.applyChatTemplate(chatHistory, true);
+    const modelInput = tokenizer.applyChatTemplate(chatHistory, true, undefined, tools);
```
yatarkan: If tools are constant, we could set `tools` to `chatHistory` directly with `chatHistory.setTools(tools)` to setup default tools behavior for this history. In that case we will not need to apply chat template with `tokenizer.applyChatTemplate()` manually in each conversation iteration.

PR #2952, opened 2025-11-03
[GENAI]Introduce add_extension to genai.

xipingyan:
## Description
This pull request introduces a new, unified mechanism for adding and managing OpenVINO extensions (both as file paths and as in-memory extension objects) via pipeline properties. The changes include a C++ API for specifying extensions, utilities for extracting and registering them, integration into the main pipeline classes, and comprehensive tests in both C++ and Python to ensure robust extension handling.
Document was updated in [PR#3547](https://github.com/openvinotoolkit/openvino.genai/pull/3547)

**Core functionality for extension management:**

* Added `openvino/genai/extensions.hpp` with the `extensions` helper function and `ExtensionList` type, enabling users to specify extensions as file paths or `ov::Extension` objects in a way compatible with pipeline constructors.
* Implemented utility functions `extract_extensions` and `extract_extensions_to_core` in `utils.hpp/cpp` to extract extensions from property maps and register them with the OpenVINO Core singleton.

**Pipeline integration:**

* Updated `ContinuousBatchingPipeline`, `LLMPipeline`, and `VLMPipeline` to call `extract_extensions_to_core` before model loading, ensuring extensions are registered prior to model instantiation.

**Python API support:**

* Enhanced the Python-to-C++ property conversion logic to support the new "extensions" property, accepting both string paths and `ov.Extension` objects. 

Ticket: CVS-176587

## Checklist:
- [x] I have made corresponding changes to the documentation. [PR#3547](https://github.com/openvinotoolkit/openvino.genai/pull/3547)

### DIFF
--- a/.github/skills/genai-model-checker/SKILL.md
+++ b/.github/skills/genai-model-checker/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: genai-model-checker
+description: "Validate a newly supported optimum-intel model with OpenVINO GenAI. Use when: checking new model support, verifying model export to OpenVINO IR, running GenAI inference test with llm_bench, benchmarking model accuracy with who-what-benchmark."
+argument-hint: "model_id and task (e.g. tencent/HY-MT1.5-1.8B text-generation-with-past)"
+---
+
+# GenAI Model Checker
+
+Validates that a HuggingFace model exported via optimum-intel works correctly with OpenVINO GenAI pipelines and passes accuracy benchmarks.
+
+## When to Use
+
+- A new model was added to optimum-intel and needs GenAI validation
+- Verify a HuggingFace model exports to OpenVINO IR and runs inference
+- Check model accuracy after conversion using who-what-benchmark
+
+## Inputs
+
+The user must provide:
+
+- **model_id**: HuggingFace model identifier (e.g. `tencent/HY-MT1.5-1.8B`)
+- **task**: optimum-cli export task. Supported values:
+  - `text-generation-with-past`
+  - `image-text-to-text`
+  - `text-to-image`
+  - `image-to-image`
+  - `feature-extraction`
+  - `text-classification`
+  - `text-to-video`
+  - `automatic-speech-recognition`
+
+## Prerequisites
+
+Activate the Python virtual environment before running any commands.
+
+1. **Locate the virtual environment** — check for common directories at the repository root: `.venv/`, `venv/`, `env/`. Use `list_dir` to find it. If none is found, ask the user for its location.
+2. **Activate** based on the current platform:
+   - **Linux/macOS**: `source <venv_path>/bin/activate`
+   - **Windows (cmd)**: `<venv_path>\Scripts\activate.bat`
+   - **Windows (PowerShell)**: `<venv_path>\Scripts\Activate.ps1`
+
+## Procedure
+
+### Step 1: Run check_model.py
+
+Run the checker script from the repository root:
+
+```
+python3 .github/skills/genai-model-checker/scripts/check_model.py \
+    --model-id <model_id> \
+    --task <export_task> \
+    --work-dir /tmp/genai-model-check
+```
+
+Run `python3 .github/skills/genai-model-checker/scripts/check_model.py --help` for the full argument reference including defaults.
+
+#### Skip flags (for re-runs after a fix)
+
+When a previous run already passed some steps (e.g. export succeeded but inference test failed), use skip flags to avoid repeating expensive passed steps:
+
+- `--skip-export` — reuse existing IR in `<work-dir>/model_ir` instead of re-exporting (avoids re-downloading weights)
+- `--skip-llm-bench` — skip the llm_bench inference test
+- `--skip-wwb` — skip the who-what-benchmark accuracy check
+
+Do **not** use skip flags on the first run. Only use them when retrying after a targeted fix.
+
+### Step 2: Interpret Results
+
+The script logs progress for each step and exits with code 0 (pass) or non-zero (fail).
+
+**Pass criteria:**
+
+- Export: exit code 0
+- Inference test (llm_bench): exit code 0, metrics line logged
+- WWB accuracy (three sub-steps, all must pass):
+  1. HF ground truth generation: exit code 0
+  2. Optimum target evaluation: similarity ≥ `SIMILARITY_THRESHOLD`
+  3. GenAI target evaluation: similarity ≥ `SIMILARITY_THRESHOLD`
+
+  Note: the WWB step is skipped automatically for `automatic-speech-recognition` (no WWB support).
+
+**Log files:** each tool writes its own dedicated log; paths are printed during execution. When a step fails, read the corresponding log for the full traceback and context before drawing any conclusions.
+
+### Step 3: Report Results
+
+Report results to the user. If all steps pass, indicate success and provide performance metrics and accuracy data. If a step fails, analyze tool logs and provide a summary with the failed tool log path.
+
+### Security
+
+- **NEVER** install any packages. Assume the environment is pre-configured.
+- **NEVER** invoke `optimum-cli`, `wwb`, or `llm_bench` directly. Always go through `check_model.py`.
+- **NEVER** modify `model_id` — pass it exactly as provided by the user.
--- a/.github/skills/genai-model-checker/scripts/check_model.py
+++ b/.github/skills/genai-model-checker/scripts/check_model.py
@@ -1,0 +1,450 @@
+#!/usr/bin/env python3
+# Copyright (C) 2024-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""End-to-end validation of a HuggingFace model with OpenVINO GenAI."""
+
+import argparse
+from dataclasses import dataclass
+import logging
+import os
+import re
+import shlex
+import subprocess  # nosec B404
+import sys
+import time
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+FAIL_TAIL_LINES = 10  # Lines of output to show on failure (full output goes to per-step log file)
+
+# Embedded task mapping: optimum-cli export task → (llm_bench --task, wwb --model-type)
+# wwb_type=None means WWB step is skipped for that task.
+TASK_MAPPING = {
+    "text-generation-with-past": ("text_gen", "text"),
+    "image-text-to-text": ("visual_text_gen", "visual-text"),
+    "text-to-image": ("text-to-image", "text-to-image"),
+    "image-to-image": ("image-to-image", "image-to-image"),
+    "feature-extraction": ("text_embed", "text-embedding"),
+    "text-classification": ("text_rerank", "text-reranking"),
+    "text-to-video": ("text-to-video", "text-to-video"),
+    "automatic-speech-recognition": ("speech_to_text", None),
+}
+
+
+WWB_SIMILARITY_THRESHOLD = 0.95
+
+
+def log_header(args: argparse.Namespace, bench_task: str, wwb_task: str, work_dir: str, log_file: str) -> None:
+    logger.info("GenAI Model Checker")
+    logger.info("  model_id:    %s", args.model_id)
+    logger.info("  task:        %s", args.task)
+    logger.info("  device:      %s", args.device)
+    logger.info("  bench_task:  %s", bench_task)
+    logger.info("  wwb_type:    %s", wwb_task)
+    logger.info("  work_dir:    %s", work_dir)
+    logger.info("  log_file:    %s", log_file)
+    logger.info("Python: %s", sys.version)
+    logger.info("Executable: %s", sys.executable)
+    for pkg in ["openvino", "openvino_genai", "openvino_tokenizers", "optimum-intel", "transformers", "diffusers"]:
+        version = subprocess.run(
+            [sys.executable, "-c", f"import importlib.metadata; print(importlib.metadata.version('{pkg}'))"],
+            capture_output=True,
+            text=True,
+            timeout=10,
+        )
+        logger.info("  %-25s %s", pkg, version.stdout.strip() if version.returncode == 0 else "NOT INSTALLED")
+
+
+@dataclass
+class ToolResult:
+    name: str
+    success: bool
+    err_msg: str | None = None
+
+    def raise_if_failed(self) -> None:
+        if not self.success and self.err_msg:
+            raise RuntimeError(self.err_msg)
+
+
+class ToolWrapper:
+    """Utility to run a command-line tool and capture output for logging."""
+
+    def __init__(self, name: str, commands_list: list, work_dir: Path):
+        self.name = name
+        self.commands_list = commands_list
+        self.work_dir = work_dir
+        self.work_dir.mkdir(parents=True, exist_ok=True)
+
+        self.logger_prefix = f"[{self.name}]"
+        self.log_path = self.work_dir / f"{self.name}.log"
+        self.log_path.touch(exist_ok=True)
+
+    def _post_run_hook(self, result: subprocess.CompletedProcess) -> ToolResult:
+        """Process results"""
+        if result.returncode == 0:
+            return ToolResult(name=self.name, success=True)
+
+        err_messages = []
+        err_messages.append(f"{self.logger_prefix}: failed with exit code {result.returncode}")
+        err_messages.append(f"{self.logger_prefix}: Last lines from {self.log_path}:")
+        for line in result.stdout[-FAIL_TAIL_LINES:]:
+            err_messages.append(f"  {line}")
+        err_msg = "\n".join(err_messages)
+        return ToolResult(name=self.name, success=False, err_msg=err_msg)
+
+    def run(self) -> ToolResult:
+        command = shlex.join(self.commands_list)
+        logger.info(f"{self.logger_prefix}: {command}")
+        logger.info(f"{self.logger_prefix}: Log path: {self.log_path.resolve()}")
+
+        with open(self.log_path, "w") as lf:
+            lf.write(f"Command: {command}\n")
+            lf.write(f"work_dir: {self.work_dir.resolve() if self.work_dir else Path(os.getcwd()).resolve()}\n")
+            lf.write("Tool output:\n")
+            lf.flush()
+
+            start = time.perf_counter()
+            # Using Popen directly to stream output line-by-line to log file
+            proc = subprocess.Popen(
+                self.commands_list,
+                cwd=os.getcwd(),
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                text=True,
+                bufsize=10,
+            )
+
+            tool_logs: list[str] = []
+            for line in proc.stdout:
+                lf.write(line)
+                lf.flush()
+                tool_logs.append(line.rstrip("\n"))
+            proc.wait()
+
+            elapsed = time.perf_counter() - start
+
+        result = subprocess.CompletedProcess(self.commands_list, returncode=proc.returncode, stdout=tool_logs)
+        logger.info(f"{self.logger_prefix}: took {elapsed:.2f} seconds")
+
+        return self._post_run_hook(result)
+
+
+class OptimumExportTool(ToolWrapper):
+    def __init__(self, model_id: str, task: str, model_dir: Path, work_dir: Path):
+        cmd = [
+            "optimum-cli",
+            "export",
+            "openvino",
+            "-m",
+            model_id,
+            str(model_dir),
+            "--task",
+            task,
+        ]
+        super().__init__(name="optimum_export", commands_list=cmd, work_dir=work_dir)
+
+
+class LlmBenchTool(ToolWrapper):
+    def __init__(self, model_dir: Path, task: str, device: str, work_dir: Path):
+        llm_bench_script_path = Path("tools/llm_bench/benchmark.py")
+
+        if not llm_bench_script_path.is_file():
+            raise FileNotFoundError(f"llm_bench script not found: {llm_bench_script_path}")
+        cmd = [
+            sys.executable,
+            str(llm_bench_script_path),
+            "-m",
+            str(model_dir),
+            "-d",
+            device,
+            "-n",
+            "1",
+            "--task",
+            task,
+        ]
+        super().__init__(name="llm_bench", commands_list=cmd, work_dir=work_dir)
+
+    def _post_run_hook(self, result):
+        if result.returncode != 0:
+            return super()._post_run_hook(result)
+        last_llm_bench_line = result.stdout[-1]
+        logger.info(f"{self.logger_prefix}: llm_bench metrics: {last_llm_bench_line}")
+        return ToolResult(name=self.name, success=True)
+
+
+def _log_csv_listing(logger_prefix: str, directory: Path) -> None:
+    """Log CSV files in directory so the agent can inspect them."""
+    csv_files = sorted(directory.glob("*.csv")) if directory.is_dir() else []
+    if csv_files:
+        logger.info("%s: CSV files in %s:", logger_prefix, directory)
+        for f in csv_files:
+            logger.info("%s:   %s (%d bytes)", logger_prefix, f.name, f.stat().st_size)
+    else:
+        logger.info("%s: No CSV files found in %s", logger_prefix, directory)
+
+
+class HFWWBGroundTruthTool(ToolWrapper):
+    def __init__(self, model_id: str, task: str, work_dir: Path, num_samples: int, device: str):
+        cmd = [
+            "wwb",
+            "--base-model",
+            model_id,
+            "--gt-data",
+            str(work_dir / "gt.csv"),
+            "--model-type",
+            task,
+            "--device",
+            device,
+            "--num-samples",
+            str(num_samples),
+            "--hf",  # Use HuggingFace backend for ground truth generation
+        ]
+        super().__init__(name="wwb_hf_ground_truth", commands_list=cmd, work_dir=work_dir)
+
+    def _post_run_hook(self, result: subprocess.CompletedProcess) -> ToolResult:
+        tool_result = super()._post_run_hook(result)
+        _log_csv_listing(self.logger_prefix, self.work_dir)
+        return tool_result
+
+
+def parse_wwb_metrics_value(stdout: list[str]) -> float | None:
+    # example output, parse similarity
+    # INFO:whowhatbench.wwb:Metrics for model: model-check/model_ir
+    # INFO:whowhatbench.wwb:   similarity
+    # 0    0.957204
+    if len(stdout) < 1:
+        return None
+    metric_value_str = stdout[-1].strip().split()[1]
+    return float(metric_value_str)
+
+
+class OptimumWWBTargetEvaluationTool(ToolWrapper):
+    def __init__(self, model_dir: Path, task: str, work_dir: Path, num_samples: int, device: str):
+        cmd = [
+            "wwb",
+            "--target-model",
+            str(model_dir),
+            "--gt-data",
+            str(work_dir / "gt.csv"),
+            "--model-type",
+            task,
+            "--device",
+            device,
+            "--num-samples",
+            str(num_samples),
+            # Optimum backend used by default
+            "--output",
+            str(work_dir / "optimum"),
+        ]
+        super().__init__(name="wwb_optimum_target_eval", commands_list=cmd, work_dir=work_dir)
+
+    def _post_run_hook(self, result: subprocess.CompletedProcess) -> ToolResult:
+        _log_csv_listing(self.logger_prefix, self.work_dir / "optimum")
+        if result.returncode != 0:
+            return super()._post_run_hook(result)
+
+        metrics_value = parse_wwb_metrics_value(result.stdout)
+        if metrics_value is None:
+            return ToolResult(
+                name=self.name,
+                success=False,
+                err_msg=f"{self.logger_prefix}: Failed to parse WWB metrics value from output.",
+            )
+
+        if metrics_value < WWB_SIMILARITY_THRESHOLD:
+            return ToolResult(
+                name=self.name,
+                success=False,
+                err_msg=(
+                    f"{self.logger_prefix}: WWB similarity {metrics_value:.4f} is below threshold of "
+                    f"{WWB_SIMILARITY_THRESHOLD:.2f}."
+                ),
+            )
+        logger.info(
+            f"{self.logger_prefix}: optimum similarity {metrics_value:.4f} meets threshold of {WWB_SIMILARITY_THRESHOLD:.2f}."
+        )
+        return ToolResult(name=self.name, success=True)
+
+
+class GenAIWWBTargetEvaluationTool(ToolWrapper):
+    def __init__(self, model_dir: Path, task: str, work_dir: Path, num_samples: int, device: str):
+        cmd = [
+            "wwb",
+            "--target-model",
+            str(model_dir),
+            "--gt-data",
+            str(work_dir / "gt.csv"),
+            "--model-type",
+            task,
+            "--device",
+            device,
+            "--num-samples",
+            str(num_samples),
+            "--genai",  # Use GenAI backend for target evaluation
+            "--output",
+            str(work_dir / "genai"),
+        ]
+        super().__init__(name="wwb_genai_target_eval", commands_list=cmd, work_dir=work_dir)
+
+    def _post_run_hook(self, result: subprocess.CompletedProcess) -> ToolResult:
+        _log_csv_listing(self.logger_prefix, self.work_dir / "genai")
+        if result.returncode != 0:
+            return super()._post_run_hook(result)
+
+        metrics_value = parse_wwb_metrics_value(result.stdout)
+        if metrics_value is None:
+            return ToolResult(
+                name=self.name,
+                success=False,
+                err_msg=f"{self.logger_prefix}: Failed to parse WWB metrics value from output.",
+            )
+
+        if metrics_value < WWB_SIMILARITY_THRESHOLD:
+            return ToolResult(
+                name=self.name,
+                success=False,
+                err_msg=(
+                    f"{self.logger_prefix}: WWB similarity {metrics_value:.4f} is below threshold of "
+                    f"{WWB_SIMILARITY_THRESHOLD:.2f}."
+                ),
+            )
+        logger.info(
+            f"{self.logger_prefix}: genai similarity {metrics_value:.4f} meets threshold of {WWB_SIMILARITY_THRESHOLD:.2f}."
+        )
+        return ToolResult(name=self.name, success=True)
+
+
+def _setup_exception_logging() -> None:
+    def handle_exception(exc_type, exc_value, exc_traceback):
+        if issubclass(exc_type, KeyboardInterrupt):
+            sys.__excepthook__(exc_type, exc_value, exc_traceback)
+            return
+        logger.critical("Unhandled exception", exc_info=(exc_type, exc_value, exc_traceback))
+
+    sys.excepthook = handle_exception
+
+
+def _setup_logging(log_file: Path) -> None:
+    file_handler = logging.FileHandler(log_file, mode="w")
+    file_handler.setFormatter(logging.Formatter("%(asctime)s %(levelname)s: %(message)s"))
+    console_handler = logging.StreamHandler()
+    console_handler.setFormatter(logging.Formatter("%(message)s"))
+    logging.basicConfig(level=logging.INFO, handlers=[console_handler, file_handler])
+    _setup_exception_logging()
+
+
+def _get_arguments() -> argparse.Namespace:
+    def model_id_validator(model_id: str) -> str:
+        MODEL_ID_PATTERN = re.compile(r"^[A-Za-z0-9_.\-]+/[A-Za-z0-9_.\-]+$")
+        if not MODEL_ID_PATTERN.match(model_id):
+            raise argparse.ArgumentTypeError(
+                f"Invalid model_id format: {model_id}\n"
+                "Expected format: org-name/model-name (alphanumeric, hyphens, dots, underscores)"
+            )
+        return model_id
+
+    class _HelpFormatter(argparse.RawDescriptionHelpFormatter, argparse.ArgumentDefaultsHelpFormatter):
+        pass
+
+    parser = argparse.ArgumentParser(
+        description=(
+            "End-to-end validation of a HuggingFace model with OpenVINO GenAI.\n\n"
+            "Steps:\n"
+            "  1. Export model to OpenVINO IR via optimum-cli.\n"
+            "  2. Inference test via llm_bench (1 iteration).\n"
+            f"  3. Accuracy check via who-what-benchmark (similarity threshold: {WWB_SIMILARITY_THRESHOLD}):\n"
+            "       a. Generate ground truth with HuggingFace backend.\n"
+            "       b. Evaluate with Optimum backend.\n"
+            "       c. Evaluate with GenAI backend."
+        ),
+        epilog=(
+            "Examples:\n"
+            "  python check_model.py \\\n"
+            "      --model-id tencent/HY-MT1.5-1.8B \\\n"
+            "      --task text-generation-with-past \\\n"
+            "      --work-dir /tmp/genai-model-check\n\n"
+            "  # Reuse existing IR, skip accuracy check:\n"
+            "  python check_model.py \\\n"
+            "      --model-id tencent/HY-MT1.5-1.8B \\\n"
+            "      --task text-generation-with-past \\\n"
+            "      --skip-export --skip-wwb"
+        ),
+        formatter_class=_HelpFormatter,
+    )
+    parser.add_argument(
+        "--model-id",
+        type=model_id_validator,
+        required=True,
+        help="HuggingFace model identifier (e.g. tencent/HY-MT1.5-1.8B)",
+    )
+    parser.add_argument(
+        "--task",
+        required=True,
+        choices=sorted(TASK_MAPPING.keys()),
+        help="optimum-cli export task (e.g. text-generation-with-past)",
+    )
+
+    parser.add_argument("--work-dir", default="/tmp/genai-model-check", help="Working directory for outputs")
+    parser.add_argument("--device", default="CPU", help="Inference device")
+    parser.add_argument(
+        "--skip-export", action="store_true", help="Skip model export (reuse existing IR in work-dir/model_ir)"
+    )
+    parser.add_argument("--skip-llm-bench", action="store_true", help="Skip llm_bench inference test")
+    parser.add_argument("--skip-wwb", action="store_true", help="Skip who-what-benchmark accuracy check")
+    parser.add_argument("--num-samples", type=int, default=4, help="Number of WWB samples")
+    return parser.parse_args()
+
+
+def main():
+    args = _get_arguments()
+
+    bench_task, wwb_task = TASK_MAPPING[args.task]
+
+    work_dir = Path(args.work_dir)
+    model_dir = work_dir / "model_ir"
+    work_dir.mkdir(parents=True, exist_ok=True)
+
+    log_file = work_dir / "check_model.log"
+    _setup_logging(log_file)
+
+    log_header(args, bench_task, wwb_task, work_dir, log_file)
+
+    # Step 1: Export
+    if args.skip_export:
+        logger.info("Skipping model export. Reusing existing IR: %s", model_dir)
+    else:
+        optimum_export_work_dir = work_dir / "optimum_export"
+        result = OptimumExportTool(args.model_id, args.task, model_dir, optimum_export_work_dir).run()
+        result.raise_if_failed()
+
+    # Step 2: Inference test
+    if args.skip_llm_bench:
+        logger.info("Skipping llm_bench test")
+    else:
+        llm_bench_work_dir = work_dir / "llm_bench"
+        result = LlmBenchTool(model_dir, bench_task, args.device, llm_bench_work_dir).run()
+        result.raise_if_failed()
+
+    # Step 3: WWB accuracy
+    if args.skip_wwb or wwb_task is None:
+        logger.info("Skipping wwb accuracy check")
+    else:
+        wwb_work_dir = work_dir / "wwb"
+        hf_gt_result = HFWWBGroundTruthTool(args.model_id, wwb_task, wwb_work_dir, args.num_samples, args.device).run()
+        hf_gt_result.raise_if_failed()
+
+        optimum_result = OptimumWWBTargetEvaluationTool(
+            model_dir, wwb_task, wwb_work_dir, args.num_samples, args.device
+        ).run()
+        optimum_result.raise_if_failed()
+
+        genai_result = GenAIWWBTargetEvaluationTool(
+            model_dir, wwb_task, wwb_work_dir, args.num_samples, args.device
+        ).run()
+        genai_result.raise_if_failed()
+
+
+if __name__ == "__main__":
+    main()
--- a/tools/llm_bench/task/text_generation.py
+++ b/tools/llm_bench/task/text_generation.py
@@ -23,10 +23,25 @@ FW_UTILS = {'pt': llm_bench_utils.pt_utils, 'ov': llm_bench_utils.ov_utils}
 DEFAULT_OUTPUT_TOKEN_SIZE = 512
 
 
-def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list, md5_list,
-                        prompt_index, bench_hook, tokens_len, streaming, model_precision, proc_id, mem_consumption):
-    set_seed(args['seed'])
-    input_text_list = [input_text] * args['batch_size']
+def run_text_generation(
+    input_text,
+    num,
+    model,
+    tokenizer,
+    args,
+    iter_data_list,
+    md5_list,
+    prompt_index,
+    bench_hook,
+    tokens_len,
+    streaming,
+    model_precision,
+    proc_id,
+    mem_consumption,
+    prefix,
+):
+    set_seed(args["seed"])
+    input_text_list = [input_text] * args["batch_size"]
     if args["output_dir"] is not None and num == 0:
         for bs_index, in_text in enumerate(input_text_list):
             llm_bench_utils.output_file.output_input_text(in_text, args, model_precision, prompt_index, bs_index, proc_id)
@@ -58,6 +73,7 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
     # This is too small because test prompt may contain 4096 tokens which leaves no space for new tokens.
     # Override it to preserve max_new_tokens.
     max_length = 2**64 - 1
+    log.info("%s Text generation start: %s", prefix, datetime.datetime.now().isoformat())
     start = time.perf_counter()
     if streaming:
         if args['infer_count'] is not None and args['end_token_stopping'] is False:
@@ -110,6 +126,7 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
                 **additional_args
             )
     end = time.perf_counter()
+    log.info("%s Text generation end: %s", prefix, datetime.datetime.now().isoformat())
     generation_time = end - start
     memory_metrics = mem_consumption.iter_stop_and_collect_data(num)
 
@@ -192,7 +209,7 @@ def run_text_generation(input_text, num, model, tokenizer, args, iter_data_list,
         bench_hook.clear_time_infer_list()
 
 
-def genai_generate(streaming, model, tokens_len, gen_config, empty_lora, input_data, batch_size):
+def genai_generate(streaming, model, tokens_len, gen_config, empty_lora, input_data, batch_size, prefix):
     import openvino_genai
     import openvino as ov
     cb_pipeline = isinstance(model, openvino_genai.ContinuousBatchingPipeline)
@@ -200,6 +217,7 @@ def genai_generate(streaming, model, tokens_len, gen_config, empty_lora, input_d
         input_data = [ov.Tensor([input]) for input in input_data.input_ids.data]
         gen_config = [gen_config] * batch_size
 
+    log.info("%s Text generation start: %s", prefix, datetime.datetime.now().isoformat())
     start = time.perf_counter()
     if streaming:
         text_print_streamer = get_genai_chunk_streamer()(model.get_tokenizer(), tokens_len)
@@ -230,6 +248,7 @@ def genai_generate(streaming, model, tokens_len, gen_config, empty_lora, input_d
         else:
             generation_result = model.generate(input_data, gen_config)
     end = time.perf_counter()
+    log.info("%s Text generation end: %s", prefix, datetime.datetime.now().isoformat())
     generated_tokens = []
     if cb_pipeline:
         for res in generation_result:
@@ -242,9 +261,24 @@ def genai_generate(streaming, model, tokens_len, gen_config, empty_lora, input_d
     return generated_tokens, perf_metrics, end - start
 
 
-def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data_list, md5_list, prompt_index,
-                              streamer, tokens_len, streaming, model_precision, proc_id, mem_consumption):
-    input_text_list = [input_text] * args['batch_size']
+def run_text_generation_genai(
+    input_text,
+    num,
+    model,
+    tokenizer,
+    args,
+    iter_data_list,
+    md5_list,
+    prompt_index,
+    streamer,
+    tokens_len,
+    streaming,
+    model_precision,
+    proc_id,
+    mem_consumption,
+    prefix,
+):
+    input_text_list = [input_text] * args["batch_size"]
     if args["output_dir"] is not None and num == 0:
         for bs_index, in_text in enumerate(input_text_list):
             llm_bench_utils.output_file.output_input_text(in_text, args, model_precision, prompt_index, bs_index, proc_id)
@@ -327,8 +361,9 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
         gen_config.num_assistant_tokens = int(args['num_assistant_tokens'])
         config_info += f"max_ngram_size {gen_config.max_ngram_size}, num_assistant_tokens {gen_config.num_assistant_tokens}"
         log.info(config_info)
-    generated_tokens, perf_metrics, generation_time = genai_generate(streaming, model, tokens_len, gen_config,
-                                                                     args["empty_lora"], input_data, args['batch_size'])
+    generated_tokens, perf_metrics, generation_time = genai_generate(
+        streaming, model, tokens_len, gen_config, args["empty_lora"], input_data, args["batch_size"], prefix
+    )
     if streaming:
         tokenization_time.append(np.mean(perf_metrics.raw_metrics.detokenization_durations) / 1000)
         generated_text = tokenizer.decode(generated_tokens)
@@ -405,9 +440,24 @@ def run_text_generation_genai(input_text, num, model, tokenizer, args, iter_data
         metrics_print.print_generated(num, warm_up=(num == 0), generated=generated_text[0], prompt_idx=prompt_index)
 
 
-def run_text_generation_genai_with_stream(input_text, num, model, tokenizer, args, iter_data_list, md5_list,
-                                          prompt_index, streamer, tokens_len, streaming, model_precision, proc_id, mem_consumption):
-    input_text_list = [input_text] * args['batch_size']
+def run_text_generation_genai_with_stream(
+    input_text,
+    num,
+    model,
+    tokenizer,
+    args,
+    iter_data_list,
+    md5_list,
+    prompt_index,
+    streamer,
+    tokens_len,
+    streaming,
+    model_precision,
+    proc_id,
+    mem_consumption,
+    prefix,
+):
+    input_text_list = [input_text] * args["batch_size"]
     if args["output_dir"] is not None and num == 0:
         for bs_index, in_text in enumerate(input_text_list):
             llm_bench_utils.output_file.output_input_text(in_text, args, model_precision, prompt_index, bs_index, proc_id)
@@ -479,9 +529,12 @@ def run_text_generation_genai_with_stream(input_text, num, model, tokenizer, arg
         gen_config.num_assistant_tokens = int(args['num_assistant_tokens'])
         config_info += f"max_ngram_size {gen_config.max_ngram_size}, num_assistant_tokens {gen_config.num_assistant_tokens}"
         log.info(config_info)
+
+    log.info("%s Text generation start: %s", prefix, datetime.datetime.now().isoformat())
     start = time.perf_counter()
     generated_tokens = model.generate(input_data, gen_config, streamer=streamer).tokens
     end = time.perf_counter()
+    log.info("%s Text generation end: %s", prefix, datetime.datetime.now().isoformat())
     generation_time = end - start
     memory_metrics = mem_consumption.iter_stop_and_collect_data(num)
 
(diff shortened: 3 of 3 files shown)


src/cpp/include/openvino/genai/generation_config.hpp
Wovchena, line 1: Please take a different approach.

Introduce a new `ov::Property<std::vector<std::string>>` that will correspond to `ov::Core::add_extension()`. If this property is given in pipeline constructor, pass its content to the corresponding `ov::Core`.

`ov::Core` should no longer be singleton. Any pipeline that introduces support for `add_extension()` should manage its own `ov::Core`. This requires https://github.com/openvinotoolkit/openvino/pull/32738 to be merged. To allow other `ov::Core`'s functions in future, GenAI should introduce a private common function that calls the needed `ov::Core`'s function given properties (it should be only `add_extension()` at this moment). Every pipeline needs to call this helper.

Don't modify all pipelines at once. Split the changes into multiple PRs.

Let me know if you have other though

Wovchena (changes requested):

src/cpp/include/openvino/genai/generation_config.hpp
xipingyan: I list all candidates:
```
ContinuousBatchingPipeline
LLMPipeline
VLMPipeline
WhisperPipeline
Image2ImagePipeline
InpaintingPipeline
SD3Transformer2DModel
T5EncoderModel
Text2ImagePipeline
UNet2DConditionModel
TextEmbeddingPipeline
TextRerankPipeline
Text2SpeechPipeline
```
So I will prioritize to enable `LLMPipeline` and `VLMPipeline` in the first PR.
Other question about test case,
Actually, I don't know how to add appropriate  test case, because we need to pass a library(dll or so) path.
What do you think about current enabled test case? Only pass tokenizer library path and check if we can catch any exceptions. @Wovchena

Wovchena: This idea for test is nice. I can't come up with a different test case

xipingyan: We can't get any exception when we pass correct tokenizer library, so I think we can pass a fake path, we can get openvino threw exception log, just check this log, it may be better.

Wovchena: What log are you referring to?
```py
    with pytest.raises(RuntimeError) as exc_info:
        VLMPipeline(models_path, "CPU", properties)
```
looks correct

xipingyan: because we pass faked path, so openvino will throw exception with specific log, I just catch this exception and compare this log. so this log is from openvino, if openvino update the threw log, maybe trigger this test fail.
is this reasonable? @Wovchena @sunxiaoxia2022 @peterchen-intel

Wovchena: It's reasonable. There seem to be no other way

Wovchena: There's going to be a discussion with arch team this Monday. I'll update you after that

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1508,3 +1508,14 @@ def get_nanollava_processor():
     genai_text = genai_output.texts[0]
 
     assert optimum_text == genai_text
+
+@pytest.mark.precommit
```
Wovchena:
```suggestion
```
It was removed. Same for tests/python_tests/test_continuous_batching.py and tests/python_tests/test_llm_pipeline.py

src/cpp/include/openvino/genai/generation_config.hpp
Wovchena: > There's going to be a discussion with arch team this Monday. I'll update you after that

Moved to Wednesday

tests/python_tests/test_vlm_pipeline.py
sunxiaoxia2022: Ok, removed it.

src/cpp/src/utils.cpp
```diff
@@ -718,6 +718,21 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
...
 };
 
+void add_extensions_to_core(ov::AnyMap& properties) {
+    auto it = properties.find(EXTENSIONS_ARG_NAME);
+    if (it != properties.end()) {
+#ifdef _WIN32
+        auto extensions = it->second.as<std::vector<std::wstring>>();
+#else
+        auto extensions = it->second.as<std::vector<std::string>>();
+#endif
```
yangwang201911: Should we need this platform-specific path conversion in GenAI? OpenVINO's `add_extension()` API already handles it.

https://github.com/openvinotoolkit/openvino/blob/55aefe5a933dd4dfcbec642e85b3e6b7c469ec3c/src/inference/include/openvino/runtime/core.hpp#L469

sunxiaoxia2022: If we don't specify platform, that is parsing all extensions to `std::string` type, then we will only call `void add_extension(const std::string& library_path)`, and will never call `void add_extension(const std::wstring& library_path)`.

src/cpp/src/utils.cpp
```diff
@@ -718,6 +718,21 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
     return {properties, attention_backend};
 };
 
+void add_extensions_to_core(ov::AnyMap& properties) {
```
sgonorov: Maybe it's safer to add mutex synchronization here just in case?

sgonorov: Do we have any safeguards against loading one extension multiple times?

sgonorov: Maybe we should also have `remove_extension` functionality?

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1620,3 +1620,12 @@ def get_nanollava_processor():
     genai_text = genai_output.texts[0]
 
     assert optimum_text == genai_text
+
+def test_vlm_pipeline_add_extension():
+    models_path = _get_ov_model(MODEL_IDS[0])
+
+    properties = {"EXTENSIONS": ["fake_path"]}
```
sgonorov: Should we store extensions as a part of property? I think it's better move them outside. If we're loading extensions into one pipeline, other pipelines will get access to it. Can't we make this more explicit? It's not a clear behavior.

src/cpp/src/utils.cpp
```diff
@@ -718,6 +718,21 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
...
+    auto it = properties.find(EXTENSIONS_ARG_NAME);
+    if (it != properties.end()) {
+#ifdef _WIN32
```
sgonorov: I think type alias will do better here:
```
#ifdef _WIN32
using ExtensionPath = std::wstring;
#else
using ExtensionPath = std::string;
#endif
```

tests/python_tests/test_continuous_batching.py
```diff
@@ -690,3 +690,17 @@ def test_eagle3_sd_string_inputs(main_model, main_device, draft_model, draft_dev
 
     # Compare results:
     compare_generation_results([prompt], ref_gen_results, ov_gen_results, ov_generation_config)
+
+
+def test_continuous_batching_add_extension():
```
sgonorov: Do we have successful loading tests? I can see only negative tests as for now.

tests/python_tests/test_vlm_pipeline.py
wangleis: After GenAI create own OV core for each pipeline, other pipelines will not get access to it. This change for multiple OV Cores in GenAI will be handled in other PR.

@Wovchena Could you comment on this?

src/cpp/src/utils.cpp
sunxiaoxia2022: I think we don't need `remove_extension`. `add_extensions_to_core` will be called if "EXTENSIONS" exists in properties, Otherwise `add_extensions_to_core` will not be called. I don't know  `remove_extension` will be called in what case.

sunxiaoxia2022: A mutex synchronization has been added to `Core::add_extension` from OV. The code is as follows:
https://github.com/openvinotoolkit/openvino/blob/master/src/inference/src/dev/core_impl.cpp#L1445
So we don't need add it again.

src/cpp/src/utils.cpp
sunxiaoxia2022: Ok, updated.

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: The purpose of this test case is to verify that the interface can enter the OV process, meaning the interface is working correctly. Exception test can also verify this function. 
This test case has been discussed with @Wovchena in above comments.

tests/python_tests/test_vlm_pipeline.py
Wovchena: Agree with Wanglei

src/cpp/src/utils.cpp
Wovchena: Use `filesystem::path` overload as said in CVS-176661. This should eliminate `ifdef` branching. This will be covered on arch review tomorrow. @xipingyan, doublecheck if you are invited

xipingyan: Very good, `std::filesystem::path` can handle file system paths across different operating systems, the important thing is that  `std::filesystem::path` also provided lots of API to handle kinds of processing about Path.
@sunxiaoxia2022 , please enable it .

sunxiaoxia2022: Updated, please take a look.

src/cpp/src/utils.cpp
sunxiaoxia2022: I tested this case with OV. The result is ok, That is loading one extension multiple times will not lead to incorrect output. 
Ops in extension were called once during inference.
But the ops in extension were registered multiple times indeed. So I created this issue [CVS-179111](https://jira.devtools.intel.com/browse/CVS-179111).

src/cpp/src/utils.cpp
```diff
@@ -718,6 +718,29 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
     return {properties, attention_backend};
 };
 
+void add_extensions_to_core(ov::AnyMap& properties) {
+    auto it = properties.find(EXTENSIONS_ARG_NAME);
+    if (it != properties.end()) {
+        if (it->second.is<std::vector<std::string>>()) {
```
Wovchena: It should be `vector<varian<path, shared_ptr<Extension>>>`

Wovchena: Add property like functions to public header. See arch review for details

sunxiaoxia2022: @Wovchena Updated, please take a look.

sunxiaoxia2022: Updated.

sgonorov (approved):

src/cpp/include/openvino/genai/add_extension.hpp
```diff
@@ -0,0 +1,55 @@
...
+    return extensions(
+        std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>{extension_list.begin(),
+                                                                                         extension_list.end()});
+}
+
+/**
+ * @brief A helper allowing extensions(initializer_list{"path"}) instead of explicit std::variant.
+ */
+template <class T, class = std::enable_if_t<std::is_constructible<std::filesystem::path, const T&>::value>>
+inline std::pair<std::string, ov::Any> extensions(std::initializer_list<T> paths) {
```
Wovchena: Why would someone need an explicit `initializer_list`? `vector<std::filesystem::path>` overload already handles everything this overload can handle

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +735,16 @@ def test_dynamic_split_fuse_for_speculative_decoding():
...
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
+    scheduler_config = SchedulerConfig()
+
+    properties = {"extensions": ["fake_path"]}
+
+    with pytest.raises(RuntimeError) as exc_info:
+        ContinuousBatchingPipeline(models_path, scheduler_config, "CPU", properties)
+    assert "Cannot find entry point to the extension library" in str(exc_info.value)
```
rkazants: just have only negative tests. Please cover normal flow by tests as well.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,55 @@
...
+}
+
+/**
+ * @brief A helper allowing extensions({extensions}) instead of explicit std::variant.
+ */
+inline std::pair<std::string, ov::Any> extensions(const std::vector<std::shared_ptr<ov::Extension>>& extension_list) {
+    return extensions(
+        std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>{extension_list.begin(),
+                                                                                         extension_list.end()});
+}
```
rkazants, line 56: Did you check this path? Also, please check it in tests

src/cpp/src/utils.cpp
```diff
@@ -718,6 +718,22 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
     return {properties, attention_backend};
 };
 
+void add_extensions_to_core(ov::AnyMap& properties) {
```
rkazants:
```suggestion
void add_extensions_to_core(const ov::AnyMap& properties) {
```

src/cpp/src/utils.hpp
```diff
@@ -286,6 +287,23 @@ bool explicitly_requires_paged_attention(const ov::AnyMap& properties, bool is_n
...
+ *   - a std::filesystem::path pointing to an extension library file, or
+ *   - a std::shared_ptr<ov::Extension> representing an already constructed OpenVINO extension.
+ *
+ * Each element in this vector is added to the global OpenVINO core instance as a custom extension. After processing,
+ * the "extensions" key is removed from the @p properties map. This function is used to dynamically add custom
+ * extensions to the OpenVINO core at runtime.
+ *
+ * @param properties Properties map that may contain the "extensions" key with a vector of extension specifications.
+ */
+void add_extensions_to_core(ov::AnyMap& properties);
```
rkazants:
```suggestion
void add_extensions_to_core(const ov::AnyMap& properties);
```

src/cpp/include/openvino/genai/add_extension.hpp
rkazants:
```suggestion
inline std::pair<std::string, ov::Any> extensions(const std::initializer_list<T>& paths) {
```

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,55 @@
...
+std::pair<std::string, ov::Any> extensions(T&& vector) {
+    return {"extensions", ov::Any(std::forward<T>(vector))};
+}
+
+/**
+ * @brief A helper allowing extensions({"path"}) instead of explicit std::variant.
+ */
+inline std::pair<std::string, ov::Any> extensions(const std::vector<std::filesystem::path>& paths) {
+    return extensions(
+        std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>{paths.begin(), paths.end()});
```
rkazants, line 46: afaik, it is bad to have definition in header file. This is bad for dynamic linking on MacOS systems, for example.
Also, it is not good to use `inline` in header files. Besides, the method name `extensions` is quite bad without reflection action

src/cpp/include/openvino/genai/add_extension.hpp
Wovchena: `initializer_list`s should always be passed by value because it's just a pair of a pointer and size

src/cpp/include/openvino/genai/extensions.hpp
Wovchena: Need more info regarding linking. This is header only implementation so no linking should happen.

`inline` allows multiple definitions. This is exactly the case for header only functions because the file may be included multiple times

src/cpp/src/utils.cpp
sunxiaoxia2022: `properties` will be changed in this function, so shouldn't add `const`.

src/cpp/include/openvino/genai/extensions.hpp
sunxiaoxia2022: Path is checked in ov::add_extension() api. So no checking here.

src/cpp/src/utils.hpp
sunxiaoxia2022: properties will be changed in this function, so no need to add const.

src/cpp/src/utils.cpp
rkazants: that's bad, the method says about changing `core` but it also changes `properties`.

src/cpp/include/openvino/genai/add_extension.hpp
sunxiaoxia2022: Just add a case extensions({"path1", "path2"}). But this case may be no needed. I removed it.

Wovchena: Doesn't `vector<std::filesystem::path>` overload handle that?

src/cpp/include/openvino/genai/extensions.hpp
rkazants: If application will include this method, it will have one symbol for this method inside application binary and methods inside ov_genai library will have another symbol in the binary. It is a sort of ABI incompatibility issue.

Wovchena: That's what `inline` is supposed to handle

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,43 @@
+// Copyright (C) 2023-2026 Intel Corporation
```
rkazants, line 1: header is called as `add_extension.hpp` this is public API but logically it should be named as `extension.hpp`

src/cpp/include/openvino/genai/extensions.hpp
rkazants: Did you check that it works on MacOS? Now the testing is quite limited in GHA so we even cannot check it. Still not believable because how `inline` helps to inject definition of this method into the library.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,43 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+#include <type_traits>
+#include <utility>
+
+#include "openvino/runtime/core.hpp"
+
```
rkazants, line 15: Also per our agreement, we should have changes in pipeline API to have property with extenstions to be registered. Here I see that the API is different.

src/cpp/include/openvino/genai/extensions.hpp
Wovchena: > Did you check that it works on MacOS?

I asked that first:

> Need more info regarding linking

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: Add a normal test.

src/cpp/include/openvino/genai/extensions.hpp
sunxiaoxia2022: Ok, updated.

src/cpp/include/openvino/genai/extensions.hpp
sunxiaoxia2022: Support "std::filesystem::path" and "std::shared_ptr<ov::Extension>" two type parameters which are supported from add_extenstion() in ov::core.

src/cpp/src/utils.cpp
sunxiaoxia2022: Split `add_extensions_to_core` into two functions.

src/cpp/include/openvino/genai/add_extension.hpp
sunxiaoxia2022: Can not support: `extensions({"path_extension1", "path_extension2"})`
This can work: `extensions(std::vector<std::filesystem::path>{"path_extension1", "path_extension2"})`

tests/cpp/test_add_extenstions.cpp
```diff
@@ -0,0 +1,42 @@
...
+    EXPECT_EQ(extract_extensions(properties), pathExtentions);
+}
+
+TEST(TestAddExtensions, test_add_extensions_to_core) {
+    std::string exe_path = "";
+#ifdef __linux__
+    exe_path = "../openvino_genai/libopenvino_tokenizers.so";
+#elif defined(_WIN32)
+    exe_path = "../openvino_genai/libopenvino_tokenizers.dll";
+#endif
```
Wovchena: Mac is missing. Is it intentional?

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +735,29 @@ def test_dynamic_split_fuse_for_speculative_decoding():
...
+    scheduler_config = SchedulerConfig()
+
+    if sys.platform == "win32":
+        ext_path = "libopenvino_tokenizers.so"
+    elif sys.platform == "darwin":
+        ext_path = "libopenvino_tokenizers.dylib"
+    elif sys.platform == "linux":
+        ext_path = "libopenvino_tokenizers.so"
+    else:
+        ext_path = ""
```
Wovchena: Make `else` throw. It's strange if there's something else

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +735,29 @@ def test_dynamic_split_fuse_for_speculative_decoding():
...
+    models_path = download_and_convert_model(model_id).models_path
+
+    scheduler_config = SchedulerConfig()
+
+    if sys.platform == "win32":
+        ext_path = "libopenvino_tokenizers.so"
+    elif sys.platform == "darwin":
+        ext_path = "libopenvino_tokenizers.dylib"
+    elif sys.platform == "linux":
+        ext_path = "libopenvino_tokenizers.so"
```
Wovchena: Add positive test version to tests/python_tests/test_llm_pipeline.py and tests/python_tests/test_vlm_pipeline.py

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +735,29 @@ def test_dynamic_split_fuse_for_speculative_decoding():
...
+    else:
+        ext_path = ""
+
+    if ext_path != "":
+        properties_linux = {"extensions": [ext_path]}
+        ContinuousBatchingPipeline(models_path, scheduler_config, "CPU", properties_linux)
```
Wovchena:
```suggestion
        ContinuousBatchingPipeline(models_path, scheduler_config, "CPU", extensions=[ext_path])
```
"linux" in var name was strange anyway

tests/python_tests/test_continuous_batching.py
Wovchena: How the paths to the files are actually found? If it works just by luck, follow C++'s approach in `ov_tokenizers_module_path()`

tests/cpp/test_add_extenstions.cpp
sunxiaoxia2022: Updated

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: Updated

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: Updated.

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: Updated.

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: Updated.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace genai {
+
+using PathExtensions = std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>;
+
+/**
+ * @brief enable extensions property serves for users to experiment with models and operators that are not officially
+ * supported.
+ *  And create LLMPipeline instance with this config.
+ */
+static constexpr ov::Property<PathExtensions> extension{"extensions"};
```
Wovchena:
```suggestion
static constexpr ov::Property<PathExtensions> extensions{"extensions"};
```
I suspect it won't compile with this patch if you try calling `extensions()` in one of samples. If this is the case, the Property should be removed. If it's not the case, the template overload should be removed

sunxiaoxia2022: Ok, removed it.

src/cpp/src/utils.cpp
```diff
@@ -718,6 +718,26 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
...
+
+void add_extensions_to_core(const PathExtensions& extensions) {
+    for (const auto& extension : extensions) {
+        if (std::holds_alternative<std::filesystem::path>(extension)) {
+            singleton_core().add_extension(std::get<std::filesystem::path>(extension));
+        } else {
+            singleton_core().add_extension(std::get<std::shared_ptr<ov::Extension>>(extension));
+        }
+    }
+}
```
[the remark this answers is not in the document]
sunxiaoxia2022: thread synchronization is done in ov::add_extension()

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1925,3 +1925,17 @@ def test_cdpruner_continuous_batching(
...
 
     assert results[0].texts[0].strip() != "", "Result should not be empty"
+
+
+def test_vlm_pipeline_add_extension():
+    models_path = _get_ov_model(MODEL_IDS[0])
+
+    if openvino_tokenizers._ext_path.exists():
+        properties = {"extensions": [str(openvino_tokenizers._ext_path)]}
+        VLMPipeline(models_path, "CPU", config=properties)
```
Wovchena:
```suggestion
    properties = {"extensions": [str(openvino_tokenizers._ext_path)]}
    VLMPipeline(models_path, "CPU", config=properties)
```
If `openvino_tokenizers._ext_path` doesn't exist, the set up is incorrect so it should be an error

tests/python_tests/test_llm_pipeline.py
```diff
@@ -905,3 +906,18 @@ def test_pipelines_generate_with_streaming(
...
         mock_streamer.assert_called()
+
+
+def test_llm_pipeline_add_extension():
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
+    if _ext_path.exists():
+        properties = {"extensions": [str(_ext_path)]}
+        ov_genai.LLMPipeline(models_path, "CPU", **properties)
```
Wovchena:
```suggestion
    properties = {"extensions": [str(_ext_path)]}
    ov_genai.LLMPipeline(models_path, "CPU", **properties)
```

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +736,20 @@ def test_dynamic_split_fuse_for_speculative_decoding():
...
+
+def test_continuous_batching_add_extension():
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
+    scheduler_config = SchedulerConfig()
+
+    if _ext_path.exists():
+        properties = {"extensions": [str(_ext_path)]}
+        ContinuousBatchingPipeline(models_path, scheduler_config, "CPU", properties)
```
Wovchena:
```suggestion
    properties = {"extensions": [str(_ext_path)]}
    ContinuousBatchingPipeline(models_path, scheduler_config, "CPU", properties)
```

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,30 @@
...
+    // Use intentionally non-existent, platform-agnostic extension paths to trigger error handling.
+    ov::AnyMap properties2 = {ov::genai::extensions(
+        std::vector<std::filesystem::path>{"non_existent_extension1", "non_existent_extension2"})};
+    ov::AnyMap properties3 = {ov::genai::extensions(std::vector<std::shared_ptr<ov::Extension>>{})};
+
+    auto extensions2 = extract_extensions(properties2);
+    auto extensions3 = extract_extensions(properties3);
+    EXPECT_THROW(add_extensions_to_core(extensions2), ov::Exception);
+    EXPECT_NO_THROW(add_extensions_to_core(extensions3));
+}
```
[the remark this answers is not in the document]
Wovchena: You should be able to test `nullptr`s

tests/python_tests/test_vlm_pipeline.py
sunxiaoxia2022: Ok, updated.

tests/python_tests/test_llm_pipeline.py
sunxiaoxia2022: done.

tests/python_tests/test_continuous_batching.py
sunxiaoxia2022: done.

tests/cpp/test_add_extensions.cpp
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/cpp/src/utils.hpp
```diff
@@ -103,6 +103,7 @@ void read_anymap_param(const ov::AnyMap& config_map, const std::string& name, T&
 const std::string STREAMER_ARG_NAME = "streamer";
 const std::string CONFIG_ARG_NAME = "generation_config";
 const std::string DRAFT_MODEL_ARG_NAME = "draft_model";
+const std::string EXTENSIONS_ARG_NAME = "extensions";
```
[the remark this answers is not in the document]
sunxiaoxia2022: not change the name for consistency

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -92,7 +92,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
     auto is_prompt_lookup_enabled = extract_prompt_lookup_from_config(properties_without_draft_model);
     auto eagle_rt_info = extract_eagle3_mode_from_config(draft_model_desr.properties, models_path);
 
-    auto model = utils::read_model(models_path, properties);
+    utils::add_extensions_to_core(properties_without_draft_model);
+    auto model = utils::read_model(models_path, properties_without_draft_model);
     auto [properties_without_draft_model_without_gguf, enable_save_ov_model] = utils::extract_gguf_properties(properties_without_draft_model);
     properties_without_draft_model_without_gguf[ov::cache_model_path.name()] = models_path;
```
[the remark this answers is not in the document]
sunxiaoxia2022: Done.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * @endcode
+ */
+inline std::pair<std::string, ov::Any> extensions(const std::vector<std::filesystem::path>& paths) {
+    return extensions(
+        std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>{paths.begin(), paths.end()});
+}
+
+/**
+ * @brief A helper allowing extensions({extensions}) instead of explicit std::variant.
+ */
```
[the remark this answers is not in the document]
sunxiaoxia2022: Updated.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <variant>
+
+#include "openvino/runtime/core.hpp"
+
+namespace ov {
+namespace genai {
+
+using PathExtensions = std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>;
+
+/**
```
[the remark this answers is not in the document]
sunxiaoxia2022: Updated.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+using namespace ov::genai::utils;
+
+TEST(TestAddExtensions, test_extract_extensions) {
+    ov::AnyMap properties1 = {
+        ov::genai::extensions(std::vector<std::filesystem::path>{"path_extension1", "path_extension2"})};
+    ov::AnyMap properties2 = {
+        ov::genai::extensions(std::vector<std::shared_ptr<ov::Extension>>{nullptr, nullptr})};
+    ov::genai::PathExtensions pathExtensions1{"path_extension1", "path_extension2"};
+    ov::genai::PathExtensions pathExtensions2{nullptr, nullptr};
```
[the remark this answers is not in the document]
sunxiaoxia2022: nullptr should be reserved.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,58 @@
...
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "openvino/runtime/core.hpp"
+
+namespace ov {
+namespace genai {
+
+using PathExtensions = std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>;
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

src/cpp/src/utils.cpp
```diff
@@ -791,6 +791,26 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
...
+
+void add_extensions_to_core(const PathExtensions& extensions) {
+    for (const auto& extension : extensions) {
+        if (std::holds_alternative<std::filesystem::path>(extension)) {
+            singleton_core().add_extension(std::get<std::filesystem::path>(extension));
+        } else {
+            singleton_core().add_extension(std::get<std::shared_ptr<ov::Extension>>(extension));
+        }
+    }
```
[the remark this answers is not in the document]
sunxiaoxia2022: singleton_core will be changed.

src/cpp/src/llm/pipeline.cpp
```diff
@@ -204,6 +204,8 @@ ov::genai::LLMPipeline::LLMPipeline(
 
     bool is_npu_requested = ov::genai::utils::is_npu_requested(device, user_properties);
     auto [properties, attention_backend] = utils::extract_attention_backend(user_properties, is_npu_requested);
+    auto extensions = utils::extract_extensions(properties);
+    utils::add_extensions_to_core(extensions);
```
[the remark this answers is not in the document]
sunxiaoxia2022: Updated

src/cpp/src/utils.hpp
```diff
@@ -290,6 +292,22 @@ bool explicitly_requires_paged_attention(const ov::AnyMap& properties, bool is_n
...
+ * @brief Extracts the "extensions" key from the provided properties map and returns the corresponding
+ * list of extensions.
+ *
+ * The "extensions" entry, if present, is expected to be an ov::Any containing a
+ * std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>, where each element is either:
+ *   - a std::filesystem::path pointing to an extension library file, or
+ *   - a std::shared_ptr<ov::Extension> representing an already constructed OpenVINO extension.
+ *
+ * @param properties Properties map that may contain the "extensions" key with a vector of extension specifications.
+ * @return A ExtensionList object representing the extracted extensions.
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,34 @@
...
+TEST(TestAddExtensions, test_extract_extensions) {
+    ov::AnyMap properties1 = {
+        ov::genai::extensions(std::vector<std::filesystem::path>{"path_extension1", "path_extension2"})};
+    ov::AnyMap properties2 = {
+        ov::genai::extensions(std::vector<std::shared_ptr<ov::Extension>>{nullptr, nullptr})};
+    ov::genai::ExtensionList ExtensionList1{"path_extension1", "path_extension2"};
+    ov::genai::ExtensionList ExtensionList2{nullptr, nullptr};
+
+    EXPECT_EQ(extract_extensions(properties1), ExtensionList1);
+    EXPECT_EQ(extract_extensions(properties2), ExtensionList2);
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -100,9 +101,11 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(
     auto draft_model_desr = utils::extract_draft_model_from_config(properties_without_draft_model);
     auto is_prompt_lookup_enabled = extract_prompt_lookup_from_config(properties_without_draft_model);
     auto eagle_rt_info = utils::eagle3::extract_eagle3_info_from_config(draft_model_desr.properties, models_path);
-    auto model = utils::read_model(models_path, properties_without_draft_model);
+
+    utils::extract_extensions_to_core(properties_without_draft_model);
     auto [properties_without_draft_model_without_gguf, enable_save_ov_model] = utils::extract_gguf_properties(properties_without_draft_model);
     properties_without_draft_model_without_gguf[ov::cache_model_path.name()] = models_path;
+    auto model = utils::read_model(models_path, properties_without_draft_model_without_gguf);
```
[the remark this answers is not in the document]
sunxiaoxia2022: Updated

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -51,9 +51,10 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
     auto is_prompt_lookup_enabled = extract_prompt_lookup_from_config(properties_without_draft_model);
     auto eagle_rt_info = utils::eagle3::extract_eagle3_info_from_config(draft_model_desr.properties, models_path);
 
-    auto model = utils::read_model(models_path, properties);
+    utils::extract_extensions_to_core(properties_without_draft_model);
     auto [properties_without_draft_model_without_gguf, enable_save_ov_model] = utils::extract_gguf_properties(properties_without_draft_model);
     properties_without_draft_model_without_gguf[ov::cache_model_path.name()] = models_path;
+    auto model = utils::read_model(models_path, properties_without_draft_model_without_gguf);
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -2431,6 +2431,18 @@ def test_cdpruner_continuous_batching_chat_history(
     assert baseline3 == pruned3, f"Turn 3 mismatch: baseline='{baseline3}', pruned='{pruned3}'"
```
rkazants, line 2441: please add test with real extensions. For this, you need to update test cmake file and build shared library with extensions. For example, in IR you can change operation type to some custom op type for which you create extension in shared library.

sunxiaoxia2022: Ok, updated. Please take a look, thank you!

src/cpp/src/utils.cpp
```diff
@@ -800,6 +800,27 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
...
+
+void extract_extensions_to_core(ov::AnyMap& properties) {
+    ExtensionList extensions = extract_extensions(properties);
+    for (const auto& extension : extensions) {
+        if (std::holds_alternative<std::filesystem::path>(extension)) {
+            singleton_core().add_extension(std::get<std::filesystem::path>(extension));
+        } else {
+            singleton_core().add_extension(std::get<std::shared_ptr<ov::Extension>>(extension));
+        }
+    }
```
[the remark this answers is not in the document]
sunxiaoxia2022: singleton_core() will be removed later. So not change here.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,58 @@
...
+namespace ov {
+namespace genai {
+
+using ExtensionList = std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>;
+
+/**
+ * @brief Wrap paths and Extensions into an ov::AnyMap-compatible pair that can be passed to pipeline constructors
+ * like LLMPipeline or VLMPipeline, ensuring that the extensions are loaded into the OpenVINO Core instance before
+ * model loading.
+ */
```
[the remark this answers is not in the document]
sunxiaoxia2022: Updated.

tests/python_tests/test_vlm_pipeline.py
peterchen-intel: @rkazants WIP to add C++ extension library and create LLM to test it.  Do you have any reference for creating a simple text generation LLM with custom OP type?

tests/python_tests/utils/ov_genai_pipelines.py
```diff
@@ -356,6 +357,30 @@ def should_skip_npuw_tests():
...
+def get_custom_add_extension_path() -> Path | None:
+    env_path = os.getenv("OPENVINO_CUSTOM_ADD_EXTENSION_PATH")
+    if env_path:
+        candidate = Path(env_path)
+        if candidate.exists():
+            return candidate
+
+    repo_root = Path(__file__).resolve().parents[3]
+    print("repo_root:", repo_root)
+    local_candidates = [
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed these codes

tests/cpp/custom_op/custom_add.cpp
```diff
@@ -0,0 +1,65 @@
...
+        {
+            ptr_out[i] = ptr_in[i] + pBias[0];
+        }
+    }
+    else
+    {
+        std::cout << "Error: Not implemented for data type: " << in.get_element_type() << std::endl;
+    }
+
+    return true;
```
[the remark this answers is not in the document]
sunxiaoxia2022: remove this file

tests/cpp/custom_op/ov_extension.cpp
```diff
@@ -0,0 +1,11 @@
+#include <openvino/core/extension.hpp>
+#include <openvino/core/op_extension.hpp>
+#include <openvino/frontend/extension.hpp>
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed this file.

tests/cpp/custom_op/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+set(SRC custom_add.cpp ov_extension.cpp)
+
+add_library(${TARGET_NAME} MODULE ${SRC})
+
+set_target_properties(${TARGET_NAME} PROPERTIES
+    PREFIX ""
+    OUTPUT_NAME "openvino_custom_add_extension"
+    LIBRARY_OUTPUT_DIRECTORY "$<TARGET_FILE_DIR:${TEST_TARGET_NAME}>"
+    RUNTIME_OUTPUT_DIRECTORY "$<TARGET_FILE_DIR:${TEST_TARGET_NAME}>"
+)
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -2435,3 +2436,22 @@ def test_vlm_prompt_lookup_functionality(cat_tensor):
...
+def test_vlm_pipeline_add_extension_path():
+    models_path = _get_ov_model(MODEL_IDS[0])
+
+    custom_extension_path = get_custom_add_extension_path()
+    print("custom_extension_path: ", custom_extension_path)
+    if custom_extension_path is not None:
+        properties = {"extensions": [str(custom_extension_path)]}
+    else:
+        assert "Cannot find custom_extension_path"
+    VLMPipeline(models_path, "CPU", config=properties)
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -2435,3 +2436,22 @@ def test_vlm_prompt_lookup_functionality(cat_tensor):
...
         "Result should be the same when prompt_lookup is enabled and disabled."
     )
+
+
+def test_vlm_pipeline_add_extension_path():
+    models_path = _get_ov_model(MODEL_IDS[0])
+
+    custom_extension_path = get_custom_add_extension_path()
+    print("custom_extension_path: ", custom_extension_path)
+    if custom_extension_path is not None:
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,45 @@
...
+using namespace ov::genai::utils;
+
+static std::filesystem::path get_custom_extension_library_name() {
+#if defined(_WIN32)
+    return "openvino_custom_add_extension.dll";
+#elif defined(__APPLE__)
+    return "openvino_custom_add_extension.dylib";
+#else
+    return "openvino_custom_add_extension.so";
+#endif
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/custom_op/custom_add.hpp
```diff
@@ -0,0 +1,24 @@
+#include <openvino/core/extension.hpp>
+#include <openvino/op/op.hpp>
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/custom_op/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+        execute_process(
+            COMMAND ${Python3_EXECUTABLE} -c "from openvino.utils import get_cmake_path; print(get_cmake_path(), end='')"
+            OUTPUT_VARIABLE OpenVINO_DIR_PY
+            ERROR_QUIET)
+    endif()
+endif()
+
+message(status "OpenVINO_DIR_PY=${OpenVINO_DIR_PY}")
+find_package(OpenVINO REQUIRED PATHS "${OpenVINO_DIR_PY}")
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/custom_op/custom_add.hpp
```diff
@@ -0,0 +1,24 @@
...
+        std::shared_ptr<ov::Node> clone_with_new_inputs(const ov::OutputVector &new_args) const override;
+        bool visit_attributes(ov::AttributeVisitor &visitor) override;
+
+        bool evaluate(ov::TensorVector &outputs, const ov::TensorVector &inputs) const override;
+        bool has_evaluate() const override;
+
+    private:
+        // float _bias = 0;
+    };
+};
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/custom_op/custom_add.cpp
```diff
@@ -0,0 +1,65 @@
...
+        auto *ptr_out = reinterpret_cast<float *>(out.data());
+        for (size_t i = 0; i < total; i++)
+        {
+            ptr_out[i] = ptr_in[i] + pBias[0];
+        }
+    }
+    else
+    {
+        std::cout << "Error: Not implemented for data type: " << in.get_element_type() << std::endl;
+    }
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/python_tests/utils/ov_genai_pipelines.py
```diff
@@ -344,6 +345,30 @@ def generate_and_compare(
...
+        repo_root / "build" / "bin" / "openvino_custom_add_extension.so",
+        repo_root / "build" / "bin" / "openvino_custom_add_extension.dylib",
+        repo_root / "build" / "bin" / "openvino_custom_add_extension.dll",
+        repo_root / "bin" / "openvino_custom_add_extension.so",
+        repo_root / "bin" / "libopenvino_custom_add_extension.so",
+    ]
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -2449,3 +2450,22 @@ def test_vlm_prompt_lookup_functionality(cat_tensor):
...
+        properties = {"extensions": [str(custom_extension_path)]}
+    else:
+        assert "Cannot find custom_extension_path"
+    VLMPipeline(models_path, "CPU", config=properties)
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/custom_op/custom_add.cpp
```diff
@@ -0,0 +1,65 @@
...
+    return true;
+}
+
+bool MyAdd::evaluate(ov::TensorVector &outputs, const ov::TensorVector &inputs) const
+{
+    std::cout << "== MyAdd::evaluate is called." << std::endl;
+    float *inpData = reinterpret_cast<float *>(const_cast<void*>(inputs[0].data()));
+    if (inputs[1].get_element_type() != ov::element::f32)
+        OPENVINO_THROW("Unexpected bias type: " + inputs[1].get_element_type().to_string());
+    float *pBias = reinterpret_cast<float *>(const_cast<void*>(inputs[1].data()));
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/custom_op/custom_add.cpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    // visitor.on_attribute("bias", this->_bias);
+    return true;
+}
+
+bool MyAdd::evaluate(ov::TensorVector &outputs, const ov::TensorVector &inputs) const
+{
+    std::cout << "== MyAdd::evaluate is called." << std::endl;
+    float *inpData = reinterpret_cast<float *>(const_cast<void*>(inputs[0].data()));
+    if (inputs[1].get_element_type() != ov::element::f32)
```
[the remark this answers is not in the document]
sunxiaoxia2022: removed.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "utils.hpp"
+
+using namespace ov::genai::utils;
+
+TEST(TestAddExtensions, test_extract_extensions) {
+    ov::AnyMap properties_path = {ov::genai::extensions(std::vector<std::filesystem::path>{"non_existent_path"})};
+    auto op = std::make_shared<ov::frontend::OpExtension<>>("Relu", "MyRelu");
+    ov::AnyMap properties_op = {ov::genai::extensions(std::vector<std::shared_ptr<ov::Extension>>{op})};
+    ov::genai::ExtensionList extensionList_path{"non_existent_path"};
+    ov::genai::ExtensionList extensionList_op{op};
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +737,23 @@ def test_dynamic_split_fuse_for_speculative_decoding():
 
 def test_dynamic_split_fuse_for_eagle3():
     compare_results_for_dynamic_split_fuse_config("Qwen/Qwen3-1.7B", "AngelSlim/Qwen3-1.7B_eagle3")
+
+
+def test_continuous_batching_add_extension_custom_op():
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

tests/python_tests/test_continuous_batching.py
```diff
@@ -735,3 +737,23 @@ def test_dynamic_split_fuse_for_speculative_decoding():
...
+    scheduler_config = SchedulerConfig()
+
+    properties = {"extensions": [OpExtension("Relu", "MyRelu")]}
+    ContinuousBatchingPipeline(models_path, scheduler_config, "CPU", properties)
+
+
+def test_continuous_batching_add_extension():
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

tests/python_tests/test_llm_pipeline.py
```diff
@@ -876,3 +878,19 @@ def test_pipelines_generate_with_streaming(
         mock_streamer.assert_not_called()
     else:
         mock_streamer.assert_called()
+
+
+def test_llm_pipeline_add_extension_custom_op():
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

tests/python_tests/test_llm_pipeline.py
```diff
@@ -876,3 +878,19 @@ def test_pipelines_generate_with_streaming(
...
+    properties = {"extensions": [OpExtension("Relu", "MyRelu")]}
+    ov_genai.LLMPipeline(models_path, "CPU", **properties)
+
+
+def test_llm_pipeline_add_extension():
+    model_id = "katuni4ka/tiny-random-phi3"
+    models_path = download_and_convert_model(model_id).models_path
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -200,6 +205,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(
     auto draft_model_desr = utils::extract_draft_model_from_config(properties_without_draft_model);
     auto is_prompt_lookup_enabled = extract_prompt_lookup_from_config(properties_without_draft_model);
     auto model_pair = utils::get_model_weights_pair(models_map, "language");
+
+    utils::extract_extensions_to_core(properties_without_draft_model);
     auto model = utils::singleton_core().read_model(model_pair.first, model_pair.second);
```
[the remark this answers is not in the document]
sunxiaoxia2022: not this PR change. do not update.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+TEST(TestAddExtensions, test_extract_extensions_to_core) {
+    // Use intentionally non-existent, platform-agnostic extension paths to trigger error handling.
+    ov::AnyMap properties_path = {ov::genai::extensions(std::vector<std::filesystem::path>{"non_existent_path"})};
+    ov::AnyMap properties_op = {ov::genai::extensions(
+        std::vector<std::shared_ptr<ov::Extension>>{std::make_shared<ov::frontend::OpExtension<>>("Relu", "MyRelu")})};
+
+    EXPECT_THROW(extract_extensions_to_core(properties_path), ov::Exception);
+    EXPECT_NO_THROW(extract_extensions_to_core(properties_op));
+}
```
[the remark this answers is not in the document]
sunxiaoxia2022: no needed.

tests/cpp/custom_op/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+        execute_process(
+            COMMAND ${Python3_EXECUTABLE} -c "from openvino.utils import get_cmake_path; print(get_cmake_path(), end='')"
+            OUTPUT_VARIABLE OpenVINO_DIR_PY
+            ERROR_QUIET)
+    endif()
+endif()
+
+message(STATUS "OpenVINO_DIR_PY=${OpenVINO_DIR_PY}")
+find_package(OpenVINO REQUIRED PATHS "${OpenVINO_DIR_PY}")
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -100,6 +101,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(
     auto draft_model_desr = utils::extract_draft_model_from_config(properties_without_draft_model);
     auto is_prompt_lookup_enabled = extract_prompt_lookup_from_config(properties_without_draft_model);
     auto eagle_rt_info = utils::eagle3::extract_eagle3_info_from_config(draft_model_desr.properties, models_path);
+
+    utils::extract_extensions_to_core(properties_without_draft_model);
     auto model = utils::read_model(models_path, properties_without_draft_model);
     auto [properties_without_draft_model_without_gguf, enable_save_ov_model] = utils::extract_gguf_properties(properties_without_draft_model);
```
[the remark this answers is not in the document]
sunxiaoxia2022: These codes are not change from this PR. So no need to update.

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -51,7 +51,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
     auto is_prompt_lookup_enabled = extract_prompt_lookup_from_config(properties_without_draft_model);
     auto eagle_rt_info = utils::eagle3::extract_eagle3_info_from_config(draft_model_desr.properties, models_path);
 
-    auto model = utils::read_model(models_path, properties);
+    utils::extract_extensions_to_core(properties_without_draft_model);
+    auto model = utils::read_model(models_path, properties_without_draft_model);
     auto [properties_without_draft_model_without_gguf, enable_save_ov_model] = utils::extract_gguf_properties(properties_without_draft_model);
```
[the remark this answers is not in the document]
sunxiaoxia2022: These codes are not change from this PR.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,58 @@
...
+#pragma once
+#include <filesystem>
+#include <memory>
+#include <string>
+#include <type_traits>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "openvino/runtime/core.hpp"
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/cpp/src/utils.cpp
```diff
@@ -800,6 +800,27 @@ std::pair<ov::AnyMap, std::string> extract_attention_backend(const ov::AnyMap& e
     return {properties, attention_backend};
 };
 
+ExtensionList extract_extensions(ov::AnyMap& properties) {
+    auto it = properties.find(EXTENSIONS_ARG_NAME);
+    ExtensionList extensions = {};
+    if (it != properties.end()) {
+        extensions = it->second.as<std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>>();
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/custom_op/custom_add.hpp
```diff
@@ -0,0 +1,24 @@
...
+#pragma once
+
+#include <openvino/core/extension.hpp>
+#include <openvino/op/op.hpp>
+
+namespace TemplateExtension {
+class MyAdd : public ov::op::Op {
+public:
+    // OP name.
+    OPENVINO_OP("MyAdd");
```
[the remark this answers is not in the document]
sunxiaoxia2022: no needed, OPENVINO_OP() requires only a single argument.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -876,3 +915,151 @@ def test_pipelines_generate_with_streaming(
...
+        raise FileNotFoundError(f"Build directory was not found: {build_dir}")
+
+    for file_name in file_names:
+        matches = sorted(build_dir.rglob(file_name))
+        if matches:
+            return matches[0]
+
+    raise FileNotFoundError(
+        f"Could not find compiled custom extension '{extension_name}'. "
+        f"Searched EXTENSION_LIB_PATH='{os.getenv('EXTENSION_LIB_PATH')}' and build directory '{build_dir}'. "
```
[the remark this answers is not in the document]
sunxiaoxia2022: using EXTENSION_LIB_PATH in CI test, using fix path in local test. BUILD_DIR is also defined in CI test which should not be used in local test. So no need to change here.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -876,3 +915,151 @@ def test_pipelines_generate_with_streaming(
...
+    xml_text = xml_text[: match.start(1)] + updated_tag + xml_text[match.end(1) :]
+    ir_xml_path.write_text(xml_text, encoding="utf-8")
+
+
+def get_extension_model(model_path: str, temp_dir: Path) -> Path:
+    source_path = Path(model_path)
+    extension_path = temp_dir / f"{source_path.name}_extension"
+    ir_xml_path = extension_path / "openvino_model.xml"
+    if ir_xml_path.exists():
+        return extension_path
```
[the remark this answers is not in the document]
sunxiaoxia2022: I think no need to add these logic, because other files were copied over.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -876,3 +915,151 @@ def test_pipelines_generate_with_streaming(
...
+    )
+
+
+@pytest.mark.parametrize("llm_model", ["katuni4ka/tiny-random-phi3"], indirect=True)
+def test_llm_pipeline_add_extension(llm_model: OVConvertedModelSchema) -> None:
+    properties = {"extensions": [str(_ext_path)]}
+    ov_genai.LLMPipeline(llm_model.models_path, "CPU", **properties)
+
+    properties = {"extensions": [OpExtension("Relu", "MyRelu")]}
+    ov_genai.LLMPipeline(llm_model.models_path, "CPU", **properties)
```
rkazants: please also call and check generation results

rkazants: please check that IR really contains "MyRelu" op. After that, please call generate and check generation results.

as-suvorov:
@sunxiaoxia2022 Regarding https://github.com/openvinotoolkit/openvino.genai/actions/runs/23183510718/job/67364623954?pr=2952 fail.
I can find such fail in the current PR only. I also see modifications done to workflow and cmake event after some reverts.
I still assume that workflow fails is caused by PR changes. I don't have a quick fix from my side.
The proposal is to continue investigation. Consider reverting changes to workflow or/and cmake, you can maybe open a test PR for investigation purposes.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -2457,3 +2458,13 @@ def test_vlm_prompt_lookup_functionality(cat_tensor):
...
+
+
+def test_vlm_pipeline_add_extension():
+    models_path = _get_ov_model(MODEL_IDS[0])
+
+    properties = {"extensions": [str(openvino_tokenizers._ext_path)]}
+    VLMPipeline(models_path, "CPU", config=properties)
+
+    properties = {"extensions": [OpExtension("Relu", "MyRelu")]}
+    VLMPipeline(models_path, "CPU", config=properties)
```
rkazants: Please also check results of generation. Besides, it needs to check that IR contains "MyRelu" operation that is passed to 'VLMPipeline'

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,112 @@
...
+        }
+        return std::make_shared<CustomAdd>(new_args.at(0), new_args.at(1));
+    }
+
+    bool visit_attributes(ov::AttributeVisitor&) override {
+        return true;
+    }
+
+    bool evaluate(ov::TensorVector& outputs, const ov::TensorVector& inputs) const override {
+        std::cout << "custom add evaluate\n";
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,112 @@
...
+    CustomAdd(const ov::Output<ov::Node>& lhs, const ov::Output<ov::Node>& rhs) : ov::op::Op({lhs, rhs}) {
+        constructor_validate_and_infer_types();
+    }
+
+    void validate_and_infer_types() override {
+        const ov::element::Type input_type_0 = get_input_element_type(0);
+        const ov::element::Type input_type_1 = get_input_element_type(1);
+        if (!input_type_0.compatible(input_type_1)) {
+            throw std::runtime_error("CustomAdd input element types must match");
+        }
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+        set_output_type(0, input_type_0, get_input_partial_shape(0));
+    }
+
+    std::shared_ptr<ov::Node> clone_with_new_inputs(const ov::OutputVector& new_args) const override {
+        if (new_args.size() != 2) {
+            throw std::runtime_error("CustomAdd expects exactly 2 new inputs");
+        }
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    bool evaluate(ov::TensorVector& outputs, const ov::TensorVector& inputs) const override {
+        std::cout << "custom add evaluate\n";
+        if (inputs.size() != 2) {
+            throw std::runtime_error("CustomAdd expects exactly 2 inputs");
+        }
+        if (outputs.size() != 1) {
+            throw std::runtime_error("CustomAdd expects exactly 1 output");
+        }
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -2,16 +2,18 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
+import shutil
 import pytest
 import torch
 import gc
 import sys
 import os
+import ctypes
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -2,16 +2,18 @@
...
 import json
 import logging
 import numpy as np
 from pathlib import Path
-from typing import Literal, Callable
+from typing import Literal, Callable, Optional
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/custom_op/custom_add.hpp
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/python/py_utils.cpp
```diff
@@ -21,6 +21,7 @@
 #include "openvino/genai/whisper_generation_config.hpp"
 #include "openvino/genai/whisper_pipeline.hpp"
 #include "openvino/genai/rag/text_embedding_pipeline.hpp"
+#include "openvino/runtime/core.hpp"
```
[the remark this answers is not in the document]
sunxiaoxia2022: can not remove.

tests/python_tests/test_llm_pipeline.py
sunxiaoxia2022: Because there is no "Relu" operation in this test model, so changed "Add" to "CustomAdd" in this model, and checked this operation, then compare results of two generations. Please take a look!

tests/python_tests/test_vlm_pipeline.py
sunxiaoxia2022: Updated test_vlm_pipeline_add_extension and test_continuous_batching_add_extension like test_llm_pipeline_add_extension, please take a look.

src/cpp/include/openvino/genai/extensions.hpp
```diff
@@ -0,0 +1,59 @@
...
+namespace ov {
+namespace genai {
+
+using ExtensionList = std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>>;
+
+/**
+ * @brief Wrap paths and Extensions into an ov::AnyMap-compatible pair that can be passed to pipeline constructors
+ * like LLMPipeline or VLMPipeline, ensuring that the extensions are loaded into the OpenVINO Core instance before
+ * model loading.
+ */
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/python_tests/test_llm_pipeline.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
+import shutil
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/python/py_utils.cpp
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/cpp/include/openvino/genai/extensions.hpp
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/python/py_utils.cpp
```diff
@@ -194,6 +195,38 @@ ov::Any py_object_to_any(const py::object& py_obj, std::string property_name) {
                 }
             }
             return parsers;
+        } else if (property_name == "extensions") {
+            auto property_list = py_obj.cast<py::list>();
+            std::vector<std::variant<std::filesystem::path, std::shared_ptr<ov::Extension>>> extensions;
+            py::object pathlib_path = py::module_::import("pathlib").attr("Path");
+            py::object op_extension_ctor = py::module_::import("openvino").attr("OpExtension");
+
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+        if (element_type == ov::element::f32) {
+            return evaluate_typed<float>(outputs[0], inputs[0], inputs[1]);
+        }
+        if (element_type == ov::element::f16) {
+            return evaluate_typed<ov::float16>(outputs[0], inputs[0], inputs[1]);
+        }
+
+        OPENVINO_ASSERT(false, "CustomAdd unsupported element type");
```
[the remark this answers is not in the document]
sunxiaoxia2022: done.

src/python/py_utils.cpp
```diff
@@ -194,6 +196,33 @@ ov::Any py_object_to_any(const py::object& py_obj, std::string property_name) {
...
+                if (py::isinstance<ov::Extension>(item)) {
+                    extensions.push_back(item.cast<std::shared_ptr<ov::Extension>>());
+                    continue;
+                }
+                if (py::isinstance<py::type>(item)) {
+                    try {
+                        auto py_ext = op_extension_ctor(item);
+                        extensions.push_back(py_ext.cast<std::shared_ptr<ov::Extension>>());
+                        continue;
+                    } catch (const py::error_already_set&) {
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

src/cpp/src/continuous_batching/pipeline.cpp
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/cpp/test_add_extensions.cpp
```diff
@@ -0,0 +1,106 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <gtest/gtest.h>
+
+#include <openvino/core/op_extension.hpp>
+#include <openvino/openvino.hpp>
+
+#include "openvino/genai/extensions.hpp"
+#include "utils.hpp"
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

tests/python_tests/test_continuous_batching.py
```diff
@@ -9,6 +9,8 @@
 from pathlib import Path
 from shutil import rmtree
 
+import openvino as ov
+from openvino.frontend import OpExtension
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated.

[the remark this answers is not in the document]
sunxiaoxia2022: updated.

as-suvorov:
Linux Tools Release tests should be fixed by: https://github.com/openvinotoolkit/openvino.genai/pull/3580

peterchen-intel:
https://github.com/openvinotoolkit/openvino.genai/pull/3579 This is also required.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -769,6 +773,7 @@ VLMPipeline::VLMPipeline(
         m_pimpl = std::make_unique<VLMPipelineImpl>(models_dir, device, properties);
     } else {
         auto properties_copy = properties;
+        utils::extract_extensions_to_core(properties_copy);
         auto language_model_path = models_dir / "openvino_language_model.xml";
         auto language_model = utils::singleton_core().read_model(language_model_path, {}, properties_copy);
```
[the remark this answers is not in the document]
sunxiaoxia2022: udpated.

PR #2961, opened 2025-11-04
Readme improvements

MaximProshin:
## Description
Readme improvements and minor fixes

Ticket: NA

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ See [here](https://openvinotoolkit.github.io/openvino_notebooks/?search=Visual-l
 
 </details>
 
-<a id="image-to-text"></a>
+<a id="text-to-image"></a>
 
 ## Performing image generation
 


README.md
```diff
@@ -37,6 +54,8 @@ Library efficiently supports LoRA adapters for Text and Image generation scenari
 
 All scenarios are run on top of OpenVINO Runtime that supports inference on CPU, GPU and NPU. See [here](https://docs.openvino.ai/2025/about-openvino/release-notes-openvino/system-requirements.html) for platform support matrix.
 
+<a id="optimization-methods"></a>
+
 ## Supported Generative AI optimization methods
```
as-suvorov, line 59: md can navigate to headings without additional `<a>` tags: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#section-links

Example:
`- [Supported Generative AI optimization methods](#supported-generative-ai-optimization-methods)`

MaximProshin: @as-suvorov I think both ways are ok. With my approach you can change the headers and don't need to change the related links after.  If you insist that the "headers-based" approach is more convenient, I can apply it.

as-suvorov (approved):

README.md
as-suvorov: There is a good md support in vs code extension. There is autocomplete for headers in the document:
image
Also you can rename header with F2 shortcut, then appropriate links will be updated automatically.
I'm ok with proposed approach as well, so up to you.

PR #2864, opened 2025-10-20
Add threaded callback for image generation pipeline

JohnLeFeng:
## Description
Add a sample for decode intermediate latent in parallel based on heterogeneous image generation pipeline.

PR #2962, opened 2025-11-04
Update README for threaded callback of image generation pipeline

JohnLeFeng:
## Description

Update README for threaded callback of image generation pipeline. 

> Note: following PR of #2864


Wovchena:
build_jenkins

PR #2986, opened 2025-11-10
[NPU] Add `batch_size` support for embedding model

mengweiguo:
## Description

The model `qwen3-embedding-0.6B` failed on `wwb` test on NPU due to dynamic `batch size`. This PR adds `batch_size` option support for this model in `llm-benchmark` and `wwb`.

[CVS-176378](https://jira.devtools.intel.com/browse/CVS-176378)

## Checklist:

### DIFF
--- a/samples/python/image_generation/image2image.py
+++ b/samples/python/image_generation/image2image.py
@@ -26,13 +26,20 @@ def main():
 
     image = read_image(args.image)
 
-    image_tensor = pipe.generate(args.prompt, image,
-        strength=0.8 # controls how initial image is noised after being converted to latent space. `1` means initial image is fully noised
+    def callback(step, num_steps, latent):
+        print(f"Step {step + 1}/{num_steps}")
+        return False
+
+    image_tensor = pipe.generate(
+        args.prompt,
+        image,
+        strength=0.8,
+        callback=callback
     )
 
     image = Image.fromarray(image_tensor.data[0])
     image.save("image.bmp")
 
 
-if '__main__' == __name__:
+if __name__ == '__main__':
     main()
--- a/samples/python/image_generation/inpainting.py
+++ b/samples/python/image_generation/inpainting.py
@@ -28,11 +28,15 @@ def main():
     image = read_image(args.image)
     mask_image = read_image(args.mask)
 
-    image_tensor = pipe.generate(args.prompt, image, mask_image)
+    def callback(step, num_steps, latent):
+        print(f"Step {step + 1}/{num_steps}")
+        return False
+
+    image_tensor = pipe.generate(args.prompt, image, mask_image, callback=callback)
 
     image = Image.fromarray(image_tensor.data[0])
     image.save("image.bmp")
 
 
-if '__main__' == __name__:
+if __name__ == '__main__':
     main()
--- a/samples/python/image_generation/text2image.py
+++ b/samples/python/image_generation/text2image.py
@@ -17,16 +17,21 @@ def main():
     device = 'CPU'  # GPU can be used as well
     pipe = openvino_genai.Text2ImagePipeline(args.model_dir, device)
 
+    def callback(step, num_steps, latent):
+        print(f"Step {step + 1}/{num_steps}")
+        return False
+
     image_tensor = pipe.generate(
         args.prompt,
         width=512,
         height=512,
         num_inference_steps=20,
-        num_images_per_prompt=1)
+        num_images_per_prompt=1,
+        callback=callback)
 
     image = Image.fromarray(image_tensor.data[0])
     image.save("image.bmp")
 
 
-if '__main__' == __name__:
+if __name__ == '__main__':
     main()
--- a/src/cpp/src/image_generation/threaded_callback.hpp
+++ b/src/cpp/src/image_generation/threaded_callback.hpp
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <thread>
+#include <variant>
 
 #include "synchronized_queue.hpp"
 
@@ -33,7 +34,7 @@ public:
             return CallbackStatus::STOP;
         }
 
-        m_squeue.push({step, num_steps, latent});
+        m_squeue.push(std::make_tuple(step, num_steps, latent));
 
         return CallbackStatus::RUNNING;
     }
@@ -44,7 +45,7 @@ public:
         }
 
         m_status = CallbackStatus::STOP;
-        m_squeue.empty();
+        m_squeue.push(std::monostate());
 
         if (m_worker_thread && m_worker_thread->joinable()) {
             m_worker_thread->join();
@@ -58,18 +59,23 @@ public:
 private:
     std::function<bool(size_t, size_t, ov::Tensor&)> m_callback = nullptr;
     std::shared_ptr<std::thread> m_worker_thread = nullptr;
-    SynchronizedQueue<std::tuple<size_t, size_t, ov::Tensor>> m_squeue;
+    SynchronizedQueue<std::variant<std::tuple<size_t, size_t, ov::Tensor>, std::monostate>> m_squeue;
 
     std::atomic<CallbackStatus> m_status = CallbackStatus::RUNNING;
 
     void _worker() {
         while (m_status == CallbackStatus::RUNNING) {
-            // wait for queue pull
-            auto [step, num_steps, latent] = m_squeue.pull();
+            auto item = m_squeue.pull();
 
-            if (m_callback(step, num_steps, latent)) {
-                m_status = CallbackStatus::STOP;
-                m_squeue.empty();
+            if (auto callback_data = std::get_if<std::tuple<size_t, size_t, ov::Tensor>>(&item)) {
+                auto& [step, num_steps, latent] = *callback_data;
+                const auto should_stop = m_callback(step, num_steps, latent);
+                
+                if (should_stop) {
+                    m_status = CallbackStatus::STOP;
+                }
+            } else if (std::get_if<std::monostate>(&item)) {
+                break;
             }
         }
     }
--- a/src/cpp/src/logger.cpp
+++ b/src/cpp/src/logger.cpp
@@ -1,0 +1,164 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "logger.hpp"
+
+#include <cctype>
+#include <chrono>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
+#include <ctime>
+#include <iostream>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace ov::genai {
+
+ov::log::Level get_openvino_env_log_level() {
+    const char* env = std::getenv("OPENVINO_LOG_LEVEL");
+    if (!env) {
+        return ov::log::Level::ERR;
+    }
+    try {
+        std::string env_str(env);
+        size_t idx = 0;
+        const auto env_var_value = std::stoi(env_str, &idx);
+        if (idx != env_str.size()) {
+            return ov::log::Level::ERR;
+        }
+        return static_cast<ov::log::Level>(env_var_value);
+    } catch (...) {
+        return ov::log::Level::ERR;
+    }
+}
+
+ov::log::Level get_cur_log_level() {
+    static ov::log::Level cur_log_level = get_openvino_env_log_level();
+    return cur_log_level;
+}
+
+Logger::Logger() {
+    log_level.store(get_cur_log_level(), std::memory_order_relaxed);
+}
+
+void Logger::do_log(ov::log::Level level, const char* file, int line, const std::string& msg) {
+    write_message(level, file, line, msg);
+}
+
+void Logger::log_format(ov::log::Level level, const char* file, int line, const char* format, ...) {
+    va_list args;
+    va_start(args, format);
+    try {
+        log_format_impl(level, file, line, format, args);
+    } catch (const std::exception& ex) {
+        va_end(args);
+        throw std::runtime_error(std::string{"Logger format error: "} + ex.what());
+    } catch (...) {
+        va_end(args);
+        throw std::runtime_error("Logger format error: unknown exception");
+    }
+    va_end(args);
+}
+
+void Logger::set_log_level(ov::log::Level level) {
+    log_level.store(level, std::memory_order_relaxed);
+}
+
+bool Logger::should_log(ov::log::Level level) const {
+    const auto current_level = log_level.load(std::memory_order_relaxed);
+    return current_level != ov::log::Level::NO && level != ov::log::Level::NO && level <= current_level;
+}
+
+void Logger::write_message(ov::log::Level level, const char* file, int line, const std::string& msg) {
+    std::lock_guard<std::mutex> lock(log_mutex);
+    std::ostream& out = (level == ov::log::Level::ERR) ? std::cerr : std::cout;
+    format_prefix(out, level, file, line);
+    out << msg;
+    if (msg.empty() || msg.back() != '\n') {
+        if (level == ov::log::Level::ERR) {
+            out << std::endl;
+        } else {
+            out << '\n';
+        }
+    }
+}
+
+void Logger::log_format_impl(ov::log::Level level, const char* file, int line, const char* format, va_list args) {
+    std::string formatted = format_from_variadic(format, args);
+    write_message(level, file, line, formatted);
+}
+
+std::string Logger::format_from_variadic(const char* format, va_list args) const {
+    if (!format) {
+        return {};
+    }
+    va_list args_copy;
+    va_copy(args_copy, args);
+    int required = std::vsnprintf(nullptr, 0, format, args_copy);
+    va_end(args_copy);
+    if (required < 0) {
+        throw std::runtime_error("Logger format error: invalid format string or arguments");
+    }
+    if (required == 0) {
+        return {};
+    }
+    std::vector<char> buffer(static_cast<size_t>(required) + 1u);
+    int result = std::vsnprintf(buffer.data(), buffer.size(), format, args);
+    if (result < 0) {
+        throw std::runtime_error("Logger format error: failed to format message");
+    }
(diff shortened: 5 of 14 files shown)


tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
```diff
@@ -189,7 +189,14 @@ def evaluate(self, data_gold, data_prediction):
             with open(prediction, 'rb') as f:
                 prediction_data = np.load(f)
 
-            cos_sim = F.cosine_similarity(torch.from_numpy(gold_data), torch.from_numpy(prediction_data))
+            if gold_data.shape[0] != prediction_data.shape[0]:
+                print(f"Warning: gold_data rows = {gold_data.shape[0]}, prediction_data rows = {prediction_data.shape[0]}")
+
+            min_len = min(gold_data.shape[0], prediction_data.shape[0])
```
rkazants: why it is needed? why are reference batch and predicted batch different?

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -232,7 +232,7 @@ def parse_args():
         "If the base/target model is a local path, gguf-file should be just the filename (e.g., 'model.gguf'). "
         "If the base/target model is a HuggingFace model ID, gguf-file should be a relative path.",
     )
-
+    parser.add_argument('-bs', '--batch_size', type=int, default=None, help='Batch size value')
```
rkazants: please add test on CPU at least in precommit

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -683,6 +683,7 @@ def create_genai_text_embed_model(model_path, device, memory_data_collector, **k
     if padding_side:
         config.padding_side = padding_side
 
+    config.batch_size = kwargs.get("batch_size", config.batch_size)
```
rkazants, line 686: shall it be documented and added to help. Tests?

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -504,6 +504,7 @@ def load_embedding_genai_pipeline(model_dir, device="CPU", ov_config=None, **kwa
     config.max_length = EMBED_DEFAULT_MAX_LENGTH
     config.normalize = kwargs.get("embeds_normalize", False)
     config.pad_to_max_length = True
```
as-suvorov, line 520: @sbalandi I think we need to set it to `false` by defalt and add `pad_to_max_length` argument to wwb.
Most of the cases It needed for NPU device to fix model shapes. For CPU device if input prompt is shorter than `EMBED_DEFAULT_MAX_LENGTH` we just introduce unnesesary additional computations.
For the momemnt `EMBED_DEFAULT_MAX_LENGTH=200` do you know average test dataset prompt lengths?

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
mengweiguo: When `batch_size` is set, it might generate token with the subset of `dataset_documents` as input, like something below.
~~~
   docs_to_embed = dataset_documents[: config.batch_size] if config.batch_size else dataset_documents
~~~
however, it seems the reference dataset is generated with the entire `dataset_documents` in `wwb`

tools/who_what_benchmark/whowhatbench/wwb.py
mengweiguo: Added in `test_rag.py`

tools/llm_bench/llm_bench_utils/ov_utils.py
mengweiguo: Actually this option `batch_size` is already there in benchmark.
~~~
parser.add_argument('-bs', '--batch_size', type=int, default=1, required=False, help='Batch size value')
~~~

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
```diff
@@ -178,8 +180,12 @@ def default_gen_answer(model, tokenizer, passages, **kwargs):
             kwargs = {'padding_side': self.padding_side,
                       'pooling_type': self.pooling_type,
                       'normalize': self.normalize}
-            result = gen_answer_fn(model, self.tokenizer, data[0], **kwargs)
-            passages.append(data[0])
+
+            batch_size = self.batch_size or len(data[0])
+            data_input = data[0][:batch_size]
```
sbalandi: What will be the behavior if the chunk of input data is less than the batch size?

tools/who_what_benchmark/whowhatbench/model_loaders.py
sbalandi: min 85, max 722, avg 307, looks like we can extend , should it be extend to max or to avg ?

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
sbalandi: Sorry, I still don't understand why it's needed. Do you want to make possibility to compare ground truth data, which was generated without batch and output, which was generated with batch ? What is reference dataset ?

mengweiguo: >  Do you want to make possibility to compare ground truth data, which was generated without batch and output, which was generated with batch ? What is reference dataset ?

Yes, it might have the scenario that the reference data is generated without batch and target data is with batch. In wwb CI, it seems that the reference data is pre-generated without batch setting.

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
mengweiguo: Added the check.

tools/who_what_benchmark/whowhatbench/model_loaders.py
as-suvorov: No, I was thinking about switching `pad_to_max_length` to `False` by default. Considering avg is 307 for this dataset, let's skip it for now

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
sbalandi: I meant inside the plugin, there's no point in line `min(batch_size, len(data[0]))`, it will be taken the maximum elements the list can give. But if we set batch 10, but send 8 as input - what will plugin do ?

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
sbalandi: ok, how we could find out the situation when batch size is not applied ? maybe we can send batch size to evaluate() ?
or maybe the data can be regenerated if the point only in a CI ?

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
mengweiguo: A exception will throw if batch and data-size are not match in `text-embedding-pipeline`. I also added an assert check as below.
~~~
+            assert batch_size <= len(data[0]), \
+                f"batch_size ({batch_size}) cannot be greater than data length ({len(data[0])})"
~~~
I don't know if it is redundant.

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
mengweiguo: > ok, how we could find out the situation when batch size is not applied ? maybe we can send batch size to evaluate() ? or maybe the data can be regenerated if the point only in a CI ?

I don't think I can control the CI behavior and what can I do is to adapt to CI requirements. so I take the method "we can send batch size to evaluate()" to inference with proper data-set.

sbalandi: do you speak about genAI CI and tests for wwb ? you can update tests

tests/python_tests/test_rag.py
```diff
@@ -332,6 +333,53 @@ def test_qwen3_embedding(download_and_convert_embeddings_models, dataset_documen
...
+@pytest.mark.parametrize(
+    "config",
+    [
+        TextEmbeddingPipeline.Config(normalize=False, pooling_type=TextEmbeddingPipeline.PoolingType.LAST_TOKEN, padding_side="left", batch_size=1),
+        TextEmbeddingPipeline.Config(normalize=False, pooling_type=TextEmbeddingPipeline.PoolingType.LAST_TOKEN, batch_size=1),
+    ],
+)
+@pytest.mark.precommit
+@pytest.mark.xfail(condition=(sys.platform == "darwin"), reason="Ticket - 174635")
+def test_qwen3_embedding_set_batch(download_and_convert_embeddings_models, dataset_documents, config):
```
as-suvorov: Let's extend already implemented `test_qwen3_embedding` instead of introducing new test

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
as-suvorov: Scenario 1:
1. Reference data generated with batch size 10
2. Predictions are generated with batch size 5
3. We compare reference[:5] and predicted[:5] data and getting 0.9 similarity

Scenario 2:
1. Reference data generated with batch size 5
2. Predictions are generated with batch size 10
3. We compare reference[:5] and predicted[:5] data and getting 0.9 similarity

For scenario 2 predictions generation batch_size > 5 is no longer meaningful. We would always get same similarity.
I think this results manipulation is confusing. Could you please explain your use case for different batch sizes for reference and predictions?

tools/who_what_benchmark/whowhatbench/wwb.py
as-suvorov: It's wrong place for wwb tests. Please test wwb `--batch_size` in wwb tests.

tests/python_tests/test_rag.py
as-suvorov: Please revert genai pipeline tests and implement wwb `--batch_size` tests at https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/tests/test_cli_embeddings.py

mengweiguo: Will do this. Thanks.

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
mengweiguo: I encountered the NPU WWB failure mentioned in https://jira.devtools.intel.com/browse/EISW-191591. I just thought that the reference data may be pre-generated without batch and shared between devices(GPU,CPU,NPU) from the log https://jira.devtools.intel.com/secure/attachment/5635768/WW41-2025.4.0-20128_job_LO_NPU_acc_wwb_wwb_ref_nat_vs_genai_WIN_NPU_LNL_5.log . It may be not true. If the batch can be set when generating refer data for device, the scenario of different batch between reference and prediction will no exist.  I will look into WWB test. Thanks.

tests/python_tests/test_rag.py
mengweiguo: Reverted pipeline tests and added WWB tests. I don't know if it's proper to add tests for NPU there, so only tests for CPU is added.

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
mengweiguo: Reverted this change.

tools/who_what_benchmark/tests/test_cli_embeddings.py
```diff
@@ -92,3 +92,86 @@ def test_embeddings_basic(model_id, model_type, tmp_path):
         model_type,
         "--genai",
     ])
+
+def test_embeddings_with_batch(model_id, model_type, tmp_path):
```
as-suvorov: Did you run this test locally?
`model_id` and `model_type` parameters seems to be missing. I expect this test to fail.
Let's parametrize this test with `batch_size` parameter instead of hardcoded value

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -262,6 +262,12 @@ def parse_args():
         help="Config option assistant_confidence_threshold for Speculative decoding.",
     )
 
+    parser.add_argument(
+        '-bs', '--batch_size',
+        type=int,
+        default=None,
+        help='Batch size value')
```
as-suvorov: @sbalandi do we want to propagate `batch_size` to other types of tasks? I'm thinking if we need to make this parameter task specific like `embeds_batch_size` or potentially `rag_batch_size`

tools/who_what_benchmark/tests/test_cli_embeddings.py
mengweiguo: Updated and tests passed locally.  Is it needed to introduce `setup_module`/`teardown_module` to do model conversion only once?

tools/who_what_benchmark/whowhatbench/wwb.py
sbalandi: yes, I would make it more specific

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
sbalandi: Let's discuss this before making changes.

If I understand correctly, the plugin will crash if we say the batch is 10, but provide 7 as input.
We can't always control the input data; potentially a dataset can contain different lengths. I'd suggest not rasing exaption, but adding logic so that if real input data batch is smaller, wwb duplicate the data to the end.
For example:
batch size = 5
input passages ['a', 'b', 'c']
it not appropriate for us, so we make input data ['a', 'b', 'c', **'a', 'b'**]

@mengweiguo  @as-suvorov  What do you think about this ?

tools/who_what_benchmark/whowhatbench/wwb.py
mengweiguo: Updated. [Rename batch_size to embeds_batch_size.](https://github.com/openvinotoolkit/openvino.genai/pull/2986/commits/58e4c4d80ddcfdd02a764d90c5ae034846d948c7)

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
as-suvorov: Good point. Currently if we fix `TextEmbeddingPipeline` with `batch_size=10` the pipeline would fail if number of documents passed != 10. But it's not plugin related it's genai implementation limitation. We plan to fix it in the next release. I like the data duplication approach

mengweiguo: Updated.

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
```diff
@@ -178,8 +180,22 @@ def default_gen_answer(model, tokenizer, passages, **kwargs):
...
+            data_len = len(data[0])
+
+            if batch_size <= data_len:
+                data_input = data[0][:batch_size]
+            else:
+                # Duplicate data to reach batch_size
+                data_input = list(data[0])
+                while len(data_input) < batch_size:
+                    remaining = batch_size - len(data_input)
+                    data_input.extend(data[0][:min(remaining, data_len)])
```
sbalandi:
```suggestion
                data_input  = list(itertools.islice(itertools.cycle(data[0]), batch_size))
```
or
```suggestion
                data_input  = (data[0] * ((batch_size / len(data[0])) + 1))[:batch_size]
```

mengweiguo: Updated. May I need to add `itertools` into `tools/who_what_benchmark/requirements.txt`?  Thanks.

tools/who_what_benchmark/tests/test_cli_embeddings.py
```diff
@@ -92,3 +92,96 @@ def test_embeddings_basic(model_id, model_type, tmp_path):
         model_type,
         "--genai",
     ])
+
+@pytest.mark.parametrize(
+    ("model_id", "model_type", "batch_size"),
+    [
+        pytest.param("BAAI/bge-small-en-v1.5", "text-embedding", 1, marks=pytest.mark.xfail(
```
sbalandi: LGTM, just nitpick:
I would left just cases with Qwen3-Embedding-0.6B as we test here batch size, so we need one with small value with 1 and one case with value bigger then def batch size(10) which 12, so checking several times batch size 1 is ok, but doesn't make sense

tools/who_what_benchmark/whowhatbench/embeddings_evaluator.py
sbalandi: itertools is a built-in module in Python, so no

as-suvorov (approved):

tools/who_what_benchmark/tests/test_cli_embeddings.py
mengweiguo: Updated.

sbalandi:
build_jenkins

sbalandi:
@mengweiguo thanks for the changes !

mengweiguo:
@sbalandi @as-suvorov could you help merge if you feel it's fine? Thanks.

mengweiguo:
Text-embedding tests passed. The error happened on `test_cli_image.py`.
~~~
tools/who_what_benchmark/tests/test_cli_embeddings.py::test_embeddings_basic[BAAI/bge-small-en-v1.5-text-embedding] PASSED [  2%]
tools/who_what_benchmark/tests/test_cli_embeddings.py::test_embeddings_basic[Qwen/Qwen3-Embedding-0.6B-text-embedding] PASSED [  5%]
tools/who_what_benchmark/tests/test_cli_embeddings.py::test_embeddings_with_batch[Qwen/Qwen3-Embedding-0.6B-text-embedding-1] PASSED [  7%]
tools/who_what_benchmark/tests/test_cli_embeddings.py::test_embeddings_with_batch[Qwen/Qwen3-Embedding-0.6B-text-embedding-12] PASSED [ 10%]
~~~

sbalandi:
build_jenkins

PR #2996, opened 2025-11-11
[Whisper] Rely on timestamps parsing to detect end of chunk

as-suvorov:
## Description
For the short_form audio (< 30 seconds) implementation was uncoditionally set first processed chunk as the only one. This is not correct. Switched to timestamps parsing to detect end of the chunk.

CVS-173554

## Checklist:

### DIFF
--- a/tests/python_tests/test_whisper_pipeline.py
+++ b/tests/python_tests/test_whisper_pipeline.py
@@ -469,7 +469,7 @@ def test_return_timestamps_short_form(model_descr, sample_from_dataset):
 
 
 @pytest.mark.parametrize("model_descr", get_whisper_models_list(tiny_only=True))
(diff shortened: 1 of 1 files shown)


src/cpp/src/whisper/whisper.cpp
Wovchena, line 1: Is a test addition applicable here?

For example prepend silence to `How are you doing`. The first decoded chunk will be empty and the second (meaningful) one is dropped because of the bug.

as-suvorov: The test is applicable here. Most probably It's not going to work the way you described. But we might cover this case with 2 sentences audio from short form dataset or with `How are you doing` sample repeated twice. I wasn't plan to add test for this PR as I wanted to align our tests with transformers in general. But I'm not sure when it's going to happen. I'll add test for cover this change.

as-suvorov: Test added

PR #2937, opened 2025-10-29
Move SDPAToPA-related functionality to runtime

CuriousPanCake:
## Description
Move the SDPAToPA-related functionality to the transformation to avoid decoupling the code that needs to be a part of a single component.
This would help to make the transformation more isolated and testable.

Ticket: [CVS-165769](https://jira.devtools.intel.com/browse/CVS-165769)

Signed-off-by: Andrii Staikov <andrii.staikov@intel.com>

PR #3003, opened 2025-11-12
Pin OpenVINO runtime commit to avoid circular dependency issues

CuriousPanCake:
Pin OpenVINO runtime commit to avoid circular dependency issues

- Tickets:
[CVS-165769](https://jira.devtools.intel.com/browse/CVS-165769)

Signed-off-by: Andrii Staikov <andrii.staikov@intel.com>

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -43,7 +43,7 @@ jobs:
       with:
         platform: ubuntu22
         commit_packages_to_provide: wheels
-        revision: latest_available_commit
+        revision: fcf7c2964cf460ecfcb039f748d1f4028626d58c
         # Set specific revision and uncomment to use OV from its PR build:
         # branch_name: master
         # event_name: pull_request
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -89,7 +89,7 @@ jobs:
       with:
         platform: windows
         commit_packages_to_provide: wheels,openvino_node_npm_package.zip
-        revision: latest_available_commit
+        revision: fcf7c2964cf460ecfcb039f748d1f4028626d58c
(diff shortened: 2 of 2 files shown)


CuriousPanCake:
Required for merging https://github.com/openvinotoolkit/openvino/pull/32570
Otherwise, the GenAI code is going to be broken

Wovchena:
What particular OV change is going to break GenAI? I though OV is going to extend its API, not shrink.

There's also probably backward incompatibility introduced by OV. Suppose someone has their own implementation which happens to match current OV GenAI exactly.

CuriousPanCake:
You're right, we'll have more code in runtime, less in GenAI, but

1. Merging the [runtime PR](https://github.com/openvinotoolkit/openvino/pull/32570) is not possible without pining the GenAI version. Otherwise, once it is merged, the GenAI code is not going to work correct: you're going to always have exceptions at the PA initialization stage hence falling back to the SDPA implementation.

2. Similarly, it's not possible to merge the [GenAI changes](https://github.com/openvinotoolkit/openvino.genai/pull/2937) first as runtime is not supporting them yet: the CI is going to be red and the GenAI pipelines are not going to work properly.

That's why we thought it's going to be good to 
1. Pin the runtime version
2. Merge OV runtime changes
3. Merge GenAI changes

Please, let me know if you need more information

Wovchena:
(2) is clear. My questions are about (1) (still not answered)

mryzhov:
> (2) is clear. My questions are about (1) (still not answered)
When we merge the new transformation in OV then models will be converted (sdpa->pa) by moc transformation and in genai pipeline the old transformation will fail on this model.

CuriousPanCake:
Talked with Vladimir offline, he approves the changes

PR #3005, opened 2025-11-13
[DO-NOT-REVIEW]Enable custom VIT of qwen2.5 VL.

xipingyan:
Only for internal verification.
Need to set ENV: `ENABLE_CUSTOM_VIT=1`


rkazants (changes requested):
Do not merge. Please share ticket and model card id so that we will prepare PR into optimum-intel and you will check your PR on our IR.

xipingyan:
> Do not merge. Please share ticket and model card id so that we will prepare PR into optimum-intel and you will check your PR on our IR.

Sorry, @rkazants @Wovchena , This PR is only for Customer verification. The customer thinks our default VIT of Qwen2.5-Vl can't meet their requirement, and the customer has their own VIT implementation, but they also want to use GenAI's History Chat feature(multiple loop chat) for LLM part of VLLM.
So I draft this PR, please don't review and do not merge(labels has been added).

PR #2980, opened 2025-11-06
pre-commit introduction

sgonorov:
## Add Pre-commit Hooks for Code Quality and Consistency

This PR introduces pre-commit hooks to automate code quality checks and formatting across the codebase. The hooks run automatically on staged files before each commit.

### How to Run Pre-commit Manually

**Important:** Pre-commit only lints changed/staged files, not the entire codebase.

```bash
# Install pre-commit hooks (one-time setup)
pre-commit install

# Run on all staged files
pre-commit run

# Run on all files in the repository
pre-commit run --all-files

# Run on specific files
pre-commit run --files path/to/file1.cpp path/to/file2.py
```

### Prerequisites

Before using pre-commit hooks, ensure the following tools are installed:

```bash
# Required for C++ linting and formatting
sudo apt install clang-format clang-tidy  # Linux
# or
brew install clang-format llvm  # macOS
```

#### **General Hooks** (pre-commit-hooks)
- Fixes trailing whitespace, line endings (LF), and ensures files end with newlines
- Prevents commits to master branch and catches merge conflicts
- Detects private keys and large files (>1MB)
- Validates JSON, YAML, TOML, and Python AST syntax

**Benefits:** Prevents basic formatting inconsistencies and security issues before they enter the codebase.

#### **Ruff** (Python)
Fast Python linter and formatter that replaces multiple tools (flake8, black, isort, etc.).

**Benefits:** Catches common Python errors, enforces style consistency, and automatically fixes issues. 10-100x faster than traditional Python linters.

#### **clang-format** (C/C++)
Automatic C++ code formatter for consistent style.

**Benefits:** Eliminates style debates and ensures uniform formatting across the C++ codebase.

#### **clang-tidy** (C/C++)
Static analysis tool for C++ that catches bugs, performance issues, and style violations.

**Benefits:** Detects potential bugs, memory leaks, and anti-patterns. Enforces modern C++ best practices and improves code safety.

#### **mypy** (Python)
Static type checker for Python.

**Benefits:** Catches type-related bugs before runtime, improves code documentation through type hints, and enhances IDE autocomplete.

#### **absolufy-imports** (Python)
Converts relative imports to absolute imports.

**Benefits:** Makes imports more explicit and prevents import resolution issues in larger codebases.

#### **Prettier** (JSON/YAML/Markdown/JS/TS)
Opinionated code formatter for web-related files.

**Benefits:** Ensures consistent formatting for configuration and documentation files.

#### **cmake-format** (CMake)
[...]

PR #3008, opened 2025-11-13
Minimal pre-commit implementation

sgonorov:
Minimal pre-commit configuration and CI integration.

For now only fixes whitespaces/end-of-file and some other small checks. 

Will be used a base for adding other linters.

For the maxed-out Formatting/Linting see PR: https://github.com/openvinotoolkit/openvino.genai/pull/2980

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
     rev: v6.0.0
     hooks:
       - id: trailing-whitespace
-        exclude: '\.py$'
+        exclude: '\.(py|c|cpp|h|hpp)$'
       - id: end-of-file-fixer
       - id: check-merge-conflict
       - id: check-case-conflict


.github/workflows/lint.yml
```diff
@@ -0,0 +1,87 @@
...
+      - master
+  push:
+    branches:
+      - master
+
+env:
+  OV_BRANCH: master
+
+jobs:
+  openvino_download:
```
as-suvorov: is openvino required for genai linting? For `compile_commands.json` generation?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,87 @@
...
+          name: ${{ needs.openvino_download.outputs.ov_artifact_name }}
+          path: ${{ env.OV_INSTALL_DIR }}
+          merge-multiple: true
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Set up Node.js
```
as-suvorov: why nodejs needed?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,20 @@
+repos:
```
as-suvorov, line 1: Am I right that it would be required to fix linting issues for any changed files?

.github/workflows/lint.yml
sgonorov: Ah, yes, forgot to remove it - not yet required. But it's for compile_commands.json generation, yes. For clang-tidy.

.github/workflows/lint.yml
sgonorov: For prettier formatter- we have some JS code and other things in repo.  But it's also disabled now, so i'll remove it too.

.pre-commit-config.yaml
sgonorov: Yes, only for changed ones.

as-suvorov: So If I change 2 lines in several hundred loc file I would be required to fix lint for the whole file. Then feature changes would endup hidden in the linting changes. Is it correct?

sgonorov: Yes, that's the way. We can't partially format file sadly.
But for now it's mostly whitespaces as a starting point so it won't affect much. Later I'll try to make a separate PRs with properly formatted files and new formatters/linters introduction.

sgonorov: BTW, for python there's a formatter called Darker which is compatible with Ruff format, but applies changes only to changed lines. When we get to it, maybe we'll use it instead of simply enabling Ruff across the codebase.

as-suvorov: I guess whitespaces only rules are ok for now. But I'm not sure about more complex rules.
I see 3 options:
1. Enable rules. Expect PRs to fix linting along with the feature changes.
    This would be unconvenient for both, reviewer and author. Especialy for external contributors. We could ask to fix linting in a separate first to split linting and feature changes. But still confusing for external contributors.
2. Enable rule by rule with appropriate files linting fixes. I guess this is how you want to proceed.
4. Enable rules and implement filter pattern to filter out not linted files. Then we could incrementially fix linting and update filter.

Any more ideas on this?
Do you want to proceed with option 2? If yes, maybe we could start with this plan in this PR and fix whitespaces?

Wovchena: 2 is the best

sgonorov: I've added Darker and yes, path 2 is the good one. Whitespaces is in around 200+ files. So i'll make it as a separate PR after this one.

pyproject.toml
```diff
@@ -59,3 +59,16 @@ requires = [
     "cmake~=3.24.0; platform_system == 'Darwin' and platform_machine == 'arm64'",
 ]
 build-backend = "py_build_cmake.build"
+
+[tool.ruff]
+line-length = 100
```
as-suvorov: Please align rules with openvino: https://github.com/openvinotoolkit/openvino/blob/master/pyproject.toml#L45

sgonorov: Why we should also follow 160 line-length rule? Are there any specific reasons? I consider avoiding it as much as possible - such a high value defeats formatting purpose and makes code much harder to read.

Wovchena: Consistency. You can ask OV to change that

sgonorov: https://github.com/openvinotoolkit/openvino/pull/32987 - will change to 120, it's not perfect but it's much better. I'll just wait for this PR to be merged.

sgonorov: Change implemented in OV - now line length equals 120. I've also updated the PR.

.pre-commit-config.yaml
Wovchena: Darker uses just ruff. There are hooks like trailing-whitespace. I suspect they still have the problem of changing the whole file

sgonorov: It ignores python files. I can add CPP-related file to exception for now too if needed.

Wovchena: What "it" refers to?

as-suvorov (approved):

.pre-commit-config.yaml
sgonorov: Whitespace hook - for now it ignores only Pyhton files. We can add cpp files also.

Wovchena: Add cpp to fix https://github.com/openvinotoolkit/openvino.genai/pull/3008#discussion_r2522836097 Any reason not to do so?

sgonorov: Added. Then i'll create some migratiion PRs where fix all the whitespaces and remove these ignores.

PR #3012, opened 2025-11-13
[JS] Update tokenizer methods

Retribution98:
## Description
- Using Tensor from openvino-node in Tokenizer encode and decode
- Extended the Tensor API
- Added constructors for Tensor
- Moved Tokenizer to the separate TS file
- Use BigInt for tokenId
- Store openvino-node addon in the AddondData to manipulate with entity from the core part
- Aligned tests with the new API and verify tokenizer and binding behavior.
- Updated benchmark sample with using Tensor encode
- Update documentation - https://retribution98.github.io/openvino.genai/

## Ticket
CVS-174909

## Checklist:

### DIFF
--- a/src/js/src/helper.cpp
+++ b/src/js/src/helper.cpp
@@ -326,7 +326,9 @@ ov::Tensor js_to_cpp<ov::Tensor>(const Napi::Env& env, const Napi::Value& value)
     auto tensor_prototype = get_prototype_from_ov_addon(env, "Tensor");
     OPENVINO_ASSERT(tensor_wrap.InstanceOf(tensor_prototype), "Passed argument is not of type Tensor");
 
-    auto native_tensor_func = tensor_wrap.Get("__getExternalTensor").As<Napi::Function>();
+    Napi::Value get_external_tensor_val = tensor_wrap.Get("__getExternalTensor");
+    OPENVINO_ASSERT(get_external_tensor_val.IsFunction(), "Tensor object does not have a '__getExternalTensor' function. This may indicate an incompatible or outdated openvino-node version.");
+    auto native_tensor_func = get_external_tensor_val.As<Napi::Function>();
     Napi::Value native_tensor_value = native_tensor_func.Call(tensor_wrap, {});
     OPENVINO_ASSERT(native_tensor_value.IsExternal(), "__getExternalTensor() did not return an External object.");
 


Retribution98:
This PR is blocked by both [PR#32802](https://github.com/openvinotoolkit/openvino/pull/32802) and [PR#32850](https://github.com/openvinotoolkit/openvino/pull/32850) in openvino repository

src/js/lib/tokenizer.ts
```diff
@@ -0,0 +1,186 @@
...
+  /**
+   * Encodes a list of paired prompts into tokenized inputs.
+   * Input format is same as for HF paired input [[prompt_1, prompt_2], ...].
+   * @param prompts - list of paired prompts to encode
+   * @param options - encoding options
+   * @returns TokenizedInputs object containing input_ids and attention_mask tensors.
+   */
+  encode(prompts: [string, string][], options?: EncodeOptions): TokenizedInputs;
+
+  // TODO: move decode options to another interface
```
almilosz: 👍

src/js/src/addon.cpp
```diff
@@ -20,6 +20,28 @@ void init_class(Napi::Env env,
     exports.Set(class_name, prototype);
 }
 
+Napi::Value init_ov_addon(const Napi::CallbackInfo& info) {
```
almilosz:
```suggestion
void init_ov_addon(const Napi::CallbackInfo& info) {
```

src/js/src/helper.cpp
```diff
@@ -123,6 +129,23 @@ std::vector<std::string> js_to_cpp<std::vector<std::string>>(const Napi::Env& en
...
+
+        std::vector<int64_t> vector;
+        vector.reserve(arrayLength);
+        for (uint32_t i = 0; i < arrayLength; ++i) {
+            vector.push_back(js_to_cpp<int64_t>(env, array[i]));
+        }
+        return vector;
+    } else {
+        OPENVINO_THROW("Passed argument must be of type Array.");
+    }
```
```suggestion
     OPENVINO_ASSERT(value.IsArray(), "Passed argument must be of type Array.");
    auto array = value.As<Napi::Array>();
    size_t arrayLength = array.Length();

    std::vector<int64_t> vector;
    vector.reserve(arrayLength);
    for (uint32_t i = 0; i < arrayLength; ++i) {
        vector.push_back(js_to_cpp<int64_t>(env, array[i]));
    }
    return vector;
```

src/js/src/tokenizer.cpp
```diff
@@ -118,3 +124,157 @@ Napi::Value TokenizerWrapper::get_pad_token_id(const Napi::CallbackInfo& info) {
...
+            }
+            if (options.Has("maxLength")) {
+                tokenization_params["max_length"] = static_cast<size_t>(options.Get("maxLength").ToNumber().Int64Value());
+            }
+            if (options.Has("paddingSide")) {
+                tokenization_params["padding_side"] = options.Get("paddingSide").ToString().Utf8Value();
+            }
+            
+            last_text_arg_idx--;
+        }
```
almilosz: I would create a helper function for parsing these args. It'll be easier to read `encode()`

src/js/tests/tokenizer.test.js
```diff
@@ -181,33 +181,178 @@ describe("tokenizer", async () => {
...
     assert.strictEqual(typeof token, "number");
   });
+
+  it("getChatTemplate return string", () => {
+    const template = tokenizer.getChatTemplate();
+    assert.strictEqual(typeof template, "string");
+  });
+
+  it("setChatTemplate updates template", () => {
+    const originalTemplate = tokenizer.getChatTemplate();
```
almilosz:
```suggestion
   it("setChatTemplate updates template", () => {
    const originalTemplate = tokenizer.getChatTemplate();
    assert.strictEqual(typeof originalTemplate, "string");
```

yatarkan:
Docs also can be updated with new JS tokenizer functionality - in [Node.js Bindings](https://openvinotoolkit.github.io/openvino.genai/docs/bindings/node-js) page and in [Tokenization Guide](https://openvinotoolkit.github.io/openvino.genai/docs/guides/tokenization) (add js code example)

src/js/src/addon.cpp
Retribution98: This function is used as a JavaScript function, so we usually return a Napi::Value even if nothing is returned. I understand your confusion because it is called like `init_class`, but they have different usages. To be clearer, I will rename this function to `set_ov_node`.

src/js/lib/tokenizer.ts
```diff
@@ -0,0 +1,186 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { Tensor } from "openvino-node";
+import { ChatHistory } from "./addon.js";
+
+/**
+ * TokenizedInputs contains input_ids and attention_mask tensors.
+ * This is the result of encoding prompts using the Tokenizer.
+ */
+export interface TokenizedInputs {
```
apaniukov, line 13: There is also `std::optional<ov::Tensor> token_type_ids`, that used mostly in the rerank pipeline. It might be critical when pair input is used.

src/js/lib/tokenizer.ts
```diff
@@ -0,0 +1,186 @@
...
+   * Maximum length of the sequence.
+   * If undefined, the value will be taken from the IR.
+   */
+  maxLength?: number;
+
+  /**
+   * Side to pad the sequence, can be 'left' or 'right'.
+   * If undefined, the value will be taken from the IR.
+   */
+  paddingSide?: "left" | "right";
```
apaniukov: Can be None, which means that the default padding side (which is stored in IR) is used.

Retribution98: PaddingSide can be undefined, in which case the default value from the core part is used.

src/js/tests/tokenizer.test.js
almilosz: I think we don't need to have a separate test to check return type of `getChatTemplate()`. It'll be enough to get this error from e.g. `setChatTemplate updates template`.

src/js/src/addon.cpp
almilosz: I believe there are functions that return Napi::Value even if they are void, but I do not think we should repeat that pattern. Here it `pop_back` is used as js function and it's void https://github.com/openvinotoolkit/openvino.genai/blob/cf4bc0b577d2be4fca8d6a76b925e6289e762d36/src/js/src/chat_history.cpp#L55

Retribution98: Agree, it has been updated and let's use this approach in the next changes.

Retribution98:
@almilosz @apaniukov 
Please take another look.

PR #3020, opened 2025-11-14
[ImageGeneration] Use device for AutoencoderKL in sample

as-suvorov:
## Description
Use device for AutoencoderKL in sample

### DIFF
--- a/samples/cpp/image_generation/stable_diffusion_export_import.cpp
+++ b/samples/cpp/image_generation/stable_diffusion_export_import.cpp
@@ -38,7 +38,7 @@ void dedicated_models_export_import(const std::filesystem::path& root_dir) {
     auto unet = ov::genai::UNet2DConditionModel(root_dir / "unet", device);
     unet.export_model(blob_path / "unet");
 
-    auto vae = ov::genai::AutoencoderKL(root_dir / "vae_decoder", "CPU", ov::AnyMap{});
+    auto vae = ov::genai::AutoencoderKL(root_dir / "vae_decoder", device, ov::AnyMap{});
     vae.export_model(blob_path);
     // AutoencoderKL can be composed with decoder and encoder models
     // exported/


samples/cpp/image_generation/stable_diffusion_export_import.cpp
```diff
@@ -38,7 +38,7 @@ void dedicated_models_export_import(const std::filesystem::path& root_dir) {
     auto unet = ov::genai::UNet2DConditionModel(root_dir / "unet", device);
     unet.export_model(blob_path / "unet");
 
-    auto vae = ov::genai::AutoencoderKL(root_dir / "vae_decoder", "CPU", ov::AnyMap{});
+    auto vae = ov::genai::AutoencoderKL(root_dir / "vae_decoder", device, ov::AnyMap{});
     vae.export_model(blob_path);
```
Wovchena, line 42: Python still needs a fix

as-suvorov: `AutoencoderKL` has device in python https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/image_generation/stable_diffusion_export_import.py#L47

Wovchena: Why is `compile()` explicit there?

as-suvorov: I don't remember exactly. I guess it's ctor overload resolution issue

Wovchena: In Python? That should be something possible to fix in that case. Please check

as-suvorov: In both, python and c++. I'll take a look, but not sure when

Wovchena: C++ tests just passed

as-suvorov: It passed because of empty properties
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/image_generation/stable_diffusion_export_import.cpp#L41

PR #3026, opened 2025-11-17
Const-related fixes on tesor creation

sgonorov:
Fixes const-related casting issues in tests and code.

### DIFF
--- a/src/cpp/src/image_generation/models/autoencoder_kl.cpp
+++ b/src/cpp/src/image_generation/models/autoencoder_kl.cpp
@@ -32,7 +32,7 @@ public:
         OPENVINO_ASSERT(shape[0] == 1, "Batch size must be 1");
         shape[1] /= 2;
 
-        m_mean = ov::Tensor(parameters.get_element_type(), shape, parameters.data());
+        m_mean = ov::Tensor(parameters.get_element_type(), shape, const_cast<void*>(parameters.data()));
         m_std = ov::Tensor(m_mean.get_element_type(), shape);
         ov::Tensor logvar(parameters.get_element_type(), shape, m_mean.data<float>() + m_mean.get_size());
 
--- a/src/cpp/src/llm/pipeline_stateful.cpp
+++ b/src/cpp/src/llm/pipeline_stateful.cpp
@@ -325,7 +325,7 @@ EncodedResults StatefulLLMPipeline::generate(
 
     // Tail of previous output in chat mode is missing in KV cache.
     if (is_chat_conversation && m_chat_input_type == ov::genai::utils::GenerationChatInputsType::ENCODED_INPUTS) {
-        ov::Tensor new_chat_tokens = ov::Tensor{ov::element::i64, {1, m_tokenized_chat_history.size()}, m_tokenized_chat_history.data()};
+        ov::Tensor new_chat_tokens = ov::Tensor{ov::element::i64, {1, m_tokenized_chat_history.size()}, const_cast<int64_t*>(m_tokenized_chat_history.data())};
         ov::genai::align_kv_cache_and_history(new_chat_tokens, m_kv_cache_state);
 
         auto encoded_input = get_chat_encoded_input(new_chat_tokens, m_kv_cache_state);
@@ -414,7 +414,7 @@ EncodedResults StatefulLLMPipeline::generate(
             tokenized_chat_hist.reserve(state.size() + input_ids.get_size());
             std::copy(state.begin(), state.end(), std::back_inserter(tokenized_chat_hist));
             std::copy(input_ids.data<int64_t>(), input_ids.data<int64_t>() + input_ids.get_size(), std::back_inserter(tokenized_chat_hist));
-            sequence_group = std::make_shared<SequenceGroup>(request_id,  ov::Tensor(ov::element::i64, {1, tokenized_chat_hist.size()}, tokenized_chat_hist.data()), config, block_size);
+            sequence_group = std::make_shared<SequenceGroup>(request_id,  ov::Tensor(ov::element::i64, {1, tokenized_chat_hist.size()}, const_cast<int64_t*>(tokenized_chat_hist.data())), config, block_size);
         } else {
             size_t seq_len = input_ids.get_shape().at(1);
             size_t batch_offset = request_id * seq_len;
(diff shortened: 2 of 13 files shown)


tests/cpp/speculative_decoding.cpp
```diff
@@ -387,7 +387,7 @@ TEST_F(CBForSDTest, remove_and_replace_tokens__two_sequence) {
 
 TEST_F(CBForSDTest, add_tokens__two_sequence) {
     std::vector<int64_t> input_vector{0, 1, 2, 3, 4};
-    ov::Tensor input_tensor(ov::element::i64, ov::Shape{1, 5}, input_vector.data());
+    ov::Tensor input_tensor(ov::element::i64, ov::Shape{1, 5}, const_cast<int64_t*>(input_vector.data()));
```
Wovchena, line 390: Was that also broken? Looks strange that it needs `const_cast()`

CuriousPanCake: Would it be correct to `static_cast<const int64_t*>()` and all the other cases where `const_cast` is used?

sgonorov: > Would it be correct to `static_cast<const int64_t*>()` and all the other cases where `const_cast` is used?

Our goal is get rid of `const`, so this won't work.


> Was that also broken? Looks strange that it needs `const_cast()`

I've looked through all of the possible places where this issue can arise.

CuriousPanCake:
Can we expedite the merge of this PR as don't want to have the GenAI repo to have OV version pinned for a long time. Thanks for understanding.

tests/cpp/speculative_decoding.cpp
Wovchena: Tensor constructor has overloads for `void* host_ptr` with and without `const`. What issue is possible here?
I'm going to merge this. Please get back to it after the merge

src/cpp/src/image_generation/models/autoencoder_kl.cpp
```diff
@@ -32,7 +32,7 @@ class DiagonalGaussianDistribution {
         OPENVINO_ASSERT(shape[0] == 1, "Batch size must be 1");
         shape[1] /= 2;
 
-        m_mean = ov::Tensor(parameters.get_element_type(), shape, parameters.data());
+        m_mean = ov::Tensor(parameters.get_element_type(), shape, const_cast<void*>(parameters.data()));
```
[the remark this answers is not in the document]
praasz: Why const cast is required ?
IF reason is about exception for data member it looks like tensor in parameters is created from const pointer (as read-only view) I think it should be corrected at place when input tensor is created this looks like some kind WA to silent the exception but still new tensor will allow write read-only data.

src/cpp/src/visual_language/phi3_vision/classes.cpp
```diff
@@ -66,11 +66,11 @@ ov::Tensor HD_transform(const ov::Tensor& uint8, size_t num_crops) {
     if (trans) {
         src = clip_image_u8{int(height), int(width), {uint8_data, uint8_data + uint8.get_size()}};
         bilinear_resize(src, dst, new_h, new_w);
-        return padding_336(ov::Tensor{ov::element::u8, {1, new_w, new_h, 3}, dst.buf.data()});
+        return padding_336(ov::Tensor{ov::element::u8, {1, new_w, new_h, 3}, const_cast<uint8_t*>(dst.buf.data())});
     }
     src = clip_image_u8{int(width), int(height), {uint8_data, uint8_data + uint8.get_size()}};
     bilinear_resize(src, dst, new_w, new_h);
-    return padding_336(ov::Tensor{ov::element::u8, {1, new_h, new_w, 3}, dst.buf.data()});
+    return padding_336(ov::Tensor{ov::element::u8, {1, new_h, new_w, 3}, const_cast<uint8_t*>(dst.buf.data())});
```
praasz: Maybe the `dst.buf.data()` should be not a const.
The const_cast looks like WA and maybe root of problem should be fixed if not possible because some external lib API issue, then  is good to comment why source is constness is removed. Now it looks we allow modify read-only data

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -636,8 +636,8 @@ std::unique_ptr<CircularBufferQueue<ov::InferRequest>> create_vision_encoder_ire
     for (auto& v : a_image_scale)
         v = 1.0f / (v * 255.0f);
 
-    auto image_mean = ov::op::v0::Constant(ov::element::f32, ov::Shape{1, a_image_mean.size(), 1, 1}, a_image_mean.data());
-    auto image_scale = ov::op::v0::Constant(ov::element::f32, ov::Shape{1, a_image_scale.size(), 1, 1}, a_image_scale.data());
+    auto image_mean = ov::op::v0::Constant(ov::element::f32, ov::Shape{1, a_image_mean.size(), 1, 1}, const_cast<float*>(a_image_mean.data()));
+    auto image_scale = ov::op::v0::Constant(ov::element::f32, ov::Shape{1, a_image_scale.size(), 1, 1}, const_cast<float*>(a_image_scale.data()));
```
praasz: Why const cast is required, Constant accept const void* pointer

sgonorov: Ah, yeah, missed it. Fixed.

src/cpp/src/visual_language/phi3_vision/classes.cpp
sgonorov: Ah, yeah, you're right. Fixed it 👍

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -325,7 +325,7 @@ EncodedResults StatefulLLMPipeline::generate(
 
     // Tail of previous output in chat mode is missing in KV cache.
     if (is_chat_conversation && m_chat_input_type == ov::genai::utils::GenerationChatInputsType::ENCODED_INPUTS) {
-        ov::Tensor new_chat_tokens = ov::Tensor{ov::element::i64, {1, m_tokenized_chat_history.size()}, m_tokenized_chat_history.data()};
+        ov::Tensor new_chat_tokens = ov::Tensor{ov::element::i64, {1, m_tokenized_chat_history.size()}, const_cast<int64_t*>(m_tokenized_chat_history.data())};
```
praasz, line 328: In general the `const_cast` should be removed or comment why is required and why is safe to used it(tensor will be used still as read-only), otherwise it will introduce hack to allow write to read-only data.

It apply to all usage of `const_cast`.

PR #3034, opened 2025-11-18
Fix for deadlock in python callback

sgonorov:
Fixes deadlock [CVS-176777](https://jira.devtools.intel.com/browse/CVS-176777) and adds more samples and test cases.

### DIFF
--- a/samples/python/image_generation/inpainting.py
+++ b/samples/python/image_generation/inpainting.py
@@ -20,7 +20,6 @@ def main():
     parser.add_argument('prompt')
     parser.add_argument('image')
     parser.add_argument('mask')
-    parser.add_argument('--show-progress', action='store_true', help='Show generation progress')
     args = parser.parse_args()
 
     device = 'CPU'  # GPU can be used as well
@@ -33,11 +32,7 @@ def main():
         print(f"Step {step + 1}/{num_steps}")
         return False
 
-    kwargs = {}
-    if args.show_progress:
-        kwargs['callback'] = callback
-
-    image_tensor = pipe.generate(args.prompt, image, mask_image, **kwargs)
+    image_tensor = pipe.generate(args.prompt, image, mask_image, callback=callback)
 
     image = Image.fromarray(image_tensor.data[0])
     image.save("image.bmp")
--- a/samples/python/image_generation/text2image.py
+++ b/samples/python/image_generation/text2image.py
@@ -12,7 +12,6 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('model_dir')
     parser.add_argument('prompt')
-    parser.add_argument('--show-progress', action='store_true', help='Show generation progress')
     args = parser.parse_args()
 
     device = 'CPU'  # GPU can be used as well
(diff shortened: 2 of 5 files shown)


samples/python/image_generation/inpainting_with_callback.py
Wovchena, line 1: Can one of existing samples be updated instead?

src/cpp/src/synchronized_queue.hpp
```diff
@@ -44,4 +51,17 @@ class SynchronizedQueue
         std::unique_lock<std::mutex> lock(m_mutex);
         return m_queue.empty();
     }
+
+    void interrupt() {
```
Wovchena: LLMs are fine without `interrupt()`. Does that mean they are broken too?

sgonorov: Yes, as far as i understand they are synchronous so no need for this.

samples/python/image_generation/inpainting_with_callback.py
sgonorov: Merged them 👍

src/cpp/src/synchronized_queue.hpp
as-suvorov: They are not: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching/threaded_streamer.hpp

src/cpp/src/image_generation/threaded_callback.hpp
```diff
@@ -44,7 +44,7 @@ class ThreadedCallbackWrapper {
         }
 
         m_status = CallbackStatus::STOP;
-        m_squeue.empty();
+        m_squeue.interrupt();
```
as-suvorov: Is it correct that with `interrupt` the callback won't be called for the remaining queue values ?

src/cpp/src/synchronized_queue.hpp
sgonorov: Ah, yes, continuous batching uses `std::monostate` to handle this. I'll make it that way.

src/cpp/src/image_generation/threaded_callback.hpp
sgonorov: Yes, looks like it. I've made and pushed a small fix for this, so now it should be working fine.

as-suvorov: Would it work if there are 2 values remain after interrupt call?
Also `SynchronizedQueue` update with `try_pull` is confusing. Not clear when to use `pull` and when `try_pull`. Queue interface becomes not clean.

sgonorov: Maybe you're right. It becomes more complicated than I've expected. I'll make it similar to CB implementation.

as-suvorov (approved):

samples/python/image_generation/inpainting.py
```diff
@@ -20,6 +20,7 @@ def main():
     parser.add_argument('prompt')
     parser.add_argument('image')
     parser.add_argument('mask')
+    parser.add_argument('--show-progress', action='store_true', help='Show generation progress')
```
Wovchena: Keep samples args minimal. Just draw the bar without asking

Wovchena (changes requested):

samples/python/image_generation/text2image.py
```diff
@@ -12,21 +12,30 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('model_dir')
     parser.add_argument('prompt')
+    parser.add_argument('--show-progress', action='store_true', help='Show generation progress')
```
likholat: Could you please also add a callback to the image2image sample
https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/image_generation/image2image.py

sgonorov: Done 👍

samples/python/image_generation/inpainting.py
sgonorov: Done 👍

Wovchena (approved):

PR #3045, opened 2025-11-19
Update C chat samples with using ChatHistory class

zhaohb:


src/c/include/openvino/genai/c/chat_history.h
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * @brief Create a ChatHistory instance from a JSON array string.
+ * @param messages_json A JSON string containing an array of message objects.
+ * @param history A pointer to the newly created ov_genai_chat_history.
+ * @return ov_genai_chat_history_status_e A status code, return OK(0) if successful.
+ */
+OPENVINO_GENAI_C_EXPORTS ov_genai_chat_history_status_e ov_genai_chat_history_create_from_json(
+    const char* messages_json,
+    ov_genai_chat_history** history);
```
yatarkan: Other functions have `ov_genai_chat_history` param in the first place - let's align

src/c/src/chat_history.cpp
```diff
@@ -0,0 +1,371 @@
...
+    if (!history || !(history->object)) {
+        return OV_GENAI_CHAT_HISTORY_INVALID_PARAM;
+    }
+    try {
+        if (history->object->empty()) {
+            return OV_GENAI_CHAT_HISTORY_EMPTY;
+        }
+        history->object->pop_back();
+    } catch (const ov::Exception& e) {
+        return OV_GENAI_CHAT_HISTORY_EMPTY;
```
yatarkan: Why reported status for ov::Exception is empty, not error?

src/c/src/chat_history.cpp
```diff
@@ -0,0 +1,371 @@
...
+    } catch (const std::exception& e) {
+        return OV_GENAI_CHAT_HISTORY_ERROR;
+    } catch (...) {
+        return OV_GENAI_CHAT_HISTORY_ERROR;
+    }
```
yatarkan: Why different catch blocks are needed if all of them return the same error status?

src/c/src/chat_history.cpp
```diff
@@ -0,0 +1,371 @@
...
+    try {
+        ov::genai::JsonContainer messages = ov::genai::JsonContainer::from_json_string(std::string(messages_json));
+        if (!messages.is_array()) {
+            return OV_GENAI_CHAT_HISTORY_INVALID_JSON;
+        }
+        std::unique_ptr<ov_genai_chat_history> _history = std::make_unique<ov_genai_chat_history>();
+        _history->object = std::make_shared<ov::genai::ChatHistory>(messages);
+        *history = _history.release();
+    } catch (const ov::Exception& e) {
+        return OV_GENAI_CHAT_HISTORY_INVALID_JSON;
```
yatarkan: Shouldn't this also return error status?

src/c/include/openvino/genai/c/chat_history.h
yatarkan, line 1: In C++ ChatHistory mainly works with JsonContainer instances. I think in C we should also expose JsonContainer and use it in C implementation for ChatHistory.

yatarkan (changes requested):

src/c/src/chat_history.cpp
zhaohb: The exception handling distinguishes between ov::Exception and std::exception. 
Alternatively, returning ERROR directly is also acceptable if simpler error handling is preferred.

src/c/src/chat_history.cpp
zhaohb: So you're suggesting that **all exceptions should be caught with a single catch block and return only one error status**?

src/c/src/chat_history.cpp
zhaohb: same as above？

src/c/include/openvino/genai/c/chat_history.h
zhaohb: Excellent suggestion, will be revised.

src/c/include/openvino/genai/c/chat_history.h
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * @brief Create a ChatHistory instance from a JsonContainer (array).
+ * @param history A pointer to the newly created ov_genai_chat_history.
+ * @param messages A JsonContainer containing an array of message objects.
+ * @return ov_genai_chat_history_status_e A status code, return OK(0) if successful.
+ */
+OPENVINO_GENAI_C_EXPORTS ov_genai_chat_history_status_e ov_genai_chat_history_create_from_json_container(
+    ov_genai_chat_history** history,
+    const ov_genai_json_container* messages
```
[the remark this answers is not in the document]
yatarkan: Please address Copilot's proposal - align order of input/output params

src/c/src/chat_history.cpp
```diff
@@ -0,0 +1,269 @@
...
+}
+
+ov_genai_chat_history_status_e ov_genai_chat_history_push_back(
+    ov_genai_chat_history* history,
+    const ov_genai_json_container* message) {
+    if (!history || !(history->object) || !message || !(message->object)) {
+        return OV_GENAI_CHAT_HISTORY_INVALID_PARAM;
+    }
+    try {
+        history->object->push_back(*(message->object));
```
yatarkan, line 65: Let's check that message is json container object

src/c/src/chat_history.cpp
```diff
@@ -0,0 +1,269 @@
...
+    size_t index,
+    ov_genai_json_container** message) {
+    if (!history || !(history->object) || !message) {
+        return OV_GENAI_CHAT_HISTORY_INVALID_PARAM;
+    }
+    try {
+        if (index >= history->object->size()) {
+            return OV_GENAI_CHAT_HISTORY_OUT_OF_BOUNDS;
+        }
+        ov::genai::JsonContainer message_ref = (*history->object)[index];
```
yatarkan: Is `message_ref` actually a reference? If not let's rename accordingly. The same for methods below.

src/c/include/openvino/genai/c/json_container.h
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * @brief Create a copy of JsonContainer.
+ * @param container A pointer to the source ov_genai_json_container instance.
+ * @param copy_container A pointer to store the copied ov_genai_json_container.
+ * @return ov_genai_json_container_status_e A status code, return OK(0) if successful.
+ */
+OPENVINO_GENAI_C_EXPORTS ov_genai_json_container_status_e ov_genai_json_container_copy(
+    const ov_genai_json_container* container,
+    ov_genai_json_container** copy_container);
```
yatarkan: Let's name it a bit more explicitly for clarification
```suggestion
    const ov_genai_json_container* source,
    ov_genai_json_container** target);
```

src/c/include/openvino/genai/c/chat_history.h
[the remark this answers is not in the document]
zhaohb: Done.

src/c/src/llm_pipeline.cpp
```diff
@@ -144,6 +145,38 @@ ov_status_e ov_genai_llm_pipeline_generate(ov_genai_llm_pipeline* pipe,
...
+                                      : pipe->object->generate(*(history->object), {}, callback);
+        } else {
+            *(_results->object) = (config && config->object) ? pipe->object->generate(*(history->object), *(config->object))
+                                                             : pipe->object->generate(*(history->object));
+        }
+        if (results) {
+            *results = _results.release();
+        }
+    } catch (...) {
+        return ov_status_e::UNKNOW_EXCEPTION;
```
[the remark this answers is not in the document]
zhaohb: Keep consistent with OpenVINO's definition; no modifications here.

src/c/include/openvino/genai/c/json_container.h
```diff
@@ -0,0 +1,90 @@
...
+    OV_GENAI_JSON_CONTAINER_OUT_OF_BOUNDS = -3,
+    OV_GENAI_JSON_CONTAINER_ERROR = -4
+} ov_genai_json_container_status_e;
+
+/**
+ * @brief Create a new empty JsonContainer instance (as an object).
+ * @param container A pointer to the newly created ov_genai_json_container.
+ * @return ov_genai_json_container_status_e A status code, return OK(0) if successful.
+ */
+OPENVINO_GENAI_C_EXPORTS ov_genai_json_container_status_e ov_genai_json_container_create(ov_genai_json_container** container);
```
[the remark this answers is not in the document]
yatarkan: `JsonContainer` constructor creates object by default.

PR #3046, opened 2025-11-19
Update pybind11 to 3.0.1

almilosz:
## Description

PR in openvino https://github.com/openvinotoolkit/openvino/pull/32899


as-suvorov:
Addressed in: https://github.com/openvinotoolkit/openvino.genai/pull/3044

PR #3055, opened 2025-11-21
eagle3 cb impl with top-1 proposal

songbell:
eagle3 CB impl
Tickets: [CVS-173358](https://jira.devtools.intel.com/browse/CVS-173358)
ref code: https://github.com/SafeAILab/EAGLE

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,5 +1,7 @@
 'category: GHA':
 - '.github/**/*'
+- '.pre-commit-config.yaml'
+- 'pyproject.toml'
 
 'category: tests dependencies':
 - 'tests/python_tests/requirements.txt'
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,6 +8,8 @@ on:
     branches:
       - master
 
+permissions: read-all
+
 env:
   OV_BRANCH: master
 
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -167,6 +167,8 @@ jobs:
 
       - name: CMake Build
         run: |
+          apt update
+          apt install -y libgtk2.0-dev pkg-config ffmpeg libavformat-dev libavcodec-dev libswscale-dev libavutil-dev
           source ${{ env.OV_INSTALL_DIR }}/setupvars.sh
 
           cmake -DOpenVINODeveloperPackage_DIR=${{ env.OV_INSTALL_DIR }}/developer_package/cmake \
@@ -379,7 +381,7 @@ jobs:
       matrix:
         build-type: [Release]
     needs: [ openvino_download, genai_build_cmake ]
-    timeout-minutes: 10
+    timeout-minutes: 30
     defaults:
       run:
         shell: bash
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -146,7 +146,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-${{ runner.arch }}-ccache-${{ env.TARGET_BRANCH }}-${{ matrix.build-type }}-cpack
           path: ${{ env.CCACHE_DIR }}
-        
+
       - name: Generate product manifest
         id: create_manifest
         uses: openvinotoolkit/openvino/.github/actions/create_manifest@master
@@ -159,7 +159,7 @@ jobs:
 
       - name: Clean ccache stats
         run: ccache --zero-stats --show-config
-          
+
       - name: CMake Build
         run: |
           source ${{ env.OV_INSTALL_DIR }}/setupvars.sh
@@ -205,7 +205,7 @@ jobs:
           name: genai_tests_${{ matrix.build-type }}
           path: ${{ env.INSTALL_TESTS_DIR }}
           if-no-files-found: 'error'
-          
+
       - name: Upload manifest
         if: always()
         uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
@@ -263,7 +263,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-${{ runner.arch }}-ccache-${{ env.TARGET_BRANCH }}-wheel
           path: ${{ env.CCACHE_DIR }}\
-          
+
       - name: Set CI environment
         id: create_manifest
         uses: openvinotoolkit/openvino/.github/actions/create_manifest@master
@@ -318,7 +318,7 @@ jobs:
       matrix:
         build-type: [Release]
     needs: [ openvino_download, genai_build_cmake ]
-    timeout-minutes: 10
+    timeout-minutes: 30
     defaults:
       run:
         shell: bash
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -100,7 +100,7 @@ jobs:
       matrix:
         build-type: [Release, Debug]
     needs: [ openvino_download ]
-    timeout-minutes: 45
+    timeout-minutes: 80
     defaults:
       run:
         shell: pwsh
@@ -487,7 +487,7 @@ jobs:
       matrix:
         build-type: [Release, Debug]
     needs: [ openvino_download, genai_build_cpack ]
-    timeout-minutes: 10
+    timeout-minutes: 70
     defaults:
       run:
         shell: pwsh
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+exclude: '\.pyi$'
 repos:
   - repo: meta
     hooks:
@@ -26,3 +27,4 @@ repos:
         args: ["--formatter=ruff"]
         additional_dependencies:
           - ruff==0.14.4
+          - tomli==2.3.0
--- a/samples/cpp/visual_language_chat/CMakeLists.txt
+++ b/samples/cpp/visual_language_chat/CMakeLists.txt
@@ -1,6 +1,10 @@
 # Copyright (C) 2023-2025 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
+if (MSVC)
+  set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+endif()
+
 find_package(OpenVINOGenAI REQUIRED
     PATHS
         "${CMAKE_BINARY_DIR}"  # Reuse the package from the build.
@@ -52,6 +56,55 @@ set_target_properties(benchmark_vlm PROPERTIES
     INSTALL_RPATH_USE_LINK_PATH ON)
 
 install(TARGETS benchmark_vlm
+        RUNTIME DESTINATION samples_bin/
+        COMPONENT samples_bin
+        EXCLUDE_FROM_ALL)
+
+find_package(OpenCV QUIET)
+
+add_executable(video_to_text_chat video_to_text_chat.cpp)
+
+if(NOT OpenCV_FOUND)
+    include(FetchContent)
+
+    set(BUILD_SHARED_LIBS ON)
+    set(WITH_FFMPEG ON)
+    FetchContent_Declare(
+            opencv
+            GIT_REPOSITORY https://github.com/opencv/opencv.git
+            GIT_TAG 4.11.0
+            GIT_SHALLOW TRUE
+    )
+    FetchContent_MakeAvailable(opencv)
+
+    target_include_directories(video_to_text_chat PRIVATE
+            ${OPENCV_CONFIG_FILE_INCLUDE_DIR}
+            ${OPENCV_MODULE_opencv_core_LOCATION}/include
+            ${OPENCV_MODULE_opencv_videoio_LOCATION}/include
+            )
+
+
+    target_link_libraries(video_to_text_chat opencv_core opencv_imgcodecs opencv_imgproc opencv_videoio openvino::genai cxxopts::cxxopts)
+
+    if(LINUX)
+        set_target_properties(video_to_text_chat opencv_core opencv_imgcodecs opencv_imgproc opencv_videoio PROPERTIES
+            INSTALL_RPATH "$ORIGIN/../lib"
+            # Ensure out of box LC_RPATH on macOS with SIP
+            INSTALL_RPATH_USE_LINK_PATH ON)
+    elseif(APPLE)
+        set_target_properties(video_to_text_chat opencv_core opencv_imgcodecs opencv_imgproc opencv_videoio PROPERTIES
+            INSTALL_RPATH "@loader_path/../lib"
+            # Ensure out of box LC_RPATH on macOS with SIP
+            INSTALL_RPATH_USE_LINK_PATH ON)
+    endif()
+else()
+    set(OPENCV_LIBS ${OpenCV_LIBS})
+
+    target_include_directories(video_to_text_chat PRIVATE ${OpenCV_INCLUDE_DIRS} "${CMAKE_BINARY_DIR}")
+    target_link_libraries(video_to_text_chat PRIVATE ${OpenCV_LIBS} openvino::genai cxxopts::cxxopts)
+endif()
+
+install(TARGETS video_to_text_chat
         RUNTIME DESTINATION samples_bin/
         COMPONENT samples_bin
         EXCLUDE_FROM_ALL)
--- a/samples/cpp/visual_language_chat/README.md
+++ b/samples/cpp/visual_language_chat/README.md
@@ -3,8 +3,9 @@
 This example showcases inference of Visual language models (VLMs). The application doesn't have many configuration options to encourage the reader to explore and modify the source code. For example, change the device for inference to GPU. The sample features `ov::genai::VLMPipeline` and runs the simplest deterministic greedy sampling algorithm. There is also a Jupyter [notebook](https://github.com/openvinotoolkit/openvino_notebooks/tree/latest/notebooks/minicpm-v-multimodal-chatbot) which provides an example of Visual-language assistant.
 
 
-There are two sample files:
+There are three sample files:
  - [`visual_language_chat.cpp`](./visual_language_chat.cpp) demonstrates basic usage of the VLM pipeline.
+ - [`video_to_text_chat.cpp`](./video_to_text_chat.cpp) demonstrates video to text usage of the VLM pipeline.
  - [`benchmark_vlm.cpp`](./benchmark_vlm.cpp) shows how to benchmark a VLM in OpenVINO GenAI. The script includes functionality for warm-up iterations, generating text and calculating various performance metrics.
 
 
@@ -19,9 +20,9 @@ pip install --upgrade-strategy eager -r ../../requirements.txt
 optimum-cli export openvino --model openbmb/MiniCPM-V-2_6 --trust-remote-code MiniCPM-V-2_6
 ```
 
-## Run
+Follow [Get Started with Samples](https://docs.openvino.ai/2025/get-started/learn-openvino/openvino-samples/get-started-demos.html) to run samples.
 
-Follow [Get Started with Samples](https://docs.openvino.ai/2025/get-started/learn-openvino/openvino-samples/get-started-demos.html) to run the sample.
+## Run image-to-text chat sample:
 
 [This image](https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11) can be used as a sample image.
 
@@ -30,6 +31,17 @@ Follow [Get Started with Samples](https://docs.openvino.ai/2025/get-started/lear
 Discrete GPUs (dGPUs) usually provide better performance compared to CPUs. It is recommended to run larger models on a dGPU with 32GB+ RAM. For example, the model `llava-hf/llava-v1.6-mistral-7b-hf` can benefit from being run on a dGPU. Modify the source code to change the device for inference to the `GPU`.
 
 Refer to the [Supported Models](https://openvinotoolkit.github.io/openvino.genai/docs/supported-models/#visual-language-models-vlms) for more details.
+
+
+## Run video-to-text chat sample:
+
+A model that supports video input is required to run this sample, for example `llava-hf/LLaVA-NeXT-Video-7B-hf`.
+
+[This video](https://huggingface.co/datasets/raushan-testing-hf/videos-test/resolve/main/sample_demo_1.mp4) can be used as a sample video.
+
+`video_to_text_chat ./LLaVA-NeXT-Video-7B-hf/ sample_demo_1.mp4`
+
+Supported models with video input are listed in [this section](https://openvinotoolkit.github.io/openvino.genai/docs/use-cases/image-processing/#use-image-or-video-tags-in-prompt).
 
 ## Run benchmark:
 
@@ -61,7 +73,7 @@ Tokenization time: 1.26 ± 0.09 ms
 Detokenization time: 0.33 ± 0.05 ms
 Embeddings preparation time: 5733.85 ± 26.34 ms
 TTFT: 11246.98 ± 80.55 ms
-TPOT: 135.45 ± 4.73 ms/token 
+TPOT: 135.45 ± 4.73 ms/token
 Throughput: 7.38 ± 0.26 tokens/s
 ```
 
--- a/samples/cpp/visual_language_chat/video_to_text_chat.cpp
+++ b/samples/cpp/visual_language_chat/video_to_text_chat.cpp
@@ -1,0 +1,125 @@
+// Copyright (C) 2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include <openvino/genai/visual_language/pipeline.hpp>
+#include <opencv2/core.hpp>
+#include <opencv2/videoio.hpp>
+#include <iostream>
+#include <filesystem>
+
+namespace fs = std::filesystem;
+
+std::vector<size_t> make_indices(size_t total_frames, size_t num_frames) {
+    std::vector<size_t> indices;
+    indices.reserve(num_frames);
+
+    auto step = float(total_frames) / num_frames;
+
+    for (size_t i = 0; i < num_frames; ++i) {
+        size_t idx = std::min(size_t(i * step), total_frames - 1);
+        indices.push_back(idx);
+    }
+
+    return indices;
+}
+
+ov::Tensor load_video(const std::filesystem::path& video_path, size_t num_frames = 8) {
+    cv::VideoCapture cap(video_path.string());
+
+    if (!cap.isOpened()) {
+        OPENVINO_THROW("Could not open the video file.");
+    }
+    size_t total_num_frames = cap.get(cv::CAP_PROP_FRAME_COUNT);
+    auto indices = make_indices(total_num_frames, num_frames);
+
+    std::vector<cv::Mat> frames;
+    cv::Mat frame;
+    size_t width  = cap.get(cv::CAP_PROP_FRAME_WIDTH);
+    size_t height = cap.get(cv::CAP_PROP_FRAME_HEIGHT);
+    ov::Tensor video_tensor(ov::element::u8, ov::Shape{num_frames, height, width, 3});
+    auto video_tensor_data = video_tensor.data<uint8_t>();
+
+    size_t frame_idx = 0;
+    while (cap.read(frame)) {
+        OPENVINO_ASSERT(frame.cols == width && frame.rows == height && frame.channels() == 3);
+        if (std::find(indices.begin(), indices.end(), frame_idx) != indices.end()) {
+            memcpy(video_tensor_data, frame.data, frame.total() * 3 * sizeof(uint8_t));
+            video_tensor_data += frame.total() * 3;
+        }
+        frame_idx++;
+    }
+    OPENVINO_ASSERT(frame_idx == total_num_frames, "Frame count mismatch: expected " + std::to_string(total_num_frames) + ", got " + std::to_string(frame_idx));
+    
+    return video_tensor;
+}
+
+std::vector<ov::Tensor> load_videos(const std::filesystem::path& input_path) {
+    if (input_path.empty() || !fs::exists(input_path)) {
+        OPENVINO_THROW("Path to videos is empty or does not exist.");
+    }
+    if (fs::is_directory(input_path)) {
+        std::set<fs::path> sorted_videos{fs::directory_iterator(input_path), fs::directory_iterator()};
+        std::vector<ov::Tensor> videos;
+        for (const fs::path& dir_entry : sorted_videos) {
+            videos.push_back(load_video(dir_entry));
+        }
+        return videos;
+    }
+    return {load_video(input_path)};
+}
+
+bool print_subword(std::string&& subword) {
+    return !(std::cout << subword << std::flush);
+}
+
+int main(int argc, char* argv[]) try {
+    if (argc < 3 || argc > 4) {
+        OPENVINO_THROW(std::string{"Usage "} + argv[0] + " <MODEL_DIR> <VIDEO_FILE OR DIR_WITH_VIDEOS> <DEVICE>");
+    }
+    std::vector<ov::Tensor> videos = load_videos(argv[2]);
+
+    // GPU and NPU can be used as well.
+    // Note: If NPU is selected, only language model will be run on NPU
+    std::string device = (argc == 4) ? argv[3] : "CPU";
+    ov::AnyMap enable_compile_cache;
+    if (device == "GPU") {
+        // Cache compiled models on disk for GPU to save time on the
+        // next run. It's not beneficial for CPU.
+        enable_compile_cache.insert({ov::cache_dir("vlm_cache")});
+    }
+    ov::genai::VLMPipeline pipe(argv[1], device, enable_compile_cache);
+
+    ov::genai::GenerationConfig generation_config;
+    generation_config.max_new_tokens = 100;
+
+    std::string prompt;
+
+    pipe.start_chat();
+    std::cout << "question:\n";
+
+    std::getline(std::cin, prompt);
+    pipe.generate(prompt,
+                  ov::genai::videos(videos),
+                  ov::genai::generation_config(generation_config),
+                  ov::genai::streamer(print_subword));
+    std::cout << "\n----------\n"
+        "question:\n";
+    while (std::getline(std::cin, prompt)) {
+        pipe.generate(prompt,
+                      ov::genai::generation_config(generation_config),
+                      ov::genai::streamer(print_subword));
+        std::cout << "\n----------\n"
+            "question:\n";
+    }
+    pipe.finish_chat();
+} catch (const std::exception& error) {
+    try {
+        std::cerr << error.what() << '\n';
+    } catch (const std::ios_base::failure&) {}
+    return EXIT_FAILURE;
+} catch (...) {
+    try {
+        std::cerr << "Non-exception object thrown\n";
+    } catch (const std::ios_base::failure&) {}
+    return EXIT_FAILURE;
+}
--- a/samples/deployment-requirements.txt
+++ b/samples/deployment-requirements.txt
@@ -4,3 +4,4 @@ librosa==0.11.0  # For Whisper
 pillow==12.0.0  # Image processing for VLMs
 json5==0.12.1  # For ReAct
 pydantic==2.12.5  # For Structured output json schema
+opencv-python==4.12.0.88 # For video-to-text VLM sample
--- a/samples/python/visual_language_chat/README.md
+++ b/samples/python/visual_language_chat/README.md
@@ -2,8 +2,9 @@
 
 This example showcases inference of text-generation Vision Language Models (VLMs): `miniCPM-V-2_6` and other models with the same signature. The application doesn't have many configuration options to encourage the reader to explore and modify the source code. For example, change the device for inference to GPU. The sample features `openvino_genai.VLMPipeline` and configures it for the chat scenario. There is also a Jupyter [notebook](https://github.com/openvinotoolkit/openvino_notebooks/tree/latest/notebooks/minicpm-v-multimodal-chatbot) which provides an example of Visual-language assistant.
 
-There are two sample files:
+There are three sample files:
  - [`visual_language_chat.py`](./visual_language_chat.py) demonstrates basic usage of the VLM pipeline.
+ - [`video_to_text_chat.py`](./video_to_text_chat.py) demonstrates video to text usage of the VLM pipeline.
  - [`benchmark_vlm.py`](./benchmark_vlm.py) shows how to benchmark a VLM in OpenVINO GenAI. The script includes functionality for warm-up iterations, generating text and calculating various performance metrics.
  - [`milebench_eval_vlm.py`](./milebench_eval_vlm.py) provides MileBench validation for VLMs, enabling evaluation of image–text reasoning and visual QA tasks across multiple subsets designed to assess the MultImodal Long-contExt capabilities of MLLMs.
 
@@ -39,19 +40,28 @@ tokenizer = AutoTokenizer.from_pretrained("openbmb/MiniCPM-V-2_6")
 export_tokenizer(tokenizer, output_dir)
 ```
 
-## Run:
+Install [deployment-requirements.txt](../../deployment-requirements.txt) via `pip install -r ../../deployment-requirements.txt` to run VLM samples.
+
+## Run image-to-text chat sample:
 
 [This image](https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11) can be used as a sample image.
 
-Install [deployment-requirements.txt](../../deployment-requirements.txt) via `pip install -r ../../deployment-requirements.txt` and then, run a sample:
-
 `python visual_language_chat.py ./miniCPM-V-2_6/ 319483352-d5fbbd1a-d484-415c-88cb-9986625b7b11.jpg`
 
+See https://github.com/openvinotoolkit/openvino.genai/blob/master/src/README.md#supported-models for the list of supported models.
+
+## Run video-to-text chat sample:
+
+A model that supports video input is required to run this sample, for example `llava-hf/LLaVA-NeXT-Video-7B-hf`.
+
+[This video](https://huggingface.co/datasets/raushan-testing-hf/videos-test/resolve/main/sample_demo_1.mp4) can be used as a sample video.
+
+`python video_to_text_chat.py ./LLaVA-NeXT-Video-7B-hf/ sample_demo_1.mp4`
+
+Supported models with video input are listed in [this section](https://openvinotoolkit.github.io/openvino.genai/docs/use-cases/image-processing/#use-image-or-video-tags-in-prompt).
 
 Discrete GPUs (dGPUs) usually provide better performance compared to CPUs. It is recommended to run larger models on a dGPU with 32GB+ RAM. # TODO: examples of larger models
 Modify the source code to change the device for inference to the GPU.
-
-See https://github.com/openvinotoolkit/openvino.genai/blob/master/src/README.md#supported-models for the list of supported models.
 
 ## Run benchmark:
 
@@ -83,7 +93,7 @@ Tokenization time: 1.26 ± 0.09 ms
 Detokenization time: 0.33 ± 0.05 ms
 Embeddings preparation time: 5733.85 ± 26.34 ms
 TTFT: 11246.98 ± 80.55 ms
-TPOT: 135.45 ± 4.73 ms/token 
+TPOT: 135.45 ± 4.73 ms/token
 Throughput: 7.38 ± 0.26 tokens/s
 ```
 
--- a/samples/python/visual_language_chat/video_to_text_chat.py
+++ b/samples/python/visual_language_chat/video_to_text_chat.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+# Copyright (C) 2025 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+
+import argparse
+import numpy as np
+import cv2
+import openvino_genai
+from openvino import Tensor
+from pathlib import Path
+
+
+def streamer(subword: str) -> bool:
+    """
+
+    Args:
+        subword: sub-word of the generated text.
+
+    Returns: Return flag corresponds whether generation should be stopped.
+
+    """
+    print(subword, end="", flush=True)
+
+    # No value is returned as in this example we don't want to stop the generation in this method.
+    # "return None" will be treated the same as "return openvino_genai.StreamingStatus.RUNNING".
+
+
+def read_video(path: str, num_frames: int = 8) -> Tensor:
+    """
+
+    Args:
+        path: The path to the video.
+        num_frames: Number of frames sampled from the video.
+
+    Returns: the ov.Tensor containing the video.
+
+    """
+    cap = cv2.VideoCapture(path)
+
+    frames = []
+    total_num_frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+    indices = np.arange(0, total_num_frames, total_num_frames / num_frames).astype(int)
+
+    idx = 0
+    while cap.isOpened():
+        ret, frame = cap.read()
+        if not ret:
+            break
+        if idx in indices:
+            frames.append(np.array(frame))
+        idx += 1
+    assert idx == total_num_frames, "Frame count mismatch: expected {}, got {}".format(total_num_frames, idx)
+
+    return Tensor(frames)
+
+
+def read_videos(path: str) -> list[Tensor]:
+    entry = Path(path)
+    if entry.is_dir():
+        return [read_video(str(file)) for file in sorted(entry.iterdir())]
+    return [read_video(path)]
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("model_dir", help="Path to the model directory")
+    parser.add_argument("video_dir", help="Path to a video file.")
+    parser.add_argument("device", nargs="?", default="CPU", help="Device to run the model on (default: CPU)")
+    args = parser.parse_args()
+
+    videos = read_videos(args.video_dir)
+
+    # GPU and NPU can be used as well.
+    # Note: If NPU is selected, only the language model will be run on the NPU.
+    enable_compile_cache = dict()
+    if args.device == "GPU":
+        # Cache compiled models on disk for GPU to save time on the next run.
+        # It's not beneficial for CPU.
+        enable_compile_cache["CACHE_DIR"] = "vlm_cache"
+
+    pipe = openvino_genai.VLMPipeline(args.model_dir, args.device, **enable_compile_cache)
+
+    config = openvino_genai.GenerationConfig()
+    config.max_new_tokens = 100
+
+    pipe.start_chat()
+    prompt = input("question:\n")
+    pipe.generate(prompt, videos=videos, generation_config=config, streamer=streamer)
+
+    while True:
+        try:
+            prompt = input("\n----------\nquestion:\n")
+        except EOFError:
+            break
+        pipe.generate(prompt, generation_config=config, streamer=streamer)
+    pipe.finish_chat()
+
+
+if __name__ == "__main__":
+    main()
--- a/site/docs/use-cases/speech-recognition/index.mdx
+++ b/site/docs/use-cases/speech-recognition/index.mdx
@@ -11,7 +11,14 @@ import UsageOptionsSection from './_sections/_usage_options/index.mdx';
 <ConvertModelSection>
     Download and convert model (e.g. [openai/whisper-base](https://huggingface.co/openai/whisper-base)) to OpenVINO format from Hugging Face:
 
-    <OptimumCLI model='openai/whisper-base' outputDir='whisper_ov' trustRemoteCode />
+    <Tabs groupId="whisper-export-command">
+        <TabItem label="Default" value="default">
+            <OptimumCLI model='openai/whisper-base' outputDir='whisper_ov' trustRemoteCode />
+        </TabItem>
+        <TabItem label="INT8 Static Quantization" value="int8_quant">
+            <OptimumCLI model='openai/whisper-base' outputDir='whisper_ov_int8' trustRemoteCode quantMode='int8' dataset='librispeech' numSamples={32} />
+        </TabItem>
+    </Tabs>
 
     See all supported [Speech Recognition Models](/docs/supported-models/#speech-recognition-models-whisper-based).
 </ConvertModelSection>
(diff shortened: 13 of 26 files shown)


src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,440 @@
...
+    auto draft_model = draft_model_desc.model;
+
+    auto main_device = main_model_desc.device;
+    std::string draft_device = draft_model_desc.device.empty() ? main_model_desc.device : draft_model_desc.device;
+
+    ov::AnyMap draft_properties =
+        draft_model_desc.properties.empty() ? main_model_desc.properties : draft_model_desc.properties;
+
+    // main and draft model can have different tokenizers
+    // to do: support retokenization: 154103
```
peterchen-intel: It is not decicated for EAGLE3, but the whole speculative decoding pipeline is missing?

tests/python_tests/utils/hugging_face.py
```diff
@@ -228,9 +232,16 @@ def convert_models(
...
+<<<<<<< HEAD
+    if hf_tokenizer:
+        hf_tokenizer.save_pretrained(models_path)
+        # convert tokenizers as well
+        convert_and_save_tokenizer(hf_tokenizer, models_path, **tokenizer_kwargs)
+=======
     hf_tokenizer.save_pretrained(str(models_path))
     # convert tokenizers as well
     convert_and_save_tokenizer(hf_tokenizer, models_path)
+>>>>>>> eb64c87ebc885499f7fafb38e576dce4b7010495
```
Wovchena: Leftover

tests/python_tests/samples/test_speculative_decoding_lm.py
```diff
@@ -26,22 +46,27 @@ def test_sample_speculative_decoding_lm(self, convert_model, convert_draft_model
...
+    @pytest.mark.parametrize(
+        "convert_model, convert_draft_model, sample_args",
+        [
+            pytest.param("Qwen3-1.7B", "qwen3_1.7b_eagle3", test_prompt, marks=pytest.mark.skip(reason='CVS-171947, CVS-171943, CVS-174959')),
+        ],
+        indirect=["convert_model", "convert_draft_model"],
+    )
+    def test_sample_speculative_decoding_lm(self, convert_model, convert_draft_model, sample_args):
+        if sys.platform == 'darwin':
+            pytest.xfail("Ticket 173586")
```
Wovchena:
```suggestion
            pytest.xfail("Ticket 173586")  # Update mac.yml to run the test separately similar to linux.yml after resolution 
```

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,59 @@
...
 
 namespace {
+struct Eagle3RTInfo {
+    bool eagle3_mode = false;
+    std::vector<int> hidden_layers_list;
+    std::filesystem::path dt_mapping_table;
+};
+
+Eagle3RTInfo
+extract_eagle_mode_from_config(ov::AnyMap& config, const std::filesystem::path& models_path) {
```
Wovchena:
```suggestion
extract_eagle3_info_from_config(ov::AnyMap& config, const std::filesystem::path& models_path) {
```
That's how you named the struct

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,59 @@
...
+    Eagle3RTInfo eagle_rt_info;
+    if (config.find("eagle3_mode") != config.end()) {
+        eagle_rt_info.eagle3_mode = config.at("eagle3_mode").as<bool>();
+        config.erase("eagle3_mode");
+        if (config.find("hidden_layers_list") != config.end()) {
```
Wovchena: Need a test for that feature. I think you usually rely on configs

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,59 @@
...
+            try {
+                eagle_rt_info.hidden_layers_list = config.at("hidden_layers_list").as<std::vector<int>>();
+                config.erase("hidden_layers_list");
+            } catch (const std::exception&) {
```
Wovchena, line 41: What can trigger the exception? You know `at()` won't throw because it's under `if (config.find("hidden_layers_list") != config.end())`

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,59 @@
...
+
+Eagle3RTInfo
+extract_eagle_mode_from_config(ov::AnyMap& config, const std::filesystem::path& models_path) {
+    Eagle3RTInfo eagle_rt_info;
+    if (config.find("eagle3_mode") != config.end()) {
+        eagle_rt_info.eagle3_mode = config.at("eagle3_mode").as<bool>();
+        config.erase("eagle3_mode");
+        if (config.find("hidden_layers_list") != config.end()) {
+            try {
+                eagle_rt_info.hidden_layers_list = config.at("hidden_layers_list").as<std::vector<int>>();
```
Wovchena: You already found `hidden_layers_list` by calling `config.find("hidden_layers_list")` in `if (config.find("hidden_layers_list") != config.end())`. Factor the result of `find()` out into a var

Wovchena (changes requested):

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -266,14 +316,22 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::update
                      updated_context_len = min_candidate_len + prompt_len,
                      max_new_tokens = request->get_max_new_tokens();
         size_t generated_len = request->get_context_len() >= request->get_prompt_len() ? request->get_context_len() - request->get_prompt_len() + 1 : 0;
-        if (generated_len > 0 && result.removed_tokens_cnt > 0) {
-            request->update_processed_tokens_num(num_processed_tokens - result.removed_tokens_cnt + 1);
+        if (validate_length > 0) {
+            if (generated_len > 0) {
+                request->update_processed_tokens_num(num_processed_tokens - result.removed_tokens_cnt + 1 - (validate_length - 1));
```
sbalandi: Why is it needed to subtract validate_length ?
validate_length is eq hidden state of main model and eq to sequence len, or no ?

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -266,14 +316,22 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::update
...
         }
-        if (result.inserted_tokens_cnt > 0 && result.removed_tokens_cnt == 0) {
+        if (validate_length == 0 && result.inserted_tokens_cnt > 0 && result.removed_tokens_cnt == 0) {
             request->set_num_validated_tokens(result.inserted_tokens_cnt);
+        } else if (validate_length > 0) {
+            request->set_num_validated_tokens(validate_length - 1);  // in generation stage
         }
         // to pause `draft_model` generation in case of `generated_len >= max_new_tokens - 1` to generate last token by `main_model`
-        if (!m_is_validation_mode_enabled) {
+        if (!m_is_validation_mode_enabled && result.inserted_tokens_cnt != 0) {
```
sbalandi, line 327: Why `&& result.inserted_tokens_cnt != 0` was added ?

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,59 @@
...
+struct Eagle3RTInfo {
+    bool eagle3_mode = false;
+    std::vector<int> hidden_layers_list;
+    std::filesystem::path dt_mapping_table;
+};
+
+Eagle3RTInfo
+extract_eagle_mode_from_config(ov::AnyMap& config, const std::filesystem::path& models_path) {
+    Eagle3RTInfo eagle_rt_info;
+    if (config.find("eagle3_mode") != config.end()) {
```
sbalandi, line 33: in last shared models this property is `is_eagle3`, which version is right ?
but also it looks like you check input property here, not the model config

songbell: @xufang-lisa current it should be eagle3_mode, right?
do  you mean I should check it somewhere else?

src/cpp/src/continuous_batching/pipeline.cpp
songbell: ov::Any as may throw, to avoid unable to convert to std::vector from std::istream

src/cpp/src/continuous_batching/pipeline.cpp
songbell: applied

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
songbell: this is mapped to the stable kv idea in eagle3:https://github.com/SafeAILab/EAGLE/blob/main/eagle/model/cnets.py#L696
in the draft iterations, only the first iter is marked as stable KV (others are not verified and not reliable)
after the main validation, it will need to rewind the process token number to the stable kv position

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
songbell: to ensure the draft model is only triggered when needed. 
for example, in continuous batching vllm mode, prompt is always scheduled first
considering below:
sequence 0 generate ----> some time sequence 1 prefill joined
then sequence 1 prefill is scheduled first, at this point, sequence 0 draft should not be waked up and need to wait for next cycles.

tests/python_tests/test_continuous_batching.py
```diff
@@ -547,31 +560,43 @@ def test_dynamic_split_fuse_doesnt_affect_generated_text():
         PipelineType.SPECULATIVE_DECODING,
     ]
 )
-def test_speculative_decoding_extended_perf_metrics(pipeline_type: PipelineType):
+@pytest.mark.parametrize("main_model_id,draft_model_id, prompt", speculative_cases)
+@pytest.mark.precommit
```
songbell: @Wovchena shall I not mark precommit here?

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -24,6 +24,27 @@ inline std::string get_paged_attention_score_output_for_decoder_layer(size_t dec
     return ss.str();
 }
 
+enum HiddenStateFlags : uint8_t {
```
rkazants, line 45: please document these flags here

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -24,6 +24,27 @@ inline std::string get_paged_attention_score_output_for_decoder_layer(size_t dec
...
+    bool operator<(const SequenceKey& other) const {
+        return std::tie(request_id, grouped_sequence_id) <
+            std::tie(other.request_id, other.grouped_sequence_id);
+    }
+};
+
+struct HiddenStateRange {
+    size_t start_token_idx{};
+    size_t length{};
+};
```
rkazants, line 89: please also document these structures - why do we need them

tests/python_tests/test_continuous_batching.py
Wovchena: Correct. Remove it

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,59 @@
...
+                eagle_rt_info.hidden_layers_list = config.at("hidden_layers_list").as<std::vector<int>>();
+                config.erase("hidden_layers_list");
+            } catch (const std::exception&) {
+                OPENVINO_THROW("please check the hidden layers input");
+            }
+        } else {
+            // compute the layers from number of hidden layers
+            auto config_file_path = models_path / "config.json";
+            if (!std::filesystem::exists(config_file_path))
+                OPENVINO_THROW("cannot deduce layers for hidden layer extraction");
```
Wovchena:
```suggestion
                OPENVINO_THROW("Cannot deduce layers for hidden layer extraction because the file is missing: ", config_file_path);
```

src/cpp/src/continuous_batching/pipeline.cpp
songbell: @Wovchena  do you mean to test the generated draft model has these rt_info entry? 
@rkazants how do you view the "eagle3_mode" entry in draft model? or shall we use some other way to distinguish it in draft model?

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
sbalandi: isn't it just request->get_prompt_len() ?

src/cpp/src/continuous_batching/pipeline.cpp
sbalandi: @songbell 
if I explicitly made this check true, I still can't run provided models successfully
I try to run sample [speculative_decoding_lm.cpp](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/cpp/text_generation/speculative_decoding_lm.cpp)
If I don't set scheduler config for draft model it fails here [speculative_decoding_impl.cpp#L69](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/speculative_decoding/speculative_decoding_impl.cpp#L69) from here [speculative_decoding_impl.cpp#L75](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/speculative_decoding/speculative_decoding_impl.cpp#L75)
with error:
`map::at`
if I set it, then it fails here [cache_manager.hpp#L60](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching/cache_manager.hpp#L60)
with 
```
Exception from src/core/src/shape_util.cpp:65:
Accessing out-of-range dimension
```

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
songbell: it is prompt len + tokens whose kv caches are stable(generated with tokens validated by target model + hidden states from target model)

songbell: I update the code with a more friendly variable  name, please check again if this is clear enough @sbalandi

sbalandi: thank you, I just tried to find out what that mean, now I managed to run and see

src/cpp/src/continuous_batching/pipeline.cpp
sbalandi: looks like problem was in openvino version, I can run speculative_decoding_lm now on last build

src/cpp/src/continuous_batching/pipeline.cpp
Wovchena: Agreed to delete `else` branch

songbell: per last discussion, we will leave it here first, and after the solution finalized from optimum intel side, we will have separate PR to refine this logic @rkazants  please correct me if I misunderstand the info.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,329 @@
...
+        draft_weight_node->output(0).replace(main_weight_node->output(0));
+    } catch (const std::exception& e) {
+        GENAI_WARN(std::string("Error: failed to import embedding weights from main model to draft model. Exception: ") + e.what());
+    } catch (...) {
+        GENAI_WARN("Error: failed to import embedding weights from main model to draft model due to unknown exception.");
+    }
+}
+
+void shift_fc_from_draft_to_main(std::shared_ptr<ov::Model>& main_model, std::shared_ptr<ov::Model>& draft_model) {
+    // extract the FC transform weight from draft model
```
songbell, line 54: here is the logic to transfer the FC node in draft model to main model, to keep the hidden states output from main model the correct size for draft model to consume,  I also simplified the other transformation logic , please review again @rkazants

src/cpp/src/continuous_batching/pipeline.cpp
peterchen-intel: @songbell Already in CVS-171943

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,329 @@
...
+    // Eagle speculative decoding does not support dynamic_split_fuse mode
+    // because it requires hidden state interaction from main model to draft model
+    // until CVS-177647 is resolved
+    if (scheduler_configs.first.dynamic_split_fuse) {
+        GENAI_WARN(
+            "Note: disable dynamic split fuse for eagle3 speculative decoding"
+        );
+        scheduler_configs.first.dynamic_split_fuse = false;
+        scheduler_configs.second.dynamic_split_fuse = false;
+    }
```
popovaan, line 184: Maybe throw exception in that case? Otherwise it might give a false impression that this parameter works with eagle3 while it actually doesn't.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,329 @@
...
+        return;
+    }
+
+    try {
+        draft_weight_node->output(0).replace(main_weight_node->output(0));
+    } catch (const std::exception& e) {
+        GENAI_WARN(std::string("Error: failed to import embedding weights from main model to draft model. Exception: ") + e.what());
+    } catch (...) {
+        GENAI_WARN("Error: failed to import embedding weights from main model to draft model due to unknown exception.");
+    }
```
popovaan: Does the rest of the algorithm suppose to work if this transformation failed? Maybe raise exception if it fails?

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,329 @@
...
+                    target.replace_source_output(input_node);
+                }
+            }
+            return matmul_node->input_value(1).get_node_shared_ptr();
+        }
+        return nullptr;
+    };
+    auto fc_weights = remove_fc_and_rewire(draft_model);
+    if (!fc_weights) {
+        GENAI_WARN("Error: failed to retrieve and remove FC matmul from draft model.");
```
popovaan: Same question as above. I suggest to raise exception.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: the pipeline will work, but the draft model will work at producing wrong results, leads to low acceptance rate, changed to throw exception

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: we define users to be unaware of any eagle3 info in app level, otherwise, they have to explicitly set this flag where they need to know they want to run eagle3 mode now. I'll implement ticket CVS-177647 to close this.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: thanks, applied

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -48,6 +69,10 @@ class ModelRunner {
     // Input shape: [N, conversation length].
     // Output shape: [1, conversation length, hidden_size].
     EmbeddingsModel::Ptr m_embedding;
+    uint8_t m_hidden_state_flags = HS_NONE;
+    std::map<SequenceKey, HiddenStateRange> m_sequence_hidden_state_mapping;
+    // a container which use sequence group id and request id as key to store hidden states
+    std::map<size_t, ov::Tensor> m_initial_hidden_states; // shape: [N, seq_len, hidden_size]
```
rkazants: what is key of `size_t` type here? The comment looks incorrect "a container which use sequence group id and request id as key to store hidden states".
Please double-check

rkazants: I think we can use `unordered_map` because hash function for  standard type `size_t` is defined

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -24,6 +24,27 @@ inline std::string get_paged_attention_score_output_for_decoder_layer(size_t dec
...
+enum HiddenStateFlags : uint8_t {
+    HS_NONE      = 0,
+    HS_EXPORT    = 1 << 0,
+    HS_IMPORT    = 1 << 1,
+    HS_INTERNAL  = 1 << 2
+};
+
+struct SequenceKey {
+    size_t request_id{};
+    size_t grouped_sequence_id{};
```
rkazants, line 68: please document - what are `request_id` and `grouped_sequence_id`, and why are they needed.

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -490,6 +600,113 @@ class ModelRunner {
...
+                    hidden_size = hidden_states_shape.back();
+                    break;
+                }
+            }
+        }
+        if (hidden_size == 0) {
+            return {};
+        }
+
+        ov::Tensor hs(ov::element::f32, {total_num_tokens, 1, hidden_size});
```
rkazants, line 711: can `hidden_state` be of a different type, not `fp32`? For example, on GPU it can be `fp16`.

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -490,6 +600,113 @@ class ModelRunner {
...
+    ov::Tensor _prepare_hidden_state_input(size_t total_num_tokens,
+                                           size_t& hidden_size /*in/out*/) {
+        if (!(m_hidden_state_flags & (HS_IMPORT | HS_INTERNAL))) {
+            return {};
+        }
+
+        if (hidden_size == 0) {
+            for (const auto& kv : m_initial_hidden_states) {
+                const auto& initial_hidden_states = kv.second;
+                if (initial_hidden_states && initial_hidden_states.get_shape().size() >= 2) {
```
rkazants: can a rank of `initial_hidden_states` be 1?

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -424,6 +517,23 @@ class ModelRunner {
 
         _reset_cache_rotation_coefficients();
 
+        if (_is_hs_export()) {
+            try {
```
rkazants: please re-write this code with required conditional checking without exceptions. Exceptions mechanism quite slowdowns performance.

src/cpp/src/continuous_batching/model_runner.hpp
```diff
@@ -490,6 +600,113 @@ class ModelRunner {
...
+        auto make_roi = [](const std::vector<size_t>& shape, size_t first_dim_start, size_t first_dim_end) {
+            ov::Coordinate start(shape.size(), 0), end(shape.size(), 0);
+            start[0] = first_dim_start;
+            end[0] = first_dim_end;
+            for (size_t d = 1; d < shape.size(); ++d) {
+                start[d] = 0;
+                end[d] = shape[d];
+            }
+            return std::make_pair(start, end);
+        };
```
rkazants: It looks a duplicate. The same logic is implemented in other place. Please create helper for this and re-use.
Please clarify why do we need a crop along batch dimension.

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,60 @@
...
+    Eagle3RTInfo eagle_rt_info;
+    if (config.find("eagle3_mode") != config.end()) {
+        eagle_rt_info.eagle3_mode = config.at("eagle3_mode").as<bool>();
+        config.erase("eagle3_mode");
+        auto it = config.find("hidden_layers_list");
+        if (it != config.end()) {
+            try {
+                eagle_rt_info.hidden_layers_list = it->second.as<std::vector<int>>();
+                config.erase("hidden_layers_list");
+            } catch (const std::exception&) {
```
rkazants, line 41: please avoid try-catch mechanism. If you are using try-catch, it means that internal design is not quite thought over

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,60 @@
...
+            read_json_param(data, "num_hidden_layers", num_decoder_layers);
+            OPENVINO_ASSERT(num_decoder_layers > 3, "num_decoder_layers is too small to deduce hidden layers for extraction");
+            // The following default hidden layer selection corresponds to the EAGLE reference implementation:
+            // https://github.com/SafeAILab/EAGLE/blob/0ea94696/eagle/model/modeling_llama_kv.py#L1138
+            // These layers (2, num_decoder_layers / 2, num_decoder_layers - 3) are chosen to capture features from
+            // early, middle, and late stages of the decoder, as recommended by the EAGLE authors.
+            // If you wish to use different layers, provide the "hidden_layers_list" parameter in the config.
+            eagle_rt_info.hidden_layers_list = { 2, num_decoder_layers / 2, num_decoder_layers - 3 };
+        }
+        OPENVINO_ASSERT(eagle_rt_info.hidden_layers_list.size() == 3, "only exactly 3 layer extraction are expected in eagle3");
```
rkazants:
```suggestion
        OPENVINO_ASSERT(eagle_rt_info.hidden_layers_list.size() == 3, "Eagle3 is expected to provide exactly three layers for extraction");
```

src/cpp/src/speculative_decoding/update_request_structs.hpp
```diff
@@ -10,11 +10,17 @@ namespace ov::genai {
...
-
+    // Stores the hidden states tensor associated with the generated sequence.
+    // This field is used for the "eagle speculative" decoding algorithm,
+    // where hidden states are required to efficiently validate and extend speculative tokens.
+    // If not using eagle speculative decoding, this field may remain empty.
+    ov::Tensor hidden_states;
     GeneratedSequence(const std::vector<int64_t>& generated_token_ids,
-                    const std::vector<float>& generated_log_probs) :
+                    const std::vector<float>& generated_log_probs,
+                    const ov::Tensor generated_hidden_states = {}) :
```
rkazants:
```suggestion
                    const ov::Tensor& generated_hidden_states = {}) :
```

src/cpp/src/continuous_batching/model_runner.hpp
songbell: no in current ov impl. Since hidden state is the output, gpu will handle it as fp32.

src/cpp/src/continuous_batching/model_runner.hpp
songbell: I suppose hidden states in LLM model should at least contain [seq_len, hidden_size]? current test models doesn't run into this case, but I'll change to a more strict shape check here

src/cpp/src/continuous_batching/model_runner.hpp
songbell: yes applied.
crop along batch dimension: in continuous batching case, we need to figure out hidden state for each sequence
suppose 2 seqs with len [30,40], we have a hidden state shape [70, hidden_size], then we need to assign correct hidden state for each seq

src/cpp/src/sampling/sampler.hpp
```diff
@@ -99,6 +99,7 @@ class Sampler {
     Tokenizer m_tokenizer;
 
     ThreadPool m_thread_pool;
+    std::shared_ptr<ov::op::v0::Constant> m_draft2target_mapping; // Tensor to store draft2target mapping for eagle model
```
rkazants:
```suggestion
    std::shared_ptr<ov::op::v0::Constant> m_d2t_mapping; // Tensor to store draft2target mapping for eagle model
```
please leave a comment what this constant contain and how it is used for mapping.

src/cpp/src/sampling/sampler.hpp
```diff
@@ -125,6 +126,10 @@ class Sampler {
     // pair with map with backend name and corresponding compiler init time, and vector of compile times for each concrete grammar
     std::pair<std::map<std::string, float>, std::vector<float>> get_structured_output_times();
     void clear_structured_output_compile_times();
+
+    void set_d2t_for_decoding(std::shared_ptr<ov::op::v0::Constant>& d2t) {
```
rkazants:
```suggestion
    void set_d2t_for_decoding(const std::shared_ptr<ov::op::v0::Constant>& d2t) {
```

src/cpp/src/sampling/sampler.cpp
```diff
@@ -852,6 +852,11 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
                         }
                     }
                 }
+                if (!is_validation_mode_enabled && m_draft2target_mapping) { // compute token offset for draft model in speculative sampling
+                    ov::Tensor d2t_tensor = m_draft2target_mapping->get_tensor_view();
```
rkazants: I think it is better to use std::vector<int64_t> for storing this map instead of ov::op::Constant. That's because anyway you access a buffer of data for mapping. Using std::vector, it will be simpler code.

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -63,12 +63,48 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::get_ge
...
+    ov::Coordinate end_coord(shape.size(), 0);
+
+    for (size_t i = 0; i < shape.size(); ++i) {
+        start_coord[i] = 0;
+        if (i == seq_len_dim) {
+            end_coord[i] = new_seq_len;
+        } else {
+            end_coord[i] = shape[i];
+        }
+    }
```
rkazants: looks duplicating logic, please create helper and re-use in different places

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -252,6 +293,18 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::update
                 candidate_token_ids.resize(min_candidate_len);
                 candidate_token_log_probs.resize(min_candidate_len);
                 result.inserted_tokens_cnt = insert_tokens_to_sequence(running_sequence, candidate_token_ids, candidate_token_log_probs, logit_processor, is_update_logit_processor);
+                // handle hidden states for eagle mode
+                if (eagle_mode_enabled && !m_is_validation_mode_enabled && result.inserted_tokens_cnt > 0) { // update hidden states for draft model
+                    // at least there should be one bonus token from main
+                    auto& hidden_state = candidate_sequence.hidden_states;
+                    ov::Tensor pruned_hidden_state = truncate_hidden_state_from_end(hidden_state, result.removed_tokens_cnt);
+                    const auto& shape = pruned_hidden_state.get_shape();
+                    OPENVINO_ASSERT(shape[0] >= 1, "Unexpected hidden state shape from main model hidden state after verification.");
```
rkazants:
```suggestion
                    OPENVINO_ASSERT(shape[0] >= 1, "Unexpected hidden state shape from the main model.");
```

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
```diff
@@ -252,6 +293,18 @@ ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::update
                 candidate_token_ids.resize(min_candidate_len);
                 candidate_token_log_probs.resize(min_candidate_len);
                 result.inserted_tokens_cnt = insert_tokens_to_sequence(running_sequence, candidate_token_ids, candidate_token_log_probs, logit_processor, is_update_logit_processor);
+                // handle hidden states for eagle mode
+                if (eagle_mode_enabled && !m_is_validation_mode_enabled && result.inserted_tokens_cnt > 0) { // update hidden states for draft model
+                    // at least there should be one bonus token from main
+                    auto& hidden_state = candidate_sequence.hidden_states;
+                    ov::Tensor pruned_hidden_state = truncate_hidden_state_from_end(hidden_state, result.removed_tokens_cnt);
```
rkazants, line 289: can we avoid pruning of hidden state and use it as is?

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp
```diff
@@ -40,5 +40,50 @@ class ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl :
...
+                                                 const GenerationConfig& generation_config,
+                                                 const SchedulerConfig& scheduler_config,
+                                                 const std::string& device,
+                                                 const ov::AnyMap& plugin_config,
+                                                 bool is_validation_mode_enabled) : ContinuousBatchingForSpeculativeDecodingImpl(
+                                                     model, tokenizer, generation_config,
+                                                     scheduler_config, device, plugin_config,
+                                                     is_validation_mode_enabled) {
+                                                        eagle_mode_enabled = true;
+                                                     };
```
rkazants:
```suggestion
                                                     is_validation_mode_enabled), eagle_mode_enabled(true) {};
```
please also do something with code style. Now I see extra indentation.

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp
```diff
@@ -40,5 +40,50 @@ class ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl :
...
+
+    bool is_requests_empty();
+
+    void set_d2t_for_draft_decoding(std::shared_ptr<ov::op::v0::Constant>& d2t) {
```
rkazants:
```suggestion
    void set_d2t_for_draft_decoding(const std::shared_ptr<ov::op::v0::Constant>& d2t) {
```

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp
```diff
@@ -40,5 +40,50 @@ class ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl :
...
+        if (m_model_runner) {
+            m_model_runner->enable_hidden_state_import(is_needed);
+        }
+    }
+
+    void set_hidden_state_internal_needed(bool is_needed) {
+        if (m_model_runner) {
+            m_model_runner->enable_hidden_state_internal(is_needed);
+        }
+    }
```
rkazants, line 125: please document these methods - why are they needed.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+        return;
+    }
+
+    try {
+        draft_weight_node->output(0).replace(main_weight_node->output(0));
+    } catch (const std::exception& e) {
+        OPENVINO_THROW(std::string("Error: failed to import embedding weights from main model to draft model. Exception: ") + e.what());
+    } catch (...) {
+        OPENVINO_THROW("Error: failed to import embedding weights from main model to draft model due to unknown exception.");
+    }
```
rkazants: please avoid try-catch. It looks not accurate because it can throw new exception from `catch` block

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
+// Copyright (C) 2023-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+#include "speculative_decoding_eagle3_impl.hpp"
+#include "logger.hpp"
+
+namespace ov::genai {
+void share_embedding_weights(std::shared_ptr<ov::Model>& main_model, std::shared_ptr<ov::Model>& draft_model) {
```
rkazants:
```suggestion
void share_embedding_weights(const std::shared_ptr<ov::Model>& main_model, std::shared_ptr<ov::Model>& draft_model) {
```

you share weights from the main model to the draft model. Please leave a comment what are these weights, and why we share it to draft model

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+            if (!matmul_node) continue;
+            auto input_node = matmul_node->get_input_node_shared_ptr(0);
+            auto param_node = ov::as_type_ptr<ov::op::v0::Parameter>(input_node);
+            if (!param_node || input_node->get_friendly_name().find("hidden_states") == std::string::npos) continue;
+            // Rewire all outputs of this MatMul to use the input_node directly
+            for (auto& output : matmul_node->outputs()) {
+                for (auto& target : output.get_target_inputs()) {
+                    target.replace_source_output(input_node);
+                }
+            }
```
rkazants, line 67: not sure how it is possible just to re-connect matmul input to output. Probably matmul output is a result node of the model, right. I think you can just re-create a draft model with new set of outputs, without matmul output. It will allow to simply the model itself. And you just need to return a pointer the required matmul.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+        if (input_node && input_node->get_friendly_name().find("eagle3_hidden_states_concat") != std::string::npos) {
+            auto matmul = std::make_shared<ov::op::v0::MatMul>(input_node, fc_weights, false, true);
+            matmul->set_friendly_name("eagle3_hidden_state_fc");
+            result->input(0).replace_source_output(matmul);
+            break;
+        }
+    }
+}
+
+std::shared_ptr<ov::op::v0::Constant> extract_d2t_mapping_table(std::shared_ptr<ov::Model>& model) {
```
rkazants:
```suggestion
std::shared_ptr<ov::op::v0::Constant> extract_d2t_mapping_table(const std::shared_ptr<ov::Model>& model) {
```

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+    // extract result nodes from model
+    for (const auto& result : model->get_results()) {
+        auto input_node = result->input_value(0).get_node_shared_ptr();
+        auto constant = ov::as_type_ptr<ov::op::v0::Constant>(input_node);
+        if (constant && constant->get_friendly_name().find("d2t") != std::string::npos) {
+            return constant;
+        }
+    }
+    return nullptr;
```
rkazants, line 98: please remove this "Constant->Result" sub-graph after d2t extraction.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+}
+
+void hidden_state_transform(std::shared_ptr<ov::Model>& model,
+                                  const std::vector<int>& hidden_layers_to_abstract) {
```
rkazants: please do the following changes:
1. use explicitly `int32_t` or `int64_t`
2. document this function what is it doing and why is it needed?

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+        for (int idx : hidden_layers_to_abstract) {
+            patterns.emplace_back("layers." + std::to_string(idx) + "/"); // main description
+        }
+    } else {
+        patterns.emplace_back("midlayer"); // draft description
+    }
+
+    // Helper: check if node is a residual Add node with expected structure
+    auto is_residual_node = [](const std::shared_ptr<ov::Node>& node) -> bool {
+        if (auto add = ov::as_type_ptr<ov::op::v1::Add>(node)) {
```
rkazants:
```suggestion
        if (const auto& add = ov::as_type_ptr<ov::op::v1::Add>(node)) {
```

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+    auto main_device = main_model_desc.device;
+    std::string draft_device = draft_model_desc.device.empty() ? main_model_desc.device : draft_model_desc.device;
+
+    ov::AnyMap draft_properties =
+        draft_model_desc.properties.empty() ? main_model_desc.properties : draft_model_desc.properties;
+
+    // main and draft model use same tokenzier, but could differ in configurations
+    // for example, llama3 draft model has different eos_token_id in config.json
+    Tokenizer main_model_tokenizer = main_model_desc.tokenizer;
+    Tokenizer draft_model_tokenizer = draft_model_desc.tokenizer;
```
rkazants:
```suggestion
    const Tokenizer& main_model_tokenizer = main_model_desc.tokenizer;
    const Tokenizer& draft_model_tokenizer = draft_model_desc.tokenizer;
```

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+    // check draft_model, retrieve d2t table if exists
+    auto d2t_tensor = extract_d2t_mapping_table(draft_model);
+    update_eagle_pipeline_params(d2t_tensor);
+}
+
+ov::Tensor ContinuousBatchingPipeline::Eagle3DecodingImpl::create_draft_input_ids(const ov::Tensor& original_input_ids) {
+    auto shape = original_input_ids.get_shape();
+    if (shape.size() == 0u || shape.back() <= 1u) {
+        return ov::Tensor(original_input_ids);
+    }
```
rkazants: Am I correct that we don't expect that this `if` branch is triggered, right. May be it is better to add a check `OPENVINO_THROW` instead of returning unmeaningful outputs from this function?

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+
+    size_t original_length = shape.back();
+    size_t new_length = original_length - 1;
+
+    ov::Tensor draft_input_ids(ov::element::i64, {1, new_length});
+
+    const int64_t* src_data = original_input_ids.data<int64_t>();
+    int64_t* dst_data = draft_input_ids.data<int64_t>();
+
```
rkazants, line 244: please leave a comment - why the first token is skipped

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+    std::copy(src_data + 1, src_data + original_length, dst_data);
+
+    return draft_input_ids;
+}
+
+void ContinuousBatchingPipeline::Eagle3DecodingImpl::update_eagle_pipeline_params(std::shared_ptr<ov::op::v0::Constant>& d2t_tensor) {
```
rkazants:
```suggestion
void ContinuousBatchingPipeline::Eagle3DecodingImpl::update_eagle_pipeline_params(const std::shared_ptr<ov::op::v0::Constant>& d2t_tensor) {
```

src/cpp/src/sampling/sampler.cpp
songbell: use constant op, we can access the underlying buffer without any extra copy, thus it brings no overhead.
otherwise, I'll have to copy the d2t buffer out, seems not needed?

src/cpp/src/sampling/sampler.hpp
songbell: fixed

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,327 @@
...
+                                  const std::vector<GenerationConfig>& sampling_params) {
+        OPENVINO_ASSERT(!streamer_ptr->has_callback() ||
+                        (input_ids.size() == 1 &&
+                         (sampling_params[0].is_greedy_decoding())),
+                        "Eagle3 streaming only supports batch size=1 with greedy");
+    };
+    strategy.start_timer = [](){
+        return std::chrono::steady_clock::now();
+    };
+    strategy.stop_timer = [](TimePoint start){
```
rkazants:
```suggestion
    strategy.stop_timer = [](const TimePoint& start){
```

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.hpp
```diff
@@ -0,0 +1,53 @@
...
+class ContinuousBatchingPipeline::Eagle3DecodingImpl : public ContinuousBatchingPipeline::SpeculativeDecodingImpl {
+public:
+    template<class Impl>
+    friend std::vector<EncodedGenerationResult> generate_common(
+        Impl*,
+        const std::vector<ov::Tensor>&,
+        const std::vector<GenerationConfig>&,
+        const StreamerVariant&,
+        std::optional<std::vector<ov::Tensor>>,
+        GenerateStrategy&);
```
rkazants, line 30:
```suggestion
        GenerateStrategy&);

```

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.cpp
songbell: we either have to prune it here(by making a ROI tensor), or pruning/copying before scheduling to draft model.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: to match the algo in https://github.com/SafeAILab/EAGLE/blob/main/eagle/model/cnets.py#L683
add a comment in code as well

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: current test draft models does not have its own embed_tokens in the checkpoint. we need to share it from main model.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: fixed

src/cpp/src/speculative_decoding/continuous_batching_for_speculative_decoding_impl.hpp
songbell: eagle_mode_enabled is member in base class, and cannot use initialize list to init it.
update indenting

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: fixed

tests/python_tests/samples/test_speculative_decoding_lm.py
```diff
@@ -26,22 +46,27 @@ def test_sample_speculative_decoding_lm(self, convert_model, convert_draft_model
...
 
+Question: Can you please add 2 and 3
+A:"""
+class TestEagle3SpeculativeDecodingLM:
+    @pytest.mark.llm
+    @pytest.mark.samples
+    @pytest.mark.parametrize(
+        "convert_model, convert_draft_model, sample_args",
+        [
+            pytest.param("Qwen3-1.7B", "qwen3_1.7b_eagle3", test_prompt, marks=pytest.mark.skip(reason='CVS-171947, CVS-171943, CVS-174959')),
```
rkazants: why is this skipped if you are using developer branch for eagle 3 in GHA now?

tests/python_tests/test_continuous_batching.py
```diff
@@ -612,3 +637,34 @@ def run_extended_perf_metrics_collection(
...
+    main_model_schema = download_and_convert_model(main_model)
+    main_opt_model = main_model_schema.opt_model
+    main_hf_tokenizer = main_model_schema.hf_tokenizer
+    main_model_path = main_model_schema.models_path
+
+    draft_model_path = download_and_convert_model(draft_model).models_path
+
+    # Create OpenVINO GenAI pipeline:
+
+    ov_pipe = create_ov_pipeline(main_model_path, pipeline_type = PipelineType.SPECULATIVE_DECODING, draft_model_path = draft_model_path)
```
rkazants: did you validate eagle3 for prefix caching and rollback scenarious?

src/cpp/src/speculative_decoding/speculative_decoding_impl.hpp
```diff
@@ -11,6 +11,129 @@
...
+
+    streamer_ptr->start();
+    while (self->has_non_finished_requests()) {
+        try {
+            self->step();
+        } catch (...) {
+            self->drop_requests();
+            streamer_ptr->end();
+            std::rethrow_exception(std::current_exception());
+        }
```
rkazants, line 85: please avoid try-catch

tests/python_tests/test_continuous_batching.py
songbell: this is not enabled not, tracked in CVS-177647

tests/python_tests/samples/test_speculative_decoding_lm.py
songbell: fixed, forget to revert the skip

src/cpp/src/speculative_decoding/speculative_decoding_impl.hpp
songbell: this try catch is needed for handle resource clean.  I need to leave it here, and optimize in another PR if needed.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: no, matmul is not a result node, I checked the graph is as expected.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,330 @@
+// Copyright (C) 2023-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+#include "speculative_decoding_eagle3_impl.hpp"
+#include "logger.hpp"
+
+namespace ov::genai {
+void share_embedding_weights(const std::shared_ptr<ov::Model>& main_model, const std::shared_ptr<ov::Model>& draft_model) {
```
popovaan, line 10: Please add a comment why we need to share weights between draft and main model.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,330 @@
...
+}
+
+std::shared_ptr<ov::op::v0::Constant> extract_d2t_mapping_table(const std::shared_ptr<ov::Model>& model) {
+    // extract result nodes from model
+    for (const auto& result : model->get_results()) {
+        auto input_node = result->input_value(0).get_node_shared_ptr();
+        auto constant = ov::as_type_ptr<ov::op::v0::Constant>(input_node);
+        if (constant && constant->get_friendly_name().find("d2t") != std::string::npos) {
+            model->remove_result(result);
+            model->validate_nodes_and_infer_types();
```
popovaan, line 94: Can we run `validate_nodes_and_infer_types()` once after the loop?

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
```diff
@@ -0,0 +1,330 @@
...
+    }
+    OPENVINO_ASSERT(
+        hidden_layers_to_abstract.size() == 3 || hidden_layers_to_abstract.size() == 1,
+        "Expected exactly 1 or 3 hidden layers for extraction: 1 for draft model, 3 for main model (early/middle/late stages)."
+    );
+
+    std::vector<std::string> patterns;
+    if (hidden_layers_to_abstract.size() > 1) {
+        patterns.reserve(hidden_layers_to_abstract.size());
+        for (int idx : hidden_layers_to_abstract) {
```
popovaan:
```suggestion
        for (int32_t idx : hidden_layers_to_abstract) {
```

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: added

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: this is actually only run once. It is currently called inside the loop, but only when a matching constant is found and a result is removed.

src/cpp/src/speculative_decoding/speculative_decoding_eagle3_impl.cpp
songbell: fixed

sbalandi (approved):
LGTM

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -11,14 +11,60 @@
...
+                eagle_rt_info.hidden_layers_list = it->second.as<std::vector<int32_t>>();
+                config.erase("hidden_layers_list");
+            } catch (const std::exception&) {
+                OPENVINO_THROW("please check the hidden layers input");
+            }
+        } else {
+            // compute the layers from number of hidden layers
+            auto config_file_path = models_path / "config.json";
+            if (!std::filesystem::exists(config_file_path))
+                OPENVINO_THROW("Cannot deduce layers for hidden layer extraction because the file is missing: ", config_file_path);
```
Wovchena:
```suggestion
            OPENVINO_ASSERT(std::filesystem::exists(config_file_path), "Cannot deduce layers for hidden layer extraction because the file is missing: ", config_file_path);
```

PR #3064, opened 2025-11-25
Fix 'Can not access non-const pointer' exception when running Phi-4-multimodal model

mateusztabaka:


src/cpp/src/visual_language/phi3_vision/classes.cpp
```diff
@@ -646,17 +646,18 @@ ov::Tensor insert_image_placeholders(
     return merged;
 }
 
-std::vector<std::variant<ov::Tensor, size_t>> drop_image_placeholders(const ov::Tensor& tokens) {
+std::vector<std::variant<ov::Tensor, size_t>> drop_image_placeholders(int64_t* tokens, const ov::Shape& shape) {
     std::vector<std::variant<ov::Tensor, size_t>> chunks;
-    int64_t last_token = tokens.data<int64_t>()[0];
```
Wovchena, line 651: Is there a way to use const version of `data()`? The underlying memory isn't affected.

@sgonorov, please update your PR to include the fix if it's truly needed https://github.com/openvinotoolkit/openvino.genai/pull/3026

@mateusztabaka, as you can see I'm leaving it to @sgonorov. Let us know if you aren't OK with that

mateusztabaka: > Is there a way to use const version of data()? The underlying memory isn't affected.

Yes, however non-const pointer has to be passed to ov::Tensor constructor to make it work.

> @mateusztabaka, as you can see I'm leaving it to @sgonorov. Let us know if you aren't OK with that

I'm fine with that. @sgonorov let me know when you pull my change so I can test your branch then.

sgonorov: Fixed in PR: https://github.com/openvinotoolkit/openvino.genai/pull/3072

as-suvorov:
Clossing as addressed in https://github.com/openvinotoolkit/openvino.genai/pull/3072

PR #3072, opened 2025-11-25
Fix const in phi3 qwen and minicpm

sgonorov:
Fix const cast in phi3 qwen and minicpm

### DIFF
--- a/tests/cpp/block_manager.cpp
+++ b/tests/cpp/block_manager.cpp
@@ -15,7 +15,7 @@ TEST(TestBlockManager, general_test) {
     ov::genai::SequenceGroup::Ptr sequence_group = std::make_shared<ov::genai::SequenceGroup>(
         0,
         ov::Tensor(ov::element::i64, {
-        prompt_ids.size()}, const_cast<int64_t*>(prompt_ids.data())),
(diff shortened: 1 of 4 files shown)


src/cpp/src/visual_language/phi3_vision/classes.cpp
```diff
@@ -665,7 +665,7 @@ std::vector<std::variant<ov::Tensor, size_t>> drop_image_placeholders(const ov::
                 std::in_place_type<ov::Tensor>,
                 ov::element::i64,
                 ov::Shape{1, offset - text_start},
-                tokens.data<int64_t>() + text_start
+                const_cast<int64_t*>(tokens.data<int64_t>()) + text_start
```
Wovchena, line 669: As Pawel suggested add comments in the next PR that the cast is safe and why

Wovchena (approved):

src/cpp/src/visual_language/phi3_vision/classes.cpp
sgonorov: Done. Also removed some redundant const_casts i've introduced before by mistake.

PR #3073, opened 2025-11-25
Fix for deadlock in python callback

sgonorov:
Fix update from release branch

### DIFF
--- a/src/python/py_image_generation_pipelines.cpp
+++ b/src/python/py_image_generation_pipelines.cpp
@@ -210,8 +210,22 @@ public:
                 }
             }
 
-            TorchTensorAllocator(const TorchTensorAllocator&) = default;
-            TorchTensorAllocator& operator=(const TorchTensorAllocator&) = default;
+            TorchTensorAllocator(const TorchTensorAllocator& other)
+                : m_total_size(other.m_total_size), m_mutable_data(other.m_mutable_data) {
+                py::gil_scoped_acquire acquire;
+                m_torch_tensor = other.m_torch_tensor;
+            }
+
+            TorchTensorAllocator& operator=(const TorchTensorAllocator& other) {
+                if (this != &other) {
+                    m_total_size = other.m_total_size;
+                    m_mutable_data = other.m_mutable_data;
+                    py::gil_scoped_acquire acquire;
+                    m_torch_tensor = other.m_torch_tensor;
+                }
+                return *this;
+            }
+
             TorchTensorAllocator(TorchTensorAllocator&&) = default;
             TorchTensorAllocator& operator=(TorchTensorAllocator&&) = default;
 


src/python/py_utils.cpp
```diff
@@ -374,10 +374,21 @@ ov::Any py_object_to_any(const py::object& py_obj, std::string property_name) {
         return py::cast<std::shared_ptr<ov::genai::Generator>>(py_obj);
     } else if (py::isinstance<py::function>(py_obj) && property_name == "callback") {
         auto py_callback = py::cast<py::function>(py_obj);
+        auto shared_callback = std::shared_ptr<py::function>(
+            new py::function(py_callback),
+            [](py::function* f) {
+                if (Py_IsInitialized()) {
+                    PyGILState_STATE gstate = PyGILState_Ensure();
+                    delete f;
+                    PyGILState_Release(gstate);
```
Wovchena: What are the results of running tests without acquired GIL? Do the really fail and this code is required?

Wovchena: Why not `gil_scoped_acquire`?

src/python/py_utils.cpp
```diff
@@ -443,12 +454,25 @@ ov::genai::StreamerVariant pystreamer_to_streamer(const PyBindStreamerVariant& p
...
+                    }
+                }
+            );
+
+            auto callback_wrapped = [shared_callback](std::string subword) -> ov::genai::StreamingStatus {
                 py::gil_scoped_acquire acquire;
-                auto py_str = PyUnicode_DecodeUTF8(subword.data(), subword.length(), "replace");
-                std::optional<uint16_t> callback_output = py_callback(py::reinterpret_borrow<py::str>(py_str));
+                PyObject* py_str = PyUnicode_DecodeUTF8(subword.data(), subword.length(), "replace");
+                if (!py_str) {
```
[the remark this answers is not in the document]
sgonorov: Used PyErr_WriteUnraisable

src/python/py_utils.cpp
sgonorov: Yes, if GIL is not there it means we have undefined behaviour, it fails from time to time.

sgonorov: Replaced it with gil_scoped_acquire - I've tried using C-api to check how it works. But looks like it doesn't make any difference.

src/python/py_image_generation_pipelines.cpp
```diff
@@ -195,12 +197,24 @@ class TorchGenerator : public ov::genai::CppStdGenerator {
         class TorchTensorAllocator {
             size_t m_total_size;
             void * m_mutable_data;
-            py::object m_torch_tensor; // we need to hold torch.Tensor to avoid memory destruction
+            py::object m_torch_tensor;
```
Wovchena: Why? leave the commeny as it was

src/python/py_image_generation_pipelines.cpp
```diff
@@ -195,12 +197,24 @@ class TorchGenerator : public ov::genai::CppStdGenerator {
...
 
         public:
             TorchTensorAllocator(size_t total_size, void * mutable_data, py::object torch_tensor) :
                 m_total_size(total_size), m_mutable_data(mutable_data), m_torch_tensor(torch_tensor) { }
 
+            ~TorchTensorAllocator() {
+                if (m_torch_tensor && Py_IsInitialized()) {
```
Wovchena, line 207: m_torch_tensor is always set in constructor

Wovchena (changes requested):

src/python/py_image_generation_pipelines.cpp
sgonorov: Yes, but when we're using move it may be an empty Python object as far as i understand. So it's a little bit defensive here.

src/python/py_image_generation_pipelines.cpp
sgonorov: Restored 👍

src/python/py_image_generation_pipelines.cpp
Wovchena: is `TorchTensorAllocator` movable? It defines a constructor which should disable default move constructot

sgonorov: Yes, it was movable before i've added destructor (constructor doesn't disable generation of default move methods). When i've added destructor it became unmovable, so i've had to specify default move constructor and copy/assign methods.

Wovchena: I guess copy constructor copies a ref to `m_torch_generator`. Copy methods should be deleted to avoid this problem

sgonorov: Yes, but that's the old behaviour so i haven't touched it. Also all python objects have ref counter, so it's okay to copy them if needed. The only potential problem i see here is that we have to explicitly get GIL lock  before calling copy methods.

Wovchena: > we have to explicitly get GIL lock before calling copy methods.

Good catch! Apply that

PR #3082, opened 2025-11-27
I2I Python deadlock tests and fixes

sgonorov:
Fix a deadlock occuring when using `TorchGenerator` which was not GIL-safe.  Looks like the initial threading fix revealed the GIL bug. Previously the broken synchronization accidentally prevented the deadlock.


MaximProshin:
@aleksandr-mokrov @sgonorov , please validate it notebooks to confirm that it fixes the problem finally.

aleksandr-mokrov:
@MaximProshin, unfortunately, the wheels built in the  precommit CI are not compatible with openvino 2025.4.0.0.rc3: `openvino_genai/libopenvino_genai.so.2540: undefined symbol: _ZN2ov3Any4Base9to_stringB5cxx11Ev`.  And I was unable to find a compatible version of openvino.
However, I verified it with self-built wheels. There are no deadlocks now in InpaintingPipeline and Image2ImagePipeline.

PR #3083, opened 2025-11-27
Removed not needed code from apply_paged_attention_transformation().

popovaan:
After this PR https://github.com/openvinotoolkit/openvino/pull/32570 `SDPAToPagedAttention` sets 4D dynamic shapes and dynamic element type for KV cache and sets required runtime info with k/v_head_size and num_k/v_head, so there's no need to do it in GenAI anymore.

Also after this PR PA pipeline fails with following error during the attempt to get shape from Parameter node with dynamic dimention:
```
Exception from src/core/src/dimension.cpp:227:
Cannot get length of dynamic dimension
```

## Checklist:
- This patch fully addresses the ticket. 
- I have made corresponding changes to the documentation.


popovaan:
Closing, as there's already a PR addressing the issue: https://github.com/openvinotoolkit/openvino.genai/pull/2937

PR #3086, opened 2025-11-28
Fix const casting in non-writable numpy-based tensors

sgonorov:
Fix const casting in non-writable numpy-based tensors.

### DIFF
--- a/tests/python_tests/test_llm_pipeline.py
+++ b/tests/python_tests/test_llm_pipeline.py
@@ -184,11 +184,11 @@ def test_encoded_inputs(
     assert np.all(ov_res == hf_res)
 
 
-@pytest.mark.parametrize("llm_model", ['katuni4ka/tiny-random-phi3'], indirect=True)
+@pytest.mark.parametrize("llm_model", ["katuni4ka/tiny-random-phi3"], indirect=True)
 def test_readonly_input_tensor(ov_pipe: ov_genai.LLMPipeline) -> None:
     input_ids = np.array([[1, 4, 42]], dtype=np.int64)
     input_ids.flags.writeable = False
-    
+
     attention_mask = np.array([[1, 1, 1]], dtype=np.int64)
     attention_mask.flags.writeable = False
 


src/cpp/src/image_generation/image_processor.cpp
```diff
@@ -130,7 +130,10 @@ ov::Tensor ImageResizer::execute(ov::Tensor image, int64_t dst_height, int64_t d
     target_spatial_tensor.data<int64_t>()[0] = dst_height;
     target_spatial_tensor.data<int64_t>()[1] = dst_width;
 
-    m_request.set_tensor("image", image);
+    ov::Tensor image_copy(image.get_element_type(), image.get_shape());
+    image.copy_to(image_copy);
+
+    m_request.set_tensor("image", image_copy);
```
Wovchena: Was it broken? Should be reported as OV bug if so

Wovchena (changes requested):

src/cpp/src/image_generation/image_processor.cpp
sgonorov: No, we just needed tensor with non-const data for `set_tensor`, but we were receiving tensor with the const data. So i don't see it as broken in OV right now.

Wovchena: It's a shame that a copy is needed just to please the API. I view it as a problem. @praasz, can you take a look

praasz: I will check it

praasz: The data copy for inputs should be not required. It should be possible use Inputs tensor which are read-only.
The exception on `infer()` call is error on OV side, investigation in progress.

AlexanderKalistratov: > The data copy for inputs should be not required. It should be possible use Inputs tensor which are read-only. The exception on `infer()` call is error on OV side, investigation in progress.

@praasz 
FYI: unnecessary non-const access which I found while debugging this issue:
https://github.com/openvinotoolkit/openvino/blob/master/src/inference/src/dev/isync_infer_request.cpp#L289

```c++
    OPENVINO_ASSERT(
        std::dynamic_pointer_cast<ov::IRemoteTensor>(tensor._ptr) || tensor->data() != nullptr || is_dynamic,
        "Tensor data equal nullptr!");
}
```

Here: **`tensor->data()`**

Though there should be more places like this. Since fixing this one allow to move forward a bit, but didn't completely solve the issue

sgonorov: @AlexanderKalistratov Do we have an issue for that, so i can reference? Or should i create one?

sgonorov: @Wovchena do we still need this fix or it's better to wait for it to be resolved on OV side?

Wovchena: Wait for OV

praasz: The openvinotoolkit/openvino#33120 can be used for verification if that solve issue on GenAI side.

sgonorov: I've removed all fixes on our side and it works fine. I still consider merging this PR - it has some new simple unit tests which can be handy in order to detect these kind of errors in the future.

Wovchena (approved):

PR #3088, opened 2025-12-01
[NPU] Support NPUW for text-embedding models

mengweiguo:
## Description

The benefits handled by prefill-chunk in NPUW:
- Support long context
- Performance improvement 

Note:
1. NPUW PR: https://github.com/openvinotoolkit/openvino/pull/33076
2. The [tests](https://jira.devtools.intel.com/secure/attachment/5782028/text_embeddings.py) has been verified to work with both NPUW and GenAI updates.

CVS-177453

## Checklist:

### DIFF
--- a/src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
+++ b/src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp
@@ -12,6 +12,20 @@
 namespace ov {
 namespace genai {
 
+//
+// Enable NPU dynamic prompt input support, which allows text-embedding models to handle long-context inputs more
+// effectively. The following properties can determine how NPUW behaves.
+//
+// NPUW_LLM_PREFILL_CHUNK_SIZE:
+//     Controls the chunk size for prompt prefill, which determines the granularity of input dynamism.
+//     Default value: 1024.
+//
+// NPUW_F16IC:
+//     Forces subgraph interconnect tensors to use FP16 precision when they would otherwise be FP32, provided that the
+//     partitioning pipeline is enabled. Setting this property to False may improve accuracy.
+//     Default value: True
+//
+
 using EmbeddingResult = std::variant<std::vector<float>, std::vector<int8_t>, std::vector<uint8_t>>;
 using EmbeddingResults =
     std::variant<std::vector<std::vector<float>>, std::vector<std::vector<int8_t>>, std::vector<std::vector<uint8_t>>>;
--- a/src/cpp/src/rag/npu/text_embedding_pipeline.hpp
+++ b/src/cpp/src/rag/npu/text_embedding_pipeline.hpp
@@ -9,20 +9,6 @@
 namespace ov {
 namespace genai {
 
-//
-// Enable NPU dynamic prompt input support, which allows text-embedding models to handle long-context inputs more
-// effectively.
-//
-// NPUW_LLM_PREFILL_CHUNK_SIZE:
-//     Controls the chunk size for prompt prefill, which determines the granularity of input dynamism.
-//     Default value: 1024.
-//
-// NPUW_F16IC:
-//     Forces subgraph interconnect tensors to use FP16 precision when they would otherwise be FP32, provided that the
-//     partitioning pipeline is enabled. Setting this property to False may improve accuracy.
-//     Default value: True
-//
-
 InferRequest create_text_embedding_npu_request(std::shared_ptr<ov::Model>& model,
                                                const TextEmbeddingPipeline::Config& config,
                                                const ov::AnyMap& properties,


src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -211,32 +223,55 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
         }
 
         if (m_config.pad_to_max_length) {
             m_tokenization_params.insert({pad_to_max_length.name(), *m_config.pad_to_max_length});
+            is_fixed_size &= m_config.pad_to_max_length.value();
+        } else {
+            is_fixed_size = false;
         }
 
+        bool is_padding_on_left = false;
```
as-suvorov: `is_padding_on_left` used in NPU branch only, let's move it under NPU `if`

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -211,32 +223,55 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
-
-        if (device == "NPU") {
-            OPENVINO_ASSERT(!model->is_dynamic(),
-                            "NPU device does not support dynamic shapes. In order to fix model shape, set batch_size, "
-                            "max_length and pad_to_max_length in the configuration.");
-        }
-
-        model = apply_postprocessing(model, m_config);
-
+        bool is_fixed_size = true;
```
as-suvorov:
```suggestion
        bool is_seq_len_fixed = true;
```

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -211,32 +223,55 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
+
+            auto kv_pos = ov::genai::utils::get_kv_axes_pos(model);
+            utils::KVDesc kv_desc;
+            ov::AnyMap local_config;
+            local_config["NPUW_TEXT_EMBED_POST_TYPE"] = get_post_type_string(config);
+            if (m_config.max_length.has_value()) {
+                local_config["MAX_PROMPT_LEN"] = m_config.max_length.value();
+            }
+            std::tie(compiled_model, kv_desc) =
+                utils::compile_decoder_for_npu_text_embedding(model, properties, kv_pos, local_config);
```
as-suvorov: Let's move NPU specific properties processing into compile_decoder_for_npu_text_embedding.
```suggestion
                utils::compile_decoder_for_npu_text_embedding(model, properties, kv_pos, m_config);
```
or
```suggestion
                utils::compile_decoder_for_npu_text_embedding(model, properties, kv_pos, m_config.pooling_type, m_config.max_length);
```

src/cpp/src/utils.cpp
```diff
@@ -124,12 +143,14 @@ inline bool is_paged_attention_available() {
...
 }
 
-} // anonymous
+}  // namespace
 
 namespace ov {
 namespace genai {
 namespace utils {
 
+enum class ModelType { Standard, Whisper, TextEmbedding };
```
as-suvorov: nitpick, not clear what standard model type is
```suggestion
enum class ModelType { Default, Whisper, TextEmbedding };
```

as-suvorov:
@mengweiguo You applied formatting for updated files. We have several PRs to enable formatting. I'm a bit concerned if your changes would match our fromatting rules.
If you want to apply formatting for updated files please align your PR with https://github.com/openvinotoolkit/openvino.genai/pull/3008 pre commit rules. Otherwise please revert formatting changes.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done. Take `m_config` as parameter.

src/cpp/src/utils.cpp
mengweiguo: Done.

mengweiguo:
> Otherwise please revert formatting changes.

Reverted. Thanks.

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -211,32 +211,47 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
+        if (should_reshape_non_npu || should_reshape_npu) {
+            reshape_model(model);
+        }
+
+        ov::CompiledModel compiled_model;
+        if (device == "NPU" && model->is_dynamic()) {
+            bool is_padding_on_left = m_config.padding_side.has_value() && m_config.padding_side.value() == "left";
+            if (is_padding_on_left && is_seq_len_fixed &&
+                config.pooling_type != TextEmbeddingPipeline::PoolingType::MEAN) {
+                OPENVINO_THROW("Padding on left is only supported for the mean post-processing type");
```
as-suvorov: Just to double check: do you support last token pooling with right padding (token positions calculated from attention_mask) inside plugin?

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -666,7 +666,7 @@ def create_genai_text_embed_model(model_path, device, memory_data_collector, **k
 
     pooling_type = kwargs.get("emb_pooling_type")
     max_length = kwargs.get("emb_max_length")
-    padding_side = kwargs.get("embedding_padding_side")
+    padding_side = kwargs.get("emb_padding_side")
```
as-suvorov, line 670: @sbalandi argument names were not aligned, thus option was lost. Consider arguments review and potentially introduce types.
@mengweiguo thanks!

as-suvorov (approved):

as-suvorov:
@mengweiguo we also need tests for NPUW, could you please implement simple test at: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_rag.py

mengweiguo:
> could you please implement simple test

Sure, will do that.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Yes, this case is supported inside plugin.

mengweiguo:
Tests on the NPU have been added. Because the outputs from NPU and CPU do not match exactly, I increased the threshold for NPU tests.

tests/python_tests/test_rag.py
```diff
@@ -351,6 +352,113 @@ def test_qwen3_embedding(emb_model, dataset_documents, config):
...
+            pad_to_max_length = True,
+            pooling_type=TextEmbeddingPipeline.PoolingType.MEAN,
+            padding_side="left"
+        ),
+    ],
+)
+@pytest.mark.xfail(condition=(sys.platform == "darwin"), reason="Ticket - 174635")
+def test_qwen3_embedding_npu(emb_model, config):
+    # Use the entire TEXT_DATASET as a single document instead of chunks
+    flat_documents = [TEXT_DATASET]
```
as-suvorov: Should we test multibatch as well?

tests/python_tests/test_rag.py
```diff
@@ -351,6 +352,113 @@ def test_qwen3_embedding(emb_model, dataset_documents, config):
...
+        device="CPU"
+    )
+    embeddings_genai_npu = run_text_embedding_genai(
+        emb_model.models_path,
+        flat_documents,
+        config,
+        "embed_documents",
+        device="NPU"
+    )
+    validate_embedding_results(embeddings_genai_npu, embeddings_genai_cpu, 0.25)
```
as-suvorov: Quite a big error threshold, is it expected?

tests/python_tests/test_rag.py
mengweiguo: NPU doesn't support multi-batch.

tests/python_tests/test_rag.py
mengweiguo: I previously used the static text-embedding running on NPU as a reference to compare outputs with the CPU, but the difference is also somewhat large.

tests/python_tests/test_rag.py
as-suvorov: Is it NPUW postprocessing limitation?
We have a test for NPUW with CPU fallback wich worked with multibatch: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_rag.py#L504

as-suvorov: I also forgot to mention - CI has no real NPU device, we have to use CPU fallback

mengweiguo: 1. The batch is unrolled or detached before go into NPUW.  And I also need a big chunk size( >200) to enable prefill-chunk in NPUW, so take the entire prompt as a single input.
2. Got it. I will switch to CPU fallback. Thanks.

as-suvorov: 1. I didn't get this. TextEmbeddingPipeline public API supports multiple documents to be passed to embed method. This documents can be of any length. What would happen if several small documents passed to embedding pipeline on NPU device?

mengweiguo: The NPU plugin will try to unroll or detach batch.  During inference, iterate over the documents and feed them to the NPU. So this is not true batch processing.
The documents can be of any length. I just need a long chunk to cover my test case.

mengweiguo: Changed. I introduced the threshold parameter to make the tests adaptable to different scenarios.

mengweiguo:
@as-suvorov hi, sorry to interrupt you. I have a update on post-processing.
1. The post-processing operations cannot be appended to the embedding model when entering NPUW. Previously, I duplicated the post-processing code within NPUW, but this approach is difficult to maintain and makes extending post types in the future challenging.
2. To address this, I’ve created a separate post-processing model to handle these operations in GenAI.  We have the flow below:

 ~~~
input Token -> Embedding model -> output Token  -> post model -> final output 
~~~

With this change, post-processing is now centralized in a single location. Could you please review it once more? Thanks!

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -125,31 +126,71 @@ std::shared_ptr<op::Op> get_last_token_pooling_op(std::shared_ptr<Model> model,
...
 }
 
+std::shared_ptr<ov::Model> create_post_model(std::shared_ptr<ov::Model> model,
+                                            const TextEmbeddingPipeline::Config& config,
+                                            ov::Dimension::value_type max_prompt_size) {
+    auto output_node = model->outputs()[0];
+    auto output_shape = output_node.get_partial_shape();
+    auto input_param =
+        std::make_shared<ov::op::v0::Parameter>(output_node.get_element_type(), ov::PartialShape{1, max_prompt_size, output_shape[2]});
+    set_node_name(input_param, "input_ids");
```
as-suvorov:
```suggestion
    set_node_name(input_param, "embedding_hidden_state");
```

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -321,6 +385,28 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
+    ov::Tensor post_model_infer(ov::Tensor input) {
+        if (m_post_request) {
+            m_post_request.set_tensor("input_ids", input);
+
+            auto attention_mask_tensor = m_post_request.get_tensor("attention_mask");
+
+            std::copy_n(m_attention_mask.data<int64_t>(),
+                    m_attention_mask.get_size(),
+                    attention_mask_tensor.data<int64_t>());
+            if (m_attention_mask.get_size() < attention_mask_tensor.get_size()) {
```
as-suvorov: When this is the case?

as-suvorov:
> @as-suvorov hi, sorry to interrupt you. I have a update on post-processing.
> 
> 1. The post-processing operations cannot be appended to the embedding model when entering NPUW. Previously, I duplicated the post-processing code within NPUW, but this approach is difficult to maintain and makes extending post types in the future challenging.
> 2. To address this, I’ve created a separate post-processing model to handle these operations in GenAI.  We have the flow below:
> 
> ```
> input Token -> Embedding model -> output Token  -> post model -> final output 
> ```
> 
> With this change, post-processing is now centralized in a single location. Could you please review it once more? Thanks!

1. Why post-processing operations cannot be appended for NPUW? This 2 model split can make things a bit more complex for future features. We might want to use async infer queue for embedding pipeline.
2. Nitpick, there are no tokens after embedding model: `input Token -> Embedding model -> embeddings  -> post model -> final output `

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: NPUW can support long context with prefill-chunk. 
For example, 8K tokens input may not work on NPU. So we split 8K into eight chunks of 1K each and reshape model to 1K input. Then 8 iterations complete the inference.
For the case of `max-length`=8K and real input 7.5K, the output of embedding from NPUW is 8K and the input for post model is also 8K(`attention_mask_tensor`=8K) while the real mask is only 7.5K. So need to fill the dirty slots  with 0.

It also explains why post ops can't append to embedding model because the runtime model is not matched with the original  static model. 

I think the post model does not affect asynchronous inference for embedding. The post model now uses synchronous inference because the model is small enough.

as-suvorov: Can we add attention mask normalization for NPUW (7.5k -> 8k) as a step in postprocessing and then merge the whole postprocessing into model?

mengweiguo: No. The 7.5K value is learned during the inference stage, not during model compilation.

as-suvorov: We can get tensor shape in postprocessing

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: For example, 8K tokens input may not work on NPU. So we split 8K into eight chunks of 1K each and reshape model to 1K input. Then 8 iterations complete the inference. 
The output shape of Embedding is 1x1024x1024, and the entire output shape should be 1x8192x1024. The post model can append to the embedding with 1x8192x1024, but not the one with 1x1024x1024.

as-suvorov: Please add comment to the code expalining why this if is needed. Would simplify understanding

mengweiguo: Done

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -678,7 +678,8 @@ def create_genai_text_embed_model(model_path, device, memory_data_collector, **k
             config.pooling_type = openvino_genai.TextEmbeddingPipeline.PoolingType.CLS
     if max_length is not None:
         config.max_length = max_length
-        config.pad_to_max_length = True
+
+    config.pad_to_max_length = kwargs.get("emb_pad_to_max_length", False)
```
mengweiguo: Should we set the default value to `True` so that it may remain compatible with previous behavior? @as-suvorov

as-suvorov: Good point. Let's set it to `True` but only if `max_length is not None` to preserve behavior.

mengweiguo: Change `embedding_pad_to_max_length` to `embedding_disable_pad_to_max_length`.
1. `embedding_pad_to_max_length` is not used before.
2. Go into NPUW only when `max_length` is True and `pad_to_max_length` is False. If `pad_to_max_length` follows `max_length`, we cannot test NPUW embedding using benchmark.py
3. By default, `embedding_disable_pad_to_max_length` is set to False. To test NPUW embedding, explicitly add the `--embedding_disable_pad_to_max_length` flag.

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -211,32 +251,51 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
+        if (should_reshape_non_npu || should_reshape_npu) {
+            reshape_model(model);
+        }
+
+        ov::CompiledModel compiled_model;
+        if (device == "NPU" && model->is_dynamic()) {
+            bool is_padding_on_left = m_config.padding_side.has_value() && m_config.padding_side.value() == "left";
+            if (is_padding_on_left && is_seq_len_fixed &&
+                config.pooling_type != TextEmbeddingPipeline::PoolingType::MEAN) {
+                OPENVINO_THROW("Padding on left is only supported for the MEAN pooling type");
```
as-suvorov:
```suggestion
                OPENVINO_THROW("Padding on left is only supported for the MEAN pooling type for dynamic inputs model. In order to fix model shape, set batch_size, "
                            "max_length and pad_to_max_length in the configuration.");
```

tools/llm_bench/benchmark.py
```diff
@@ -210,6 +210,7 @@ def get_argprser():
                         help="Pooling type CLS or MEAN for encoders, LAST_TOKEN for decoders. "
                              "Different post-processing is applied depending on the padding side. Applicable only for text embeddings")
     parser.add_argument("--embedding_normalize", action="store_true", help="Normalize embeddings. Applicable only for text embeddings")
+    parser.add_argument("--embedding_disable_pad_to_max_length", action="store_true", help="Disable padding embeddings. Applicable only for text embeddings")
```
as-suvorov: Can we get rid of negation: `embedding_pad_to_max_length`?

mengweiguo: I did have a try but didn't find out a good method.

The `benchmark `previously ran `text_embed` on the NPU using the minimal set of arguments below, not explicitly setting `--embedding_pad_to_max_length`. It assume this argument is set Ture following the `--embedding_max_length` setting.

~~~
benchmark.py   -d NPU --task "text_embed" -m Qwen3-Embedding-0.6B --embedding_max_length 4096 
~~~

Then the changes of embedding pipeline doesn't impact on the behavior of this command. And work with NPUW with command below.
~~~
benchmark.py   -d NPU --task "text_embed" -m Qwen3-Embedding-0.6B --embedding_max_length 4096 --embedding_disable_pad_to_max_length
~~~

If  keep `embedding_pad_to_max_length`, we don't have method to explicitly set `embedding_pad_to_max_length` False since  it follows the `--embedding_max_length` setting.  Otherwise we have to separate `embedding_max_length` and `embedding_max_length`.  However, original NPU command needs to explicitly add `--embedding_pad_to_max_length`.

~~~
benchmark.py   -d NPU --task "text_embed" -m Qwen3-Embedding-0.6B --embedding_max_length 4096  --embedding_pad_to_max_length
~~~

PS: I don't like the negation too ...

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done. Thanks.

tools/llm_bench/benchmark.py
as-suvorov: From what I see in the implementation you already separated setting `max_length` and `pad_to_max_length`. Isn't so?
And regarding original command and NPU, I guess now pipeline expected to work without setting pad_to_max length? So we can actually change this default safely

mengweiguo: Originally, the default value of `embedding_pad_to_max_length` is False.  
~~~
parser.add_argument("--embedding_pad_to_max_length", action="store_true", help="Pad embeddings. Applicable only for text embeddings")
~~~

If set the default value True,  we might have definition below.
~~~
parser.add_argument("--embedding_pad_to_max_length", action="store_false", help="Pad embeddings. Applicable only for text embeddings")

or
 
parser.add_argument("--embedding_pad_to_max_length", action="store_true", default=True, help="Pad embeddings. Applicable only for text embeddings")
~~~

These 2 statements looks weird.  May we have other better approach?

as-suvorov: Originally you mean in this PR?
I'm a bit lost, what is wrong with this definition?
```
parser.add_argument("--embedding_pad_to_max_length", action="store_false", help="Pad embeddings. Applicable only for text embeddings")
```

mengweiguo: When the command includes `--embedding_pad_to_max_length`, the actual behavior is to disable padding, which seems opposite to what the argument name suggests. 
I'm OK if this definition is fine. Will update this argument.

as-suvorov: The default tokenizer behavior is not pad to max length. Or am I missing something? Where we disable padding with `--embedding_pad_to_max_length=True`?

mengweiguo: On NPU, dynamic inputs needs to reshape to static before. So must set `embedding_max_length` and  `embedding_pad_to_max_length` is set True.  

But if want to work with NPUW, we need to set `embedding_pad_to_max_length` False. So if set `embedding_pad_to_max_length` default value True, we need to be able to set `embedding_pad_to_max_length` False.

Here `action="store_false` means `--embedding_pad_to_max_length` is set false if it appears in command line.

as-suvorov: We aligned here, I still see no issues with:
```
parser.add_argument("--embedding_pad_to_max_length", action="store_false", help="Pad embeddings. Applicable only for text embeddings")
```
And still don't get this statement:
`When the command includes --embedding_pad_to_max_length, the actual behavior is to disable padding, which seems opposite to what the argument name suggests.`

For the public API we need to be clear with argument names `--embedding_pad_to_max_length` set to `False` by default make sense to me.

mengweiguo: Done. Thanks a lot.

as-suvorov:
@mengweiguo Please fix linting: https://github.com/openvinotoolkit/openvino.genai/actions/runs/20851351356/job/59909449192?pr=3088
Tests are also failing: https://github.com/openvinotoolkit/openvino.genai/actions/runs/20851351387/job/59910419179?pr=3088

mengweiguo:
> Please fix linting: https://github.com/openvinotoolkit/openvino.genai/actions/runs/20851351356/job/59909449192?pr=3088 Tests are also failing: https://github.com/openvinotoolkit/openvino.genai/actions/runs/20851351387/job/59910419179?pr=3088

Hi, the tests failure is expected as it needs the corresponding OV. The Lint error is not related to this PR. Perhaps I need to rebase.

as-suvorov:
@mengweiguo GenAI gha takes latest OV commit from master: https://github.com/openvinotoolkit/openvino.genai/blob/master/.github/workflows/linux.yml#L96
Were corresponding changes merged to OV master?

mengweiguo:
> @mengweiguo GenAI gha takes latest OV commit from master: https://github.com/openvinotoolkit/openvino.genai/blob/master/.github/workflows/linux.yml#L96 Were corresponding changes merged to OV master?

Not yet.

mengweiguo:
@as-suvorov The [OV-PR](https://github.com/openvinotoolkit/openvino/pull/33076) is already merged in master branch. Could you help run CI? Thanks.

mengweiguo:
@as-suvorov  Fixed lint and tests. Could you run it again?   Thanks.

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -45,6 +45,11 @@ bool has_token_type_ids_input(const T& inputs) {
     return false;
 }
 
+void set_node_name(std::shared_ptr<ov::Node> node, const std::string& name) {
```
rkazants:
```suggestion
void set_node_name(const std::shared_ptr<ov::Node>& node, const std::string& name) {
```

rkazants: why this is needed? Does it mean that IR comes with wrong name?

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -211,32 +251,53 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
+
+            auto kv_pos = ov::genai::utils::get_kv_axes_pos(model);
+            utils::KVDesc kv_desc;
+            std::tie(compiled_model, kv_desc) =
+                utils::compile_decoder_for_npu_text_embedding(model, properties, kv_pos, m_config);
+
+            auto post_model = create_post_model(model, m_config);
+            auto post_compiled_model = core.compile_model(post_model, "CPU");
+            m_post_request = post_compiled_model.create_infer_request();
+        } else {
```
rkazants: let us separate NPU logic into a separate sub-directory of rag

tools/llm_bench/benchmark.py
```diff
@@ -210,28 +210,65 @@ def get_argprser():
...
+        "--embedding_padding_side",
+        choices=["left", "right"],
+        default=None,
+        help="Side to use for padding 'left' or 'right'. Applicable only for text embeddings",
+    )
+    parser.add_argument(
+        "--reranking_max_length",
+        type=int,
+        default=None,
+        help="Max length for text reranking. Input text will be padded or truncated to specified value",
```
rkazants, line 379: please do not re-format this code because now it is not clear what code is changed?
Did you add anything new here?

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: New nodes have been created and need to be assigned the correct names.

tools/llm_bench/benchmark.py
mengweiguo: The only change is the addition of a new argument, `embedding_disable_pad_to_max_length`. All other modifications are due to linting.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: The NPU-specific codebase is relatively small. Separating the NPU implementation will result in some code duplication.

Here is my plan to separate NPU.
1. Create a new file `src/cpp/src/rag/text_embedding_util.cpp` and move all post-operation code into this file.
2.  Create a new file `src/cpp/src/rag/npu/text_embedding_pipeline.cpp` and move npu processing code to this file.
3.  Here’s what `TextEmbeddingPipelineImpl` will look like in the end.
~~~
...
if (device == "NPU") {
   m_request = create_text_embedding_npu_request(...);
   m_post_request = create_text_embedding_npu_post_request(...);
} else {
 ....
}
~~~

What do you think? Thanks

tools/llm_bench/benchmark.py
rkazants: linting is not needed in this PR. If we want to do it, we will make it in separate PR dedicated to re-formatting. Now just format your new code as existing one. It will help to highlight your changes in PR.

src/cpp/src/rag/text_embedding_pipeline.cpp
rkazants: I am fine with your plan.
Now we have a new component in GenAI NPU, so let us localize NPU specific code in sub-directory then.

src/cpp/src/rag/text_embedding_pipeline.cpp
rkazants: let us move this into text_embedding_util.cpp file then.

tools/llm_bench/benchmark.py
mengweiguo: The Lint check would fail if without the other modification and I’m not sure if that’s acceptable.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done. Could you take a look? Thanks.

Wovchena:
build_jenkins

tools/llm_bench/benchmark.py
```diff
@@ -210,28 +210,65 @@ def get_argprser():
...
-                        help="Apply chat template for LLM. By default chat template is not applied. It's better to use with --disable_prompt_permutation,"
-                             " otherwise the prompt will be modified after applying the chat template, so the structure of chat template will not be kept.")
-    parser.add_argument("--speaker_embeddings", type=str, default=None,
-                        help="Path to .bin or .pt file with speaker embeddings for text to speech scenarios")
-    parser.add_argument("--vocoder_path", type=str, default=None,
-                        help="Path to vocoder  for text to speech scenarios")
-    parser.add_argument("-vf", "--video_frames", type=int, default=None,
-                        help="controller of video frames to process (required frame number if positive or decimation factor if negative)")
+    parser.add_argument(
+        "--embedding_disable_pad_to_max_length",
```
sbalandi: options with negation are difficult to understand and lead to confusion in the future, let's remove `disable`, move it to `--embedding_pad_to_max_length` and update the behavior. 

after PR merge, validation team will update scripts for embedding scenario, it would be great to get it done before next Wednesday(before next validation cycle started), disccessed with Kowalczyk, Pawel

FYI @as-suvorov

as-suvorov: @sbalandi Thank you!

mengweiguo: hi @sbalandi  As I discussed with @as-suvorov before, it is not possible to remain fully compatible with the previous behavior if we use `--embedding_pad_to_max_length` instead of `--embedding_disable_pad_to_max_length`.

Previously, when running the embedding model on NPU, `--embedding_max_length` had to be set as `config.pad_to_max_length` followed `config.max_length`. This allowed us to reshape a dynamic model into a static one. This also implied that the default value of `pad_to_max_length` on NPU was True.

Currently, the model only falls back to NPUW on NPU when it fails to become a static model. I use `config.pad_to_max_length` as a switch:

When `config.max_length = true && config.pad_to_max_length = true`, the model runs as a static model.
When `config.max_length = true && config.pad_to_max_length = false`, the model runs with NPUW.

If we switch to `--embedding_pad_to_max_length`, its default value becomes False, and users must explicitly enable it when needed. For static model scenarios, users will need to specify both `--embedding_max_length` and `--embedding_pad_to_max_length`. This introduces the need for an extra option, `--embedding_pad_to_max_length`. Please take note.

If it's OK, I would like to update this option. Thanks.

sbalandi: yes, it is ok, validation team will update the scripts, it have been discussed with @as-suvorov as well 
if you know of any other places where the cmds needs to be updated after changes, other than validation, please let me know

mengweiguo: Done.

sbalandi: please, update also optimum implementation https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/task/text_embeddings.py#L23 , it constantly pad to max length now

@as-suvorov , could you please help to determine, which behavior corresponds to the config.pad_to_max_lenght False: False/do_not_pad or True/longest (https://huggingface.co/docs/transformers/pad_truncation#padding-and-truncation) ?

as-suvorov: @sbalandi It should be `max_length`

mengweiguo: Updated.

sbalandi: have checked, if pad_to_max_length is not True, padding for AutoTokenizer should be set to True(which means padding to the actual maximum prompts length), it is already the case

what should be updated:
```
    tokenizer_kwargs = {'truncation': True, 'padding_side': args.get('emb_padding_side', 'right')}
    # if padding True, prompts will pad to maximum length of actual prompts
    tokenizer_kwargs['padding'] = 'max_length' if args.pad_to_max_length else True
    max_lenght = args.get('emb_max_length')
    if max_lenght is not None:
        tokenizer_kwargs.update({ 'max_length': max_lenght})
```

mengweiguo: Done. Retain the default value of  `padding` True if  `pad_to_max_length` is false.

mengweiguo:
Tests failed due to network error. @Wovchena Could you help re-run Jenkins?  Thanks.

mengweiguo:
The failure of the `Whisper test` is due to a `MaxRetryError` triggered when attempting to reach `huggingface.co`.

src/cpp/src/rag/text_embedding_utils.hpp
Wovchena, line 1: This is a public header. I don't think you intended to publish this. Move it out of `src/cpp/include/`

src/cpp/src/rag/npu/text_embedding_pipeline.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+std::optional<InferRequest> create_text_embedding_npu_post_request(std::shared_ptr<ov::Model>& model,
+                                                                   const TextEmbeddingPipeline::Config& config) {
+    if (model->is_dynamic()) {
+        ov::Core core = utils::singleton_core();
+        auto post_model = utils::create_post_model(model, config);
+        auto post_compiled_model = core.compile_model(post_model, "CPU");
+        return post_compiled_model.create_infer_request();
+    } else {
+        return std::nullopt;
```
Wovchena: You can return `InferRequest{}` removing `optinal` because the return value is copied into `m_post_request` which stays empty if optional is null

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -281,44 +172,49 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
...
     AnyMap m_tokenization_params;
     std::optional<size_t> m_max_position_embeddings;
+    ov::Tensor m_attention_mask;
 
-    void reshape_model(std::shared_ptr<Model>& model) {
-        ov::PartialShape target_shape{ov::Dimension::dynamic(), ov::Dimension::dynamic()};
-
-        if (m_config.batch_size.has_value()) {
-            target_shape[0] = ov::Dimension(*m_config.batch_size);
+    ov::Tensor post_model_infer(ov::Tensor input) {
```
Wovchena:
```suggestion
    ov::Tensor post_model_infer(const ov::Tensor& input) {
```

tests/python_tests/test_rag.py
```diff
@@ -349,6 +356,230 @@ def test_qwen3_embedding(emb_model, dataset_documents, config):
...
+            ),
+            16,
+            6e-3,
+            "embed_query",
+        ),
+    ],
+)
+@pytest.mark.xfail(condition=(sys.platform == "darwin"), reason="Ticket - 174635")
+def test_qwen3_embedding_npu(emb_model, dataset_documents, config, chunk_size, threshold, task):
+    NPU_FALLBACK_PROPERTIES = {"NPUW_DEVICES": "CPU", "NPUW_F16IC": "False", "NPUW_LLM_PREFILL_CHUNK_SIZE": chunk_size}
```
Wovchena, line 572: You may want to document these properties. There's https://docs.openvino.ai/2025/openvino-workflow-generative/inference-with-genai/inference-with-genai-on-npu.html for NPUW_LLM_PREFILL_CHUNK_SIZE but it focuses on LLMs. NPUW_F16IC isn't covered there. Describe them in `src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp` unless we come up with a better place

src/cpp/src/rag/text_embedding_utils.hpp
mengweiguo: Move it to directory `src/cpp/src/rag/`.

src/cpp/src/rag/npu/text_embedding_pipeline.cpp
mengweiguo: Done. Thanks.

src/cpp/src/rag/text_embedding_pipeline.cpp
mengweiguo: Done.

tests/python_tests/test_rag.py
mengweiguo: Added in file `src/cpp/include/openvino/genai/rag/npu/text_embedding_pipeline.hpp`.

Wovchena: `src/cpp/include/openvino/genai/rag/npu/text_embedding_pipeline.hpp` is public but it shouldn't be. The user interface is represented by `NPUW_LLM_PREFILL_CHUNK_SIZE` and others

mengweiguo: Move `src/cpp/include/openvino/genai/rag/npu/text_embedding_pipeline.hpp` to `src/cpp/src/rag/npu/text_embedding_pipeline.hpp`

mengweiguo: And move property description to `src/cpp/include/openvino/genai/rag/text_embedding_pipeline.hpp`

mengweiguo:
The model converted by the latest OV is slightly different with the previous one. Need to fix it in NPUW.

dmatveev:
> The model converted by the latest OV is slightly different with the previous one. Need to fix it in NPUW.

Please sync with us on the components update, assume we can use older transformers/optimum if required.

mengweiguo:
The tests related to NPU failed because the default NPU compiler type is set to `PREFER_PLUGIN` instead of `DRIVER` in the recent OV PR https://github.com/openvinotoolkit/openvino/pull/33684 . And it seems  NPUW doesn't parse the property `NPU_COMPILER_TYPE`. I can run tests successfully via env parameter `export IE_NPU_COMPILER_TYPE=DRIVER` locally.
Perhaps we should pass the parsed property to NPUW in plugin.

mengweiguo:
The PR https://github.com/openvinotoolkit/openvino/pull/33779 can fix this issue. @AlexanderKalistratov

mengweiguo:
The `Node.js bindings tests` are fine on Linux/Windows except on MacOS.  This Tests are on CPU rather than NPU.  

The issue was macOS-specific. 
1. Race condition during cleanup: The previous code called `embedDocuments()` without capturing the result, causing Node.js to begin its exit sequence while native operations were still in progress.
2. Stricter mutex handling on macOS: macOS's libc++ has stricter validation of mutex states, detecting when destructors try to lock mutexes during concurrent cleanup
3. Premature resource destruction: Without properly awaiting and storing the result, native resources were being destroyed while still in use.

I’m not sure whether I should be the one to fix it.

mengweiguo:
All failed tests can be seen in post-merge CI of the latest merged PR https://github.com/openvinotoolkit/openvino.genai/pull/3219. So I believe the test failures in this PR is not related to my change.

Wovchena:
@Retribution98, please take a look at
> libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument

https://github.com/openvinotoolkit/openvino.genai/actions/runs/21305434780/job/61441485017?pr=3088

@as-suvorov, please take a look at
> ImportError: DLL load failed while importing _core: The specified module could not be found.

https://github.com/openvinotoolkit/openvino.genai/actions/runs/21305434783/job/61441502260?pr=3088

I asked @sgonorov to fix
> RuntimeError: Check 'merged_generation_config.guidance_rescale <= 0' failed at /home/runner/_work/openvino.genai/openvino.genai/src/src/cpp/src/video_generation/ltx_pipeline.hpp:529:
Parameter 'guidance_rescale' is not currently supported by LTX Pipeline. Please, contact OpenVINO GenAI developers.

as-suvorov:
> @Retribution98, please take a look at
> 
> > libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument
> 
> https://github.com/openvinotoolkit/openvino.genai/actions/runs/21305434780/job/61441485017?pr=3088
> 
> @as-suvorov, please take a look at
> 
> > ImportError: DLL load failed while importing _core: The specified module could not be found.
> 
> https://github.com/openvinotoolkit/openvino.genai/actions/runs/21305434783/job/61441502260?pr=3088
> 
> I asked @sgonorov to fix
> 
> > RuntimeError: Check 'merged_generation_config.guidance_rescale <= 0' failed at /home/runner/_work/openvino.genai/openvino.genai/src/src/cpp/src/video_generation/ltx_pipeline.hpp:529:
> > Parameter 'guidance_rescale' is not currently supported by LTX Pipeline. Please, contact OpenVINO GenAI developers.

Potential fix: https://github.com/openvinotoolkit/openvino.genai/pull/3222
The issue is with import: `from transformers import pipeline`. It chains `VideoClassificationPipeline` and `from av._core import time_base, library_versions, ffmpeg_version_info` which fails on missing DLL. I'm not sure what have changed in CI, replaced pipeline with ASR pipeline import.

as-suvorov:
@Wovchena @Retribution98 
nodejs tests skip: https://github.com/openvinotoolkit/openvino.genai/pull/3227

mengweiguo:
It seems Whisper pipeline is broken due to OV update.
All tests related to my change passed. @Wovchena could you help merge? Thanks.

mengweiguo:
@dmatveev could you help merge?  The CI failures are known issues. Thanks.

PR #3095, opened 2025-12-03
[JS] Remove deprecated generate behavior.

Retribution98:
## Description

Since version 2026.0.0, LLMPipeline.generate() returns DecodedResults by default.
To get a string representation, call the toString() method on DecodedResults.

CVS-176152

## Checklist:


almilosz:
Are any changes necessary for `openvino-langchain`?

Retribution98:
> Are any changes necessary for `openvino-langchain`?

Yes, changes will be needed after the release of the major version of OV and the update of the openvino-langchain to use it. Until then, it will work correctly.

PR #3096, opened 2025-12-04
llm_bench: use built in sum

Wovchena:


Wovchena:
I implemented memory measurement for multiple Core instances and just noticed a shorter way to implement `get_rss_memory()`

PR #3111, opened 2025-12-12
Fix pre-commit Darker deps and pyi ignoring

sgonorov:
- Added tomli to dependencies of darker in order to support Python 3.10
- Excluded pyi files from formatting in order not to break CMake stubgen pipelines

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,6 +8,8 @@ on:
     branches:
       - master
 
+permissions: read-all
+
 env:
   OV_BRANCH: master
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -26,3 +27,4 @@ repos:
         args: ["--formatter=ruff"]
         additional_dependencies:
           - ruff==0.14.4
+          - tomli
```
Wovchena:
```suggestion
          - tomli==2.3.0
```

PR #3112, opened 2025-12-12
[JS] VLMPipeline implementation

Retribution98:
## Description
This PR introduces comprehensive support for the JavaScript bindings in the VLMPipeline.

VLMPipeline is implemented similarly to LLMPipeline, but with some improvements:
- Error handling in `vLmPerformInferenceThread` was modified for a much better user experience. It no longer throws fatal errors. Now, the thread uses a callback to reject the promise and throw a catchable exception to the user. If that is not possible, an error message is logged.
- Streamer and callback were separated to create clearer code. These functions are used in different cases and have different semantics. A callback is always used to resolve a promise with the final result or reject it with an error. The streamer is only used if it is provided by the user to get chunks.
- A wrapper property, `is_initializing`, was added to prevent the pipeline from being loaded incorrectly twice.
- A wrapper property, `is_generating`, was added to prevent the incorrect attempt to call generate in parallel. This leads to an InferRequest exception, which may confuse the user.
- VLMPerfMetrics should extend PerfMetrics. This was refactored to avoid code duplication. The Node addon API doesn't support inheritance from the box, so a [workaround](https://mmomtchev.medium.com/c-class-inheritance-with-node-api-and-node-addon-api-c180334d9902) was used to resolve this issue.
- The optional parameters for generate and stream were grouped into the options argument. This hook allows the use of named arguments and simplifies calls.

CVS-172789

## Checklist:
- [x] I have made corresponding changes to the documentation. https://retribution98.github.io/openvino.genai/

### DIFF
--- a/src/js/tests/vlmPipeline.test.js
+++ b/src/js/tests/vlmPipeline.test.js
@@ -12,11 +12,7 @@ const MODEL_PATH = process.env.VLM_MODEL_PATH || `./tests/models/${models.VLM.sp
 
 // Skip tests on macOS due to insufficient memory
 describe("VLMPipeline", { skip: process.platform === "darwin" }, () => {
-  let pipeline,
-    testImage1,
-    testImage2,
-    testVideo1,
-    testVideo2 = null;
+  let pipeline, testImage1, testImage2, testVideo1, testVideo2;
 
   before(async () => {
     pipeline = await VLMPipeline(MODEL_PATH, "CPU");
@@ -122,7 +118,10 @@ describe("VLMPipeline", { skip: process.platform === "darwin" }, () => {
     // Property from base PerformanceMetrics
     const numTokens = result.perfMetrics.getNumGeneratedTokens();
     assert.ok(typeof numTokens === "number", "getNumGeneratedTokens should return number");
-    assert.ok(0 < numTokens <= 10, "Number of tokens should be between 0 and max_new_tokens");
+    assert.ok(
+      0 < numTokens && numTokens <= 10,
+      "Number of tokens should be between 0 and max_new_tokens",
+    );
     // VLM-specific properties
     const prepareEmbeddings = result.perfMetrics.getPrepareEmbeddingsDuration();
     assert.ok(


src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+                            }
+                        } catch (std::exception& err) {
+                            streamer_exceptions.push_back(err.what());
+                            resultPromise.set_value(ov::genai::StreamingStatus::CANCEL);
+                        }
+                    });
+
+                if (status != napi_ok) {
+                    streamer_exceptions.push_back("The streamer callback BlockingCall failed with the status: " +
+                                                  status);
```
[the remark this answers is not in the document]
Retribution98: It doesn't give a user-friendly string.

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+                            env.Null(),                         // Error should be null in normal case
+                            to_vlm_decoded_result(env, result)  // Return DecodedResults as the final result
+                        });
+                    } catch (std::exception& err) {
+                        report_error("The final callback failed. Details:\n" + std::string(err.what()));
+                    }
+                });
+
+            if (status != napi_ok) {
+                report_error("The final BlockingCall failed with status " + status);
```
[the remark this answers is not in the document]
Retribution98: It doesn't give a user-friendly string.

site/docs/use-cases/image-processing/_sections/_run_model/_code_example_js.mdx
```diff
@@ -0,0 +1,41 @@
...
+    const { width, height, channels } = metadata;
+    const imageBuffer = await img.raw().toBuffer();
+    console.log("Read image: ", width, height, channels);
+    return new ov.Tensor(ov.element.u8, [height, width, channels], imageBuffer);
+}
+
+async function readImages(imagePath) {
+    const stats = statSync(imagePath);
+    if (stats.isDirectory()) {
+        const files = readdirSync(imagePath).sort();
```
yatarkan: Should we use asynchronous fs operations instead of sync versions (`statSync`, `readdirSync`) as `readImages` function is also async?

site/docs/use-cases/image-processing/_sections/_run_model/_code_example_js.mdx
```diff
@@ -0,0 +1,41 @@
...
+    }
+    return [await readImage(imagePath)];
+}
+
+const images = await readImages("./images");
+
+const pipe = await VLMPipeline(modelPath, "${props.device || 'CPU'}");
+
+const result = await pipe.generate(prompt, {
+    images: images,
```
yatarkan:
```suggestion
    images,
```

src/js/lib/addon.ts
```diff
@@ -58,9 +61,36 @@ export interface TextEmbeddingPipelineWrapper {
...
+    device: string,
+    ovProperties: VLMPipelineProperties,
+    callback: (err: Error | null) => void,
+  ): void;
+  generate(
+    prompt: string,
+    images: Tensor[] | undefined,
+    videos: Tensor[] | undefined,
+    streamer: ((chunk: string) => StreamingStatus) | undefined,
+    generationConfig: GenerationConfig | undefined,
```
yatarkan: Why these arguments are explicitly include undefined type? They can't be optional?
And callback arguments is not marked as undefined intentionally?

src/js/lib/decodedResults.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+    let result = "";
+    for (let i = 0; i < this.texts.length - 1; ++i) {
+      result += `${this.scores[i].toFixed(6)}: ${this.texts[i]}\n`;
+    }
+    result += `${this.scores[this.scores.length - 1].toFixed(
+      6,
+    )}: ${this.texts[this.texts.length - 1]}`;
+
+    return result;
```
yatarkan: [nitpick] Can be a bit simplified for more readability
```suggestion
    const lines = this.scores.map((score, i) => `${score.toFixed(6)}: ${this.texts[i]}`);
    return lines.join('\n');
```

src/js/lib/perfMetrics.ts
```diff
@@ -0,0 +1,120 @@
...
+  std: number;
+};
+
+/** Structure holding summary of statistical values */
+export type SummaryStats = {
+  mean: number;
+  std: number;
+  min: number;
+  max: number;
+};
```
yatarkan: Should it reuse MeanStdPair type?
```suggestion
export type SummaryStats = MeanStdPair & {
  min: number;
  max: number;
};
```

src/js/lib/pipelines/vlmPipeline.ts
```diff
@@ -0,0 +1,227 @@
...
+      async next() {
+        // If there is data in the queue, return it
+        // Otherwise, return a promise that will resolve when data is available
+        const data = queue.shift();
+
+        if (data !== undefined) {
+          const { done, subword } = data;
+
+          return { value: subword, done: done };
+        }
```
yatarkan:
```suggestion
        if (data) {
          return { value: data.subword, done: data.done };
        }
```

src/js/lib/pipelines/vlmPipeline.ts
```diff
@@ -0,0 +1,227 @@
...
+   * @param options.generationConfig - Generation configuration parameters.
+   * @param options.streamer - Optional streamer callback called for each chunk.
+   * @returns Resolves with decoded results once generation finishes.
+   */
+  async generate(
+    prompt: string,
+    options: {
+      images?: Tensor[];
+      videos?: Tensor[];
+      generationConfig?: GenerationConfig;
```
yatarkan: I would move these options to separate type (e.g. `VLMGenerateOptions`) and reuse in `stream()` and `generate()` methods (in case of generate() - extend type with streamer arg)

src/js/src/helper.cpp
```diff
@@ -337,6 +338,28 @@ ov::Tensor js_to_cpp<ov::Tensor>(const Napi::Env& env, const Napi::Value& value)
...
+    if (value.IsArray()) {
+        auto array = value.As<Napi::Array>();
+        size_t length = array.Length();
+        tensors.reserve(length);
+        for (uint32_t i = 0; i < length; ++i) {
+            tensors.push_back(js_to_cpp<ov::Tensor>(env, array[i]));
+        }
+    } else if (value.IsObject()) {
+        // Single tensor
+        tensors.push_back(js_to_cpp<ov::Tensor>(env, value));
```
yatarkan: Does it mean that API like `images=ov::Tensor` is supported?
I would force users to follow `images=std::vector<ov::Tensor>` API for `images` argument, e.g. in JS `generate(..., { images: [image_tensor], ...})`.

src/js/tests/models.js
```diff
@@ -2,4 +2,5 @@ export const models = {
   LLM: "OpenVINO/Llama-3.1-8B-Instruct-FastDraft-150M-int8-ov",
   InstructLLM: "OpenVINO/Qwen2.5-1.5B-Instruct-int4-ov",
   Embedding: "OpenVINO/bge-base-en-v1.5-fp16-ov",
+  VLLM: "OpenVINO/Qwen2-VL-7B-Instruct-int4-ov",
```
yatarkan: Typo?
```suggestion
  VLM: "OpenVINO/Qwen2-VL-7B-Instruct-int4-ov",
```

yatarkan: BTW, isn't this model too heavy for tests? Probably there are lighter VLM models or even tiny-random versions

src/js/tests/vlmPipeline.test.js
```diff
@@ -0,0 +1,164 @@
...
+    });
+
+    assert.strictEqual(result.texts.length, 1, "Should generate comparison");
+  });
+
+  it("should generate text with video input", async () => {
+    const testVideo = createTestVideoTensor();
+
+    const result = await pipeline.generate("Describe what happens in this video.", {
+      images: [],
```
yatarkan: Is images arg required here?

src/js/tests/utils.js
```diff
@@ -45,3 +46,46 @@ async function saveFile(file, response) {
...
+  }
+
+  return new ov.Tensor("u8", [height, width, channels], data);
+}
+
+// Helper function to create a synthetic video tensor (4 frames, 32x32x3)
+export function createTestVideoTensor() {
+  const frames = 4;
+  const height = 32;
+  const width = 32;
```
yatarkan: I would move height, width and frames to function params to be able to create different test assets. The same is for test image tensor.

src/js/tests/vlmPipeline.test.js
```diff
@@ -0,0 +1,164 @@
...
+        temperature: 0,
+      },
+    });
+
+    assert.ok(result, "Should return result");
+    assert.ok(result.perfMetrics, "Should have perfMetrics");
+    // Property from base PerformanceMetrics
+    const numTokens = result.perfMetrics.getNumGeneratedTokens();
+    assert.ok(typeof numTokens === "number", "getNumGeneratedTokens should return number");
+    assert.ok(numTokens > 0, "Should generate at least one token");
```
yatarkan: `numTokens` should be between 0 and `max_new_tokens` from `generationConfig`

src/js/tests/vlmPipeline.test.js
```diff
@@ -0,0 +1,164 @@
...
+    assert.ok(
+      result instanceof VLMDecodedResults,
+      "Result should be instance of VLMDecodedResults",
+    );
+    assert.ok(result.texts.length > 0, "Should generate some output");
+  });
+
+  it("should generate text with images", async () => {
+    const testImage1 = createTestImageTensor();
+    const testImage2 = createTestImageTensor();
```
yatarkan: I would create test images with different sizes (and videos if needed) outside the test case and reuse them in several `it(...)` test cases.

site/docs/use-cases/image-processing/_sections/_usage_options/index.mdx
```diff
@@ -6,7 +6,7 @@ import Streaming from '@site/docs/use-cases/_shared/_streaming.mdx';
 ## Additional Usage Options
 
 :::tip
-Check out [Python](https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/python/visual_language_chat) and [C++](https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/cpp/visual_language_chat) visual language chat samples.
+Check out [Python](https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/python/visual_language_chat), [C++](https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/cpp/visual_language_chat) and [JavaScript](https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/js/visual_language_chat) visual language chat samples.
```
yatarkan: Let's add link to JS VLM sample later, once it is added to the repo to prevent dead link in deployed docs

src/js/tests/models.js
Retribution98: Yes, it is really large to run tests, but the JS test scripts only support models that have already been converted, so download them from HF/OpenVINO. Unfortunately, I couldn't find a smaller model there.

I am going to change it in another PR.

yatarkan: Let's try [OpenVINO/InternVL2-1B-int4-ov](https://huggingface.co/OpenVINO/InternVL2-1B-int4-ov)

site/docs/use-cases/image-processing/_sections/_run_model/_code_example_js.mdx
Retribution98: Yes, we can. It has been updated

src/js/lib/pipelines/vlmPipeline.ts
Retribution98: Agree, is updated

src/js/src/helper.cpp
Retribution98: I thought about this option, but I also decided to keep only Tensor[] for images and videos.
Thank you for catch this case, I remove it

src/js/tests/vlmPipeline.test.js
Retribution98: Nope, is removed

src/js/tests/utils.js
Retribution98: Done :)

src/js/tests/vlmPipeline.test.js
Retribution98: Is updated

src/js/tests/vlmPipeline.test.js
Retribution98: Agree, done

site/docs/use-cases/image-processing/_sections/_usage_options/index.mdx
Retribution98: Agree

site/docs/use-cases/image-processing/_sections/_run_model/_code_example_js.mdx
```diff
@@ -0,0 +1,41 @@
...
+import { stat, readdir } from "node:fs/promises";
+import sharp from "sharp";
+import path from "node:path";
+
+async function readImage(imagePath) {
+    const img = sharp(imagePath);
+    const metadata = await img.metadata();
+    const { width, height, channels } = metadata;
+    const imageBuffer = await img.raw().toBuffer();
+    console.log("Read image: ", width, height, channels);
```
yatarkan: Do we need console log in read image function?

site/docs/use-cases/image-processing/_sections/_run_model/_code_example_js.mdx
```diff
@@ -0,0 +1,41 @@
+import CodeBlock from '@theme/CodeBlock';
+
+<CodeBlock language="javascript" showLineNumbers>
+{`import { addon as ov } from "openvino-node";
+import { VLMPipeline } from "openvino-genai-node";
+import { stat, readdir } from "node:fs/promises";
+import sharp from "sharp";
```
yatarkan, line 7: Sharp seems to be an external JS library. But I didn't found it is used in our code or tests, only in docs code example. Do we want to guide users to use this dependency as a recommended way?

Retribution98: [Sharp](https://www.npmjs.com/package/sharp) is a popular, up-to-date library for efficiently working with images.
We don't use it for testing because we don't read images from files in tests. I suggest using this library for samples, which will be added separately.

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+        }
+        context->native_thread = std::thread(vlmPerformInferenceThread, context);
+    } catch (const std::exception& ex) {
+        Napi::Error::New(env, ex.what()).ThrowAsJavaScriptException();
+        *this->is_generating = false;
+    }
+    return env.Undefined();
+}
+
+Napi::Value VLMPipelineWrapper::start_chat(const Napi::CallbackInfo& info) {
```
almilosz, line 216:
```suggestion
void VLMPipelineWrapper::start_chat(const Napi::CallbackInfo& info) {
```

site/docs/use-cases/image-processing/_sections/_run_model/_code_example_js.mdx
Retribution98: No, I removed it.

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+        OPENVINO_ASSERT(this->pipe, "VLMPipeline is not initialized");
+        auto tokenizer = this->pipe->get_tokenizer();
+        return TokenizerWrapper::wrap(env, tokenizer);
+    } catch (const std::exception& ex) {
+        Napi::Error::New(env, ex.what()).ThrowAsJavaScriptException();
+    }
+    return env.Undefined();
+}
+
+Napi::Value VLMPipelineWrapper::set_chat_template(const Napi::CallbackInfo& info) {
```
almilosz, line 261: Please remove return value from void functions
```suggestion
void VLMPipelineWrapper::set_chat_template(const Napi::CallbackInfo& info) {
```

src/js/src/helper.cpp
```diff
@@ -539,3 +570,20 @@ Napi::Function get_prototype_from_ov_addon(const Napi::Env& env, const std::stri
 
     return ctor_val.As<Napi::Function>();
 }
+
+Napi::Object to_decoded_result(const Napi::Env& env, const ov::genai::DecodedResults& results) {
+    Napi::Object obj = Napi::Object::New(env);
+    obj.Set("texts", cpp_to_js<std::vector<std::string>, Napi::Value>(env, results.texts));
+    obj.Set("scores", cpp_to_js<std::vector<float>, Napi::Value>(env, results.scores));
+    obj.Set("perfMetrics", PerfMetricsWrapper::wrap(env, results.perf_metrics));
+    obj.Set("subword", Napi::String::New(env, results));
```
almilosz, line 579: Please check if `subword` is used in context of `DecodedResults`
I think it can be a leftover to keep backward compatibility.

src/js/src/vlm_pipeline/init_worker.cpp
```diff
@@ -0,0 +1,32 @@
...
+      device(device),
+      properties(properties) {};
+
+void VLMInitWorker::Execute() {
+    *this->is_initializing = true;
+    this->pipe = std::make_shared<ov::genai::VLMPipeline>(this->model_path, this->device, this->properties);
+};
+
+void VLMInitWorker::OnOK() {
+    *this->is_initializing = false;
```
almilosz, line 25: is this variable needed also for LLMPipeline? If yes please add it in separate PR

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+                       {InstanceMethod("init", &VLMPipelineWrapper::init),
+                        InstanceMethod("generate", &VLMPipelineWrapper::generate),
+                        InstanceMethod("getTokenizer", &VLMPipelineWrapper::get_tokenizer),
+                        InstanceMethod("startChat", &VLMPipelineWrapper::start_chat),
+                        InstanceMethod("finishChat", &VLMPipelineWrapper::finish_chat),
+                        InstanceMethod("setChatTemplate", &VLMPipelineWrapper::set_chat_template),
+                        InstanceMethod("setGenerationConfig", &VLMPipelineWrapper::set_generation_config)});
+}
+
+Napi::Value VLMPipelineWrapper::init(const Napi::CallbackInfo& info) {
```
almilosz, line 145:
```suggestion
void VLMPipelineWrapper::init(const Napi::CallbackInfo& info) {
```

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+        context->callback =
+            Napi::ThreadSafeFunction::New(env,
+                                          callback,                     // JavaScript function called asynchronously
+                                          "VLM_generate_callback",      // Name
+                                          0,                            // Unlimited queue
+                                          1,                            // Only one thread will use this initially
+                                          [context, this](Napi::Env) {  // Finalizer used to clean threads up
+                                              context->native_thread.join();
+                                              delete context;
+                                          });
```
almilosz, line 200: This Finalizer is repeated for each pipeline. We can extract it like here and reuse it: https://github.com/openvinotoolkit/openvino/blob/805d41a1c7ae209619651e57863d961ee32c65c2/src/bindings/js/node/src/infer_request.cpp#L197

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,287 @@
...
+            // If there were exceptions from the streamer, report them all as a single error and finish without result
+            std::string combined_error = "Streamer exceptions occurred:\n";
+            for (size_t i = 0; i < streamer_exceptions.size(); ++i) {
+                combined_error += "[" + std::to_string(i + 1) + "] " + streamer_exceptions[i] + "\n";
+            }
+            report_error(combined_error);
+        } else {
+            // If no exceptions from streamer, call the final callback with the result
+            napi_status status =
+                context->callback.BlockingCall([result, &report_error](Napi::Env env, Napi::Function jsCallback) {
```
almilosz, line 109: here `result` is passed by value. Such copy may affect performance. Maybe use `shared_ptr` or capture?

src/js/lib/utils.ts
```diff
@@ -344,3 +344,7 @@ export type SchedulerConfig = {
 export type LLMPipelineProperties = {
   schedulerConfig?: SchedulerConfig;
 };
+
+export type VLMPipelineProperties = {
+  schedulerConfig?: SchedulerConfig;
+} & Record<string, unknown>;
```
almilosz, line 350: I think it's safer to have strict definition for allowed properties. Are you sure we want to add here `Record<string, unknown>` ?

almilosz (commented):
LGTM, a few comments

src/js/lib/utils.ts
Retribution98: JS can push any objects, so I suppose we cannot make it safer through typing. We manage this from the CPP part.
I believe types should help, not hinder. It would be great if we could provide the exact type, but we don't have this type here or a goal to create it now. So, I would like to use more open types in these cases.

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
Retribution98: VALIDATE_ARGS_COUNT now returns Env().Undefined(). It doesn't allow it to be done now.  :(
Let's fix it later.

src/js/src/helper.cpp
Retribution98: We will remove it when LLMPipeline is refactored

PR #3127, opened 2025-12-17
[wwb] Fix smolvlm

sbalandi:
## Description
model: HuggingFaceTB/SmolVLM2-256M-Video-Instruct
AutoModel chooses SmolVLMModel. This class doesn't have generate and has another format of output, so firstly fails with:
`AttributeError: 'SmolVLMModel' object has no attribute 'generate'`
if fix this issue, it will fail on crop_question step

CVS-###

Fixes #(issue)

## Checklist:


as-suvorov:
Please fix linting: https://github.com/openvinotoolkit/openvino.genai/actions/runs/20312969768/job/58348985003?pr=3127
image

@sgonorov So darker seems to apply not only whitespaces and newline rules. It's either max_length only or a full linting rule set. fyi

PR #3128, opened 2025-12-18
Removed building of protobuf in video_to_text sample.

popovaan:
## Description

CVS-178602

## Checklist:

### DIFF
--- a/samples/cpp/visual_language_chat/CMakeLists.txt
+++ b/samples/cpp/visual_language_chat/CMakeLists.txt
@@ -104,7 +104,7 @@ if(NOT OpenCV_FOUND)
             GIT_TAG 4.11.0
             GIT_SHALLOW TRUE
     )
-    FetchContent_MakeAvailable(opencv) 
+    FetchContent_MakeAvailable(opencv)
 
     target_include_directories(video_to_text_chat PRIVATE
             ${OPENCV_CONFIG_FILE_INCLUDE_DIR}


samples/cpp/visual_language_chat/CMakeLists.txt
```diff
@@ -69,13 +69,16 @@ if(NOT OpenCV_FOUND)
 
     set(BUILD_SHARED_LIBS ON)
     set(WITH_FFMPEG ON)
+    set(BUILD_PROTOBUF OFF CACHE BOOL "" FORCE)
+    set(PROTOBUF_UPDATE_FILES OFF CACHE BOOL "" FORCE)
+    set(WITH_PROTOBUF OFF CACHE BOOL "" FORCE)
```
Wovchena, line 72: Can you check if there's something else built but not needed

mryzhov: see https://github.com/openvinotoolkit/openvino.genai/pull/3130

Wovchena: That excludes just protobuff

samples/cpp/visual_language_chat/CMakeLists.txt
```diff
@@ -69,13 +69,16 @@ if(NOT OpenCV_FOUND)
 
     set(BUILD_SHARED_LIBS ON)
     set(WITH_FFMPEG ON)
+    set(BUILD_PROTOBUF OFF CACHE BOOL "" FORCE)
+    set(PROTOBUF_UPDATE_FILES OFF CACHE BOOL "" FORCE)
```
mryzhov: We don't need to set them manually off, WITH_PROTOBUF=OFF should be enough

samples/cpp/visual_language_chat/CMakeLists.txt
mryzhov: do you need something more? ) never mind, protobuff from openvino is incompatible with genai's one,

popovaan: I think the suggestion was to turn off all other modules not needed by the sample. I will check the opencv flags and turn off all what we don't need.

samples/cpp/visual_language_chat/CMakeLists.txt
popovaan: Thanks, removed.

PR #3129, opened 2025-12-18
[VLM] Create empty videos vector with batch

as-suvorov:
## Description
Empty videos vector was created with batch 1. It fails if multiple prompts passed. Create empty_videos_vector with prompts batch size
Use ov_genai image placeholders in `milebench_eval_vlm.py` sample.

CVS-178649

Fixes #(issue)

## Checklist:


l-bat (approved):
Thanks for fixing this! Now [milebench_eval_vlm.py](https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/visual_language_chat/milebench_eval_vlm.py) should work with batch > 1 for any VLM model.

PR #3132, opened 2025-12-19
[JS] Add support of Parser and default implementations

Retribution98:
## Description
This PR adds parsers to the JavaScript API.
The base parser can be a JavaScript object with parse methods. The user object is wrapped in a JSParser entity and uses TSFN.BlockingCall to execute it.
Child classes provide a wrapper that uses the core entity to parse the message. To boost performance, it doesn't use the JSParser and instead uses the native parser. If the user would like to extend it from the JavaScript side, it must be wrapped in a JSParser, just like the base parser.
The following child classes are supported:
- ReasoningParser
- DeepSeekR1ReasoningParser
- Phi4ReasoningParser
- Llama3PythonicToolParser
- Llama3JsonToolParser.

The IncrementalParser is not supported now because it must be set up in the Streamer class, which the JS API does not support.

CVS-176151

## Checklist:

### DIFF
--- a/samples/export-requirements.txt
+++ b/samples/export-requirements.txt
@@ -1,7 +1,7 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly
 openvino-tokenizers[transformers]~=2026.0.0.0.dev
-optimum-intel[nncf]==1.26.1
+optimum-intel[nncf]==1.27.0
 numpy==1.26.4; platform_system == "Darwin" and platform_machine == "x86_64"
 safetensors==0.6.2; platform_system == "Darwin" and platform_machine == "x86_64"
 einops==0.8.1  # For Qwen
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -8713,38 +8713,39 @@
       }
     },
     "node_modules/express": {
-      "version": "4.21.2",
-      "resolved": "https://registry.npmjs.org/express/-/express-4.21.2.tgz",
-      "integrity": "sha512-28HqgMZAmih1Czt9ny7qr6ek2qddF4FclbMzwhCREB6OFfH+rXAnuNCwo1/wFvrtbgsQDb4kSbX9de9lFbrXnA==",
+      "version": "4.22.1",
+      "resolved": "https://registry.npmjs.org/express/-/express-4.22.1.tgz",
+      "integrity": "sha512-F2X8g9P1X7uCPZMA3MVf9wcTqlyNp7IhH5qPCI0izhaOIYXaW9L535tGA3qmjRzpH+bZczqq7hVKxTR4NWnu+g==",
+      "license": "MIT",
       "dependencies": {
         "accepts": "~1.3.8",
         "array-flatten": "1.1.1",
-        "body-parser": "1.20.3",
-        "content-disposition": "0.5.4",
+        "body-parser": "~1.20.3",
+        "content-disposition": "~0.5.4",
         "content-type": "~1.0.4",
-        "cookie": "0.7.1",
-        "cookie-signature": "1.0.6",
+        "cookie": "~0.7.1",
+        "cookie-signature": "~1.0.6",
         "debug": "2.6.9",
         "depd": "2.0.0",
         "encodeurl": "~2.0.0",
         "escape-html": "~1.0.3",
         "etag": "~1.8.1",
-        "finalhandler": "1.3.1",
-        "fresh": "0.5.2",
-        "http-errors": "2.0.0",
+        "finalhandler": "~1.3.1",
+        "fresh": "~0.5.2",
+        "http-errors": "~2.0.0",
         "merge-descriptors": "1.0.3",
         "methods": "~1.1.2",
-        "on-finished": "2.4.1",
+        "on-finished": "~2.4.1",
         "parseurl": "~1.3.3",
-        "path-to-regexp": "0.1.12",
+        "path-to-regexp": "~0.1.12",
         "proxy-addr": "~2.0.7",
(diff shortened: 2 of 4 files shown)


src/js/src/addon.cpp
```diff
@@ -63,4 +86,4 @@ Napi::Object init_module(Napi::Env env, Napi::Object exports) {
 }
 
 // Register the addon with Node.js
-NODE_API_MODULE(openvino-genai-node, init_module)
+NODE_API_MODULE(openvino - genai - node, init_module)
```
yatarkan: What does it change?

src/js/src/parser.cpp
```diff
@@ -0,0 +1,199 @@
...
+#include "include/parser.hpp"
+
+#include <future>
+
+#include "include/addon.hpp"
+#include "include/helper.hpp"
+
+JSParser::JSParser(Napi::Env env, Napi::Object jsParser) {
+    js_parser_ref = Napi::Persistent(jsParser);
+    js_parser_ref.SuppressDestruct();
```
almilosz: Is `SuppressDestruct` needed here? 
 `ObjectReference`  without suppressDestruct is used as class property in openvino and it's not destructed too early. 
 Also, did you notice if any program/snippet  hangs for a few seconds at the end?

src/js/src/parser.cpp
```diff
@@ -0,0 +1,199 @@
...
+    Napi::Function parseFn = jsParser.Get("parse").As<Napi::Function>();
```
almilosz:
```suggestion
    Napi::Function js_parse_fn = jsParser.Get("parse").As<Napi::Function>();
```
Let's make the name show that this function is later called in `JSParser::parse`

src/js/src/parser.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                                parseFn,            // js callback
+                                                "js_parser_parse",  // resource name
+                                                0,                  // unlimited queue
+                                                1                   // initialThreadCount
+    );
+};
+
+void JSParser::parse(ov::genai::JsonContainer& message) {
+    std::promise<ov::genai::JsonContainer> promise;
+    parser_tsfn.BlockingCall([&message, &promise, this](Napi::Env env, Napi::Function jsCallback) {
```
almilosz:
```suggestion
    parser_tsfn.BlockingCall([&message, &promise, this](Napi::Env env, Napi::Function js_parse_fn) {
```

src/js/tests/parsers.test.js
```diff
@@ -0,0 +1,333 @@
+// Copyright (C) 2018-2025 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
```
almilosz, line 3: In the logs there is message:
```# Catch in the thread: 'Check 'value.IsObject()' failed at src/js/src/helper.cpp:348:
# Parser must be a JS object with a 'parse' method
# '
# Catch in the thread: 'Check 'obj.Has("parse")' failed at src/js/src/helper.cpp:358:
# Parser object must have a 'parse' method
# '
# Catch in the thread: 'Check 'parse_method.IsFunction()' failed at src/js/src/helper.cpp:360:
# 'parse' property of Parser object must be a function
# '`
Please take a look

File: src/js/lib/decodedResults.ts [hunk 6]
```diff
@@ -34,6 +41,7 @@ export class DecodedResults {
   texts: string[];
   scores: number[];
   perfMetrics: PerfMetrics;
+  parsed?: Record<string, unknown>[];
```
almilosz:
```suggestion
  parsed: Record<string, unknown>[];
```

src/js/src/helper.cpp
```diff
@@ -338,6 +342,40 @@ ov::Tensor js_to_cpp<ov::Tensor>(const Napi::Env& env, const Napi::Value& value)
...
+std::vector<std::shared_ptr<ov::genai::Parser>> js_to_cpp<std::vector<std::shared_ptr<ov::genai::Parser>>>(
+    const Napi::Env& env,
+    const Napi::Value& value) {
+    OPENVINO_ASSERT(value.IsArray(), "Passed argument must be of type Array.");
+    Napi::Array arr = value.As<Napi::Array>();
+    std::vector<std::shared_ptr<ov::genai::Parser>> parsers;
+    parsers.reserve(arr.Length());
+    for (uint32_t i = 0; i < arr.Length(); ++i) {
+        Napi::Value element = arr[i];
+        parsers.push_back(js_to_cpp<std::shared_ptr<ov::genai::Parser>>(env, element));
```
```suggestion
        parsers.push_back(js_to_cpp<std::shared_ptr<ov::genai::Parser>>(env, arr[i]));
```

src/js/tests/parsers.test.js
Retribution98: This is due to incorrect behavior in the way LLMPipeline.generate() handles errors (src/js/llm_pipeline/llm_pipeline_wrapper.cpp:160). For now, I suggest ignoring the error because I am going to fix it as part of the LLMPipeline refactoring.

src/js/src/addon.cpp
Retribution98: It was changed by automatic linting. I reverted the change and will address the linting issue separately. Thank you for pointing this out.

src/js/src/parser.cpp
Retribution98: No, `SuppressDestruct` is not required here. Thank you for pointing this out.

> Also, did you notice if any program/snippet hangs for a few seconds at the end?

No, I didn't notice any hangs at the end of the scripts.

src/js/src/parser.cpp
Retribution98: Sure, thanks

yatarkan (approved):

PR #3139, opened 2025-12-23
[wwb] Move models loading to conftest with AtomicDownloadManager

sbalandi:
## Description
Update WWB tests to reduce model loading amount/problem and unify the API for tool's tests

Fixes #(issue)

## Checklist:

### DIFF
--- a/tools/who_what_benchmark/tests/conftest.py
+++ b/tools/who_what_benchmark/tests/conftest.py
@@ -92,11 +92,7 @@ def convert_model(model_name: str) -> str:
         logger.info(f"Conversion command: {' '.join(command)}")
         retry_request(lambda: subprocess.run(command, check=True, text=True, capture_output=True))
 
-    try:
-        manager.execute(convert)
-    except subprocess.CalledProcessError as error:
-        logger.exception(f"optimum-cli returned {error.returncode}. Output:\n{error.stderr}")
-        raise
+    manager.execute(convert)
     return str(model_path)
 
 
@@ -123,13 +119,10 @@ def run_wwb(args: list[str], env=None):
     base_env = {"TRANSFORMERS_VERBOSITY": "debug", "PYTHONIOENCODING": "utf-8", **os.environ}
     if env:
         base_env.update(env)
-    try:
-        return subprocess.check_output(
-            command,
-            stderr=subprocess.STDOUT,
-            encoding="utf-8",
-            env=base_env,
-        )
-    except subprocess.CalledProcessError as error:
-        logger.error(f"'{' '.join(map(str, command))}' returned {error.returncode}. Output:\n{error.output}")
-        raise
+
+    return subprocess.check_output(
+        command,
+        stderr=subprocess.STDOUT,
+        encoding="utf-8",
+        env=base_env,
+    )


tools/who_what_benchmark/tests/test_cli_embeddings.py
```diff
@@ -1,191 +1,182 @@
...
+        ("Qwen/Qwen3-Embedding-0.6B", 1),
+        ("Qwen/Qwen3-Embedding-0.6B", 12),
     ],
 )
-def test_embeddings_with_batch(model_id, model_type, batch_size, tmp_path):
+def test_embeddings_with_batch(model_id, batch_size, tmp_path):
     if sys.platform == "win32":
         pytest.xfail("Ticket 178790")
-
     GT_FILE = tmp_path / f"gt_batch_{batch_size}.csv"
```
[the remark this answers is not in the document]
sbalandi: no, it is used

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+    try:
+        manager.execute(convert)
+    except subprocess.CalledProcessError as error:
+        logger.exception(f"optimum-cli returned {error.returncode}. Output:\n{error.output}")
+        raise
+    return str(model_path)
+
+
+@pytest.fixture(scope="session", autouse=True)
+def module_teardown():
```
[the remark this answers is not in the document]
sbalandi: renamed

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+            "openvino",
+            "--model",
+            model_name,
+            "--weight-format",
+            "fp16",
+            *convert_args,
+            str(temp_path),
+        ]
+        logger.info(f"Conversion command: {' '.join(command)}")
+        retry_request(lambda: subprocess.run(command, check=True, text=True, capture_output=True))
```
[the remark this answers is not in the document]
sbalandi: error.stdout -> error.stderr

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+def get_ov_cache_converted_models_dir():
+    return get_ov_cache_dir() / "converted_models"
+
+
+def convert_model(model_name):
+    models_dir = get_ov_cache_converted_models_dir()
+
+    model_id = model_name.split("/")[1]
```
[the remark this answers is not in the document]
sbalandi: updated

tools/who_what_benchmark/tests/test_cli_embeddings.py
```diff
@@ -1,191 +1,182 @@
...
+            "--model-type",
+            "text-embedding",
+            "--genai",
+        ]
+    )
 
 
 @pytest.mark.parametrize(
-    ("model_id", "model_type", "batch_size"),
+    ("model_id", "batch_size"),
```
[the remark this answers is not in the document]
sbalandi: restored

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+import shutil
+import subprocess  # nosec B404
+
+from pathlib import Path
+from typing import Any, Dict
+
+sys.path.insert(0, f"{os.path.dirname(__file__)}/../../../tests/")
+from python_tests.utils.atomic_download import AtomicDownloadManager  # noqa
+from python_tests.utils.constants import get_ov_cache_dir  # noqa
+from python_tests.utils.network import retry_request  # noqa
```
as-suvorov: This is a tricky point. Here you bind wwb to genai. But from packages separation wwb is intended to be standalone. I think we need to find another approach here. We either need to have a common utils or to copy code.

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+            str(temp_path),
+        ]
+        logger.info(f"Conversion command: {' '.join(command)}")
+        retry_request(lambda: subprocess.run(command, check=True, text=True, capture_output=True))
+
+    try:
+        manager.execute(convert)
+    except subprocess.CalledProcessError as error:
+        logger.exception(f"optimum-cli returned {error.returncode}. Output:\n{error.stderr}")
+        raise
```
as-suvorov: what is the point to re-raise with logging?

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+    try:
+        return subprocess.check_output(
+            command,
+            stderr=subprocess.STDOUT,
+            encoding="utf-8",
+            env=base_env,
+        )
+    except subprocess.CalledProcessError as error:
+        logger.error(f"'{' '.join(map(str, command))}' returned {error.returncode}. Output:\n{error.output}")
+        raise
```
as-suvorov: Same here regarding re-raise

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+    "tiny-random-flux": {"name": "optimum-intel-internal-testing/tiny-random-flux", "convert_args": []},
+    "tiny-random-flux-fill": {"name": "optimum-intel-internal-testing/tiny-random-flux-fill", "convert_args": []},
+}
+
+
+def get_ov_cache_converted_models_dir():
+    return get_ov_cache_dir() / "converted_models"
+
+
+def convert_model(model_name):
```
as-suvorov:
```suggestion
def convert_model(model_name: str) -> Path:
```

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+        ]
+        logger.info(f"Conversion command: {' '.join(command)}")
+        retry_request(lambda: subprocess.run(command, check=True, text=True, capture_output=True))
+
+    try:
+        manager.execute(convert)
+    except subprocess.CalledProcessError as error:
+        logger.exception(f"optimum-cli returned {error.returncode}. Output:\n{error.stderr}")
+        raise
+    return str(model_path)
```
as-suvorov, line 96:
```suggestion
    return Path(model_path)
```

tools/who_what_benchmark/tests/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+
+    if os.environ.get("CLEANUP_CACHE", "false").lower() == "true":
+        if ov_cache_dir.exists():
+            logger.info(f"Removing temporary directory: {ov_cache_dir}")
+            shutil.rmtree(ov_cache_dir)
+        else:
+            logger.info(f"Skipped temporary directory cleanup because it doesn't exist: {ov_cache_dir}")
+
+
+def run_wwb(args, env=None):
```
as-suvorov:
```suggestion
def run_wwb(args: list[str], env=None):
```

tools/who_what_benchmark/tests/conftest.py
sbalandi: to see the output in log in case of fails, but don't see output if everything is ok

tools/who_what_benchmark/tests/conftest.py
sbalandi: Why Path ? We need str in tests: https://github.com/sbalandi/openvino.genai/blob/c3097ebc0394755806d27fb3f214b90ae46b57cb/tools/who_what_benchmark/tests/test_cli_embeddings.py#L46

tools/who_what_benchmark/tests/conftest.py
sbalandi: updated

tools/who_what_benchmark/tests/conftest.py
sbalandi: On the one hand, these are tests, not the tool itself, so I'm not sure that this is really important avoid binding, but I don't have confident arguments, so I copied it into a separate file to tools/wwb repo part

tools/who_what_benchmark/tests/conftest.py
as-suvorov: Can't you see the error stack trace in output without re-raise and explicit logging?

tools/who_what_benchmark/tests/conftest.py
as-suvorov: The variable name `model_path`, and there is a dedicated variable type for that.
For this case, I agree, let's keep it as a string

tools/who_what_benchmark/tests/conftest.py
as-suvorov: Let's proceed with a copy. Will see if it's work well.

as-suvorov (changes requested):
Please double check if error message is shown in output without re-raise

sbalandi:
> Please double check if error message is shown in output without re-raise

I checked, you're right, subprocess won't print if the return code is 0, but it will if an error occurs. No additional wrapping is needed

PR #3160, opened 2026-01-03
Add XAttn to test cases in longbench

vshampor:
Ticket: 175120

### DIFF
--- a/tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
+++ b/tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
@@ -8,7 +8,13 @@ from pathlib import Path
 import datasets
 from tqdm import tqdm
 
-from openvino_genai import ContinuousBatchingPipeline, GenerationConfig, CacheEvictionConfig, AggregationMode
+from openvino_genai import (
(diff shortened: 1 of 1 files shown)


tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
```diff
@@ -188,7 +194,8 @@ class LongBenchTestData:
     LongBenchTestData("samsum", 4, 1.6, 2.5),
     LongBenchTestData("trec", 3.2, 2.0, 3.3),
 ], ids=["samsum", "trec"])
-def test_optimized_generation_longbench(test_struct):
+@pytest.mark.parametrize("use_xattention", [False, True])
```
Wovchena, line 197: Avoid extending the existing runs. It already takes 2.5 hours without the change: https://github.com/openvinotoolkit/openvino.genai/actions/runs/20805399448/job/59769537992

You can introduce an extra entry in matrix in `linux.yml` (and other OS) running the new test separately

vshampor: Feel free to add a commit to this branch to the effect of what you have just suggested.

PR #3172, opened 2026-01-08
CVS-178941: Align VLM resize pre-processing with optimum

RyanMetcalfeInt8:
## Description
1. Replaces bilinear/bicubic resize implementations with new variant which matches pillow.
2. For VLM's that use OpenVINO for pre-processing, align interpolation modes to closer match pillow. Note: This also covers cases where optimum/transformers uses torchvision, as torchvision's implementation produces results very similar to pillow (see ticket for more details).
3. Add new VLM tests where resize needs to take place -- GenAI output must match optimum.

CVS-178941

## Checklist:

### DIFF
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -1926,6 +1926,7 @@ def parametrize_optimum_vs_genai(models: list[str] | None = None) -> Callable[[C
         models = MODEL_IDS
 
     params = []
+
     def append_test_case(
         *,
         model_id: str,


yatarkan:
Please provide [WWB](https://github.com/openvinotoolkit/openvino.genai/tree/master/tools/who_what_benchmark) results comparison (master branch vs your branch with resize fixes) to see the accuracy improvements for affected models

RyanMetcalfeInt8:
Was finally able to reproduce this CI failure:
```
test_vlm_pipeline_match_optimum_with_resize[qwen2.5-vl/SDPA/image-999x666]
```
Note that from_cache_or_download() will download cat.jpg from the URL, read it in as a PIL image, and save a re-encoded variant of it in cache (while returning the original variant). I realized that I was only able to reproduce this issue after I cleared the pytest cache each time. I can submit another PR for the fix for this, as it's outside the scope of this..

Anyway, by default this test fails (with using the right cat.jpg) -- but it passes if I set `VISION_PREPROCESS=CPP`, so it indicates that there is some delta in the pre-processing OV graph.

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
...
+                outp[0] = clip8_from_fixed(ss0);
+                outp[1] = clip8_from_fixed(ss1);
+                outp[2] = clip8_from_fixed(ss2);
             }
         }
     }
 }
 
+void bicubic_resize(const clip_image_u8& img, clip_image_u8& dst, int target_width, int target_height) {
+    resize_pillow_like(img, dst, target_width, target_height, 2.0, pillow_bicubic_filter);
```
rkazants, line 259: Why don't we use `Interpolate` op from our opset? If you see mismatch between OV Interpolate results and PIL/torch, we need to solve it on OV side first. Our plan is to inject pre-processing OV graph into the main graph.

RyanMetcalfeInt8: As per the JIRA ticket findings, `Interpolate` op is much slower (especially with .eval), and has a bigger delta against PIL than this implementation. 

I agree that OV-side should ultimately get updated (as per steps outlined in the ticket), but for now, replacing the existing bicubic / bilinear resize function with one that has bit-for-bit match / improved performance is a good stepping stone until OV is updated -- at that point we can replace these with a call to `Interpolate` .eval().

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -88,44 +90,32 @@ ov::Tensor unpad_image(const ov::Tensor& tensor, const ImageSize& original_size)
...
-            for (int h = 0; h < unpadded_height_dim; ++h) {
-                std::copy(
-                    tensor.data<float>() + (e * current_height * current_width + (padding + h) * current_width),
-                    tensor.data<float>() + (e * current_height * current_width + (padding + h) * current_width + current_width),
-                    unpadded_tensor.data<float>() + (e * unpadded_height_dim * current_width + h * current_width)
-                );
-            }
-        }
+
+        auto unpadded_height = current_height - padding * 2;
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -88,44 +90,32 @@ ov::Tensor unpad_image(const ov::Tensor& tensor, const ImageSize& original_size)
...
-            for (int h = 0; h < current_height; ++h) {
-                std::copy(
-                    tensor.data<float>() + (e * current_height * current_width + h * current_width + padding),
-                    tensor.data<float>() + (e * current_height * current_width + h * current_width + padding + unpadded_width_dim),
-                    unpadded_tensor.data<float>() + (e * current_height * unpadded_width_dim + h * unpadded_width_dim)
-                );
-            }
-        }
+
+        auto unpadded_width = current_width - padding * 2;
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -408,4 +396,27 @@ ov::Tensor InputsEmbedderLLaVANext::get_inputs_embeds(const std::string& unified
...
+    ov::Tensor encoded_image_token =
+        m_tokenizer.encode(m_vlm_config.im_start, ov::genai::add_special_tokens(false)).input_ids;
+    int64_t image_token_id = encoded_image_token.data<int64_t>()[encoded_image_token.get_size() - 1];
+
+    auto for_inputs_embeds_ids = ov::Tensor(input_ids.get_element_type(), input_ids.get_shape());
+    input_ids.copy_to(for_inputs_embeds_ids);
+
+    // zero out values that are equal to image_token_id
+    auto* pids = for_inputs_embeds_ids.data<int64_t>();
+    for (auto i = 0; i < input_ids.get_size(); i++) {
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
...
-    dst.buf.resize(3 * target_width * target_height);
+struct Coeffs1D {
+    int outSize = 0;
+    int ksize = 0;
+    std::vector<int> bounds_xmin;   // size outSize
+    std::vector<int> bounds_count;  // size outSize
+    std::vector<int32_t> kk;        // size outSize * ksize, fixed-point
+};
+
+// Pillow uses fixed point with PRECISION_BITS = (32 - 8 - 2).
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: done, just added a ref instead.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1573,47 +1591,96 @@ def get_nanollava_processor():
...
+def test_vlm_pipeline_match_optimum_with_resize(
+    request, ov_pipe_model: VlmModelInfo, has_image: bool, has_video: bool, image_input_resolution
+):
+    resized_image = None
+    resized_video = None
+    if has_image:
+        resized_image = request.getfixturevalue("cat_image")
+        resized_image = resized_image.resize(image_input_resolution)
+
+    # TODO: Resize video?
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: it's fine.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1573,47 +1591,96 @@ def get_nanollava_processor():
...
+        resized_video = request.getfixturevalue("synthetic_video_32x32")
+
+    run_compare_genai_optimum(ov_pipe_model, resized_image, resized_video)
+
+
+GENAI_VS_OPTIMUM_IMAGE_INPUT_RESOLUTIONS = [
+    pytest.param((100, 77), id="100x77"),
+    pytest.param((999, 666), id="999x666"),
+    pytest.param((1920, 1080), id="1920x1080"),
+    # This one is problematic for qwen2vl/2.5, triggers DefaultCPUAllocator: not enough memory errors. Disable for now.
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: It already specifies that there is some memory allocation issue. These details are enough (it's at least all that I know atm)

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -88,44 +90,32 @@ ov::Tensor unpad_image(const ov::Tensor& tensor, const ImageSize& original_size)
...
-            for (int h = 0; h < unpadded_height_dim; ++h) {
-                std::copy(
-                    tensor.data<float>() + (e * current_height * current_width + (padding + h) * current_width),
-                    tensor.data<float>() + (e * current_height * current_width + (padding + h) * current_width + current_width),
-                    unpadded_tensor.data<float>() + (e * unpadded_height_dim * current_width + h * current_width)
-                );
-            }
-        }
+
+        size_t unpadded_height = current_height - padding * 2;
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: done.

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -88,44 +90,32 @@ ov::Tensor unpad_image(const ov::Tensor& tensor, const ImageSize& original_size)
...
-            for (int h = 0; h < current_height; ++h) {
-                std::copy(
-                    tensor.data<float>() + (e * current_height * current_width + h * current_width + padding),
-                    tensor.data<float>() + (e * current_height * current_width + h * current_width + padding + unpadded_width_dim),
-                    unpadded_tensor.data<float>() + (e * current_height * unpadded_width_dim + h * unpadded_width_dim)
-                );
-            }
-        }
+
+        size_t unpadded_width = current_width - padding * 2;
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: done.

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -408,4 +402,27 @@ ov::Tensor InputsEmbedderLLaVANext::get_inputs_embeds(const std::string& unified
     return utils::merge_text_and_image_embeddings_llava(input_ids, text_embeds, image_embeds, image_token_id);
 }
 
-} // namespace ov::genai
\ No newline at end of file
+// ref:
+// https://github.com/huggingface/optimum-intel/blob/v1.27.0/optimum/intel/openvino/modeling_visual_language.py#L1423
+ov::Tensor InputsEmbedderLLaVANext::get_text_embeddings_llava_next(const ov::Tensor input_ids) {
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: done

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -88,44 +90,38 @@ ov::Tensor unpad_image(const ov::Tensor& tensor, const ImageSize& original_size)
     float original_aspect_ratio = static_cast<float>(original_width) / original_height;
     float current_aspect_ratio = static_cast<float>(current_width) / current_height;
 
-    ov::Tensor unpadded_tensor;
-
+    ov::Shape unpadded_tensor_shape;
+    ov::Coordinate view_begin, view_end;
     if (original_aspect_ratio > current_aspect_ratio) {
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: The variable names are chosen to match reference implementation here: https://github.com/huggingface/transformers/blob/v4.57.3/src/transformers/models/llava_next/modeling_llava_next.py#L128

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
...
-    return std::max(lower, std::min(x, upper));
+    // Quantize to fixed point with symmetric rounding.
+    c.kk.resize(static_cast<size_t>(outSize) * ksize);
+    for (size_t i = 0; i < prekk.size(); ++i) {
+        const double v = prekk[i];
+        if (v < 0.0)
+            c.kk[i] = static_cast<int32_t>(-0.5 + v * (1 << PRECISION_BITS));
+        else
+            c.kk[i] = static_cast<int32_t>(0.5 + v * (1 << PRECISION_BITS));
+    }
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: I don't think comment is necessary here. It's quite obvious that 0.5 is half away from zero.

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
...
+        for (int y = 0; y < inH; ++y) {
+            for (int xx = 0; xx < outW; ++xx) {
+                const int xmin = cx.bounds_xmin[xx];
+                const int count = cx.bounds_count[xx];
+                const int32_t* k = &cx.kk[static_cast<size_t>(xx) * cx.ksize];
+
+                // Pillow uses rounding bias: 1<<(PRECISION_BITS-1).
+                int ss0 = 1 << (PRECISION_BITS - 1);
+                int ss1 = 1 << (PRECISION_BITS - 1);
+                int ss2 = 1 << (PRECISION_BITS - 1);
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: it's fine.

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -408,4 +402,28 @@ ov::Tensor InputsEmbedderLLaVANext::get_inputs_embeds(const std::string& unified
...
+    // zero out values that are equal to image_token_id
+    auto* pids = for_inputs_embeds_ids.data<int64_t>();
+    for (size_t i = 0; i < input_ids.get_size(); i++) {
+        int64_t val = pids[i];
+        pids[i] = (val == image_token_id) ? 0 : val;
+    }
+
+    ov::Tensor text_embeds = m_embedding->infer(req, for_inputs_embeds_ids);
+    return text_embeds;
+}
```
yatarkan: Do we really need this method and setting token ids to zero for image tokes before getting text embeddings? These related text embeddings are replaced with image embeddings later in `utils::merge_text_and_image_embeddings_llava()`.
Does this affect existing preprocessing somehow?

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1362,20 +1373,38 @@ def conversation_requests(
...
+):
     if items is None:
         items = MODELS_TO_TAG
 
+    xfail = xfail or {}
+
+    params = []
+    ids = []
+    for item in items:
+        model_id, tag = item[0], item[1]
```
yatarkan: Second item in tuple is not tag, it is attention backend that is passed to `ov_pipe_model`

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1362,20 +1373,38 @@ def conversation_requests(
...
+        params,
         indirect=["ov_pipe_model"],
-        ids=[f"{item[0]}/{item[1]}" for item in items]
+        ids=ids,
     )
 
 
-@model_and_tag_parametrize(TAG_INSERTED_BY_TEMPLATE)
+@model_and_tag_parametrize(
+    TAG_INSERTED_BY_TEMPLATE, xfail={("optimum-intel-internal-testing/tiny-random-llava", "PA"): "CVS-179090"}
```
yatarkan: The test seems to working on master right now. Is it failing in your PR?

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1739,52 +1757,101 @@ def get_nanollava_processor():
...
+@pytest.mark.parametrize(
+    "ov_pipe_model,has_image,has_video",
+    GENAI_VS_OPTIMUM_CASES,
+    indirect=["ov_pipe_model"],
+)
+# Some models fail when using OV graph pre-processing.
+# Force VISION_PREPROCESS to be set to CPP for now.
+@pytest.mark.vision_preprocess("CPP")
+def test_vlm_pipeline_match_optimum_with_resize(
+    request, ov_pipe_model: VlmModelInfo, has_image: bool, has_video: bool, image_input_resolution
```
yatarkan:
```suggestion
    request: pytest.FixtureRequest, ov_pipe_model: VlmModelInfo, has_image: bool, has_video: bool, image_input_resolution: tuple[int, int]
```

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1739,52 +1757,101 @@ def get_nanollava_processor():
...
+        resized_video = request.getfixturevalue("synthetic_video_32x32")
+
+    run_compare_genai_optimum(ov_pipe_model, resized_image, resized_video)
+
+
+GENAI_VS_OPTIMUM_IMAGE_INPUT_RESOLUTIONS = [
+    pytest.param((100, 77), id="100x77"),
+    pytest.param((999, 666), id="999x666"),
+    pytest.param((1920, 1080), id="1920x1080"),
+    # This one is problematic for qwen2vl/2.5, triggers DefaultCPUAllocator: not enough memory errors. Disable for now.
```
yatarkan: Need a ticket for this use case if it worth checking such resolution

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1739,52 +1757,101 @@ def get_nanollava_processor():
...
+def test_vlm_pipeline_match_optimum_with_resize(
+    request, ov_pipe_model: VlmModelInfo, has_image: bool, has_video: bool, image_input_resolution
+):
+    resized_image = None
+    resized_video = None
+    if has_image:
+        resized_image = request.getfixturevalue("cat_image")
+        resized_image = resized_image.resize(image_input_resolution)
+
+    # TODO: Resize video?
```
yatarkan: Let's address it as well, there is function for resizing video in the same file.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1739,52 +1757,101 @@ def get_nanollava_processor():
...
+@pytest.mark.parametrize("image_input_resolution", GENAI_VS_OPTIMUM_IMAGE_INPUT_RESOLUTIONS)
+@pytest.mark.parametrize(
+    "ov_pipe_model,has_image,has_video",
+    GENAI_VS_OPTIMUM_CASES,
+    indirect=["ov_pipe_model"],
+)
+# Some models fail when using OV graph pre-processing.
+# Force VISION_PREPROCESS to be set to CPP for now.
+@pytest.mark.vision_preprocess("CPP")
+def test_vlm_pipeline_match_optimum_with_resize(
```
yatarkan:
```suggestion
def test_vlm_pipeline_exact_match_optimum(
```
or
```suggestion
def test_vlm_pipeline_match_optimum_with_resolutions(
```
In order not to mix up names with previous preresized test

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1739,52 +1757,101 @@ def get_nanollava_processor():
...
 
     assert optimum_text == genai_text
 
 
+@pytest.mark.parametrize(
+    "ov_pipe_model,has_image,has_video",
+    GENAI_VS_OPTIMUM_CASES,
+    indirect=["ov_pipe_model"],
+)
+def test_vlm_pipeline_match_optimum_preresized(request, ov_pipe_model: VlmModelInfo, has_image: bool, has_video: bool):
```
yatarkan:
```suggestion
def test_vlm_pipeline_match_optimum_preresized(request: pytest.FixtureRequest, ov_pipe_model: VlmModelInfo, has_image: bool, has_video: bool):
```

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1739,52 +1757,101 @@ def get_nanollava_processor():
...
+]
+
+
+@pytest.mark.parametrize("image_input_resolution", GENAI_VS_OPTIMUM_IMAGE_INPUT_RESOLUTIONS)
+@pytest.mark.parametrize(
+    "ov_pipe_model,has_image,has_video",
+    GENAI_VS_OPTIMUM_CASES,
+    indirect=["ov_pipe_model"],
+)
+# Some models fail when using OV graph pre-processing.
```
yatarkan: It would be great to have a list of these failing models with OV graph pre-processing somewhere to track.
For example we can have these model groups here and parametrize the test with both vision_preprocess options.

tests/python_tests/test_vlm_pipeline.py
yatarkan: Now we have two tests for GenAI vs Optimum scenario - pre-resized (`test_vlm_pipeline_match_optimum_preresized`) and exact match (new `test_vlm_pipeline_match_optimum_with_resize`). Both are testing the same models list `GENAI_VS_OPTIMUM_CASES`, but the second is using `VISION_PREPROCESS=CPP`.
@Wovchena  Does it makes sense to exclude models that exactly match optimum from pre-resized test?
Maybe we should test 3 groups of models:
1. Exact match + CPP vision pre-processing
2. Exact match + OV graph vision pre-processing
3. The rest models with pre-resizing

src/cpp/src/visual_language/llava_next/classes.cpp
RyanMetcalfeInt8: I'll need to check / confirm that this does make a difference somehow. It's at least what optimum does: https://github.com/huggingface/optimum-intel/blob/v1.27.0/optimum/intel/openvino/modeling_visual_language.py#L1423

It of course makes a difference in terms of comparing text_embeddings at this stage between optimum & genai -- but perhaps they reduce to the same thing after `merge_text_and_image_embeddings_llava`.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: Correct, it works on master. After resize implementation changes, it starts failing. I concluded that we are basically getting lucky on master.. as per that CVS ticket, this model will produce different input_ids when chat_template is applied internally vs. manually within the test (even on master this is true, but generated output happens to match)

yatarkan: Probably it can be caused by setting token ids to zero for image tokes - as this is the only change related to llava next model as far as I see now.
Resize implementation changes should not affect this test as it uses the same vision pre processing for both cases (internal and external chat template application).

RyanMetcalfeInt8: From my analysis, it's caused by `add_special_tokens` being different between the two cases being compared(from here: https://github.com/openvinotoolkit/openvino.genai/blob/843b463d9940dac59f8633f6602158abd6ff461e/src/cpp/src/visual_language/inputs_embedder.cpp#L101) 

If I hack that to force `add_special_tokens` to false (or true), then the test passes -- as tokenization becomes more aligned.

src/cpp/src/visual_language/llava_next/classes.cpp
```diff
@@ -77,6 +77,8 @@ EncodedImage VisionEncoderLLaVANext::encode(const ov::Tensor& image, const ov::A
 
 namespace {
 
+// ref:
+// https://github.com/huggingface/transformers/blob/v4.57.3/src/transformers/models/llava_next/modeling_llava_next.py#L109
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: I've set a specific tag so that it will always give the exact code that I referenced. Replacing this with `main` doesn't make sense -- the line number will become inaccurate as things get refactored.

src/cpp/src/visual_language/llava_next/classes.cpp
RyanMetcalfeInt8: After further study, I've found that this doesn't have an impact. I've removed it.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: Just to note that, this is still (expectedly) failing after removing the logic which zeros out token ids. And actually, it wasn't going through that logic anyway, as this is a llava model (not llava-next or llava-next-video).

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: So I guess there should be a separate set of video resolutions which get parameterized into multiple test cases?

In other words, for models that support images & video, they should run IMAGE_RESOLUTIONS X VIDEO_RESOLUTIONS cases? 

Or should we simply resize video to match image resolution?

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: fixed.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: fixed.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: fixed -- changed name to `test_vlm_pipeline_match_optimum_with_resolutions`

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: fixed.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: Yeah, let me try to re-work these GenAI vs. optimum cases with this in mind.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1889,51 +1801,236 @@ def get_nanollava_processor():
...
+
+    # Note: phi4-mm & nanoLLaVA are skipped in _get_ov_model,
+    # phi3-vision is marked as xfail in _get_ov_model
+}
+
+# For these models, we will add both CPP and GRAPH pre-processing tests.
+MODELS_THAT_SUPPORT_GRAPH_PREPROCESSING = [
+    "optimum-intel-internal-testing/tiny-random-llava-next-video"
+    "optimum-intel-internal-testing/tiny-random-phi3-vision"
+    "optimum-intel-internal-testing/tiny-random-phi-4-multimodal"
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1889,51 +1801,236 @@ def get_nanollava_processor():
...
+    "optimum-intel-internal-testing/tiny-random-qwen2vl",
+    "optimum-intel-internal-testing/tiny-random-qwen2.5-vl",
+]
+
+# For these models, we will only add GRAPH pre-processing tests.
+MODELS_THAT_DO_NOT_SUPPORT_CPP_PREPROCESSING = [
+    "optimum-intel-internal-testing/tiny-random-phi-4-multimodal"
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: please see latest changes.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: Please see latest changes. I reworked things so that all genai vs. optimum cases are parametrized via a new `parametrize_optimum_vs_genai` (old `GENAI_VS_OPTIMUM_CASES` where each case is added one-by-one is removed).

For all models, it will run text-only, text+image (with 3resolutions), and text+video or text+image+video (if model supports video) -- and video also has 3 test resolutions.

Both CPP and GRAPH pre-processing cases are run (if model supports them).

`OPTIMUM_VS_GENAI_MODEL_EXPECTED_FAIL_CASES` captures all expected fail cases (there are some new ones here, as we're running more models / more test cases with these changes).

This also populates the same test cases as the older / specific `preresize` test (which is now removed) -- these will have id's with `"preresized"`.

RyanMetcalfeInt8:
@yatarkan -- Please take a look at latest changes to `test_vlm_pipeline.py`, and give some feedback (or approval😄). Thanks!

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
     return image_tensor;
 }
 
+static inline int clip_int(int v, int lo, int hi) {
```
sgonorov: This function is not used anywhere. Maybe we can remove it?

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
...
-
-                    const uint8_t Cc2 = std::min(std::max(std::round(Cc), 0.0f), 255.0f);
-                    dst.buf[(i * target_width + j) * 3 + k] = float(Cc2);
+template <typename FilterFn>
+static void resize_pillow_like(const clip_image_u8& img,
+                               clip_image_u8& dst,
+                               int target_width,
+                               int target_height,
+                               double base_support,
+                               FilterFn filter_fn) {
```
sgonorov, line 155: As we've refactoring this function, maybe make it more safe.
- width/height should be `size_t` instead of `int`
- `base_support` - i assume it can be only 1.0 or 2.0, maybe we should specify these constraints, add some docs etc
- Why do we need `dst` parameter? Maybe we should return it? Also this way we can't check for bounds, it's better to use `std::vector<uint8_t> buf` or something like this. Now it's completely unsafe, no bounds checking etc. Maybe it's better to create a struct with buffer, sizes and pixel getter/setter methods.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1889,52 +1799,257 @@ def get_nanollava_processor():
...
+
+        if model_id in VIDEO_MODEL_IDS:
+            supported_has_video_inputs = [False, True]
+
+        for attn_backend in supported_attn_backends:
+            for preprocessing in supported_preprocessing:
+                for has_image in supported_has_image_inputs:
+                    for has_video in supported_has_video_inputs:
+
+                        def add_test_case(test_id, image_resolution, video_resolution):
```
sgonorov: We can specify this function above the loop, it looks like it doesn't capture any context

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -1889,52 +1799,257 @@ def get_nanollava_processor():
...
+        if model_id in MODELS_THAT_DO_NOT_SUPPORT_CPP_PREPROCESSING:
+            supported_preprocessing = ["GRAPH"]
+
+        supported_has_image_inputs = [False, True]
+        supported_has_video_inputs = [False]
+
+        if model_id in VIDEO_MODEL_IDS:
+            supported_has_video_inputs = [False, True]
+
+        for attn_backend in supported_attn_backends:
```
sgonorov: Too nested, we can use `product` from `itertools` to fix it

src/cpp/src/visual_language/clip.cpp
```diff
@@ -24,113 +24,233 @@ ov::Tensor clip_image_f32_to_tensor(const clip_image_f32& image) {
...
+struct Coeffs1D {
+    int outSize = 0;
+    int ksize = 0;
+    std::vector<int> bounds_xmin;   // size outSize
+    std::vector<int> bounds_count;  // size outSize
+    std::vector<int32_t> kk;        // size outSize * ksize, fixed-point
+};
+
+// ref: https://github.com/python-pillow/Pillow/blob/12.1.0/src/libImaging/Resample.c#L92
+static constexpr int PRECISION_BITS = (32 - 8 - 2);
```
sgonorov, line 63: Can we add more info on what each number means?

RyanMetcalfeInt8: done.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: done.

tests/python_tests/test_vlm_pipeline.py
RyanMetcalfeInt8: done.

src/cpp/src/visual_language/clip.cpp
RyanMetcalfeInt8: > * base_support - i assume it can be only 1.0 or 2.0, maybe we should specify these constraints, add some docs etc

It doesn't *have* to be 1.0 or 2.0, but these are the two values used right now (for bilinear & bicubic, respectively). I've added some comments about it.

> * width/height should be size_t instead of int

   This PR changes the underlying resize implementation, but doesn't modify the existing APIs exposed in `clip.hpp`:
   ```
   void bicubic_resize(const clip_image_u8& img, clip_image_u8& dst, int target_width, int target_height);
   void bilinear_resize(const clip_image_u8& src, clip_image_u8& dst, int target_width, int target_height);
   ```

> * Why do we need dst parameter? Maybe we should return it? Also this way we can't check for bounds, it's better to use std::vector<uint8_t> buf or something like this. Now it's completely unsafe, no bounds checking etc. Maybe it's better to create a struct with buffer, sizes and pixel getter/setter methods.

As stated above, this PR doesn't change the existing APIs, or expected usage, which is like this:
```
clip_image_u8 resized_img;
bicubic_resize(image, resized_img, target_width, target_height);
```

`resized_img` is returned by reference to avoid deep copy of `buf` vector.

It's expected that `bilinear_resize` / `bicubic_resize` is responsible for resizing `buf` to proper size.

With all of that said, I do agree that these functions (and backing types) should be refactored (to avoid extra copies, etc.), `int` -> `size_t`, etc. For example, `clip_image_u8` could be backed by an `ov::Tensor`, or simply take `ov::Tensor` directly. T

This will require some bigger refactoring, affecting all contracts of use, which will require touching each place (across all VLM pre-processors). It's something I'd be willing to do, but it's outside of the scope of this PR.

> Now it's completely unsafe, no bounds checking etc

I did at least add some checks in input / output width and height (that they are `> 0`).

src/cpp/src/visual_language/clip.cpp
RyanMetcalfeInt8: done.

RyanMetcalfeInt8:
@sgonorov -- Please take another look.

> * Rebase it

done.

> * Add more tests, for things like 1x1 images 

At least for `1x1`, this is problematic for optimum vs. genai tests, as it seems that transformers interprets layout by examining dimension values. As so, we get issues like this:
```
The channel dimension is ambiguous. Got image shape (1, 1, 3). Assuming channels are the first dimension. Use the [input_data_format](https://huggingface.co/docs/transformers/main/internal/image_processing_utils#transformers.image_transforms.rescale.input_data_format) parameter to assign the channel dimension.
```

> or with negative dimensions

What do you mean by negative dimensions? I don't think it's possible to create such a test case, is it?

My interpretation here is that you'd like to test that these resize implementations specifically handle such corner cases.. but it's kind of tricky, as the only way to trigger use of resize is via VLM pre-processing.

sgonorov:
@RyanMetcalfeInt8  I meant we need some tests which should test that code is failing on bad inputs properly.

PR #3173, opened 2026-01-08
2026.0 deprecated APIs removal: Streamer `put` method, `bool` return for callback, ChunkStreamer

as-suvorov:
## Description

CVS-###

Fixes #(issue)

## Checklist:

### DIFF
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -16,17 +16,17 @@
         "react-dom": "18.3.1"
       },
       "devDependencies": {
-        "@docusaurus/eslint-plugin": "3.7.0",
-        "@docusaurus/module-type-aliases": "3.7.0",
-        "@docusaurus/tsconfig": "3.7.0",
-        "@docusaurus/types": "3.7.0",
-        "@eslint/eslintrc": "3.2.0",
-        "@eslint/js": "9.20.0",
-        "eslint": "9.20.1",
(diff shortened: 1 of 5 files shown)


as-suvorov:
@sbalandi I want to double check `bool` as return value deprecation: https://github.com/openvinotoolkit/openvino.genai/pull/3173/changes#diff-8239b764dc3145dead10d00ac5333c27f924b37d5b9e37e7bfb1cde555b0a161L25.
Could you remind what was the motivation to deprecate `bool` flag?
As I understand for python empty return was casted to `False` meaning continue generation. I guess it quite convenient.
For reference: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_llm_pipeline.py

as-suvorov:
> @sbalandi I want to double check `bool` as return value deprecation: https://github.com/openvinotoolkit/openvino.genai/pull/3173/changes#diff-8239b764dc3145dead10d00ac5333c27f924b37d5b9e37e7bfb1cde555b0a161L25. Could you remind what was the motivation to deprecate `bool` flag? As I understand for python empty return was casted to `False` meaning continue generation. I guess it quite convenient. For reference: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_llm_pipeline.py

So pybind already handles empty return for python, please ignore above message.

PR #3180, opened 2026-01-13
[Whisper]: add word level timestamps

as-suvorov:
## Description

CVS-174652

Fixes #(issue)

## Checklist:

### DIFF
--- a/samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
+++ b/samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include "audio_utils.hpp"
--- a/src/cpp/include/openvino/genai/whisper_generation_config.hpp
+++ b/src/cpp/include/openvino/genai/whisper_generation_config.hpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
--- a/src/cpp/src/whisper/models/statefull_decoder.cpp
+++ b/src/cpp/src/whisper/models/statefull_decoder.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2024-2025 Intel Corporation
+// Copyright (C) 2024-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include "statefull_decoder.hpp"
--- a/tests/python_tests/test_whisper_pipeline.py
+++ b/tests/python_tests/test_whisper_pipeline.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2023-2025 Intel Corporation
+# Copyright (C) 2023-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 import openvino_genai as ov_genai
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2025 Intel Corporation
+# Copyright (C) 2018-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
 
--- a/cmake/features.cmake
+++ b/cmake/features.cmake
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2025 Intel Corporation
+# Copyright (C) 2018-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
 
--- a/cmake/templates/version.cpp.in
+++ b/cmake/templates/version.cpp.in
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include "openvino/genai/version.hpp"
--- a/cmake/templates/version.hpp.in
+++ b/cmake/templates/version.hpp.in
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
--- a/cmake/version.cmake
+++ b/cmake/version.cmake
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2025 Intel Corporation
+# Copyright (C) 2018-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
 
--- a/samples/CMakeLists.txt
+++ b/samples/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2018-2025 Intel Corporation
+# Copyright (C) 2018-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 #
 if(ENABLE_SAMPLES)
--- a/samples/c/text_generation/CMakeLists.txt
+++ b/samples/c/text_generation/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2023-2025 Intel Corporation
+# Copyright (C) 2023-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 find_package(OpenVINOGenAI REQUIRED
@@ -30,4 +30,3 @@ set (SAMPLE_LIST
 foreach(sample IN LISTS SAMPLE_LIST)
     add_sample_executable(${sample})
 endforeach()
-
--- a/samples/cpp/image_generation/CMakeLists.txt
+++ b/samples/cpp/image_generation/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2023-2025 Intel Corporation
+# Copyright (C) 2023-2026 Intel Corporation
 # SPDX-License-Identifier: Apache-2.0
 
 find_package(OpenVINOGenAI REQUIRED
--- a/samples/cpp/image_generation/benchmark_image_gen.cpp
+++ b/samples/cpp/image_generation/benchmark_image_gen.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include "openvino/genai/image_generation/text2image_pipeline.hpp"
--- a/samples/cpp/image_generation/heterogeneous_stable_diffusion.cpp
+++ b/samples/cpp/image_generation/heterogeneous_stable_diffusion.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include "openvino/genai/image_generation/text2image_pipeline.hpp"
--- a/samples/cpp/image_generation/image2image.cpp
+++ b/samples/cpp/image_generation/image2image.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include "openvino/genai/image_generation/image2image_pipeline.hpp"
--- a/samples/cpp/image_generation/image2image_concurrency.cpp
+++ b/samples/cpp/image_generation/image2image_concurrency.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2025 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
 #include <vector>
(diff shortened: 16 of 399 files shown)


samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
```diff
@@ -20,19 +20,21 @@ int main(int argc, char* argv[]) try {
...
-    if (device == "NPU" || device.find("GPU") != std::string::npos) {  // need to handle cases like "GPU", "GPU.0" and "GPU.1"
+    if (device == "NPU" ||
+        device.find("GPU") != std::string::npos) {  // need to handle cases like "GPU", "GPU.0" and "GPU.1"
         // Cache compiled models on disk for GPU and NPU to save time on the
         // next run. It's not beneficial for CPU.
         ov_config = get_config_for_cache();
     }
 
-    ov::genai::WhisperPipeline pipeline(models_path, device, ov_config);
+    ov::genai::WhisperPipeline pipeline(models_path, device, ov::genai::word_timestamps(true));
```
Wovchena: `ov_config` is ignored

src/cpp/include/openvino/genai/whisper_generation_config.hpp
```diff
@@ -65,6 +65,17 @@ class OPENVINO_GENAI_EXPORTS WhisperGenerationConfig : public GenerationConfig {
     // Note that a segment of text refers to a sequence of one or more words, rather than individual words.
     bool return_timestamps = false;
 
+    // If `true` the pipeline will return word-level timestamps.
+    // When enabled ov::genai::word_timestamps(true) property should be passed to WhisperPipeline constructor:
```
Wovchena, line 69: What's the point in `WhisperGenerationConfig::word_timestamps` if `ov::genai::word_timestamps` is required anyway?

samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
Wovchena: It crashes:
./build/samples/cpp/whisper_speech_recognition/whisper_speech_recognition ./whisper-base/ ./i/how_are_you_doing_today.wav CPU
Segmentation fault (core dumped)

src/cpp/src/whisper/alignment_heads.cpp
```diff
@@ -0,0 +1,45 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "alignment_heads.hpp"
+
+namespace ov::genai {
+
+std::vector<ov::Tensor> get_whisper_alignments_heads_qks(
+    ov::InferRequest& request,
+    const std::vector<std::pair<size_t, size_t>>& alignment_heads) {
+    // [layers] * [batch, num_heads, seq_len, frame_len] -> [layers] * [batch, seq_len, frame_len]
```
Wovchena, line 11: Change the indentation so the list of args and function body are distinguishable. As an option, you can move `) {` to a separate line

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -563,6 +572,107 @@ def test_shortform(model_descr):
     )
 
 
+@pytest.fixture
+def whisper_librispeech_10_openai_tiny_reference():
+    json_path = pathlib.Path(
+        "tests/python_tests/data/whisper/librispeech_asr_dummy_10_openai_whisper_tiny_results.json"
```
Wovchena: Use path relative to `__file__`

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -563,6 +572,107 @@ def test_shortform(model_descr):
...
+    assert accuracy > 0.95
+
+
+@pytest.mark.parametrize("model_descr", get_whisper_models_list())
+@pytest.mark.parametrize(
+    "sample_from_dataset", [*get_fixture_params_for_n_whisper_dataset_samples(n=2, long_form=True)], indirect=True
+)
+@pytest.mark.xfail(condition=(sys.platform == "darwin"), reason="Ticket - 173169")
+def test_longform_audio_with_word_level_timestamps(model_descr, sample_from_dataset):
+    _, _, _, genai_pipe = read_whisper_model(model_descr, word_timestamps=True)
```
Wovchena:
```suggestion
    genai_pipe = read_whisper_model(model_descr, word_timestamps=True)[3]
```

tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py
Wovchena, line 1: `tests/python_tests/data/whisper/librispeech_asr_dummy_10_openai_whisper_tiny_results.json` should be cached and by pytest and if the cache is missing, recover it

src/cpp/src/whisper/word_level_timestamps.cpp
```diff
@@ -0,0 +1,683 @@
...
+                    for (int offset = -static_cast<int>(pad_width); offset <= static_cast<int>(pad_width); ++offset) {
+                        int neighbor_frame = static_cast<int>(frame) + offset;
+                        if (neighbor_frame >= 0 && neighbor_frame < static_cast<int>(frame_len)) {
+                            size_t index = batch * seq_len * frame_len + seq * frame_len + neighbor_frame;
+                            window.push_back(input_data[index]);
+                        }
+                    }
+                    std::nth_element(window.begin(), window.begin() + window.size() / 2, window.end());
+                    size_t output_index = batch * seq_len * frame_len + seq * frame_len + frame;
+                    output_data[output_index] = window[window.size() / 2];
```
Wovchena:
```suggestion
                    output_data[output_index] = window[(window.size() + 1) / 2];
```
Should it be rounding up everywhere since you need median?

src/cpp/src/whisper/word_level_timestamps.cpp
```diff
@@ -0,0 +1,683 @@
...
+    for (const auto& tensor : alignment_qks) {
+        const ov::Shape& shape = tensor.get_shape();
+        const size_t seq_len = shape[1];
+        const size_t frame_len = shape[2];
+
+        ov::Tensor reduced_batch_tensor{ov::element::f32, {seq_len, frame_len}};
+        auto* input_data = tensor.data<float>();
+        auto* output_data = reduced_batch_tensor.data<float>();
+
+        std::memcpy(output_data, input_data, seq_len * frame_len * sizeof(float));
```
Wovchena, line 166: Create a Tensor view instead of copy. If you need that copy, `Tensor::copy_to()` may be shorter to write

src/cpp/src/whisper/word_level_timestamps.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+    for (const auto& tensor : alignment_qks) {
+        const ov::Shape& shape = tensor.get_shape();
+        const size_t batch_size = shape[0];
+        const size_t seq_len = shape[1];
+        const size_t frame_len = shape[2];
+
+        if (n_frames > frame_len) {
+            throw std::runtime_error("Requested n_frames exceeds tensor frame length.");
+        }
```
Wovchena:
```suggestion
        OPENVINO_ASSERT(n_frames <= frame_len, "Requested n_frames exceeds tensor frame length: ", frame_len);
```

src/cpp/src/whisper/word_level_timestamps.cpp
```diff
@@ -0,0 +1,683 @@
...
+            filtered_words.push_back(word);
+        }
+    }
+
+    return filtered_words;
+}
+
+std::string trim(const std::string& text) {
+    std::string result = text;
+    result.erase(result.begin(), std::find_if(result.begin(), result.end(), [](unsigned char ch) {
```
Wovchena, line 522:
```suggestion
    result.erase(result.begin(), std::find_if(result.begin(), result.end(), [](char ch) {
```
Strings consist of `char`s

src/cpp/include/openvino/genai/whisper_generation_config.hpp
as-suvorov: The point of `WhisperGenerationConfig::word_timestamps` is to give flexibility. User can pass `ov::genai::word_timestamps` to ctor, this would prepare model for word level timestamps generation. Word level timestamps generation itself is done with additional decoder inference thus affects performance. I assume there could be scenario (ovms serving for example) where word level timestamps generation should be conditional.

src/cpp/src/whisper/alignment_heads.cpp
as-suvorov: I do not format code manually. This is how clang-format works. Moving `) {` to a new line reverted by clang-format as well as move `ov::InferRequest& request,` to previous line as well as newline right after `) {`.

tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py
as-suvorov: `tests/python_tests/data/whisper/librispeech_asr_dummy_10_openai_whisper_tiny_results.json` is in git, what is the point to cache it?

src/cpp/src/whisper/word_level_timestamps.cpp
as-suvorov: I don't think it's important. I'll check if round up gives better accuracy and apply the change if so

src/cpp/src/whisper/word_level_timestamps.cpp
as-suvorov: `unsigned` used to satisfy `isspace` behavior: https://en.cppreference.com/w/cpp/string/byte/isspace.html
From article:

> Like all other functions from [<cctype>](https://en.cppreference.com/w/cpp/header/cctype.html), the behavior of std::isspace is undefined if the argument's value is neither representable as unsigned char nor equal to [EOF](https://en.cppreference.com/w/cpp/io/c.html). To use these functions safely with plain chars (or signed chars), the argument should first be converted to unsigned char

samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
as-suvorov: reverted sample

tests/python_tests/test_whisper_pipeline.py
as-suvorov: Implemented

tests/python_tests/test_whisper_pipeline.py
as-suvorov: Done

src/cpp/src/whisper/word_level_timestamps.cpp
as-suvorov: I tried 2 approaches:
1. ROI tensor `Tensor(input_tensor, {0,0,0}, {1, seq_len, frame_len)` - doens't actually reduce dim
2. `copy_to` from ROI tensor to reduced batch tensor - doesn't work as copy_to require same shape.
Keeping `memcpy` for the moment. Maybe you can correct me

src/cpp/src/whisper/word_level_timestamps.cpp
as-suvorov: Applied

samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
Wovchena: Maybe it wasn't a bad idea to update the samples. The feature will be forgotten otherwise

tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py
Wovchena: Cache initialization will trigger `tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py` making this code alive

src/cpp/src/whisper/models/statefull_decoder.cpp
```diff
@@ -101,4 +110,14 @@ ov::Tensor WhisperStatefullDecoder::create_host_tensor(const element::Type eleme
         return ov::Tensor(element_type, shape);
     }
 }
+
+std::vector<Tensor> WhisperStatefullDecoder::get_alignments_heads_qks(
+    const std::vector<std::pair<size_t, size_t>>& alignment_heads) {
+    if (!m_decompose_cross_attention_spda_ops) {
+        OPENVINO_THROW("Encoder attention heads are not decomposed. Cannot get encoder attention QKs.");
+    }
+
```
Wovchena:
```suggestion
    OPENVINO_ASSERT(m_decompose_cross_attention_spda_ops, "Encoder attention heads are not decomposed. Cannot get encoder attention QKs.");
```

samples/cpp/whisper_speech_recognition/whisper_speech_recognition.cpp
as-suvorov: Sure, I added a separate ticket to update samples. Another for documentation. Let's address it in a further PRs

tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py
as-suvorov: You mean you want to remove `.json` from git, right? can I implement it in next PR?

src/cpp/src/whisper/models/statefull_decoder.cpp
as-suvorov: Applied

tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py
Wovchena: Yes. Yes

src/cpp/src/whisper/word_level_timestamps.cpp
Wovchena: Are there any findings?

tests/python_tests/data/whisper/generate_openai_word_level_timestamps.py
as-suvorov: Thanks, ticket created

src/cpp/src/whisper/word_level_timestamps.cpp
as-suvorov: 'window[window.size() / 2]' takes middle element for odd case and upper element for even case. I aligned implementation with np. np takes average of 2 for even case. But it actually has 0 contribution since window size is 7. And it affects only edges of tensor.

Wovchena (approved):

PR #3084, opened 2025-11-28
Conditional visual token pruning for QWen-VL models.

yangwang201911:
_This is a clean rebase of [PR#2714](https://github.com/openvinotoolkit/openvino.genai/pull/2714) to simplify the review process._ 
- Implement conditional visual token pruning for QWen-VL models.
-- Paper: [CDPruner (arXiv)](https://arxiv.org/pdf/2506.10967)
-- Code: [GitHub Repository](https://github.com/Theia-4869/CDPruner)
- Add configurations to benchmark.py and WWB tools

Tickets: CVS-173220
Related PRs:
- WWB: https://github.com/openvinotoolkit/openvino.genai/pull/2928
- LLM_Bench: https://github.com/openvinotoolkit/openvino.genai/pull/2926
- Samples: https://github.com/openvinotoolkit/openvino.genai/pull/2925
- Document: https://github.com/openvinotoolkit/openvino.genai/pull/2861

PR #3184, opened 2026-01-14
[llm_bench] conditional visual token pruning configuration

wgzintel:
Function PR https://github.com/openvinotoolkit/openvino.genai/pull/3084
Ticket: CVS-173849

### DIFF
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -16,17 +16,17 @@
         "react-dom": "18.3.1"
       },
       "devDependencies": {
-        "@docusaurus/eslint-plugin": "3.7.0",
-        "@docusaurus/module-type-aliases": "3.7.0",
-        "@docusaurus/tsconfig": "3.7.0",
-        "@docusaurus/types": "3.7.0",
-        "@eslint/eslintrc": "3.2.0",
-        "@eslint/js": "9.20.0",
-        "eslint": "9.20.1",
-        "eslint-plugin-react": "7.37.4",
+        "@docusaurus/eslint-plugin": "3.9.2",
+        "@docusaurus/module-type-aliases": "3.9.2",
+        "@docusaurus/tsconfig": "3.9.2",
+        "@docusaurus/types": "3.9.2",
+        "@eslint/eslintrc": "3.3.3",
+        "@eslint/js": "9.39.2",
+        "eslint": "9.39.2",
+        "eslint-plugin-react": "7.37.5",
         "globals": "15.15.0",
         "typescript": "5.7.3",
-        "typescript-eslint": "8.24.1"
+        "typescript-eslint": "8.53.1"
       },
       "engines": {
         "node": ">=18.0"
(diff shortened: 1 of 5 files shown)


sbalandi (approved):

tools/llm_bench/benchmark.py
```diff
@@ -118,6 +132,19 @@ def get_argprser():
...
     parser.add_argument('-bs', '--batch_size', type=int, default=1, required=False, help='Batch size value')
     parser.add_argument('--num_beams', type=int, default=1, help='Number of beams in the decoding strategy, activates beam_search if greater than 1')
+    parser.add_argument(
+        "--pruning_ratio",
+        type=ratio_type,
+        default=None,
+        help="(optional): Percentage of visual tokens to prune (valid range: 0-100). If this option is not provided, pruning is disabled.",
+    )
+    parser.add_argument(
+        "--relevance_weight",
```
Wovchena, line 143: Extend tests with the new options

sbalandi:
@rkazants could you please take a look, Is the change request still valid ?

tools/llm_bench/benchmark.py
```diff
@@ -118,6 +132,19 @@ def get_argprser():
     )
     parser.add_argument('-bs', '--batch_size', type=int, default=1, required=False, help='Batch size value')
     parser.add_argument('--num_beams', type=int, default=1, help='Number of beams in the decoding strategy, activates beam_search if greater than 1')
+    parser.add_argument(
+        "--pruning_ratio",
+        type=ratio_type,
+        default=None,
+        help="(optional): Percentage of visual tokens to prune (valid range: 0-100). If this option is not provided, pruning is disabled.",
```
rkazants: not sure that we need `(optional)...` word in the description. Optional parameter is set through parser.
Let us remove `(optional)` to align with other optional parameters. Since this is optional, I think you need to set `required=False` as well.
The same comment is for `relevance_weight`.

tools/llm_bench/benchmark.py
```diff
@@ -42,6 +42,20 @@ def num_infer_count_type(x):
     return x
 
 
+def ratio_type(value):
+    ivalue = int(value)
```
rkazants: let us strictly request this parameter to be integer. That is because you are doing implicit casting for floats etc.

tools/llm_bench/benchmark.py
```diff
@@ -42,6 +42,20 @@ def num_infer_count_type(x):
     return x
 
 
+def ratio_type(value):
```
rkazants:
```suggestion
def ratio_type(value: int):
```

tools/llm_bench/benchmark.py
```diff
@@ -42,6 +42,20 @@ def num_infer_count_type(x):
...
+    ivalue = int(value)
+    if ivalue < 0 or ivalue > 100:
+        raise argparse.ArgumentTypeError(f"pruning_ratio must be between 0 and 100, got {value}")
+    return ivalue
+
+
+def weight_0_1(value):
```
rkazants:
```suggestion
def weight_0_1(value: float):
```

tools/llm_bench/benchmark.py
peterchen-intel: Updated.

tools/llm_bench/benchmark.py
peterchen-intel: Updated.

tools/llm_bench/benchmark.py
```diff
@@ -42,6 +42,20 @@ def num_infer_count_type(x):
     return x
 
 
+def pruning_ratio_type(value: str) -> int:
```
rkazants, line 45:
```suggestion
def pruning_ratio_type(value: int):
```
does this change work?

PR #3188, opened 2026-01-14
Fix CDPruner random failures in chat mode test case

yangwang201911:
## Description
CDPruner randomly crashes in chat mode during multi-turn conversations. The root cause is that m_history stores unpruned prompts with original vision token counts, leading to token mismatch between chat history and KV cache state.

**Solution:**
The fix ensures three critical parameters stay synchronized:
- `m_history` - Update last user message by removing N `<|vision_pad|>` tokens (where N = pruned vision tokens)
- `kv_cache` - Truncate unpruned current turn, then append pruned tokens
- `m_pre_hist_lenght` - Update to final KV cache size for next iteration

Tickets: CVS-179274

### DIFF
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -2286,7 +2286,8 @@ def test_cdpruner_continuous_batching(
     car_tensor: openvino.Tensor,
 ):
     """Test CDPruner with continuous batching pipeline.
-    Verifies that pruning_ratio=1 produces the same output as pruning_ratio=0 (no pruning)."""
+    Verifies that a small non-zero pruning ratio (pruning_ratio=1, i.e. ~1% pruning)
+    produces the same output as the unpruned baseline with pruning_ratio=0."""
     # Baseline run with pruning_ratio=0
     generation_config_baseline = GenerationConfig()
     generation_config_baseline.max_new_tokens = 10
@@ -2428,7 +2429,6 @@ def test_cdpruner_continuous_batching_chat_history(
     assert baseline1 == pruned1, f"Turn 1 mismatch: baseline='{baseline1}', pruned='{pruned1}'"
     assert baseline2 == pruned2, f"Turn 2 mismatch: baseline='{baseline2}', pruned='{pruned2}'"
     assert baseline3 == pruned3, f"Turn 3 mismatch: baseline='{baseline3}', pruned='{pruned3}'"
-    assert results[0].texts[0].strip() != "", "Result should not be empty"
 
 
 def test_vlm_prompt_lookup_functionality(cat_tensor):


src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -1108,6 +1108,35 @@ ov::Tensor InputsEmbedderQwen2VL::get_inputs_embeds(const std::string& unified_p
     }
 
     ov::Tensor input_ids = get_encoded_input_ids(unified_prompt, metrics);
+
```
xipingyan: Could you move this part to common place?
I am not sure, please consider it.

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -1117,7 +1146,20 @@ ov::Tensor InputsEmbedderQwen2VL::get_inputs_embeds(const std::string& unified_p
     int64_t image_pad_token_id = m_vision_token_ids["image_pad"];
     int64_t video_pad_token_id = m_vision_token_ids["video_pad"];
```
xipingyan, line 1118: Move this part to head of `clean_historical_vision_tokens`, you can use `vision_start_token_id`  directly.

xipingyan: I think we need to process in order to keep code consistency.

src/cpp/src/visual_language/qwen2vl/classes.cpp
```diff
@@ -20,6 +20,7 @@
 #include "openvino/op/tile.hpp"
 #include "openvino/op/if.hpp"
 #include "openvino/op/concat.hpp"
+#include "logger.hpp"
```
Wovchena:
```suggestion
```

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -334,7 +334,17 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
 
         std::string decoded_results = decoded.texts.at(0);
         if (m_is_chat_conversation) {
-            m_inputs_embedder->update_chat_history(decoded_results, finish_info.streaming_finish_status);
+            // Get original prompt from history and update with pruning if needed
+            std::string original_prompt = m_history.last()["content"].get_string();
+
+            auto pruned_prompt = m_inputs_embedder->update_chat_history(decoded_results,
```
yatarkan: This also should be addressed in generate with chat history argument + in ContinuousBatching pipeline (both generate variants - with prompts and histories)

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -334,7 +334,17 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
 
         std::string decoded_results = decoded.texts.at(0);
         if (m_is_chat_conversation) {
-            m_inputs_embedder->update_chat_history(decoded_results, finish_info.streaming_finish_status);
+            // Get original prompt from history and update with pruning if needed
+            std::string original_prompt = m_history.last()["content"].get_string();
```
yatarkan: In some scenarios last message in history can be not from user when using `ChatHistory` argument in generate.
`ChatHistoryInternalState` is now tracking last user message index. Consider this when enabling vision token pruning for `ChatHistory` use case.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -334,7 +334,17 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+
+            auto pruned_prompt = m_inputs_embedder->update_chat_history(decoded_results,
+                                                                        finish_info.streaming_finish_status,
+                                                                        original_prompt);
+
+            if (pruned_prompt.has_value()) {
+                m_history.pop_back();
+                m_history.push_back({{"role", "user"}, {"content", pruned_prompt.value()}});
```
yatarkan: For new `ChatHistory` use case (as an argument in `generate()`) original user history should not be modified. Normalized message content (with expanded vision token placeholders) is stored in `ChatHistoryInternalState`. `VLMChatContext` is responsible for processing history and its internal state and it also has access to `InputsEmbedder`.

yatarkan (changes requested):

src/cpp/src/visual_language/pipeline.cpp
yangwang201911: Updated.

src/cpp/src/visual_language/pipeline.cpp
yangwang201911: Updated.

src/cpp/src/visual_language/pipeline.cpp
yangwang201911: Sure. Using metadata via `VLMChatContext::apply_pruning_to_last_message() ` for `ChatHistory` case. User's original history remains unchanged.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -2075,3 +2112,80 @@ def test_cdpruner_continuous_batching(
...
+    result2 = ov_continuous_batching_pipe_qwen2vl.generate(
+        [history], images=[[car_tensor]], generation_config=[generation_config]
+    )
+    assert result2[0].texts[0].strip() != "", "Second turn result should not be empty"
+    history.append({"role": "assistant", "content": result2[0].texts[0]})
+
+    # Third turn without image (returns GenerationResult, not VLMDecodedResults)
+    history.append({"role": "user", "content": "What did you see in total?"})
+    result3 = ov_continuous_batching_pipe_qwen2vl.generate([history], generation_config=[generation_config])
+    assert result3[0].m_generation_ids[0].strip() != "", "Third turn result should not be empty"
```
Wovchena: A model always generates something so the test won't catch any relevant problem. Configure the algo to be as close as possible to simple generation. Set `pruning_ratio` to 1 or something and require the predictions to match

src/cpp/src/visual_language/vlm_chat_context.cpp
```diff
@@ -193,4 +193,27 @@ std::string VLMChatContext::multipart_message_to_string(
...
+        return;
+    }
+
+    auto& last_metadata = messages_metadata[last_user_idx];
+
+    // Get the pruned prompt from inputs embedder if pruning is active
+    std::optional<std::string> pruned_prompt =
+        m_inputs_embedder.get_last_updated_prompt(last_metadata.normalized_content);
+
+    if (pruned_prompt.has_value() && !pruned_prompt->empty()) {
```
Wovchena:
```suggestion
    if (pruned_prompt.has_value()) {
```
I guess it's fine to set an empty string as `last_metadata.pruned_content` because it's empty by defailt

src/cpp/src/visual_language/vision_token_pruning_processor.hpp
```diff
@@ -207,24 +221,29 @@ class VisionTokenPruningProcessor {
...
+     * @param is_chat_conversation Whether in chat mode
+     * @param prev_hist_length_inout Previous history length (modified in-place if chat + pruning occurred)
      * @return std::optional<PruningResult> with pruned tensors if pruning occurred, std::nullopt otherwise
      */
     std::optional<PruningResult> execute(const PruningContext& context,
                                          ov::Tensor& position_ids,
                                          utils::KVCacheState& kv_cache_state,
-                                         size_t prev_hist_length);
+                                         bool is_chat_conversation,
+                                         size_t& prev_hist_length_inout);
```
Wovchena:
```suggestion
                                         size_t& prev_hist_length_input);
```
Do you need the word "input" at all in the name?
Or is it `in_out`?

src/cpp/src/visual_language/inputs_embedder.hpp
```diff
@@ -230,6 +233,12 @@ class InputsEmbedder {
 
         virtual void update_chat_history(const std::string& decoded_results, const ov::genai::GenerationStatus generation_finish_status);
 
+        // Get updated prompt with modified vision tokens (e.g., pruned)
+        // Base implementation returns nullopt - subclasses override if they support vision token modifications
+        virtual std::optional<std::string> get_last_updated_prompt(const std::string& original_prompt) const {
```
Wovchena:
```suggestion
        virtual std::optional<std::string> get_last_pruned_prompt(const std::string& original_prompt) const {
```
So the comments
>  Get updated prompt with modified vision tokens (e.g., pruned)

and

> gets last updated prompt after vision token pruning

aren't needed.

src/cpp/src/visual_language/inputs_embedder.hpp
```diff
@@ -230,6 +233,12 @@ class InputsEmbedder {
...
+            return std::nullopt;
```
Wovchena:
```suggestion
            OPENVINO_THROW_NOT_IMPLEMENTED("Base class get_last_pruned_prompt() isn't implemented");
```
And remove comment
> Base implementation returns nullopt - subclasses override if they support vision token modifications

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -265,6 +266,27 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
...
+                size_t last_user_idx = history_state->get_last_user_message_index();
+
+                // Get original prompt content from the last user message
+                const auto& last_user_message = m_history[last_user_idx];
+                std::string original_prompt = last_user_message["content"].get_string();
+
+                // Get the pruned prompt after CDPruner
+                std::optional<std::string> pruned_prompt = m_inputs_embedder->get_last_updated_prompt(original_prompt);
+
+                if (pruned_prompt.has_value()) {
```
Wovchena: If the prompt wasn't pruned you can still replace `m_history[last_user_idx]` with `user_message` because they are identical. At the same time it's unlikely to happen because pruning is expected to do something. You should run this code unconditionally and remove `optional` from `get_last_updated_prompt()`

src/cpp/src/visual_language/vision_token_pruning_processor.cpp
```diff
@@ -81,6 +81,83 @@ std::vector<std::vector<size_t>> VisionTokenPruningProcessor::get_last_selected_
...
+std::string VisionTokenPruningProcessor::get_last_updated_prompt(const std::string& original_prompt,
+                                                                 const std::string& vision_start_token,
+                                                                 const std::string& vision_end_token,
+                                                                 const std::string& image_pad_token,
+                                                                 const std::string& video_pad_token) const {
+    if (m_last_keep_flags.empty()) {
+        return original_prompt;
+    }
+
+    const auto& keep_flags = m_last_keep_flags;
```
Wovchena: What is `keep_flags` alias for? You could have kept using `m_last_keep_flags`. The method is `const` anyway

Wovchena (changes requested):

tests/python_tests/test_vlm_pipeline.py
yangwang201911: Updated. The test now uses `pruning_ratio=1` vs `pruning_ratio=0` baseline and asserts outputs match exactly.

src/cpp/src/visual_language/vlm_chat_context.cpp
yangwang201911: Done.

src/cpp/src/visual_language/vision_token_pruning_processor.hpp
yangwang201911: Removed `inout`.

src/cpp/src/visual_language/inputs_embedder.hpp
yangwang201911: Done.

src/cpp/src/visual_language/inputs_embedder.hpp
yangwang201911: Done.

src/cpp/src/visual_language/pipeline.cpp
yangwang201911: Updated.

src/cpp/src/visual_language/vision_token_pruning_processor.cpp
yangwang201911: Removed `keep_flags`.

src/cpp/src/visual_language/inputs_embedder.hpp
Wovchena: The function name still includes "updated". Is there a reason to prefer it over "pruned"

yangwang201911: I missed updating these names earlier. Updated now.

Wovchena:
@yatarkan's status blocks the merge

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include "visual_language/vision_registry.hpp"
 #include "visual_language/vlm_chat_context.hpp"
+#include "visual_language/chat_history_state.hpp"
```
yatarkan: chat_history_state.hpp is included in vlm_chat_context.hpp

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -278,6 +279,20 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
 
         std::string decoded_results = decoded.texts.at(0);
         if (m_is_chat_conversation) {
+            // Update m_history with pruned content if pruning is enabled
+            if (generation_config.pruning_ratio > 0) {
+                auto history_state = ChatHistoryInternalState::get_or_create(m_history, m_vision_registry);
+                size_t last_user_idx = history_state->get_last_user_message_index();
+
+                OPENVINO_ASSERT(last_user_idx < m_history.size(), "Invalid last_user_idx");
```
yatarkan: Redundant assert - `find_last_user_message_index()` iterates over `history.size()` and throws if no user message found in chat history.

src/cpp/src/visual_language/chat_history_state.hpp
```diff
@@ -38,7 +38,12 @@ struct MessageMetadata {
     // Contains vision placeholders after prompt normalization.
     // Empty for system/assistant messages.
     std::string normalized_content;
-    
+
+    // Contains content after vision token pruning (if applicable).
```
yatarkan: Do we need separate filed for pruned content? Is it possible to update `normalized_content` directly?

src/cpp/src/visual_language/pipeline.cpp
yangwang201911: Removed.

src/cpp/src/visual_language/pipeline.cpp
yangwang201911: Removed.

src/cpp/src/visual_language/chat_history_state.hpp
yangwang201911: Good point! Removed `pruned_content ` field.

yatarkan (approved):

PR #3190, opened 2026-01-14
wwb/visualtext_evaluator: try to set processor.chat_template from tokenizer if it's not already set

RyanMetcalfeInt8:
## Description

There seems to be some differences in applying chat templates between optimum & OV GenAI cases.

For example, running wwb with optimum, using `MiniCPM-V-2_6`:
With current version of wwb, `preprocess_inputs` will manually apply the chat template here: [modeling_visual_language.py#L2108](https://github.com/huggingface/optimum-intel/blob/v1.27.0/optimum/intel/openvino/modeling_visual_language.py#L2108) 

This is because processor.chat_template hasn't been set.

When running OV GenAI, it will use the chat_template from the tokenizer config.

For `MiniCPM-V-2_6`, this is a problem as the chat_template defined by the tokenizer config will add a default system prompt -- so this gets added for OV GenAI case, but not when running optimum.

To better align chat_templating, I've added some logic to set processor.chat_template (if it isn't already set, and if it's defined for the tokenizer).

With this change, we get much better similarity scores for `MiniCPM-V-2_6` (and perhaps others).

Related / found while working this ticket:
CVS-178941

## Checklist:

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
+++ b/tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
@@ -122,7 +122,10 @@ class VisualTextEvaluator(TextEvaluator):
             # OV GenAI will use the chat_template from tokenizer.
             # Optimum (within preprocess_inputs) will use the chat_template only if it has been defined for processor.
             # So, if the chat_template hasn't been set for processor, try to set it from the tokenizer.
-            if getattr(processor, "chat_template", None) is None and getattr(tokenizer, "chat_template", None) is not None:
+            if (
+                getattr(processor, "chat_template", None) is None
+                and getattr(tokenizer, "chat_template", None) is not None
+            ):
                 processor.chat_template = tokenizer.chat_template
             inputs = preprocess_inputs(prompt, image, processor, tokenizer, config=model.config, video=video)
             tokens = model.generate(


tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
```diff
@@ -118,6 +118,15 @@ def default_gen_answer(
             preprocess_inputs = MODEL_TYPE_TO_CLS_MAPPING[
                 model.config.model_type
             ].preprocess_inputs
+
+            # OV GenAI will use the chat_template from tokenizer.
+            # Optimum (within preprocess_inputs) will use the chat_template only if it has been defined for processor.
```
rkazants, line 123: After these changes we will have misalignment between GenAI and optimum-intel. It is not clear how similarity scores have been improved.

MaximProshin:
@yatarkan , please also review it. 
Actually I also saw situations when the enabled chat_template produces different WWB numbers in optimum and GenAI for the same model. When disabled, numbers are exactly the same.

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
RyanMetcalfeInt8: How is it misaligned? The intent of this change it to improve alignment by making sure that optimum actually applies the chat template, instead of falling into `else` clauses (like here: https://github.com/huggingface/optimum-intel/blob/v1.27.0/optimum/intel/openvino/modeling_visual_language.py#L2108)

PR #3193, opened 2026-01-15
[WWB] conditional visual token pruning configuration

wgzintel:
Function PR https://github.com/openvinotoolkit/openvino.genai/pull/3084
Tickets: CVS-173849

### DIFF
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -16,17 +16,17 @@
         "react-dom": "18.3.1"
       },
       "devDependencies": {
-        "@docusaurus/eslint-plugin": "3.7.0",
-        "@docusaurus/module-type-aliases": "3.7.0",
-        "@docusaurus/tsconfig": "3.7.0",
-        "@docusaurus/types": "3.7.0",
-        "@eslint/eslintrc": "3.2.0",
-        "@eslint/js": "9.20.0",
-        "eslint": "9.20.1",
-        "eslint-plugin-react": "7.37.4",
+        "@docusaurus/eslint-plugin": "3.9.2",
+        "@docusaurus/module-type-aliases": "3.9.2",
+        "@docusaurus/tsconfig": "3.9.2",
+        "@docusaurus/types": "3.9.2",
+        "@eslint/eslintrc": "3.3.3",
+        "@eslint/js": "9.39.2",
+        "eslint": "9.39.2",
+        "eslint-plugin-react": "7.37.5",
         "globals": "15.15.0",
         "typescript": "5.7.3",
-        "typescript-eslint": "8.24.1"
+        "typescript-eslint": "8.53.1"
       },
       "engines": {
         "node": ">=18.0"
@@ -155,6 +155,7 @@
       "version": "5.20.3",
       "resolved": "https://registry.npmjs.org/@algolia/client-search/-/client-search-5.20.3.tgz",
       "integrity": "sha512-9+Fm1ahV8/2goSIPIqZnVitV5yHW5E5xTdKy33xnqGd45A9yVv5tTkudWzEXsbfBB47j9Xb3uYPZjAvV5RHbKA==",
+      "peer": true,
       "dependencies": {
         "@algolia/client-common": "5.20.3",
         "@algolia/requester-browser-xhr": "5.20.3",
@@ -282,6 +283,7 @@
       "version": "7.26.9",
       "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.26.9.tgz",
       "integrity": "sha512-lWBYIrF7qK5+GjY5Uy+/hEgp8OJWOD/rpy74GplYRhEauvbHDeFB8t5hPOZxCZ0Oxf4Cc36tK51/l3ymJysrKw==",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.26.2",
@@ -1969,6 +1971,7 @@
           "url": "https://opencollective.com/csstools"
         }
       ],
+      "peer": true,
       "engines": {
         "node": ">=18"
       },
@@ -1990,6 +1993,7 @@
           "url": "https://opencollective.com/csstools"
         }
       ],
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -2066,6 +2070,7 @@
       "version": "7.1.0",
       "resolved": "https://registry.npmjs.org/postcss-selector-parser/-/postcss-selector-parser-7.1.0.tgz",
       "integrity": "sha512-8sLjZwK0R+JlxlYcTuVnyT2v+htpdrjDOKuMcOVdYjt52Lh8hWRYpxBPoKx/Zg+bcjc3wx6fmQevMmUztS/ccA==",
+      "peer": true,
       "dependencies": {
         "cssesc": "^3.0.0",
         "util-deprecate": "^1.0.2"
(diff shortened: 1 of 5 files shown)


src/python/py_generation_config.cpp
```diff
@@ -431,6 +431,7 @@ void init_generation_config(py::module_& m) {
         .def_readwrite("top_p", &GenerationConfig::top_p)
         .def_readwrite("top_k", &GenerationConfig::top_k)
         .def_readwrite("do_sample", &GenerationConfig::do_sample)
+        .def_readwrite("pruning_ratio", &GenerationConfig::pruning_ratio)
```
peterchen-intel:
```suggestion
```
Already in line438

Image

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -154,6 +154,12 @@ def parse_args():
         help="Path to the JSON file that contains SchedulerConfig for Continuous Batching Pipeline"
         "of OpenVINO GenAI API. Or a JSON string of SchedulerConfig.",
     )
+    parser.add_argument(
+        "--generate-config",
+        type=str,
+        default=None,
+        help="Path to the JSON file that contains GenerateConfig for generatingof OpenVINO GenAI API.",
```
peterchen-intel:
```suggestion
        help="Path to the JSON file that contains GenerateConfig for generatingof OpenVINO GenAI API, or a JSON string of the configuration.",
```
get_json_config supports json file path or json string

rkazants:
```suggestion
        help="Path to the JSON file that contains GenerationConfig for generating of OpenVINO GenAI API.",
```

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -154,6 +154,12 @@ def parse_args():
         help="Path to the JSON file that contains SchedulerConfig for Continuous Batching Pipeline"
         "of OpenVINO GenAI API. Or a JSON string of SchedulerConfig.",
     )
+    parser.add_argument(
+        "--generate-config",
```
rkazants:
```suggestion
        "--generation-config",
```

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -527,6 +533,15 @@ def genai_gen_visual_text(model, prompt, image, video, processor, tokenizer, max
...
         kwargs['videos'] = [ov.Tensor(np.array(video))]
 
+    if generation_config is not None:
+        import openvino_genai
+        gen_config = openvino_genai.GenerationConfig()
+        if "pruning_ratio" in generation_config:
+            pruning_ratio = generation_config.get("pruning_ratio")
+            if pruning_ratio >= 0 and pruning_ratio <= 100:
+                gen_config.pruning_ratio = int(pruning_ratio)
+        kwargs["generation_config"] = gen_config
```
sbalandi: max_new_tokens will be overridden with value from generation_config, please, add pruning_ratio as separate parameter/cmd option or update generation_config and set `do_sample`, `max_new_tokens`

I would prefer adding `pruning_ratio` as separate cmd option, because `--generation-config` seems confusing for me. It general mechanism, but it will work now only for VLM. In `text_evalutor` we have special logic with `generation_config`. So, I think we can add the `pruning_ratio` setting as a cmd parameter and plane to migrate pipelines to the `generation_config` in a separate PR or left setting the generation_config via cmd, but mention the scope of the `generation_config` work

wgzintel: Updated, pls review again.

peterchen-intel:
> please add tests
Added. @rkazants

sbalandi (approved):

tools/who_what_benchmark/tests/test_cli_cdpruner.py
```diff
@@ -0,0 +1,98 @@
+import subprocess  # nosec B404
```
rkazants, line 4: need Intel Copyright?

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -275,6 +289,20 @@ def parse_args():
...
+        help="(optional): Percentage of visual tokens to prune (valid range: 0-100). If this option is not provided, pruning is disabled.",
+    )
+    parser.add_argument(
+        "--relevance_weight",
+        type=weight_0_1,
+        default=None,
+        help="(optional): Float value from 0 to 1, control the trade-off between diversity and relevance for visual tokens pruning, "
+        "a value of 0 disables relevance weighting, while higher values (up to 1.0) emphasize relevance, "
+        "making pruning more conservative on borderline tokens.",
+    )
```
rkazants, line 307: the same comments are from llm_bench. Please apply.

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -623,7 +654,10 @@ def create_evaluator(base_model, args):
                 processor=processor,
                 crop_question=crop_question,
                 task_type=task,
-                frames_num=args.video_frames_num
+                frames_num=args.video_frames_num,
+                pruning_ratio=args.pruning_ratio,
+                relevance_weight=args.relevance_weight,
+                seqs_per_request=getattr(args, "seqs_per_request", 1),  # Default to 1 if not set.
```
rkazants: why is it needed? How user can set `seqs_per_request`?

wgzintel: not needed, removed already.

tools/who_what_benchmark/whowhatbench/wwb.py
wgzintel: Done

tools/who_what_benchmark/tests/test_cli_cdpruner.py
wgzintel: Done

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -21,6 +21,20 @@
 logger = logging.getLogger(__name__)
 
 
+def ratio_type(value: int):
```
as-suvorov: Isn't value a str? Same for `weight_0_1`
```suggestion
def ratio_type(value: str):
```

wgzintel: @as-suvorov  weight_0_1 value is float, not str

as-suvorov:
```python
import argparse


def ratio_type(value: int):
    print(type(value))
    ivalue = int(value)
    if ivalue < 0 or ivalue > 100:
        raise argparse.ArgumentTypeError(f"pruning_ratio must be between 0 and 100, got {value}")
    return ivalue


parser = argparse.ArgumentParser(
    prog="WWB CLI",
    description="This script generates answers for questions from csv file",
)
parser.add_argument(
    "--pruning_ratio",
    type=ratio_type,
    default=None,
    required=False,
    help="Percentage of visual tokens to prune (valid range: 0-100). If this option is not provided, pruning is disabled.",
)

if __name__ == "__main__":
    args = parser.parse_args()
    print(f"Pruning ratio: {args.pruning_ratio}")
```

```bash
python3 ./custom_type_test.py --pruning_ratio 50
<class 'str'>
Pruning ratio: 50
```

`ratio_type` converts `str` to `int`. Correct typing should be:
`def ratio_type(value: str) -> int:`

Please apply for llm_bench PR as well.

yangwang201911 (approved):

rkazants (approved):

PR #3196, opened 2026-01-15
Sample for Speculative Decode on NPU

AsyaPronina:
## Description
This PR brings updated `speculative_decoding_lm_npu.py` sample, that does the same as `speculative_decoding_lm.py` but for NPU. Some metrics prints were reduced to representative basis.

### DIFF
--- a/samples/python/text_generation/speculative_decoding_lm_npu.py
+++ b/samples/python/text_generation/speculative_decoding_lm_npu.py
@@ -1,32 +1,32 @@
 #!/usr/bin/env python3
-# Copyright (C) 2025 Intel Corporation
+# Copyright (C) 2026 Intel Corporation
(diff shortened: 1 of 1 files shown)


samples/python/text_generation/speculative_decoding_lm_npu.py
Wovchena, line 1: Why is a separate sample needed?

AsyaPronina: Hi! Just for the demonstrating purposes, as current sample doesn't yet accept arguments and config option and doesn't show throughput.

AsyaPronina: I can extend the original sample if you don't mind.

PR #3205, opened 2026-01-19
[wwb] Fix gemma3 fails

sbalandi:
## Description
error: AttributeError: 'Gemma3Model' object has no attribute 'generate'
fix: use class AutoModelForCausalLM 

[CVS-174651](https://jira.devtools.intel.com/browse/CVS-174651)

Fixes #(issue)

## Checklist:


sbalandi:
> Just to double check. The function is `load_visual_text_model`. Then for gemma3 you propose load it as `AutoModelForCausalLM`. Is it still gonna be visual text use case even with `AutoModelForCausalLM` class?

yes

PR #3209, opened 2026-01-21
feat: KV cache dump/restore to disk for GPU acceleration

jlee52tw:
Adds the ability to dump and restore KV cache to/from disk for the continuous batching pipeline. This feature enables significant Time-To-First-Token (TTFT) improvements (up to 12x faster) for long-context scenarios by caching pre-computed KV states.

Key features:

- SchedulerConfig: Added kv_cache_load_dir and kv_cache_dump_dir options

- CacheManager: Disk I/O operations for tensor dump/restore with metadata

- BlockManager: Block tracking and manifest management for cache state

- Pipeline: Environment variable support (GENAI_KV_CACHE_DUMP_DIR, GENAI_KV_CACHE_LOAD_DIR)

- benchmark_genai: Added --kv_cache_precision option (u8/f16/f32)

- Unit tests: 11 comprehensive test cases for KV cache persistence

Compatible with OpenVINO 2026.0.0 (uses ov::element::dynamic instead of deprecated undefined)

## Description

CVS-###

Fixes #(issue)

## Checklist:


samples/cpp/text_generation/benchmark_genai.cpp
```diff
@@ -52,37 +57,117 @@ int main(int argc, char* argv[]) try {
...
+        kv_cache_precision = ov::element::f16;
+    } else if (kv_prec_str == "f32") {
+        kv_cache_precision = ov::element::f32;
+    } else if (!kv_prec_str.empty()) {
+        std::cout << "Warning: Unknown kv_cache_precision '" << kv_prec_str << "', using device default" << std::endl;
+    }
+    
+    // IMPORTANT: KV cache precision is determined at model compile time and cached in model_cache/
+    // If you change --kv_cache_precision, you MUST delete the model_cache directory first!
+    // Otherwise, the cached model with the old precision will be loaded instead.
```
Wovchena: That should be an error

src/cpp/include/openvino/genai/scheduler_config.hpp
```diff
@@ -68,11 +69,28 @@ struct SchedulerConfig {
...
+     * When set, the pipeline will attempt to restore KV cache state from this directory
+     * during initialization, enabling fast context restoration without re-computation.
+     * Takes precedence over the OV_GENAI_LOAD_KV_DIR environment variable.
+     */
+    std::string kv_cache_load_dir;
+
+    /**
+     * Directory path to dump KV cache to disk.
+     * When set, the pipeline will save KV cache state to this directory after prefill,
+     * enabling future runs to restore the cached state for faster startup.
```
Wovchena: How is it defined what snapshot is saved and when? There may be multiple prefill requests

src/cpp/include/openvino/genai/scheduler_config.hpp
```diff
@@ -68,11 +69,28 @@ struct SchedulerConfig {
...
+     * Takes precedence over the OV_GENAI_DUMP_KV_DIR environment variable.
```
Wovchena: OpenVINO doesn't have env vars (excluding plugin specific ones) so neither should GenAI. There's only `OPENVINO_LOG_LEVEL` added. Use it instead of `OV_GENAI_VERBOSE`

samples/cpp/text_generation/benchmark_genai.cpp
```diff
@@ -52,37 +57,117 @@ int main(int argc, char* argv[]) try {
...
+    
+    // Debug: Print first 20 tokens of current prompt
+    std::cout << "Current prompt tokens (first 20): ";
+    for (size_t i = 0; i < std::min(size_t(20), prompt_token_size); i++) {
+        std::cout << input_data.input_ids.data<int64_t>()[i] << " ";
+    }
+    std::cout << std::endl;
 
+    std::cout << "Skip warmup" << std::endl;
+    /*
```
rkazants: please remove debug prints

src/cpp/include/openvino/genai/scheduler_config.hpp
Wovchena: Saving the cache to disk allows it's modification. We better provide direct API for cache manipulations to avoid users implementing workarounds with disk save/restore. As a first step it could be a semi private object that can be saved and restored. This also allows to control when the cache should be saved.

Another topic for a discussion is to define the way the cache is serialized. Alternatives that come to my mind are numpy format and maybe there's already something in OpenVINO itself.

All in all, this feature requires arch review. There should be a presentation covering the reasons for specific design choices, alternatives and solutions overview from other frameworks if they exist

Wovchena (changes requested):

samples/cpp/text_generation/benchmark_genai.cpp
```diff
@@ -53,21 +58,59 @@ int main(int argc, char* argv[]) try {
...
+    std::string pc_str = result["enable_prefix_caching"].as<std::string>();
+    bool enable_prefix_caching = (pc_str == "true" || pc_str == "1" || pc_str == "yes");
+
     ov::genai::GenerationConfig config;
     config.max_new_tokens = result["max_new_tokens"].as<size_t>();
-    config.apply_chat_template = false;
 
     ov::genai::SchedulerConfig scheduler_config;
-    scheduler_config.enable_prefix_caching = false;
+    scheduler_config.enable_prefix_caching = enable_prefix_caching;
```
sgonorov, line 69: Maybe make it customizable? In order to keep the old behaviour.

sgonorov (changes requested):
Still lots of previously addressed bugs which need attention.

PR #3213, opened 2026-01-21
add missing docstring for parsing

pavel-esir:
## Description
Adds missing docstrings for parsing. And also removed `pydantic` from requirements since it should've been added to `deployment-requirements.txt` and in any case it's is already there.

## Checklist:


Wovchena (approved):
.pyi needs to be updated

PR #3223, opened 2026-01-26
Guidance rescale removal

sgonorov:
Removed guidance rescale from WWB

### DIFF
--- a/tests/python_tests/samples/conftest.py
+++ b/tests/python_tests/samples/conftest.py
@@ -441,16 +441,13 @@ def generate_llm_bench_input_generation_jsonl(request):
     file_name, json_entries = request.param
     file_path = os.path.join(test_data, file_name)
 
-    if not os.path.exists(file_path):
-        os.makedirs(os.path.dirname(file_path), exist_ok=True)
+    os.makedirs(os.path.dirname(file_path), exist_ok=True)
 
-        with open(file_path, "w", encoding="utf-8") as f:
-            for entry in json_entries:
-                f.write(json.dumps(entry) + "\n")
+    with open(file_path, "w", encoding="utf-8") as f:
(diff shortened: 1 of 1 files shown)


tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -565,7 +565,6 @@ def genai_gen_text2video(
         num_frames=num_frames,
         frame_rate=frame_rate,
         guidance_scale=guidance_scale,
-        guidance_rescale=guidance_rescale,
```
Wovchena, line 568: Remove it from test cases instead. With this patch the value is provided but ignored. Revert after the support is added

sgonorov: Changed

tools/who_what_benchmark/whowhatbench/prompts/text_to_video_prompts.json
Wovchena, line 1: Check why the issue wasn't detected in PR that introduced it. Maybe trigger conditions should be tweaked

tools/who_what_benchmark/whowhatbench/text2video_evaluator.py
sbalandi:
```suggestion
        data_keys = ["prompt", "negative_prompt", "width", "height", "guidance_scale"]
```

tools/who_what_benchmark/whowhatbench/prompts/text_to_video_prompts.json
sbalandi: the tests for video generation was not run somehow on PR, which introduced that restriction https://github.com/openvinotoolkit/openvino.genai/actions/runs/21246748554/job/61139121502

PR #3229, opened 2026-01-27
Add close stale PR workflow

as-suvorov:
## Description
Workflow added with `debug-only: true` to test the logs first

Workflow adds `stale` tag to inactive for 60 days PR.
Worklow closes PR after 7 days with no activity with `stale` tag.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -707,7 +707,10 @@ jobs:
   genai_nodejs_tests:
     name: Node.js bindings tests
     needs: [ smart_ci, openvino_download, genai_build_nodejs ]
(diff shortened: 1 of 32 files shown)


.github/workflows/stale.yml
```diff
@@ -0,0 +1,27 @@
...
+  workflow_dispatch:
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: read
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v9
```
Wovchena: Use commit hash similar to other workflows

PR #3245, opened 2026-01-29
[JS] Refactor tests

Retribution98:
## Description
- Downloading the test model has changed. It now uses the utils from the Python tests to convert models with optimum-cli.
- GH workflows have been updated to require a new environment.
- The deprecated startChat/finishChat has been removed from the LLMPipeline tests.
- The StructuredOutput and Parsers tests have been modified to work with any LLM model. Instruct models are no longer required.
- Instructions for running the JavaScript tests have been added.

CVS-180128

## Checklist:


src/js/tests/llmPipeline.test.js
```diff
@@ -1,179 +1,230 @@
...
 });
 
-describe("LLMPipeline methods", async () => {
+describe("General LLM tests", () => {
   let pipeline = null;
 
-  await before(async () => {
-    pipeline = await LLMPipeline(MODEL_PATH, "CPU");
-
-    await pipeline.startChat();
```
almilosz, line 45: Please keep a 1 or 2 tests until `startChat` is removed completely from js api.
Does JS API emit any warning that `startChat` is deprecated?

Wovchena: I think you should keep all of the tests. Deprecation means something **is going to** be removed. Not that it's invalid API

Retribution98: > I think you should keep all of the tests. Deprecation means something **is going to** be removed. Not that it's invalid API

Yes, it's still a valid API, but these tests use outdated logic. Some time ago, LLMPipeline could not be used without startChat/finishChat, so these were added to each test case. This is no longer necessary, so they could be removed.
I also added new tests that check startChat/finishChat directly.

Retribution98: @almilosz Please take another look

src/js/tests/llmPipeline.test.js
```diff
@@ -87,80 +86,208 @@ describe("generation parameters validation", () => {
...
+    pipeline = await LLMPipeline(LLM_PATH, "CPU");
   });
 
-  after(async () => {
-    await pipeline.finishChat();
-  });
+  describe("LLMPipeline methods", () => {
```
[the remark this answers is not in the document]
yatarkan: `describe("LLMPipeline methods",` already exists above, let's reorganize/rename

src/js/tests/module.test.js
```diff
@@ -87,80 +86,208 @@ describe("generation parameters validation", () => {
...
+    }
+    assert.ok(chunks.length > 0);
+    // We need to keep previous messages between tests to avoid error for SDPA backend (macOS in CI)
+    chatHistory.push({ role: "assistant", content: chunks.join("") });
+  });
+
+  it("startChat() and finishChat() complete without error", async () => {
     await pipeline.startChat();
+    await pipeline.finishChat();
   });
```
yatarkan: Next test (`it("startChat(systemMessage) then generate() then finishChat()",`) seems to be testing this case. Do we really need this test then?

src/js/tests/bindings.test.js
yatarkan, line 1: Does it make sense to rename the file (e.g. `llmPipeline.test.js`) to align with other use cases tests?

src/js/tests/bindings.test.js
```diff
@@ -1,41 +1,25 @@
...
+import { describe, it, before } from "node:test";
 
-const MODEL_PATH = process.env.MODEL_PATH || `./tests/models/${models.LLM.split("/")[1]}`;
+const { LLM_PATH } = process.env;
+
+if (!LLM_PATH) {
+  throw new Error("Please set LLM_PATH environment variable to run the tests.");
+}
 
 describe("bindings", () => {
```
yatarkan: The same comment for renaming

```suggestion
describe("LLMPipeline", () => {
```

src/js/tests/vlmPipeline.test.js
```diff
@@ -143,11 +150,6 @@ describe("VLMPipeline", { skip: process.platform === "darwin" }, () => {
 
   it("should start and finish chat", async () => {
```
yatarkan: Similar comment as for LLMPipeline - do we need a test that simply calls `startChat()` and `finishChat()` without generate?
For instance, we have a python test checking that output is similar for both startChat() API and ChatHistory API.

src/js/tests/setup.py
yatarkan, line 1: For which cases do we still need `setup.js`?

src/js/tests/module.test.js
Retribution98: Not really. I have removed it.

src/js/tests/bindings.test.js
Retribution98: Sure, thanks!

src/js/tests/vlmPipeline.test.js
Retribution98: It is the only test for startChat/finishChat, so yes, we needed it.
I updated this test case with generate to make it more complex.

src/js/tests/setup.py
Retribution98: It is removed

yatarkan: I can't see that `setup.js` file is removed in PR changes.

src/js/tests/setup.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python3
+# Copyright (C) 2023-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Setup script to download and convert models for JS tests.
+This is a Python equivalent of setup.js
```
yatarkan: If `setup.js` is removed let's avoid referencing it.
```suggestion
```

src/js/tests/bindings.test.js
yatarkan: I see `module.test.js` is moved to‎ `llmPipeline.test.js` 👍 
But how about the `bindings.test.js`? Is it needed? I thought it is already covered in `llmPipeline.test.js`

PR #3254, opened 2026-01-30
Update model and tresholds for cache eviction tests

olpipi:
## Description
Speed up kv cache eviction tests by switching to smaller model

CVS-173772

## Checklist:

### DIFF
--- a/tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
+++ b/tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
@@ -183,11 +183,14 @@ class LongBenchTestData:
     max_cache_usage_optimization_ratio: float
     avg_cache_usage_optimization_ratio: float
 
-
-@pytest.mark.parametrize("test_struct", [
-    LongBenchTestData("samsum", 4, 1.5, 2.5),
-    LongBenchTestData("trec", 3.2, 2.0, 2.8),
-], ids=["samsum", "trec"])
+@pytest.mark.parametrize(
+    "test_struct",
+    [
+        LongBenchTestData("samsum", 1.0, 1.5, 2.5),
+        LongBenchTestData("trec", 1.0, 2.0, 2.8),
+    ],
+    ids=["samsum", "trec"],
+)
 def test_optimized_generation_longbench(test_struct):
     seqs_per_request = 16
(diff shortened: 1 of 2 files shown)


tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
```diff
@@ -185,14 +185,14 @@ class LongBenchTestData:
...
-    LongBenchTestData("trec", 3.2, 2.0, 3.3),
+    LongBenchTestData("samsum", 4, 1.3, 2.5),
+    LongBenchTestData("trec", 3.2, 2.0, 3.2),
 ], ids=["samsum", "trec"])
 def test_optimized_generation_longbench(test_struct):
     seqs_per_request = 32
     device = "CPU"
     num_kv_blocks = 1000 if device == "CPU" else 500
-    model_id = "Qwen/Qwen2-0.5B-Instruct"
+    model_id = "hf-internal-testing/tiny-random-LlamaForCausalLM"
```
l-bat: Both tests not only check cache usage, but also evaluate model outputs in terms of accuracy. So, models with random weights cannot be used. I suggest moving these cache eviction tests to weekly validation.

olpipi: @l-bat we don't have some kind of weekly validation to move these tests.
Could you please suggest another model we can use in the test that is much more lighter then original Qwen/Qwen2-0.5B-Instruct?

l-bat: I suppose you can try https://huggingface.co/HuggingFaceTB/SmolLM2-135M-Instruct

olpipi: @l-bat Could you please take alook. I updated model to HuggingFaceTB/SmolLM2-135M-Instruct and changed seq num to 16

tests/python_tests/test_kv_cache_eviction/test_kv_cache_eviction_1.py
l-bat, line 262: Could you please check if we can lower the previous threshold for the difference between scores? 
Also, if any other test reference parameters should be updated in all tests?

PR #2707, opened 2025-09-05
Enable VLM lookup.

xipingyan:
1: Move `m_pipeline->generate_candidates(); `from` ContinuousBatchingPipeline::PromptLookupImpl::step() ` to `ContinuousBatchingPipeline::ContinuousBatchingImpl::step()`
2: Add new interface pass prompts_ids for embedding inputs model.

PR #3267, opened 2026-02-02
Print generation config with OPENVINO_LOG_LEVEL>=2

dkalinowski:
## Description
It is really useful to have Plugin Config and Scheduler Config printed on certain openvino log level. This change adds possibility to print Generation Config as well. We found it hard to debug differences between OpenVINO Model Server & OpenVINO GenAI and it should help us significantly.


src/cpp/src/sequence_group.hpp
```diff
@@ -348,7 +348,9 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
           m_sampling_params(sampling_params),
           m_block_size(block_size),
           m_sequence_group_type(SequenceGroupType::TOKENS),
-          m_generation_stream(GenerationStream::create()) { }
+          m_generation_stream(GenerationStream::create()) {
+        print_generation_config_info(m_sampling_params);  // the latest possible place to print generation config info in case it is modified at some point
```
atobiszei: Wouldn't it be clearer to have if (log_level) check here? Or name it print_on_debug_generation_config_info

mzegla (commented):
In some places you print "SIZE_MAX" as a string, but isn't it platform dependent? 
Wouldn't it be clearer to print it explicitly?

src/cpp/src/sequence_group.hpp
as-suvorov: Please use https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/logger.hpp for debug prints

dkalinowski: @as-suvorov do you mind if I also fix all existing printers to use logger?
- print_compiled_model_properties
- print_gguf_debug_info
- print_scheduler_config_info
?

as-suvorov: I'm thinking about copilot's comment: https://github.com/openvinotoolkit/openvino.genai/pull/3267#discussion_r2754563102
Could you please double check if we can call `to_string` only for desired log level (debug I guess) with `logger.hpp`?
If we can address this I see no isse with applying it for the rest of the prints.

dkalinowski: If we call GENAI_DEBUG directly, to_string will be called anyway. I will switch from std::cout to GENAI_DEBUG, but keep `if (env_setup_for_print_debug_info())` to only call logger when env variable `OPENVINO_LOG_LEVEL>=2` is set (same as scheduler config/plugin config now).

The log msg will look something like this:
```
[DEBUG] 2026-02-02T16:35:49z utils.cpp:594      GenerationConfig { 
  max_new_tokens: SIZE_MAX
  ...

```

src/cpp/src/generation_config.cpp
```diff
@@ -3,13 +3,16 @@
 
 #include <fstream>
 #include <limits>
+#include <sstream>
+#include <iostream>
```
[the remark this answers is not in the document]
dkalinowski: fixed

src/cpp/src/sequence_group.hpp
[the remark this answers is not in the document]
dkalinowski: there is a check

dkalinowski:
> In some places you print "SIZE_MAX" as a string, but isn't it platform dependent? Wouldn't it be clearer to print it explicitly?

thats right, changed it. it's ready for re-review

mzegla (approved):

src/cpp/src/sequence_group.hpp
Wovchena: https://github.com/openvinotoolkit/openvino.genai/pull/2707/changes/341e75eb41fce01083ab9b4b436187b5923af73c has a way of avoiding calling a function which is passed to `GENAI_DEBUG`. The PR that this commit belongs to reverted this feature because it didn't relate to the PR feature.

as-suvorov (approved):

as-suvorov:
@dkalinowski mac build failed, please take a look: https://github.com/openvinotoolkit/openvino.genai/actions/runs/21599532276/job/62242044002?pr=3267#step:10:411

src/cpp/src/sequence_group.hpp
dkalinowski: image
@Wovchena  I don't think it is good idea to use `GENAI_CHECK_LOG_LEVEL`, there's bunch of logic we want to suppress if log level is not high enough. I agree it would be fine for one liners, sadly its not one line.

src/cpp/src/sequence_group.cpp
```diff
@@ -98,5 +99,14 @@ size_t Sequence::get_hash(size_t content_length) {
...
     return _make_hash(content_len);
 }
+
+SequenceGroup::SequenceGroup(uint64_t request_id, const ov::genai::GenerationConfig& sampling_params, std::size_t block_size)
+    : m_request_id(request_id),
+      m_sampling_params(sampling_params),
+      m_block_size(block_size),
+      m_sequence_group_type(SequenceGroupType::TOKENS),
+      m_generation_stream(GenerationStream::create()) {
+    ov::genai::utils::print_generation_config_info(m_sampling_params);  // the latest possible place to print generation config info in case it is modified at some point
```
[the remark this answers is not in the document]
dkalinowski: not true

src/cpp/src/sequence_group.hpp
Wovchena: You can move the body to a separate function. In the worst case, if you need full context, it can be lambda

src/cpp/src/generation_config.cpp
```diff
@@ -362,6 +364,90 @@ void GenerationConfig::validate() const {
...
+    oss << "  pruning_ratio: " << pruning_ratio << "\n";
+    oss << "  relevance_weight: " << relevance_weight << "\n";
+
+    // Assistant generation parameters
+    oss << "  assistant_confidence_threshold: " << assistant_confidence_threshold << "\n";
+    oss << "  num_assistant_tokens: " << num_assistant_tokens << "\n";
+    oss << "  max_ngram_size: " << max_ngram_size << "\n";
+
+    // Structured output
+    oss << "  structured_output_config: " << (structured_output_config.has_value() ? "set" : "not set") << "\n";
```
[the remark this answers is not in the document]
dkalinowski: not needed

src/cpp/src/generation_config.cpp
```diff
@@ -362,6 +364,90 @@ void GenerationConfig::validate() const {
...
+
+    // Adapters
+    oss << "  adapters: " << (adapters.has_value() ? "set" : "not set") << "\n";
```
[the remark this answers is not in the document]
dkalinowski: not needed

src/cpp/src/sequence_group.hpp
dkalinowski: You are right - separated it to another function. Please re-review.

src/cpp/src/generation_config.cpp
```diff
@@ -362,6 +364,90 @@ void GenerationConfig::validate() const {
...
 
+std::string GenerationConfig::to_string() const {
+    auto stop_criteria_to_string = [](StopCriteria criteria) -> std::string {
+        switch (criteria) {
+            case StopCriteria::EARLY: return "EARLY";
+            case StopCriteria::HEURISTIC: return "HEURISTIC";
+            case StopCriteria::NEVER: return "NEVER";
+            default: return "UNKNOWN";
+        }
+    };
```
Wovchena: Copilot's note is reasonable but I'd enhance it. You don't exploit lambda features. If you don't want in place `switch`, move to anonymous namespace function

src/cpp/src/utils.cpp
```diff
@@ -527,62 +528,61 @@ bool env_setup_for_print_debug_info() {
...
 void print_scheduler_config_info(const SchedulerConfig &scheduler_config) {
-    if (!env_setup_for_print_debug_info()) {
-        return;
-    }
+    GENAI_CHECK_LOG_LEVEL(ov::log::Level::DEBUG, scheduler_config.to_string());
+}
 
-    std::cout << scheduler_config.to_string() << std::endl;
+void print_generation_config_info(const GenerationConfig &generation_config) {
+    GENAI_CHECK_LOG_LEVEL(ov::log::Level::DEBUG, generation_config.to_string());
```
Wovchena: `GENAI_CHECK_LOG_LEVEL` is just a helper. Make `GENAI_*()` use `GENAI_CHECK_LOG_LEVEL()` and use `GENAI_DEBUG()` as nothing changed

src/cpp/src/generation_config.cpp
```diff
@@ -362,6 +364,90 @@ void GenerationConfig::validate() const {
...
+    oss << "  min_new_tokens: " << min_new_tokens << "\n";
+    oss << "  echo: " << std::boolalpha << echo << "\n";
+    oss << "  logprobs: " << logprobs << "\n";
+
+    // EOS and stop conditions
+    oss << "  eos_token_id: " << eos_token_id << "\n";
+    oss << "  stop_strings: [";
+    bool first = true;
+    for (const auto& s : stop_strings) {
+        if (!first) oss << ", ";
```
Wovchena: Move to separate function. Call it `join()` (stolen from Python). Add a comment in that function to drop in favor of `std::views::join_with()` with C++23

src/cpp/src/generation_config.cpp
Wovchena, line 475: Print `StructuredOutputConfig::format_for_json(structured_output_config)`

Wovchena (changes requested):

src/cpp/src/sequence_group.hpp
Wovchena: @xipingyan, FYI. Hope it's OK your idea is reimplemented here

dkalinowski: I think this is cleaner approach, we dont need to check debug level each time, just once inside this function

src/cpp/src/generation_config.cpp
dkalinowski: `StructuredOutputConfig::format_for_json` takes string as argument, what do you exactly mean? @Wovchena

src/cpp/src/generation_config.cpp
dkalinowski: That's right, I dont need lambda here. Changed to in-place switch since we dont need it anywhere else for now. It should probably land in some internal utils if we ever need it.

src/cpp/src/utils.cpp
```diff
@@ -527,62 +528,61 @@ bool env_setup_for_print_debug_info() {
     return (env_var_value != nullptr && atoi(env_var_value) > static_cast<int>(ov::log::Level::WARNING));
 }
 
-void print_compiled_model_properties(ov::CompiledModel& compiled_Model, const char* model_title) {
-    if (!env_setup_for_print_debug_info()) {
-        return;
-    }
+static std::string _print_compiled_model_properties(ov::CompiledModel& compiled_model, const char* model_title) {
```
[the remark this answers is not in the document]
dkalinowski: done

src/cpp/src/generation_config.cpp
```diff
@@ -362,6 +364,90 @@ void GenerationConfig::validate() const {
...
+std::string GenerationConfig::to_string() const {
+    auto stop_criteria_to_string = [](StopCriteria criteria) -> std::string {
+        switch (criteria) {
+            case StopCriteria::EARLY: return "EARLY";
+            case StopCriteria::HEURISTIC: return "HEURISTIC";
+            case StopCriteria::NEVER: return "NEVER";
+            default: return "UNKNOWN";
+        }
+    };
+
```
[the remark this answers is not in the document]
dkalinowski: done

src/cpp/src/utils.cpp
dkalinowski: Good idea, changed

src/cpp/src/generation_config.cpp
dkalinowski: Changed, is this what you meant? Whats the expected timefrime to switch from cpp17 to cpp23? @Wovchena

Wovchena (approved):

src/cpp/src/generation_config.cpp
```diff
@@ -362,6 +364,89 @@ void GenerationConfig::validate() const {
...
+    // Structured output
+    oss << "  structured_output_config: " << (structured_output_config.has_value() ? "set" : "not set") << "\n";
+
+    // Adapters
+    oss << "  adapters: " << (adapters.has_value() ? "set" : "not set") << "\n";
+
+    // Chat template
+    oss << "  apply_chat_template: " << std::boolalpha << apply_chat_template << "\n";
+
+    oss << " }";
```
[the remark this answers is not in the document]
dkalinowski: Same convention is used for cache eviction, scheduler config, sparse attention

PR #3270, opened 2026-02-02
Fix Text2VideoPipeline crash with guidance_scale

sgonorov:
Fixes [CVS-179754](https://jira.devtools.intel.com/projects/CVS/issues/CVS-179754)

- Fix RuntimeError when calling `generate(guidance_scale=1.0)` after `reshape(guidance_scale>1.0)` was used
- Add batch size tracking in LTXVideoTransformer3DModel to detect compiled model expectations
- Automatically adjust batch size and enable CFG when compiled model requires it
- Handle dynamic shapes safely in model compilation

### DIFF
--- a/tests/python_tests/test_video_generation.py
+++ b/tests/python_tests/test_video_generation.py
@@ -232,8 +232,8 @@ class TestText2VideoPipelineAdvanced:
         result = pipe.generate("test prompt", num_inference_steps=2)
         assert result.video is not None
 
-    def test_generate_guidance_scale_one_after_reshape(self, video_generation_model):
-        """Test: reshape with CFG (guidance_scale>1) then generate with guidance_scale=1 should raise error."""
+    def test_generate_without_cfg_after_reshape_with_cfg(self, video_generation_model):
+        """Test: reshape with CFG then generate without CFG should raise error."""
         pipe = ov_genai.Text2VideoPipeline(video_generation_model)
         pipe.reshape(1, 9, 32, 32, 3.0)
         pipe.compile("CPU")
@@ -248,8 +248,8 @@ class TestText2VideoPipelineAdvanced:
                 num_inference_steps=2,
             )
 
-    def test_generate_guidance_scale_high_after_reshape_low(self, video_generation_model):
-        """Regression test: reshape without CFG (guidance_scale=1) then generate with CFG (guidance_scale>1)."""
+    def test_generate_with_cfg_after_reshape_without_cfg(self, video_generation_model):
+        """Regression test: reshape without CFG then generate with CFG triggers rebuild."""
         pipe = ov_genai.Text2VideoPipeline(video_generation_model)
         pipe.reshape(1, 9, 32, 32, 1.0)
         pipe.compile("CPU")


tools/who_what_benchmark/whowhatbench/text2video_evaluator.py
```diff
@@ -21,7 +21,7 @@ class Text2VideoEvaluator(BaseEvaluator):
     DEF_FRAME_RATE = 25
     DEF_WIDTH = 704
     DEF_HEIGHT = 480
-    DEF_GUIDANCE_SCALE = 3  # CVS-179754: guidance_scale == 1 triggers GenAI runtime error
```
sbalandi, line 24: if gs == 1 is supported now, let's update default cases for wwb. cases, where guidance_scale is 5, let's make guidance_scale 1

sgonorov: Nice idea, fixed

sbalandi: no, here should be 3, I mean here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/prompts/text_to_video_prompts.json

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -404,20 +427,79 @@ class Text2VideoPipeline::LTXPipeline {
...
+            if (expected_batch_size > 0) {
+                OPENVINO_ASSERT(expected_batch_size % merged_generation_config.num_videos_per_prompt == 0,
+                                "Compiled batch size must be divisible by num_videos_per_prompt");
+                requested_batch_size_multiplier =
+                    expected_batch_size / merged_generation_config.num_videos_per_prompt;
+            } else if (m_compiled_batch_size_multiplier > 0) {
+                requested_batch_size_multiplier = m_compiled_batch_size_multiplier;
+            }
+            if (requested_batch_size_multiplier > 1 && merged_generation_config.guidance_scale <= 1.0f) {
+                merged_generation_config.guidance_scale = MIN_CFG_GUIDANCE_SCALE;
```
likholat: Let's add a warning to prevent unexpected behavior, sth like this
```suggestion
        GENAI_WARN("guidance_scale <= 1 requested, but the expected batch size requires CFG."
                   "Forcing guidance_scale to {} to match the compiled model shape. "
                   "To avoid this behavior, reshape/compile the model with guidance_scale <= 1.",
                    MIN_CFG_GUIDANCE_SCALE);
        merged_generation_config.guidance_scale = MIN_CFG_GUIDANCE_SCALE;
```

sgonorov: Added

tools/who_what_benchmark/whowhatbench/text2video_evaluator.py
sgonorov: Ah, updated it 👍

src/cpp/src/video_generation/ltx_pipeline.hpp
Wovchena: Make it an error which is impossible to miss. The fix of setting guidance_scale in user code is easy

sbalandi (approved):
LGTM, wwb side

src/cpp/src/video_generation/ltx_pipeline.hpp
sgonorov: Fixed

PR #3293, opened 2026-02-06
[NPU][EAGLE3] Fix issue when Draft model generates EOS token

GuoliangShiIntel:
## Description
**Background**

Eagle3 uses speculative decoding where a draft model generates candidate tokens that are validated by a target model. During this process, one critical issue existed:

Sequence State Management: When the draft model generates an `EOS` token, the `Sampler` immediately marks the sequence as `FINISHED`. This causes `get_running_sequences()` to return empty, leading to assertion failures in subsequent iterations if the target model rejects the `EOS`.

**Solution**
1. Restore Running State After Draft `EOS`
When draft model generates `EOS`, explicitly restore sequence to `RUNNING` state after sampling.
2. Early Termination on Draft `EOS`
Break draft generation loop immediately upon detecting EOS:

[CVS-180738](https://jira.devtools.intel.com/browse/CVS-180738)

## Checklist:

### DIFF
--- a/src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
+++ b/src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
@@ -731,6 +731,12 @@ StatefulEagle3LLMPipeline::SpeculativeResult StatefulEagle3LLMPipeline::run_spec
     if (tokens_after_accept > max_new_tokens) {
         // Truncate to exactly max_new_tokens
         size_t excess_tokens = tokens_after_accept - max_new_tokens;
+        OPENVINO_ASSERT(excess_tokens < validated_tokens.size(),
+                        "excess_tokens (",
+                        excess_tokens,
+                        ") must be less than validated_tokens.size() (",
+                        validated_tokens.size(),
+                        ")");
         size_t tokens_to_keep = validated_tokens.size() - excess_tokens;
 
         validated_tokens.resize(tokens_to_keep);
--- a/tests/python_tests/test_stateful_speculative_decoding.py
+++ b/tests/python_tests/test_stateful_speculative_decoding.py
@@ -205,6 +205,16 @@ eagle3_models_and_input = [
     (
         "Qwen/Qwen3-1.7B",
         "AngelSlim/Qwen3-1.7B_eagle3",
+        """Code:
+def add(a, b):
+    return a + b
+
+Question: Can you please add 2 and 3
+A:""",
+    ),
+    (
+        "Qwen/Qwen3-1.7B",
+        "AngelSlim/Qwen3-1.7B_eagle3",
         "What is the capital of Ireland?/no_think",
     )
 ]


GuoliangShiIntel:
> Changes look not only NPU specific. Please add test(s) to reproduce this issue

Hi @rkazants

In fact, the eagle3 stateful pipeline is only for NPU devices. 

The issues addressed in this PR occur when EOS happens during the draft stage, so they're related to the input prompt. [Here](https://github.com/openvinotoolkit/openvino.genai/pull/3293/changes#diff-0a90ca86ed82df73ee87fcfad2eb75a2b4e77f5ca324a14e925b364fb6ce31a0R208) I've simply changed the existing test to use a new input prompt that reproduces the issue.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -703,8 +723,28 @@ StatefulEagle3LLMPipeline::SpeculativeResult StatefulEagle3LLMPipeline::run_spec
...
-    const size_t total_accepted_tokens = validated_tokens.size();
+    size_t tokens_to_remove = actual_draft_tokens - accepted_count;
+    size_t total_accepted_tokens = validated_tokens.size();
+
+    // Check if accepting all validated tokens would exceed max_new_tokens
+    size_t tokens_after_accept = current_generated_tokens + validated_tokens.size();
+    if (tokens_after_accept > max_new_tokens) {
+        // Truncate to exactly max_new_tokens
+        size_t excess_tokens = tokens_after_accept - max_new_tokens;
+        size_t tokens_to_keep = validated_tokens.size() - excess_tokens;
```
AsyaPronina, line 740: Can it be negative number? May we end up in such situation?

tests/python_tests/test_stateful_speculative_decoding.py
```diff
@@ -205,12 +205,7 @@ def test_extended_perf_metrics():
     (
         "Qwen/Qwen3-1.7B",
         "AngelSlim/Qwen3-1.7B_eagle3",
-        """Code:
-def add(a, b):
-    return a + b
-
-Question: Can you please add 2 and 3
-A:""",
+        "What is the capital of Ireland?/no_think",
```
rkazants, line 218: let us have both input samples in tests: existing one and yours.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: Mathematically, `tokens_to_keep` which should always be positive. However, i have added assert for it.

tests/python_tests/test_stateful_speculative_decoding.py
GuoliangShiIntel: Done, thanks

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
AsyaPronina: I mean, can the situation led to `tokens_to_keep` be a negative number and there will be very big number because of `size_t`?

GuoliangShiIntel:
@rkazants Do you have any other comments about this PR?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -578,15 +573,18 @@ EncodedResults StatefulEagle3LLMPipeline::generate_tokens(const EncodedInputs& i
     while (!eos_reached && generated_tokens < config.max_new_tokens &&
            m_target->get_sequence_length() < m_prompt_length + config.max_new_tokens &&
            streaming_status == ov::genai::StreamingStatus::RUNNING) {
-        auto result = run_speculative_iteration(input_token_count, static_cast<int64_t>(config.eos_token_id));
+        auto result = run_speculative_iteration(input_token_count,
+                                                static_cast<int64_t>(config.eos_token_id),
```
songbell, line 577: any chance draft will have multiple EOS marker?

tests/python_tests/test_stateful_speculative_decoding.py
```diff
@@ -211,6 +211,11 @@ def add(a, b):
 
 Question: Can you please add 2 and 3
 A:""",
+    ),
+    (
+        "Qwen/Qwen3-1.7B",
+        "AngelSlim/Qwen3-1.7B_eagle3",
+        "What is the capital of Ireland?/no_think",
```
songbell, line 218: btw, what will this prompt behave if not handling eos in draft?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: When the draft model produces an EOS token, `run_speculative_iteration` returns and splits into two cases:

1. If the draft EOS is accepted by the target model, the pipeline terminates.
2. If the draft EOS is rejected, a non‑EOS token from the target is fed back into the draft model for the next iteration.

Based on this flow, the current logic appears safe.
Not sure if this fully addresses your question, though.

songbell: I mean the eos here is from generation config right? will the stop_token_ids be handled well also?

Image

tests/python_tests/test_stateful_speculative_decoding.py
GuoliangShiIntel: A segmentation fault occurs because the sampler marks the sequence state as finished.
Meanwhile, the target model still attempts to read input IDs from this sequence for validation, which causes the crash.

The current fix works but may not be the ideal long‑term solution.
A cleaner approach would be to implement the top‑k strategy properly.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: Good catch. The NPU pipeline should not set `stop_token_ids`; if it does, the current pipeline will not function correctly.

A proper top‑k implementation will introduce a larger structural change, and this issue can be fully consider in a follow‑up PR. Thanks for pointing it out.

tests/python_tests/test_stateful_speculative_decoding.py
GuoliangShiIntel: Currently, Eagle3 top‑1 stateful pipeline is only for NPU device. There is accuracy issue on the CPU device. We need to identify the root cause and enable CPU support in the following task.

Jira: CVS-182703

rkazants (approved):

GuoliangShiIntel:
@rkazants @Wovchena I think this PR is ready for merge. Could you please help. Thanks

PR #3294, opened 2026-02-06
[Whisper]: Fix word timestasmps processing perf metrics

as-suvorov:
## Description

Accumulate word timestamps processing durations instead of pushing to raw metrics.

## Checklist:

### DIFF
--- a/tests/python_tests/test_whisper_pipeline.py
+++ b/tests/python_tests/test_whisper_pipeline.py
@@ -760,6 +760,9 @@ def test_perf_metrics(model_descr, sample_from_dataset):
     assert np.allclose(mean_dur, np.mean(raw_dur))
     assert np.allclose(std_dur, np.std(raw_dur))
 
+    # processing duration is a single value per generate
+    assert len(whisper_raw_metrics.word_level_timestamps_processing_durations) == 1
+
     word_ts_raw_dur = np.array(whisper_raw_metrics.word_level_timestamps_processing_durations) / 1000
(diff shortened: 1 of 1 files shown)


src/cpp/src/whisper/pipeline_static.cpp
```diff
@@ -1246,8 +1248,7 @@ WhisperDecodedResults WhisperPipeline::StaticWhisperPipeline::generate(
             const auto word_timestamps_processing_duration = ov::genai::PerfMetrics::get_microsec(
                 std::chrono::steady_clock::now() - word_timestamps_processing_start);
 
-            perf_metrics.whisper_raw_metrics.word_level_timestamps_processing_durations.emplace_back(
-                word_timestamps_processing_duration);
+            perf_metrics.whisper_raw_metrics.word_level_timestamps_processing_durations[0] += MicroSeconds(word_timestamps_processing_duration);
```
Wovchena, line 1251: `detokenization_durations` has multiple values. Why not keep multiple values for `word_level_timestamps_processing_durations`?

as-suvorov: I think this `word_level_timestamps_processing_durations` needs to be aligned with generation, feature extraction and inference duration. When I call `pref_metrics.get_word_level_timestamps_processing_duration()` I want to see total time, so I can see how big it is comparing to generation or inference times. Current approach will give mean,std for several post processing calls.
Initially I just missed that this `word_level_timestamps_processing_durations` can be populated several times due to long-form audio chunking.

PR #3164, opened 2026-01-05
kczarnec-add-metrics-to-memory-monitor

krzyczar:
## Description

CVS-180532

## Checklist:
- [ ] Tests have been updated or added to cover the new code. 
- [x] This patch fully addresses the ticket. 
- [ ] I have made corresponding changes to the documentation.

PR #3295, opened 2026-02-06
Lint: whitespaces and newlines rules applied

as-suvorov:
## Description
PR with linting changes only. The whole project was lint with:
`pre-commit run --all-files`
Additionaly tools prompts were excluded:
```yaml
exclude: |
  (?x)(
    \.pyi$|
    ^tools/llm_bench/prompts/|
    ^tools/who_what_benchmark/whowhatbench/prompts/
  )
```

Current .pre-commit-config.yml:
```yml
exclude: |
  (?x)(
    \.pyi$|
    ^tools/llm_bench/prompts/|
    ^tools/who_what_benchmark/whowhatbench/prompts/
  )
repos:
  - repo: meta
    hooks:
      - id: identity
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v6.0.0
    hooks:
      - id: trailing-whitespace
        exclude: '\.(py|c|cpp|h|hpp)$'
      - id: end-of-file-fixer
      - id: check-merge-conflict
      - id: check-case-conflict
      - id: check-symlinks
      - id: detect-private-key
      - id: mixed-line-ending
        args: ["--fix=lf"]
      - id: check-ast
      - id: check-yaml
      - id: check-toml
      - id: check-added-large-files
        args: ["--maxkb=1000"]
  - repo: https://github.com/akaihola/darker
    rev: v3.0.0
    hooks:
      - id: darker
        args: ["--formatter=ruff"]
        additional_dependencies:
          - ruff==0.14.4
          - tomli==2.3.0
```

`pre-commit run --all-files` triggered darker hook. But as it's incremental, no actual ruff formatting was applied.
PRs will still have to apply darker+ruff formatting on a new or changed lines. @sbalandi fyi

## Checklist:
- [NA] Tests have been updated or added to cover the new code. 


sgonorov:
BTW, we can replace darker with ruff format, as this affects the whole project, darker won't be needed anymore.

as-suvorov:
> BTW, we can replace darker with ruff format, as this affects the whole project, darker won't be needed anymore.

This would require another PR with `ruff format .`, which would format all python files. I don't have strong opinion here. @sgonorov @Wovchena what's your thoughts?

Wovchena:
If the problem of requiring contributors to modify untouched lines persists, separate all python files formatting is needed

as-suvorov:
> If the problem of requiring contributors to modify untouched lines persists, separate all python files formatting is needed

I'm not aware of such cases. @sbalandi do you have examples?

sbalandi:
> > If the problem of requiring contributors to modify untouched lines persists, separate all python files formatting is needed
> 
> I'm not aware of such cases. @sbalandi do you have examples?

During my PR - no. The issue, discussed today, was with PR https://github.com/openvinotoolkit/openvino.genai/pull/3164/changes , for example, with the file - tools/llm_bench/llm_bench_utils/gen_output_data.py , author add two strings, but it looks like lint required changes for all. But I'm not sure when this changes were made. Because for the latest changes I see that lint ignore `'` in not changed lines and looks like there is no problem.

But I'm a little confused. Currently, formatting only works on modified lines ? And proposal to re-enable formatting for the entire file ?
This pull request changes almost nothing in the llm_bench/wwb. If full-file formatting is enabled, formatting will trigger for tools in many places (the simplest example - https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/benchmark.py#L106).

as-suvorov:
@sbalandi Currently python formatting is done by darker and ruff. Darker detects changed lines and lint those changed lines with ruff. Such incremental approach keeps git history (especially git blame) clean.
The question was if this incremental approach works well. And lint stage requested formatting only affects changed lines. If not, proposal is to lint all py files with ruff.

PR #3314, opened 2026-02-09
[wwb][llm-bench] Fix version of required dependency in requirements.txt

sbalandi:
## Description
- fix version of modules for wwb and llm_bench
- removed separate requirements.txt for llm-bench (content from requirements_conversion.txt was added to main requirements, requirements_gptq.txt‎ was removed)
- add extras_require for minicpm-o (to discuss)

CVS-181062

## Checklist:
- [NA] Tests have been updated or added to cover the new code. 

### DIFF
--- a/tools/llm_bench/requirements.txt
+++ b/tools/llm_bench/requirements.txt
@@ -34,6 +34,7 @@ einops==0.8.2
 # For microsoft/Phi-3.5-vision-instruct
 backoff==2.2.1
 # For openbmb/MiniCPM-o-2_6
+# Please, be advised that MiniCPM-o-2_6 requires transformers>=4.50,<=4.51.3
 vector-quantize-pytorch==1.27.20
 vocos==0.1.0
 torchaudio>=2.1.0,<=2.10.0
--- a/tools/who_what_benchmark/whowhatbench/wwb.py
+++ b/tools/who_what_benchmark/whowhatbench/wwb.py
@@ -838,7 +838,6 @@ def print_rag_results(evaluator):
 def main():
     args = parse_args()
     check_args(args)
-    print("check ci")
 
     version_str = f'openvino runtime version: {ov.get_version()}'
     if args.genai:
--- a/tools/who_what_benchmark/README.md
+++ b/tools/who_what_benchmark/README.md
@@ -34,6 +34,11 @@ PIP_PRE=1 \
 PIP_EXTRA_INDEX_URL=https://storage.openvinotoolkit.org/simple/wheels/nightly \
 pip install .
 ```
+Some models require specific module versions, WWB supports installing certain configurations. For more information, please refer to `extras_require` in setup.py.
+For example:
+```
+pip install .[minicpm-o-2_6]
+```
 
 ## Usage
 ### Compare Text-generation Models (LLMs)


tools/who_what_benchmark/setup.py
```diff
@@ -43,4 +43,8 @@ def set_version(base_version: str):
     install_requires=required,
     entry_points={"console_scripts": ["wwb=whowhatbench.wwb:main"]},
     package_data={"whowhatbench": ["prompts/*.yaml", "prompts/*.json"]},
+    extras_require={
+        "minicpmo": ["torch==2.8", "torchvision==0.23", "torchaudio==2.8.0", "soundfile==0.13.1", "librosa==0.11.0",
```
sbalandi: +torchcodec

tools/who_what_benchmark/requirements.txt
```diff
@@ -1,22 +1,17 @@
...
+accelerate>=0.26.0,<=1.12.0
+transformers[sentencepiece]>=4.35.2,<=4.57.0
+sentence-transformers>=2.2.2,<=5.2.2
+openvino-genai<=2026.1.0
+optimum-intel[nncf]>=1.19.0,<=1.27.0
+pandas>=2.0.3,<=2.3.1
+numpy==1.26.4; platform_system == "Darwin" and platform_machine == "x86_64"
+numpy>=1.16.6,<=2.4.2; platform_system != "Darwin" or platform_machine != "x86_64"
+tqdm>=4.66.1,<=4.67.1
+diffusers>=0.22.0,<=0.36.2
```
sbalandi: 0.36.0

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -838,6 +838,7 @@ def print_rag_results(evaluator):
 def main():
     args = parse_args()
     check_args(args)
+    print("check ci")
```
sbalandi: will delete it once the review is completed

tools/who_what_benchmark/setup.py
```diff
@@ -43,4 +43,18 @@ def set_version(base_version: str):
     install_requires=required,
     entry_points={"console_scripts": ["wwb=whowhatbench.wwb:main"]},
     package_data={"whowhatbench": ["prompts/*.yaml", "prompts/*.json"]},
+    extras_require={
+        "minicpmo": [
```
rkazants: shall we have version of minicpmo here? Please check that newest version of minicpmo has the same deps.

sbalandi: @rkazants  openbmb/MiniCPM-o-4_5 requires transformers<=4.51.3, also hyperpyyaml , onnxruntime and minicpmo are needed 
I have added version here.

Are you planning to increase the allowded transformer version for the MiniCPM-o-4_5 ? If so, I'll add the required versions to the general requirement.txt later. If no, I would update this setup configuration

tools/who_what_benchmark/setup.py
```diff
@@ -43,4 +43,18 @@ def set_version(base_version: str):
...
+    extras_require={
+        "minicpm-o-2_6": [
+            "torch==2.10.0",
+            "torchvision==0.25.0",
+            "torchaudio==2.10.0",
+            "soundfile==0.13.1",
+            "librosa==0.11.0",
+            "torchcodec==0.7.0",
+            "vocos==0.1.0",
+            "vector_quantize_pytorch==1.23.2",
```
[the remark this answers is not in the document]
sbalandi: removed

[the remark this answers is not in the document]
sbalandi: it's expected restriction

tools/who_what_benchmark/requirements.txt
```diff
@@ -1,24 +1,27 @@
...
+lpips==0.1.4
+scikit-image<=0.26.0
+# - openbmb/MiniCPM-o-2_6
+soundfile==0.13.1
+librosa==0.11.0
+torchcodec<=0.10.0; sys_platform == "linux"
+vocos==0.1.0
+vector_quantize_pytorch==1.27.20
+torchaudio>=2.1.0,<=2.10.0
+torchvision>=0.16,<=0.25.0
```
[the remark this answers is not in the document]
sbalandi: align

tools/who_what_benchmark/setup.py
```diff
@@ -43,4 +43,18 @@ def set_version(base_version: str):
...
     entry_points={"console_scripts": ["wwb=whowhatbench.wwb:main"]},
     package_data={"whowhatbench": ["prompts/*.yaml", "prompts/*.json"]},
+    extras_require={
+        "minicpm-o-2_6": [
+            "torch==2.10.0",
+            "torchvision==0.25.0",
+            "torchaudio==2.10.0",
+            "soundfile==0.13.1",
+            "librosa==0.11.0",
+            "torchcodec==0.7.0",
```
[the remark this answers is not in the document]
sbalandi: it's expected

tools/llm_bench/requirements.txt
```diff
@@ -1,24 +1,44 @@
...
-pillow
-torch
-transformers[sentencepiece]>=4.40.0
-diffusers>=0.22.0
-#optimum is in dependency list of optimum-intel
-optimum-intel[nncf]>=1.25.0
-packaging
+openvino<=2026.1.0
+openvino-tokenizers<=2026.1.0
+openvino_genai<=2026.1.0
```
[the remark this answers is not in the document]
sbalandi: I don't have any strong opinions on what's right, then ok, have removed

tools/who_what_benchmark/setup.py
sbalandi: due to offline discussion - MiniCPM-o-4_5 will also work base on transformers<=4.51.3

tools/who_what_benchmark/setup.py
```diff
@@ -43,4 +43,13 @@ def set_version(base_version: str):
     install_requires=required,
     entry_points={"console_scripts": ["wwb=whowhatbench.wwb:main"]},
     package_data={"whowhatbench": ["prompts/*.yaml", "prompts/*.json"]},
+    extras_require={
+        "minicpm-o-2_6": [
```
as-suvorov, line 47: Why llm_bench has no minicpm extras?
Should we update docs that for particular models extras should be installed?
As I understand they are off by default?

sbalandi:
> Are versions synced between tools and genai?

yes, with ./python_tests/ and it also will be updated by [dependabot](https://github.com/openvinotoolkit/openvino.genai/blob/master/.github/dependabot.yml#L36)

tools/who_what_benchmark/setup.py
sbalandi: llm_bench also has requirements for minicpm https://github.com/openvinotoolkit/openvino.genai/pull/3314/changes#diff-1111c943b0dec29b93bde483d8a49efcf91239e8b4fae11139421c940a83c9c7R37-R41

sbalandi: This is an experiment for developers. I'm not sure it will be useful, but it won't break anything, and I find it more comfortable to have such extras, so we'll see

wwb is used by validation and developers
minicpm requires extra dependencies - some of them are don't overlap with general dependencies(vector-quantize-pytorch, vocos, etc.), and some overlap (like transformers)
Validation installs some of the modules like transformers(which should be specific version) or torch by itself, but it doesn't install some like vocos or vector-quantize-pytorch.
Therefore, I added to general requirements.txt such dependency that don't overlap with general dependencies(and is needed for validation) - vocos, vector-quantize-pytorch.
And here I add the dependencies that need to be changed specifically for this model(transformers) and those that must be compatible together(torch, torchvision), so if somebody decided to run minicpm, installation will be easy

sbalandi: llm_bench doesn't use the setup.py, so developers will have to install the required version of transformer themselves. This is generally a common practice for tools. I will add comments about this to the requiremets.txt

as-suvorov: It makes sense to me in general. But still there is no extras for minicpm for llm_bench. So we have 2 different approaches. I think we should sync them.
wwb is still a public tool. I think we should say that some models require extras install in docs.

sbalandi: readme have updated, task for separation of installation of the dependencies for tools and python_tests: [CVS-181535](https://jira.devtools.intel.com/browse/CVS-181535)

PR #3345, opened 2026-02-16
Fix load_dataset fails

as-suvorov:
## Description
HF `load_dataset` fails with 2 errors:
1. lock release fails with "OSError: [Errno 5] Input/output error"
2. RuntimeError: Deadlock: lock '/mount/caches/huggingface/lin/datasets/mount_caches_huggingface_lin_datasets_distil-whisper__meanwhile_default_0.0.0_5a6b431a268523a6603...593841065207078877.lock' is already held by a different FileLock instance in this thread. Use is_singleton=True to enable reentrant locking across instances.

Not reproduced locally.
Setting cache_dir to a new directory doesn't help.
Workaround is to use local /tmp dir.
Ticket: 181288

VLM Sample tests skipped on Linux due to hangs.
Another PR with the same issue: https://github.com/openvinotoolkit/openvino.genai/actions/runs/22105151693/job/63888905343
Ticket: 181387

KV cache eviction tests skipped due to `trust_remote_code` not supported.
Ticket: 181303
PR with fix: https://github.com/openvinotoolkit/openvino.genai/pull/3351

## Checklist:

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -574,7 +574,7 @@ jobs:
         test:
           - name: 'Whisper'
             # TODO: skip some tests temporary untill https://github.com/huggingface/datasets/issues/7647 dataset is fixed
-            cmd: 'python -m pytest -sv ./tests/python_tests/test_whisper_pipeline.py ./tests/python_tests/test_whisper_pipeline_static.py -k "not test_smoke[sample_from_dataset0 and not test_whisper_constructors[sample_from_dataset0 and not test_max_new_tokens[sample_from_dataset0 and not test_language_mode[language and not test_task_mode[sample_from_dataset0 and not test_language_autodetect[sample_from_dataset0 and not test_whisper_config_constructor and not test_language_autodetect[sample_from_dataset1 and not test_language_autodetect[sample_from_dataset2 and not test_initial_prompt_hotwords[sample_from_dataset0 and not test_random_sampling[sample_from_dataset0"'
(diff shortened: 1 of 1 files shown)


as-suvorov:
@akashchi Have you encountered a similar problem before?

as-suvorov:
> There's another failure now

Doesn't related to local dir move. Tests skipped. Double checking on master.

as-suvorov:
> > There's another failure now
> 
> Doesn't related to local dir move. Tests skipped. Double checking on master.

It is related to local dir move. `dataset_load` fails due to remote code unsupported. But mount disk have this dataset cached so test passes.

.github/workflows/linux.yml
```diff
@@ -572,7 +574,7 @@ jobs:
...
+            cmd: 'python -m pytest -sv ./tests/python_tests/test_whisper_pipeline.py ./tests/python_tests/test_whisper_pipeline_static.py -k "not test_smoke[sample_from_dataset0 and not test_whisper_constructors[sample_from_dataset0 and not test_max_new_tokens[sample_from_dataset0 and not test_language_mode[language and not test_task_mode[sample_from_dataset0 and not test_language_autodetect[sample_from_dataset0 and not test_whisper_config_constructor and not test_language_autodetect[sample_from_dataset1 and not test_language_autodetect[sample_from_dataset2 and not test_initial_prompt_hotwords[sample_from_dataset0 and not test_random_sampling[sample_from_dataset0"'
```
Wovchena: `-s` spams genai debug logs. Do we really need `-s`?

.github/workflows/linux.yml
```diff
@@ -25,6 +25,8 @@ env:
   SCCACHE_CACHE_SIZE: 30G
   SCCACHE_AZURE_KEY_PREFIX: genai/ubuntu/22_04/x64
   HF_HOME: /mount/caches/huggingface/lin
+  # mount directory leads to issue with load_dataset lock file. It's either cannot be released or there is a deadlock. Ticket: 181288
```
Wovchena, line 28: > cannot be released

> there is a deadlock

Seem to be 2 different wordings for the same problem. Can you delete the stuck file from the share?

as-suvorov: Doesn't help

.github/workflows/linux.yml
as-suvorov: removed

PR #3353, opened 2026-02-17
Get generation config

Retribution98:
## Description
- Implement the getGenerationConfig() functions for LLMPipeline and VLMPipeline.
- Extend the helper functions to convert CPP GenerationConfig to a JavaScript entity.
- Extend the parser wrappers to convert them into JavaScript entities.
- Change types to fix value overflow.
- Update the documentation.
- Add test cases.

CVS-181673

## Checklist:


src/js/tests/llmPipeline.test.js
```diff
@@ -71,6 +84,22 @@ describe("General LLM test", () => {
       assert.strictEqual(typeof tokenizer, "object");
     });
 
+    it("getGenerationConfig returns object with expected fields", async () => {
+      const config = pipeline.getGenerationConfig();
```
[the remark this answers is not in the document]
Retribution98: We have the separate test file to check GenerationConfig

src/js/src/helper.cpp
```diff
@@ -491,6 +806,27 @@ Napi::Value cpp_to_js<ov::genai::EmbeddingResults, Napi::Value>(const Napi::Env&
...
+Napi::Value cpp_to_js<size_t, Napi::Value>(const Napi::Env& env, const size_t& value) {
+    if (value <= NAPI_NUMBER_MAX_INTEGER) {
+        return Napi::Number::New(env, value);
+    }
+    return Napi::BigInt::New(env, value);
+}
+
+template <>
+Napi::Value cpp_to_js<float, Napi::Value>(const Napi::Env& env, const float& value) {
+    return Napi::Number::New(env, std::round(static_cast<double>(value) * 1e7) / 1e7);
```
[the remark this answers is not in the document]
Retribution98: It doesn't work

src/js/src/helper.cpp
```diff
@@ -114,14 +309,61 @@ template <>
...
     if (value.IsNumber()) {
-        return value.As<Napi::Number>().Int64Value();
-    } 
+        const Napi::Number number_value = value.As<Napi::Number>();
+        OPENVINO_ASSERT(is_js_integer(env, number_value), "Passed argument must be an integer.");
+        const auto value = number_value.Int64Value();
+        return value;
+    }
     bool lossless;
     auto result = value.As<Napi::BigInt>().Int64Value(&lossless);
```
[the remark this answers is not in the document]
Retribution98: If the value was greater than NAPI_NUMBER_MAX_INTEGER, it would lose precision and BigInt would be used.

src/js/src/helper.cpp
```diff
@@ -114,14 +309,61 @@ template <>
...
+    return value.As<Napi::Number>().DoubleValue();
+}
+
+template <>
+size_t js_to_cpp<size_t>(const Napi::Env& env, const Napi::Value& value) {
+    OPENVINO_ASSERT(value.IsNumber() || value.IsBigInt(), "Passed argument must be of type Number or BigInt.");
+    if (value.IsNumber()) {
+        const Napi::Number number_value = value.As<Napi::Number>();
+        OPENVINO_ASSERT(is_js_integer(env, number_value), "Passed argument must be an integer.");
+        const auto int_value = number_value.Int64Value();
```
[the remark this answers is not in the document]
Retribution98: If the value was greater than NAPI_NUMBER_MAX_INTEGER, it would lose precision and BigInt would be used.

src/js/include/parser.hpp
```diff
@@ -15,6 +15,7 @@ class JSParser : public ov::genai::Parser {
         js_parser_ref.Unref();
     };
     void parse(ov::genai::JsonContainer& message) override;
+    Napi::Object GetJSObject(Napi::Env env) const;
```
almilosz:
```suggestion
    Napi::Object get_js_object(Napi::Env env) const;
```

src/js/src/llm_pipeline/llm_pipeline_wrapper.cpp
```diff
@@ -261,3 +263,28 @@ Napi::Value LLMPipelineWrapper::get_tokenizer(const Napi::CallbackInfo& info) {
...
+    }
+    return env.Undefined();
+}
+
+Napi::Value LLMPipelineWrapper::set_generation_config(const Napi::CallbackInfo& info) {
+    auto env = info.Env();
+    try {
+        OPENVINO_ASSERT(this->pipe, "LLMPipeline is not initialized");
+        VALIDATE_ARGS_COUNT(info, 1, "setGenerationConfig()");
+        ov::genai::GenerationConfig config = js_to_cpp<ov::genai::GenerationConfig>(env, info[0]);
```
almilosz:
```suggestion
        const auto config = js_to_cpp<ov::genai::GenerationConfig>(env, info[0]);
```

src/js/src/llm_pipeline/llm_pipeline_wrapper.cpp
```diff
@@ -261,3 +263,28 @@ Napi::Value LLMPipelineWrapper::get_tokenizer(const Napi::CallbackInfo& info) {
     }
     return env.Undefined();
 }
+
+Napi::Value LLMPipelineWrapper::get_generation_config(const Napi::CallbackInfo& info) {
+    auto env = info.Env();
+    try {
+        OPENVINO_ASSERT(this->pipe, "LLMPipeline is not initialized");
+        ov::genai::GenerationConfig config = this->pipe->get_generation_config();
+        return cpp_to_js<ov::genai::GenerationConfig, Napi::Value>(env, config);
```
almilosz:
```suggestion
        return cpp_to_js<ov::genai::GenerationConfig, Napi::Value>(env, this->pipe->get_generation_config());
```

src/js/src/llm_pipeline/llm_pipeline_wrapper.cpp
```diff
@@ -261,3 +263,28 @@ Napi::Value LLMPipelineWrapper::get_tokenizer(const Napi::CallbackInfo& info) {
...
+    return env.Undefined();
+}
+
+Napi::Value LLMPipelineWrapper::set_generation_config(const Napi::CallbackInfo& info) {
+    auto env = info.Env();
+    try {
+        OPENVINO_ASSERT(this->pipe, "LLMPipeline is not initialized");
+        VALIDATE_ARGS_COUNT(info, 1, "setGenerationConfig()");
+        ov::genai::GenerationConfig config = js_to_cpp<ov::genai::GenerationConfig>(env, info[0]);
+        this->pipe->set_generation_config(config);
```
almilosz:
```suggestion
        this->pipe->set_generation_config(js_to_cpp<ov::genai::GenerationConfig>(env, info[0]));
```

src/js/tests/llmPipeline.test.js
almilosz, line 88: Please add a test that checks the type of config properties, e.g. `num_beam_groups?: number | bigint;`
How the user is supposed to work with such property? Can the type be assumed?

src/js/tests/llmPipeline.test.js
```diff
@@ -71,6 +84,22 @@ describe("General LLM test", () => {
...
+      const result = await pipeline.generate("What is OpenVINO?", { ...config, max_new_tokens: 5 });
+      assert.strictEqual(result.texts.length, 1);
+    });
+
+    it("setGenerationConfig updates config, getGenerationConfig returns updated values", async () => {
+      const config = pipeline.getGenerationConfig();
+      config.max_new_tokens = 5;
+      pipeline.setGenerationConfig(config);
+      const newConfig = pipeline.getGenerationConfig();
+      assert.deepEqual(config, newConfig);
```
almilosz, line 99: Please add an assertion that checks default and new value of `max_new_tokens`
```suggestion
     const original = pipeline.getGenerationConfig();
      const originalMaxNewTokens = original.max_new_tokens;
      ...
      assert.deepEqual(config, newConfig);
      assert.notEqual(newConfig.max_new_tokens, originalMaxNewTokens);
```

src/js/tests/llmPipeline.test.js
Retribution98: I don't expect users to work with any values in the generation configuration. I believe the main reason for using getGenerationConfig is to display or store the current configuration values.
The number is the default type of value. BigInt is only used for values that cannot be expressed as a general number. If the user wants to check the type for any reason, they could use the default approach:
```typescript
if (typeof config.num_beam_groups === "number") {
}
```

src/js/tests/llmPipeline.test.js
Retribution98: applied

src/js/include/parser.hpp
Retribution98: done

src/js/src/llm_pipeline/llm_pipeline_wrapper.cpp
almilosz: Please also check other places where you return value from `js_to_cpp`. e.g. vlm_pipeline

src/js/tests/llmPipeline.test.js
almilosz: If we return generationConfig, we cannot assume that it won't be used for comparisons or arithmetic. I’m still not convinced about returning `number | bigint` here. Did you check how it will behave in TypeScript?

If type is sometimes one and sometimes the other based on runtime values it creates confusion.


Beside if it's mainly for display and storing I don't see why not to return it as BigInt, but accept both number and bigint as incoming values

Retribution98: The `GenerationConfig` type is the same for both the get and set functions. If we allow both types to be set, it should be `number | bigint`.
TypeScript allows comparison of `number` and `bigint` values. However, arithmetic operations are not allowed for `bigint` and users don't see a difference between `bigint` and `number | bigint` when getting the value.

src/js/lib/utils.ts
```diff
@@ -247,32 +247,37 @@ export type RandomSamplingsGenerationConfig = {
...
   temperature?: number;
   /** the number of highest probability vocabulary tokens to keep for top-k-filtering. */
-  top_k?: number;
+  top_k?: number | bigint;
   /** if set to float < 1, only the smallest set of most probable tokens with probabilities that add up to top_p or higher are kept for generation. */
   top_p?: number;
   /** initializes random generator. */
-  rng_seed?: number;
-  /** the number of sequences to generate from a single prompt. */
-  num_return_sequences?: number;
```
almilosz, line 256: Is `num_return_sequences` removed intentionally?

src/js/src/helper.cpp
```diff
@@ -15,55 +17,247 @@ constexpr const char* CPP_SCHEDULER_CONFIG_KEY = "scheduler_config";
...
+        for (const Napi::Value& v : values) {
+            if (v.IsNumber()) {
+                if (!is_js_integer(env, v.As<Napi::Number>()))
+                    has_non_integer_number = true;
+                else {
+                    const int64_t sval = v.As<Napi::Number>().Int64Value();
+                    if (sval < 0 || static_cast<uint64_t>(sval) > static_cast<uint64_t>(SIZE_MAX))
+                        all_fit_size_t = false;
+                }
+            } else if (v.IsBigInt()) {
```
almilosz: Please create helper for converting `Napi::Value` to `BigInt`

Retribution98: I use get_cpp_type to specify the C++ type that should be used for the value.

src/js/lib/utils.ts
Retribution98: Yes, it was duplicated

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -276,12 +277,22 @@ Napi::Value VLMPipelineWrapper::set_generation_config(const Napi::CallbackInfo&
...
     }
     return env.Undefined();
 }
+
+Napi::Value VLMPipelineWrapper::get_generation_config(const Napi::CallbackInfo& info) {
+    auto env = info.Env();
+    try {
+        OPENVINO_ASSERT(this->pipe, "VLMPipeline is not initialized");
+        ov::genai::GenerationConfig config = this->pipe->get_generation_config();
+        return cpp_to_js<ov::genai::GenerationConfig, Napi::Value>(env, config);
```
almilosz, line 293:
```suggestion
        return cpp_to_js<ov::genai::GenerationConfig, Napi::Value>(env, this->pipe->get_generation_config());
```

src/js/src/vlm_pipeline/vlm_pipeline_wrapper.cpp
```diff
@@ -276,12 +277,22 @@ Napi::Value VLMPipelineWrapper::set_generation_config(const Napi::CallbackInfo&
     try {
         OPENVINO_ASSERT(this->pipe, "VLMPipeline is not initialized");
         VALIDATE_ARGS_COUNT(info, 1, "setGenerationConfig()");
-        auto config_map = js_to_cpp<ov::AnyMap>(env, info[0]);
-        ov::genai::GenerationConfig config;
-        config.update_generation_config(config_map);
+        ov::genai::GenerationConfig config = js_to_cpp<ov::genai::GenerationConfig>(env, info[0]);
         this->pipe->set_generation_config(config);
```
almilosz, line 281:
```suggestion
this->pipe->set_generation_config(js_to_cpp<ov::genai::GenerationConfig>(env, info[0]));
```

PR #3367, opened 2026-02-21
fix(vlm): MAX_PROMPT_LEN crashes VLMPipeline on NPU with "Unsupported property by CPU plugin"

0ldev:
This creates an impossible situation: the pipeline tells you to set a property
that it cannot accept.

**Root cause:** In [VLMPipelineImpl](cci:1://file:///c:/Users/B3T0/Documents/openvino.genai-1/src/cpp/src/visual_language/pipeline.cpp:83:4-165:5) constructor ([pipeline.cpp](cci:7://file:///c:/Users/B3T0/Documents/openvino.genai-1/src/c/src/vlm_pipeline.cpp:0:0-0:0)), `properties_copy`
is used as the source for both `lm_properties` (→ NPU LM subgraph) and
`embedder_properties` (→ CPU vision encoder). Since `lm_properties` is a **copy**,
[pop_int_and_cast](cci:1://file:///c:/Users/B3T0/Documents/openvino.genai-1/src/cpp/src/utils.cpp:54:0-73:1) inside [compile_decoder_for_npu](cci:1://file:///c:/Users/B3T0/Documents/openvino.genai-1/src/cpp/src/utils.hpp:192:0-195:92) removes `MAX_PROMPT_LEN` from
the copy only — leaving it in `properties_copy`, where it later contaminates
`embedder_properties` and reaches the CPU plugin.

**Fix:** Extract `MAX_PROMPT_LEN` and `MIN_RESPONSE_LEN` from `properties_copy`
immediately after `DEVICE_PROPERTIES` is popped, before either sub-component
property map is derived. Re-inject them into `lm_properties` only when
`m_is_npu` is true.

The `DEVICE_PROPERTIES` nested routing path was already correct and is unaffected.

## Checklist:
  > The flat-kwargs NPU path requires NPU hardware in CI. A follow-up test
  > `test_vlm_npu_max_prompt_len_flat_kwargs` should be added to
  > [tests/python_tests/test_vlm_pipeline.py](cci:7://file:///c:/Users/B3T0/Documents/openvino.genai-1/tests/python_tests/test_vlm_pipeline.py:0:0-0:0) once the CI NPU environment
  > is confirmed. The existing `DEVICE_PROPERTIES`-based NPU tests are unaffected.
- [x] No public API surface changed; fix restores the documented intended behavior.

### DIFF
--- a/src/cpp/src/visual_language/pipeline.cpp
+++ b/src/cpp/src/visual_language/pipeline.cpp
@@ -146,6 +146,13 @@ private:
         auto device_properties = utils::pop_or_default<ov::AnyMap>(
             properties_copy, ov::device::properties.name(), { }
         );
+
+        // Extract NPU-specific KV-cache properties from the top-level map
+        // before they get forwarded to the CPU vision encoder, which would
+        // reject them as unsupported.
+        auto npu_max_prompt_len = utils::pop_option(properties_copy, "MAX_PROMPT_LEN");
+        auto npu_min_response_len = utils::pop_option(properties_copy, "MIN_RESPONSE_LEN");
+
         // Otherwise, the same properties are used for all models and devices
         auto lm_properties = device_properties.empty()
             ? properties_copy
@@ -163,6 +170,11 @@ private:
         if (m_is_npu) {
             embedder_device = "AUTO";
             utils::KVDesc kv_desc;
+            // Re-inject extracted NPU properties into lm_properties only
+            if (npu_max_prompt_len.has_value())
+                lm_properties["MAX_PROMPT_LEN"] = npu_max_prompt_len.value();
+            if (npu_min_response_len.has_value())
+                lm_properties["MIN_RESPONSE_LEN"] = npu_min_response_len.value();
             update_npu_properties(models_dir, lm_properties);
             std::tie(compiled_language_model, kv_desc) = utils::compile_decoder_for_npu(language_model, lm_properties, kv_pos);
             m_max_prompt_len = kv_desc.max_prompt_len;
--- a/src/cpp/src/visual_language/pipeline.cpp
+++ b/src/cpp/src/visual_language/pipeline.cpp
@@ -147,12 +147,6 @@ private:
             properties_copy, ov::device::properties.name(), { }
         );
 
-        // Extract NPU-specific KV-cache properties from the top-level map
-        // before they get forwarded to the CPU vision encoder, which would
-        // reject them as unsupported.
-        auto npu_max_prompt_len = utils::pop_option(properties_copy, "MAX_PROMPT_LEN");
-        auto npu_min_response_len = utils::pop_option(properties_copy, "MIN_RESPONSE_LEN");
-
         // Otherwise, the same properties are used for all models and devices
         auto lm_properties = device_properties.empty()
             ? properties_copy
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ if(UNIX AND NOT (APPLE OR ANDROID OR CYGWIN))
 endif()
 
 project(OpenVINOGenAI
-        VERSION 2026.1.0.0
+        VERSION 2025.4.1.0
```
as-suvorov: Please revert version

as-suvorov (changes requested):

CMakeLists.txt
0ldev: Version reverted

yatarkan (approved):

yatarkan:
@dmatveev Could you please take a look at the suggested fix from NPU perspective?

0ldev:
@dmatveev @Wovchena Please, a moment of your time to review this PR before it closes automatically for inactivity.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -144,6 +144,13 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
         auto device_properties = utils::pop_or_default<ov::AnyMap>(
             properties_copy, ov::device::properties.name(), { }
         );
+
+        // Extract NPU-specific KV-cache properties from the top-level map
+        // before they get forwarded to the CPU vision encoder, which would
+        // reject them as unsupported.
+        auto npu_max_prompt_len = utils::pop_option(properties_copy, "MAX_PROMPT_LEN");
+        auto npu_min_response_len = utils::pop_option(properties_copy, "MIN_RESPONSE_LEN");
```
as-suvorov, line 154: Let's move under `m_is_npu` condition as used only there

as-suvorov (changes requested):

src/cpp/src/visual_language/pipeline.cpp
0ldev: Done, moved inside `m_is_npu`.

0ldev:
@as-suvorov The requested change (moving MAX_PROMPT_LEN/MIN_RESPONSE_LEN extraction inside the `m_is_npu` block) has been addressed. Could you re-review when you get a chance? Thanks!

0ldev:
@as-suvorov Following up — the test `test_vlm_npu_max_prompt_len_flat_kwargs` (added in 332cd02) covers the flat top-level kwargs path. It uses `NPUW_DEVICES=CPU` so no real NPU hardware is needed in CI. Would you be able to re-review?

as-suvorov (approved):

Wovchena:
build_jenkins

dmatveev:
`MAX_PROMPT_LEN` should not be passed to CPU plugin when the configuration done right: https://docs.openvino.ai/2026/openvino-workflow-generative/inference-with-genai/inference-with-genai-on-npu.html#passing-config-to-vlms

AlexanderKalistratov:
@0ldev thanks for the contribution. But in the current state this PR must not be merged. There is a way how to pass configs described in [official docs](https://docs.openvino.ai/2026/openvino-workflow-generative/inference-with-genai/inference-with-genai-on-npu.html#passing-config-to-vlms). And btw you could find it in tests too: [test_vlm_pipeline.py#L270](https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/test_vlm_pipeline.py#L270)

There is much more options exists for NPU than `MAX_PROMPT_LEN` and `MIN_RESPONSE_LEN`. Extracting only these two creates more issues than solving problems.

Still since VLM configuration causes a lot of confusion something should be done.

AlexanderKalistratov:
@0ldev could you please test latest version of ov and ov.genai?
In my local test everything works even without device-specific configs.

PR #3376, opened 2026-02-23
Fix Coverity static analysis defects (Feb)

sgonorov:
## Description

Fix three Coverity static analysis defects across speculative decoding, LLM pipeline, and Whisper components.

- **CID 6415047 (INTEGER_OVERFLOW)**: Add boundary guards before decrementing unsigned indices in DTW backtracking to prevent `size_t` underflow - [CVS-181400](https://jira.devtools.intel.com/projects/CVS/issues/CVS-181400)
- **CID 6419148 (MISSING_LOCK)**: Access `m_draft_generations_mutex` field directly instead of through accessor in `generate_common` template, so Coverity can correlate the lock with the guarded data - [CVS-181402](https://jira.devtools.intel.com/projects/CVS/issues/CVS-181402)
- **CID 6415437 (NULL_FIELD)**: Add defensive assertion on `main_model_descr.model` before passing it to `StatefulLLMPipeline` constructor - [CVS-181401](https://jira.devtools.intel.com/projects/CVS/issues/CVS-181401)

### DIFF
--- a/src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
+++ b/src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
@@ -196,7 +196,6 @@ public:
     Tokenizer& tokenizer() { return m_tokenizer; }
     const Tokenizer& tokenizer() const { return m_tokenizer; }
 
-    std::mutex& draft_generations_mutex() { return m_draft_generations_mutex; }
 };
 
 }  // namespace ov::genai


src/cpp/src/llm/pipeline.cpp
```diff
@@ -151,6 +151,7 @@ static std::unique_ptr<LLMPipelineImplBase> create(const std::shared_ptr<ov::Mod
 
     auto main_model_descr =
         ov::genai::ModelDesc(model, tokenizer, device, properties_without_draft_model, {}, generation_config);
+    OPENVINO_ASSERT(main_model_descr.model, "Model descriptor must contain a valid model");
 
     if (draft_model_descr.model != nullptr) {
```
sbalandi: May this condition be removed then?

src/cpp/src/whisper/word_level_timestamps.cpp
```diff
@@ -269,12 +269,23 @@ std::vector<std::pair<size_t, size_t>> dtw_and_backtrace(const std::vector<std::
         path.push_back({path_i, path_j});
         int t = trace[i][j];
         if (t == 0) {
-            i--;
-            j--;
+            if (i == 0 || j == 0) {
+                OPENVINO_THROW("Invalid DTW trace: diagonal move at matrix boundary");
+            }
```
as-suvorov: It's preferable to use asserts over throws. Below as well
```suggestion
    OPENVINO_ASSERT(i > 0 && j > 0, "Invalid DTW trace: diagonal move at matrix boundary")
```

src/cpp/src/llm/pipeline.cpp
sgonorov: I can remove `!= nullptr` so it will be more generic.

src/cpp/src/whisper/word_level_timestamps.cpp
sgonorov: Fixed

src/cpp/src/llm/pipeline.cpp
sbalandi: it is useless here, it always be true, because if not, it will be raised in OPENVINO _ASSERT, which you added

as-suvorov: Assert is for `main_model`. `if` for `draft_model`, it should not be removed.

sbalandi: aa, sorry, yes !

src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
```diff
@@ -57,7 +57,7 @@ std::vector<EncodedGenerationResult> generate_common(
 
     std::vector<GenerationHandle> main_generations;
     {
-        std::lock_guard<std::mutex> lock(self->draft_generations_mutex());
+        std::lock_guard<std::mutex> lock(self->m_draft_generations_mutex);
```
as-suvorov, line 62: Not sure if it's intended that draft_generations_mutex() can be overloaded by subclasses.
As I see with this change `draft_generations_mutex` is no longer used, let's maybe delete it then?

PR #3383, opened 2026-02-24
Fix GIL crash in image generation pipelines on repeated generate() calls

sgonorov:
Fix `pybind11::handle::dec_ref()` GIL crash that occurs on the second `generate()` call in image generation pipelines (Text2Image, Image2Image, Inpainting) when a Python-backed generator (e.g. `TorchGenerator`) is passed.

- Introduce `GilSafeGeneratorWrapper` in `py_utils.cpp` that acquires the GIL before destroying the underlying Python generator, preventing unsafe `Py_DECREF` when the old generator is overwritten during subsequent `generate()` calls with `py::gil_scoped_release` active
- Add GIL-safe destructor to `TorchGenerator` as defense-in-depth for its `py::object` members
- Remove unused `params_have_torch_generator` helper
- Work around `ov::Core::read_model` infinite recursion caused by a template overload in OpenVINO 2026.1.0 `core.hpp` by using `std::as_const()` on non-const `AnyMap` arguments across tokenizer, whisper, and RAG pipelines

Issue [CVS-181541](https://jira.devtools.intel.com/projects/CVS/issues/CVS-181541)

### DIFF
--- a/src/cpp/src/rag/text_embedding_pipeline.cpp
+++ b/src/cpp/src/rag/text_embedding_pipeline.cpp
@@ -4,6 +4,8 @@
 #include "openvino/genai/rag/text_embedding_pipeline.hpp"
 
 #include <fstream>
+#include <utility>
+
 #include <nlohmann/json.hpp>
 
 #include "json_utils.hpp"
--- a/tests/python_tests/test_image_generation_multi_call.py
+++ b/tests/python_tests/test_image_generation_multi_call.py
@@ -2,52 +2,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import pytest
-import subprocess  # nosec B404
-import logging
-from pathlib import Path
 import numpy as np
 import openvino as ov
 import openvino_genai as ov_genai
-
-from utils.constants import get_ov_cache_converted_models_dir
-from utils.atomic_download import AtomicDownloadManager
-from utils.network import retry_request
-
-logger = logging.getLogger(__name__)
-
-MODEL_ID = "tiny-random-latent-consistency"
-MODEL_NAME = "echarlaix/tiny-random-latent-consistency"
-
-
-@pytest.fixture(scope="module")
-def image_generation_model():
-    models_dir = get_ov_cache_converted_models_dir()
-    model_path = Path(models_dir) / MODEL_ID / MODEL_NAME
-
-    manager = AtomicDownloadManager(model_path)
-
-    def convert_model(temp_path: Path) -> None:
-        command = [
-            "optimum-cli",
-            "export",
-            "openvino",
-            "--model",
-            MODEL_NAME,
-            "--trust-remote-code",
-            "--weight-format",
-            "fp16",
-            str(temp_path),
-        ]
-        logger.info(f"Conversion command: {' '.join(command)}")
-        retry_request(lambda: subprocess.run(command, check=True, text=True, capture_output=True))
-
-    try:
-        manager.execute(convert_model)
-    except subprocess.CalledProcessError as error:
-        logger.exception(f"optimum-cli returned {error.returncode}. Output:\n{error.output}")
-        raise
-
-    return str(model_path)
 
 
 def get_random_image(height: int = 64, width: int = 64) -> ov.Tensor:
--- a/src/cpp/src/tokenizer/tokenizer_impl.cpp
+++ b/src/cpp/src/tokenizer/tokenizer_impl.cpp
@@ -2,6 +2,9 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "tokenizer/tokenizer_impl.hpp"
+
+#include <utility>
+
 #include "add_second_input_pass.hpp"
 #include "sampling/structured_output/structured_output_controller.hpp"
 #include "openvino/genai/version.hpp"
--- a/src/cpp/src/whisper/models/statefull_decoder.cpp
+++ b/src/cpp/src/whisper/models/statefull_decoder.cpp
@@ -3,6 +3,8 @@
 
 #include "statefull_decoder.hpp"
 
+#include <utility>
+
 #include "openvino/op/softmax.hpp"
 #include "openvino/pass/manager.hpp"
 #include "utils.hpp"
--- a/src/cpp/src/whisper/pipeline.cpp
+++ b/src/cpp/src/whisper/pipeline.cpp
@@ -3,6 +3,8 @@
 
 #include <algorithm>
 #include <filesystem>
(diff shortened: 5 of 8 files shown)


src/python/py_image_generation_pipelines.cpp
Wovchena, line 1: @likholat should have been summoned automatically because of this file modification. Please update CODEOWNERS (not requesting review intentionally so you could test your update). tests/python_tests/test_image_generation_multi_call.py should also be included into the triggers

.github/workflows/windows.yml
```diff
@@ -674,7 +674,7 @@ jobs:
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).continuous_batching.test }}
             timeout: 360
           - name: 'LLM & VLM'
-            cmd: 'python -m pytest -s -v tests/python_tests/test_llm_pipeline.py tests/python_tests/test_llm_pipeline_static.py tests/python_tests/test_vlm_pipeline.py tests/python_tests/test_structured_output.py tests/python_tests/test_image_generation.py --override-ini cache_dir=/mount/caches/pytest/'
+            cmd: 'python -m pytest -s -v tests/python_tests/test_llm_pipeline.py tests/python_tests/test_llm_pipeline_static.py tests/python_tests/test_vlm_pipeline.py tests/python_tests/test_structured_output.py tests/python_tests/test_image_generation.py tests/python_tests/test_image_generation_multi_call.py --override-ini cache_dir=/mount/caches/pytest/'
```
Wovchena, line 677: Extend other OS. Maybe it's worth moving to a separate matrix element

src/cpp/src/rag/text_embedding_pipeline.cpp
```diff
@@ -92,7 +92,7 @@ class TextEmbeddingPipeline::TextEmbeddingPipelineImpl {
 
         ov::Core core = utils::singleton_core();
 
-        auto model = core.read_model(models_path / "openvino_model.xml", {}, properties);
+        auto model = core.read_model(models_path / "openvino_model.xml", {}, std::as_const(properties));
```
Wovchena, line 97: > Work around `ov::Core::read_model` infinite recursion caused by a template overload in OpenVINO 2026.1.0 `core.hpp` by using `std::as_const()` on non-const `AnyMap` arguments across tokenizer, whisper, and RAG pipelines

Is that about 
```C++
    template <class Path, class... Properties>
    util::EnableIfAllStringAny<std::shared_ptr<ov::Model>, Properties...> read_model(const Path& model_path,
                                                                                     const Path& bin_path,
                                                                                     Properties&&... properties) const {
        return read_model(model_path, bin_path, ov::AnyMap{std::forward<Properties>(properties)...});
    }
```
? How did it work before?

src/python/py_utils.cpp
```diff
@@ -23,6 +23,37 @@
...
 namespace py = pybind11;
+
+namespace {
+
+class GilSafeGeneratorWrapper : public ov::genai::Generator {
+    std::shared_ptr<ov::genai::Generator> m_impl;
+    py::object m_py_ref;
+
+public:
+    GilSafeGeneratorWrapper(std::shared_ptr<ov::genai::Generator> impl, py::object py_ref)
```
Wovchena:
```suggestion
    GilSafeGeneratorWrapper(std::shared_ptr<ov::genai::Generator>&& impl, py::object&& py_ref)
```

tests/python_tests/test_image_generation_multi_call.py
```diff
@@ -0,0 +1,150 @@
...
+from utils.network import retry_request
+
+logger = logging.getLogger(__name__)
+
+MODEL_ID = "tiny-random-latent-consistency"
+MODEL_NAME = "echarlaix/tiny-random-latent-consistency"
+
+
+@pytest.fixture(scope="module")
+def image_generation_model():
```
Wovchena: `tests/python_tests/test_image_generation.py` has similar fixture. Move to common

tests/python_tests/test_image_generation_multi_call.py
```diff
@@ -0,0 +1,150 @@
...
+            gen = ov_genai.CppStdGenerator(42)
+            image = pipe.generate(f"prompt {i}", generator=gen, **GENERATE_KWARGS)
+            assert image is not None
+
+    def test_multiple_generate_with_torch_generator(self, image_generation_model):
+        pytest.importorskip("torch")
+        pipe = ov_genai.Text2ImagePipeline(image_generation_model, "CPU")
+        for i in range(NUM_CALLS):
+            gen = ov_genai.TorchGenerator(42)
+            image = pipe.generate(f"prompt {i}", generator=gen, **GENERATE_KWARGS)
```
Wovchena: Make it a single test parametrized with generator. `None` would be equivalent to `test_multiple_generate_without_generator()`

tests/python_tests/test_image_generation_multi_call.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class TestInpaintingMultipleGenerations:
+    def test_multiple_generate_without_generator(self, image_generation_model):
+        pipe = ov_genai.InpaintingPipeline(image_generation_model, "CPU")
+        for i in range(NUM_CALLS):
+            image = pipe.generate(f"prompt {i}", get_random_image(), get_mask_image(), strength=0.8, **GENERATE_KWARGS)
+            assert image is not None
+
+    def test_multiple_generate_with_torch_generator(self, image_generation_model):
```
Wovchena: `CppStdGenerator` is missing. Same for `TestImage2ImageMultipleGenerations`

.github/workflows/windows.yml
sgonorov: Extended, but i think a separate matrix element is not worth it yet.

src/cpp/src/rag/text_embedding_pipeline.cpp
sgonorov: I don't know, it was failing for me during compilation. It just guarantees the const ov::AnyMap& overload is selected.

src/python/py_utils.cpp
sgonorov: Fixed

tests/python_tests/test_image_generation_multi_call.py
sgonorov: Fixed, refactored it a little bit

tests/python_tests/test_image_generation_multi_call.py
sgonorov: Fixed

PR #3394, opened 2026-02-26
[GPU][NPU] Resolve excessive memory usage in Eagle3

GuoliangShiIntel:
## Description
**Quick background:**
In the Eagle3 speculative decoding pipeline, currently the draft model's embedding weights is been derived from the target model. 

The [share_vocabulary](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/speculative_decoding/eagle3_model_transforms.cpp#L117) function does this by sharing embedding nodes directly between both models, which interferes with the shared memory optimization and prevents `ov::Model` from being properly released after compilation.

**Fix:**
Embedding weights are now cloned into the draft model instead of shared, allowing each model's memory to be released independently and reducing memory consumption by ~4 GB for 8B (4bit quantize) model.

Tickets: CVS-181133

## Checklist:

### DIFF
--- a/tests/python_tests/test_stateful_speculative_decoding.py
+++ b/tests/python_tests/test_stateful_speculative_decoding.py
@@ -4,7 +4,6 @@
 
 import pytest
 import numpy as np
-import os
 
 import openvino as ov
 import openvino_genai as ov_genai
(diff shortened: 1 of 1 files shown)


src/cpp/src/speculative_decoding/eagle3_model_transforms.cpp
```diff
@@ -113,8 +113,55 @@ void share_vocabulary(const std::shared_ptr<ov::Model>& main_model, const std::s
...
+    auto cloned_weight_node = clone_node_recursive(main_weight_node, cloned_nodes);
+
+    OPENVINO_ASSERT(cloned_weight_node,
+                    "Failed to clone embedding weight node from main model to draft model. "
+                    "This is required for Eagle3 speculative decoding.");
+
+    // Replace draft model's weight node with the cloned subgraph
+    // This avoids cross-model references by duplicating the vocabulary weights
+    draft_weight_node->output(0).replace(cloned_weight_node->output(0));
 }
```
GuoliangShiIntel, line 165: <html>
<body>
<html><head></head><body><p><strong>Self note:</strong> Why no test added?</p>
<p>Below are the system memory changes with/without this PR
<em>(Model: Qwen3-8B int4 + Eagle3 int8)</em></p>

Device | Before | After
-- | -- | --
GPU | 11.14 GB | 7.03 GB
NPU | 14.81 GB | 11.21 GB
CPU | 9.10 GB | 10.21 GB


<p>This issue only manifests on GPU and NPU — CPU behavior looks correct. Since the CI environment appears to only have CPU devices (no GPU/NPU available), have not good idea for this test.</p></body></html>
</body>
</html>

GuoliangShiIntel:
@songbell @peterchen-intel Could you please review this fix first, as it also affects the GPU pipeline?

src/cpp/src/speculative_decoding/eagle3_model_transforms.cpp
```diff
@@ -113,8 +113,55 @@ void share_vocabulary(const std::shared_ptr<ov::Model>& main_model, const std::s
...
+        auto it = cloned_nodes.find(node.get());
+        if (it != cloned_nodes.end()) {
+            return it->second;
+        }
+
+        std::shared_ptr<ov::Node> cloned;
+
+        if (auto constant = ov::as_type_ptr<ov::op::v0::Constant>(node)) {
+            // For Constant nodes, create a deep copy with new data
+            cloned = std::make_shared<ov::op::v0::Constant>(constant->get_element_type(),
```
songbell, line 135: maybe we need to understand why the extra 4G on GPU and ~3G on NPU is consumed or not released? 
for the shared output, it should cut the connection to the shared node when the ov model releases, and for the constant node, if we use the shared data instead of deep copy, the issue is still there?

GuoliangShiIntel: > maybe we need to understand why the extra 4G on GPU and ~3G on NPU is consumed or not released?

Based on debug information, i have confirmed that the `ov::Model` is not being fully released. This suggests that sharing nodes across different models may be creating reference dependencies that prevent the reference count from reaching zero. The root cause needs more investigate in OpenVINO.

> if we use the shared data instead of deep copy, the issue is still there?

Yes, the issue persists even when using shared data.

songbell: @GuoliangShiIntel the extra memory on GPU/NPU should be due to the mmap of the model. let's merge the fix first.

PR #3396, opened 2026-02-26
[NPU] Eagle 3 Topk Pipeline

GuoliangShiIntel:
## Description

[CVS-181962](https://jira.devtools.intel.com/browse/CVS-181962)

Support eagle3 stateful pipeline for NPU in GenAI

this PR should after #3451

## Checklist:

### DIFF
--- a/samples/cpp/text_generation/speculative_decoding_lm.cpp
+++ b/samples/cpp/text_generation/speculative_decoding_lm.cpp
@@ -11,32 +11,6 @@ int main(int argc, char* argv[]) try {
         throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> <DRAFT_MODEL_DIR> '<PROMPT>'");
     }
 
-    ov::genai::GenerationConfig config;
-    config.max_new_tokens = 100;
-    // Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded.
-    // Add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration.
-    // NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial
-    // value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both
-    // backends.
-    config.num_assistant_tokens = 4;
-    // Add parameter to enable speculative decoding to generate candidates by draft_model while candidate probability is higher than
-    // `assistant_confidence_threshold`.
-    // NOTE: `assistant_confidence_threshold` is supported only by ContinuousBatching backend.
-    // config.assistant_confidence_threshold = 0.4;
-
-    // Tree search parameters (for Eagle3 tree-based speculative decoding):
-    // `branching_factor` is the number of top-k candidates selected per tree node and kept globally per tree layer.
-    // `tree_depth` is the lookahead depth of the candidate tree; the draft model runs `tree_depth` iterations.
-    // NOTE: The total number of draft tokens produced by the tree is
-    // `total_draft_tokens = branching_factor^2 * (tree_depth - 1) + branching_factor`.
-    // `total_draft_tokens >= num_assistant_tokens` must hold. The top `num_assistant_tokens` candidates (by score) are
-    // selected from the tree for verification.
-    // For tree search, `num_assistant_tokens` serves as the overall number of candidate (non-root) tokens submitted to
-    // the target model for verification; total tree nodes = `num_assistant_tokens + 1` (including root).
-    // config.branching_factor = 2;
-    // config.tree_depth = 4;
-    // config.num_assistant_tokens = 8;
-
     std::string main_model_path = argv[1];
     std::string draft_model_path = argv[2];
     std::string prompt = argv[3];
@@ -47,10 +21,52 @@ int main(int argc, char* argv[]) try {
     // use NPU.
     std::string main_device = "CPU", draft_device = "CPU";
 
-    ov::genai::LLMPipeline pipe(
-        main_model_path,
-        main_device,
-        ov::genai::draft_model(draft_model_path, draft_device));
+    // Select speculative decoding approach:
+    //   "fast_draft" - sequential candidate generation with a smaller draft model
+    //   "eagle3"     - Eagle3 tree-based candidate generation
+    const std::string speculative_mode = "fast_draft";
+
+    ov::genai::GenerationConfig config;
+    config.max_new_tokens = 100;
+
+    ov::AnyMap draft_properties;
+
+    if (speculative_mode == "fast_draft") {
+        // Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold`
+        // are mutually exclusive.
+        // `num_assistant_tokens` controls how many candidates the draft model generates per iteration.
+        // NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses it as initial
+        // value and adjusts based on recent number of accepted tokens. Defaults to `5` for both backends.
+        config.num_assistant_tokens = 4;
+        // `assistant_confidence_threshold` generates candidates while probability exceeds the threshold.
+        // NOTE: supported only by ContinuousBatching backend.
+        // config.assistant_confidence_threshold = 0.4;
+    } else if (speculative_mode == "eagle3") {
+        // Eagle3 tree-based speculative decoding parameters:
+        // `branching_factor` - number of top-k candidates selected per tree node and kept globally per tree layer.
+        // `tree_depth` - lookahead depth of the candidate tree; the draft model runs `tree_depth` iterations.
+        // `num_assistant_tokens` - number of candidate (non-root) tokens submitted to the target model for
+        //   verification; total tree nodes = `num_assistant_tokens + 1` (including root).
+        // NOTE: The total draft tokens produced by the tree is:
+        //   total_draft_tokens = branching_factor^2 * (tree_depth - 1) + branching_factor
+        // Constraint: total_draft_tokens >= num_assistant_tokens must hold.
+        config.branching_factor = 4;
+        config.tree_depth = 2;
+        config.num_assistant_tokens = 7;
+
+        // NPU requires static tensor shapes at model compilation time, so compile-time upper bounds
+        // for tree search dimensions must be specified. Runtime values must not exceed these limits.
+        // On CPU/GPU these properties are ignored and can be omitted.
+        if (draft_device == "NPU") {
+            draft_properties = {{"MAX_TREE_DEPTH", config.tree_depth},
+                                {"MAX_BRANCHING_FACTOR", config.branching_factor},
+                                {"MAX_ASSISTANT_TOKENS", config.num_assistant_tokens}};
+        }
+    }
+
+    ov::genai::LLMPipeline pipe(main_model_path,
+                                main_device,
+                                ov::genai::draft_model(draft_model_path, draft_device, draft_properties));
 
     auto streamer = [](std::string subword) {
         std::cout << subword << std::flush;
@@ -66,33 +82,43 @@ int main(int argc, char* argv[]) try {
         auto main_model_metrics = sd_perf_metrics->main_model_metrics;
         std::cout << "\nMAIN MODEL " << std::endl;
         std::cout << "  Generate time: " << main_model_metrics.get_generate_duration().mean << " ms" << std::endl;
-        std::cout << "  TTFT: " << main_model_metrics.get_ttft().mean  << " ± " << main_model_metrics.get_ttft().std << " ms" << std::endl;
-        std::cout << "  TTST: " << main_model_metrics.get_ttst().mean  << " ± " << main_model_metrics.get_ttst().std << " ms/token " << std::endl;
-        std::cout << "  TPOT: " << main_model_metrics.get_tpot().mean  << " ± " << main_model_metrics.get_tpot().std << " ms/iteration " << std::endl;
-        std::cout << "  AVG Latency: " << main_model_metrics.get_latency().mean  << " ± " << main_model_metrics.get_latency().std << " ms/token " << std::endl;
-        std::cout << "  Num generated token: " << main_model_metrics.get_num_generated_tokens() << " tokens" << std::endl;
+        std::cout << "  TTFT: " << main_model_metrics.get_ttft().mean << " ± " << main_model_metrics.get_ttft().std
+                  << " ms" << std::endl;
+        std::cout << "  TTST: " << main_model_metrics.get_ttst().mean << " ± " << main_model_metrics.get_ttst().std
+                  << " ms/token " << std::endl;
+        std::cout << "  TPOT: " << main_model_metrics.get_tpot().mean << " ± " << main_model_metrics.get_tpot().std
+                  << " ms/iteration " << std::endl;
+        std::cout << "  AVG Latency: " << main_model_metrics.get_latency().mean << " ± "
+                  << main_model_metrics.get_latency().std << " ms/token " << std::endl;
+        std::cout << "  Num generated token: " << main_model_metrics.get_num_generated_tokens() << " tokens"
+                  << std::endl;
         std::cout << "  Total iteration number: " << main_model_metrics.raw_metrics.m_durations.size() << std::endl;
         std::cout << "  Num accepted token: " << sd_perf_metrics->get_num_accepted_tokens() << " tokens" << std::endl;
 
         auto draft_model_metrics = sd_perf_metrics->draft_model_metrics;
         std::cout << "\nDRAFT MODEL " << std::endl;
         std::cout << "  Generate time: " << draft_model_metrics.get_generate_duration().mean << " ms" << std::endl;
-        std::cout << "  TTFT: " << draft_model_metrics.get_ttft().mean  << " ms" << std::endl;
-        std::cout << "  TTST: " << draft_model_metrics.get_ttst().mean  << " ms/token " << std::endl;
-        std::cout << "  TPOT: " << draft_model_metrics.get_tpot().mean  << " ± " << draft_model_metrics.get_tpot().std << " ms/token " << std::endl;
-        std::cout << "  AVG Latency: " << draft_model_metrics.get_latency().mean  << " ± " << draft_model_metrics.get_latency().std << " ms/iteration " << std::endl;
-        std::cout << "  Num generated token: " << draft_model_metrics.get_num_generated_tokens() << " tokens" << std::endl;
+        std::cout << "  TTFT: " << draft_model_metrics.get_ttft().mean << " ms" << std::endl;
+        std::cout << "  TTST: " << draft_model_metrics.get_ttst().mean << " ms/token " << std::endl;
+        std::cout << "  TPOT: " << draft_model_metrics.get_tpot().mean << " ± " << draft_model_metrics.get_tpot().std
+                  << " ms/token " << std::endl;
+        std::cout << "  AVG Latency: " << draft_model_metrics.get_latency().mean << " ± "
+                  << draft_model_metrics.get_latency().std << " ms/iteration " << std::endl;
+        std::cout << "  Num generated token: " << draft_model_metrics.get_num_generated_tokens() << " tokens"
+                  << std::endl;
         std::cout << "  Total iteration number: " << draft_model_metrics.raw_metrics.m_durations.size() << std::endl;
     }
     std::cout << std::endl;
 } catch (const std::exception& error) {
     try {
         std::cerr << error.what() << '\n';
-    } catch (const std::ios_base::failure&) {}
+    } catch (const std::ios_base::failure&) {
+    }
     return EXIT_FAILURE;
 } catch (...) {
     try {
         std::cerr << "Non-exception object thrown\n";
-    } catch (const std::ios_base::failure&) {}
+    } catch (const std::ios_base::failure&) {
+    }
     return EXIT_FAILURE;
 }
--- a/samples/cpp/text_generation/speculative_decoding_lm.cpp
+++ b/samples/cpp/text_generation/speculative_decoding_lm.cpp
@@ -11,6 +11,19 @@ int main(int argc, char* argv[]) try {
         throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> <DRAFT_MODEL_DIR> '<PROMPT>'");
     }
 
+    ov::genai::GenerationConfig config;
+    config.max_new_tokens = 100;
+    // Speculative decoding generation parameters like `num_assistant_tokens` and `assistant_confidence_threshold` are mutually excluded.
+    // Add parameter to enable speculative decoding to generate `num_assistant_tokens` candidates by draft_model per iteration.
+    // NOTE: ContinuousBatching backend uses `num_assistant_tokens` as is. Stateful backend uses `num_assistant_tokens`'s copy as initial
+    // value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both
+    // backends.
+    config.num_assistant_tokens = 4;
+    // Add parameter to enable speculative decoding to generate candidates by draft_model while candidate probability is higher than
+    // `assistant_confidence_threshold`.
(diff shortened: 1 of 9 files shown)


samples/cpp/text_generation/speculative_decoding_lm.cpp
```diff
@@ -34,10 +21,52 @@ int main(int argc, char* argv[]) try {
...
+        config.branching_factor = 4;
+        config.tree_depth = 2;
+        config.num_assistant_tokens = 7;
+
+        // NPU requires static tensor shapes at model compilation time, so compile-time upper bounds
+        // for tree search dimensions must be specified. Runtime values must not exceed these limits.
+        // On CPU/GPU these properties are ignored and can be omitted.
+        if (draft_device == "NPU") {
+            draft_properties = {{"MAX_TREE_DEPTH", config.tree_depth},
+                                {"MAX_BRANCHING_FACTOR", config.branching_factor},
```
songbell: cannot reuse the generation config inside? it's not convenient user has to set these params twice

tests/python_tests/test_stateful_speculative_decoding.py
```diff
@@ -232,15 +232,34 @@ def test_eagle3_string_inputs(target_model, target_device, draft_model, draft_de
...
 
     # Run reference HF model:
-    ov_generation_config = ov_genai.GenerationConfig(max_new_tokens=30, do_sample=False)
+    ov_generation_config = ov_genai.GenerationConfig(
+        max_new_tokens=30,
+        do_sample=False,
+        branching_factor=branching_factor,
+        tree_depth=tree_depth,
+        num_assistant_tokens=num_assistant_tokens,
+    )
```
songbell: generally maybe better to parameterize the tree search params would be better? to cover: 1) default 2) top-1 3) some topk configs

src/cpp/src/speculative_decoding/eagle3_model_transforms.cpp
```diff
@@ -298,6 +306,70 @@ ov::Tensor slice_hidden_state_for_last_token(const ov::Tensor& hidden_features)
...
+
+    OPENVINO_ASSERT(attention_mask_source_node != nullptr,
+                    "Eagle3AttentionMaskTransform: Could not find attention_mask input in SDPA nodes");
+
+    // Get the shape and element type from the attention mask being used by SDPA
+    auto attention_mask_shape = attention_mask_source_node->get_output_partial_shape(0);
+    auto attention_mask_type = attention_mask_source_node->get_output_element_type(0);
+
+    // 3. Create a new parameter for eagle_tree_mask with the same shape and type
+    auto eagle_tree_mask_param = std::make_shared<ov::op::v0::Parameter>(attention_mask_type, attention_mask_shape);
```
songbell, line 479: what's the shape of attention mask shape?is it a full mask? if yes, seems eagle does not need full mask?

samples/cpp/text_generation/speculative_decoding_lm.cpp
GuoliangShiIntel: I agree the duplication is not ideal. However, there's a fundamental constraint: NPU requires compile time shape limits determined at `LLMPipeline` construction, while `GenerationConfig` is only passed at `generate()` time.

if we accept a `GenerationConfig` object in `draft_model()` — e.g. `draft_model(path, device, generation_config=config)`. Cleanest UX but requires API addition.

Since this behavior is only required for NPU, I don’t think it’s appropriate to modify the `draft_model()` API to accept `GenerationConfig`. Additionally, the draft_properties set here are not entirely the same as those used during generate(). In fact, they should be considered as greater than or equal to the requirements at generation time.

I’m not sure if you have a better idea, or we could leave it to the GenAI team to decide.

tests/python_tests/test_stateful_speculative_decoding.py
GuoliangShiIntel: good proposal, done, thanks

src/cpp/src/speculative_decoding/eagle3_model_transforms.cpp
GuoliangShiIntel: The attention_mask_shape captured here is the dynamic partial shape from the model's internal SDPA mask node (typically {1, 1, ?, ?} with symbolic/dynamic dimensions), not a fixed full-size tensor.

At runtime, the actual `eagle_tree_mask` tensor size depends on the inference phase:

1. Prefill: {1, 1, 1, 1} with value 0.0 — trivial, no overhead. Prefill submodel has built-in causal masking, so the tree mask is a no-op.
2. Generate (tree iteration): {1, 1, num_tokens, attn_len} where num_tokens = num_seqs × branch_len and attn_len = history + num_tokens (e.g. ~20). This is the only case that carries meaningful data — each row encodes which KV positions that token may attend to (blocking cross-branch attention within the flat batch).

So Eagle does not copy or fill a full kvcache_size × kvcache_size mask — it only constructs the small actual-size tensor needed for the current step. The Parameter's dynamic shape simply ensures type/rank compatibility with the SDPA node it feeds into.

Regarding Eagle, I think this stateful pipeline does not support it.

Hope this picture can help:
image

AsyaPronina: Hi @GuoliangShiIntel ! I have a question: the comment here: https://github.com/openvinotoolkit/openvino/pull/34349#discussion_r3070715836 says that we don't need past tokens for `eagle_tree_mask`, as all padded tokens will be just zeroed from the left of `eagle_tree_mask` that exists only for current tokens. So, the question: why do we use the `eagle_tree_mask` of shape `[1, 1, total_tokens, atten_len]`, if we don't need history from the left to apply Eagle3 tree draft specifics?
In both examples, we see that tokens from history are all zeroed: 

Image

Image

And, further more, they don't influence the resulting mask from `Add(causal, eagle_tree_mask)`, as we do padding from `eagle_tree_mask` to `padded_tree_mask`, where all inputs from the left of lower triangle should be zeroed. Then, causal mask does the actual work: it pricks out padding tokens.
Might be we need to pass only squared `eagle_tree_mask` of `total_tokens x total_tokens` and let NPUW to pad mask itself?

P.S. dynamic version of this mask can be left as is. For non-NPU pipelines, the full Eagle3 tree mask could be sent.

GuoliangShiIntel: Hi @AsyaPronina great proposal here; you are completely right.

The history columns in `eagle_tree_mask` are always `0.0` and redundant. NPUW's `pad_tree_mask_input()` already left-pads with zeros, achieving the same effect.

I have changed all related input builders to produce square eagle tree masks: [commit](https://github.com/openvinotoolkit/openvino.genai/pull/3396/commits/9514ab96206a7d5fad7bd68e1c25685094c4448d)

songbell (approved):
LGTM for transform part

src/cpp/src/speculative_decoding/eagle3_model_transforms.cpp
AsyaPronina: Thanks a lot, Guoliang!

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -5,47 +5,347 @@
...
+    const auto& shape = hidden_features.get_shape();
+    OPENVINO_ASSERT(shape.size() == 3 && shape[0] == 1 && shape[1] > 0, "Expected shape [1, seq_len, hidden_size]");
+    OPENVINO_ASSERT(position < shape[1], "Position ", position, " out of bounds for seq_len ", shape[1]);
+    auto [start_coord, end_coord] = ov::genai::utils::make_roi(shape, 1, position, position + 1);
+    return ov::Tensor(hidden_features, start_coord, end_coord);
+}
+
+/// @brief Copies one hidden-state row (one token) from src tensor at src_position into
+///        dst tensor at dst_position.  Both tensors must have shape [1, *, hidden_size].
+void copy_hidden_state_row(const ov::Tensor& src, size_t src_position, ov::Tensor& dst, size_t dst_position) {
```
AsyaPronina, line 42: I am not sure that it won't be redundant, but might be we want to protect `src.get_shape()[1]` and `dst.get_shape()[1]` accesses here with asserts?

GuoliangShiIntel:
Hi @AsyaPronina FYI, Based on yesterday’s version, I have uploaded a new commit related to the model input/output precision types.
In the previous version, `hidden_states, eagle_tree_mask, and logics` were hardcoded to FP32 precision, which is not a general approach.
In the current version, the precision is kept consistent with the compiled model. The changes are included in this [commit](https://github.com/openvinotoolkit/openvino.genai/pull/3396/changes/7d72b4fb8c407fdbbbf0fd1d1dd35312e5bf2155).
Sorry for the inconvenience.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -5,47 +5,412 @@
...
+                    "dst_position ",
+                    dst_position,
+                    " out of bounds for seq_len ",
+                    dst.get_shape()[1]);
+    OPENVINO_ASSERT(src.get_element_type() == dst.get_element_type(),
+                    "copy_hidden_state_row: src type ",
+                    src.get_element_type(),
+                    " != dst type ",
+                    dst.get_element_type());
+    const size_t hidden_size = src.get_shape()[2];
```
AsyaPronina: And also `src.get_shape()[2]` here?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -5,47 +5,412 @@
...
+    // context_len: prompt + previously accepted tokens already in the KV cache.
+    const size_t past_accepted_len = generated_ids.size() - num_candidates;
+    const size_t context_len = prompt_len + past_accepted_len;
+    const size_t attn_len = context_len + num_candidates;
+
+    InputTensors result;
+
+    // 1. input_ids: all N+1 candidates — root then tree nodes.
+    result.input_ids = ov::Tensor(ov::element::i64, {1, num_candidates});
+    const size_t gen_offset = generated_ids.size() - num_candidates;
```
AsyaPronina: It seems to be calculated as `past_accepted_len` before

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.hpp
```diff
@@ -135,106 +208,275 @@ class Eagle3InferWrapperBase {
...
+
+/// Compile-time shape constraints for NPU static allocation. Immutable after construction.
+/// These values represent upper bounds; runtime GenerationConfig values must not exceed them.
+struct Eagle3CompileConfig {
+    size_t max_tree_depth;
+    size_t max_branching_factor;
+    size_t max_assistant_tokens;
+
+    /// Maximum tokens the draft model processes in one generation step.
+    size_t draft_max_gen_tokens() const {
```
AsyaPronina, line 386: Is it `gen` or `input` tokens here?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -124,170 +544,190 @@ void Eagle3InferWrapperBase::release_memory() {
...
-        //   - Therefore: result_count = new_generated_len - prev_generated_len + num_tokens_to_validate
-        const size_t result_count = new_generated_len - prev_generated_len + num_tokens_to_validate;
-        std::vector<int64_t> result_tokens(generated_ids.end() - result_count, generated_ids.end());
+    sequence_group->schedule_tokens(input_token_count);
+    sequence_group->set_output_seq_len(num_candidates);
+    sequence_group->set_num_validated_tokens(num_tokens_to_validate);
 
-        record_generated_tokens(result_tokens.size());
+    m_sampler.sample({sequence_group}, logits, /*is_validation=*/true);
+    sequence_group->finish_iteration();
```
AsyaPronina: Could I please ask: why not `invoke_sampler()` reused here?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -124,170 +544,190 @@ void Eagle3InferWrapperBase::release_memory() {
...
 
 ov::Tensor Eagle3InferWrapperBase::get_logits() const {
     return m_request.get_tensor("logits");
 }
 
-ov::Tensor Eagle3InferWrapperBase::get_hidden_features() const {
-    auto hidden_state = m_request.get_tensor("last_hidden_state");
-    const auto shape = hidden_state.get_shape();
-    OPENVINO_ASSERT(shape.size() == 3 && shape[0] == 1, "Invalid hidden state shape");
+ov::Tensor Eagle3InferWrapperBase::get_hidden_features(size_t actual_seq_len) const {
```
AsyaPronina, line 670: _Nitpick: I think that given method can reuse the method below_

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +739,467 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
                     "Expected single sequence after initialization, got ",
                     m_sequence_group->num_total_seqs());
 }
 
-InferenceOutput Eagle3TargetWrapper::infer(const ov::Tensor& input_ids,
-                                           const ov::Tensor& attention_mask,
-                                           const ov::Tensor& position_ids) {
-    const size_t prompt_len = input_ids.get_shape()[1];
+InferenceOutput Eagle3TargetWrapper::infer(const InputTensors& inputs) {
+    const size_t input_len = inputs.input_ids.get_shape()[1];
```
AsyaPronina: Should we protect access with `.at(1)` here?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +739,467 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
-        OPENVINO_ASSERT(prompt_len <= m_max_prompt_len,
-                        "NPU prompt length ",
-                        prompt_len,
-                        " exceeds max ",
-                        m_max_prompt_len);
-    }
+    m_request.set_tensor("input_ids", inputs.input_ids);
+    m_request.set_tensor("attention_mask", inputs.attention_mask);
+    m_request.set_tensor("position_ids", inputs.position_ids);
+    m_request.set_tensor("eagle_tree_mask", inputs.eagle_tree_mask);
```
AsyaPronina, line 748: This input is necessary now, right?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +739,467 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
+    auto output = infer(inputs);
 
-    // 3. Sample (use sample_count for number of positions to sample from)
-    auto sampled = sample_tokens(output.logits, ctx.input_token_count, ctx.sample_count, ctx.num_tokens_to_validate);
+    // Trim padded outputs: prefill uses last 1 position, validation uses all candidates.
+    const size_t num_candidates = is_validation ? inputs.input_ids.get_shape()[1] : 1;
+    output.logits = trim_tensor_tail(output.logits, num_candidates);
+
+    const std::vector<int64_t> sampled =
+        sample_and_validate(output.logits, ctx.input_token_count, num_candidates, ctx.num_tokens_to_validate);
```
AsyaPronina, line 772: Is this method supposed to be called for validation phase only? I see ` m_sampler.sample({sequence_group}, logits, /*is_validation=*/true);` line in `sample_and_validate()`. Should we use condition to call either `sample_and_validate()` or `sample_next_tokens()`?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +739,467 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
+    OPENVINO_ASSERT(shape.size() == 2 && shape[0] == 1, "Expected input_ids shape [1, seq_len], got ", shape);
 
     const int64_t* ids_data = input_ids.data<const int64_t>();
     const size_t total_len = shape[1];
     OPENVINO_ASSERT(total_len >= 2, "Draft model requires at least 2 tokens");
 
-    // Draft model uses tokens[1:] (Eagle3 specific behavior)
-    TokenIds draft_prompt_ids(ids_data + 1, ids_data + total_len);
+    // Draft model uses tokens[1:] — Eagle3 draft skips the first prompt token because
+    // the target model's hidden state for that position is not available at draft init.
```
AsyaPronina: It seems to be not true...
Eagle3 draft skips the first prompt token, because it uses features concatenated with tokens advanced by one timestep:

(From Eagle-2, that is why only top-level features are described here):
Image

Image

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +739,467 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
-    m_request.set_tensor("position_ids", position_ids);
+    m_buffers_allocated = true;
+}
 
-    // Set hidden states (either from target model or internal)
-    OPENVINO_ASSERT(hidden_states && hidden_states.get_size() > 0, "hidden_states must be provided");
-    auto shape = hidden_states.get_shape();
-    OPENVINO_ASSERT(shape.size() == 3, "Invalid hidden states shape");
+InferenceOutput Eagle3DraftWrapper::infer(const InputTensors& inputs, const ov::Tensor& hidden_states) {
+    const size_t input_token_count = inputs.input_ids.get_shape()[1];
```
AsyaPronina: Also suggesting to protect with `.at(1)`

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +739,467 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
+        OPENVINO_ASSERT(seq_hidden && seq_hidden.get_size() > 0, "Sequence ", i, " contains invalid hidden state");
+        const auto& h_shape = seq_hidden.get_shape();
+        OPENVINO_ASSERT(h_shape.size() == 3 && h_shape[0] == 1,
+                        "Expected hidden state shape [1, branch_len, hidden_size], got: ",
+                        h_shape);
+        total_seq_len += h_shape[1];
+    }
+
+    // Use pre-allocated buffer if available and large enough, otherwise allocate.
+    const size_t hidden_size = running_sequences[0]->get_hidden_state().get_shape()[2];
```
AsyaPronina: Also suggest to use `.at()` at every places like this one.

AsyaPronina (approved):
Hi @GuoliangShiIntel !
Thanks a lot for such great code!

I have left a bit of comments: mostly on guards for out-of-range accesses.
LGTM! Approved!

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: make sense, done

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: done

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: good catch, fixed

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.hpp
GuoliangShiIntel: these are input token counts per `generate()` call (upper bound for the padded static shape that NPUW allocates). The value feeds `NPUW_LLM_MAX_GENERATION_TOKEN_LEN`. Updated the doc comments to make that explicit. Kept the `_max_gen_tokens` naming so it lines up with NPUW's `MAX_GENERATION_TOKEN_LEN` property.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: good proposal, reuse it

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: good catch, done

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: done

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: Yes. `eagle_tree_mask` is a required input on both the draft and target models: `apply_eagle3_attention_mask_transform()` rewrites the IR to add this Parameter and injects an Add node before every SDPA. Skipping the set_tensor here would leave the port unbound and `infer()` would fail.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: good question, `sample_and_validate` covers both prefill (num_tokens_to_validate=0) and validation (num_tokens_to_validate>0) for the target model, and both paths need the sampler's is_validation=true branch to avoid target side sequence forks. 
To keep this constraint explicit, `invoke_sample`r now takes an explicit is_validation flag: sample_and_validate calls it with true, sample_next_tokens (draft) calls it with false.
Splitting the target path by phase would still route through the same helper with the same flag, so there is no behavioural benefit.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: good catch, fixed it

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: done

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: done, thanks

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -299,200 +711,473 @@ void Eagle3InferWrapperBase::record_generated_tokens(size_t actual_generated_cou
...
+    // DRAFT_ITERATION: output.hidden_features has shape {1, total_tokens, hidden_size}
+    // where total_tokens = num_seqs * branch_len (flat layout, same order as input_ids).
+    // Hidden features are already trimmed to input_token_count by get_hidden_features()
+    // in infer(), so hidden_shape[1] == total_hidden_tokens (no pad offset needed).
+    // Append each sequence's LAST branch token hidden state to its accumulated buffer,
+    // growing from {1, k, H} to {1, k+1, H}.
+    const auto& hidden_shape = output.hidden_features.get_shape();
+    OPENVINO_ASSERT(hidden_shape.size() == 3 && hidden_shape[0] == 1, "Invalid hidden features shape: ", hidden_shape);
+
+    const size_t branch_len = running_sequences[0]->get_hidden_state().get_shape().at(1);
```
AsyaPronina: Final minor comment from me! `at(0)`

GuoliangShiIntel:
@AsyaPronina Thanks for your thoughtful comments. I have addressed almost all of them, and for the items that were not changed, I have left explanations in the review.

I also verified the pipeline using the latest commit, and the dataset level results show the expected behavior.

sbalandi (approved):

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: done

PR #3400, opened 2026-02-26
Change input shape for 3D position_ids of Qwen 2.5 VL with M-RoPE

CuriousPanCake:
The analysis has shown that the correct shape for the 3D position_ids tensor of Qwen 2.5 VL in the ContinuousBatching mode is not flattening, but [3, total_token_num]. It allows to preserve the correct 3D semantics for M-RoPE of the model. Instead of Reshaping on the transformation side, it is correct and logical to provide proper tensor from the GenAI side.

- Ticket: [CVS-167316](https://jira.devtools.intel.com/browse/CVS-167316)

Signed-off-by: Andrii Staikov <andrii.staikov@intel.com>

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -620,7 +620,7 @@ jobs:
             timeout: 90
           - name: 'WWB tests (nanollava)'
             cmd: |
-              python -m pip install transformers==4.48.0
+              python -m pip install transformers==4.48.0 diffusers==0.35.2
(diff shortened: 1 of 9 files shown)


CuriousPanCake:
Requires: https://github.com/openvinotoolkit/openvino/pull/34365

CuriousPanCake:
The tests will pass once we merge the runtime PR: https://github.com/openvinotoolkit/openvino/pull/34365

PR #3423, opened 2026-03-02
Whisper: WA when KVcahce is full on NPUW

eshiryae:
## Description
When model starts to repeat itself, the same token is generated and KVcache is overloaded when running on NPU(W). Not to crash in this case, force return eos_token and continue processing next chunk. It's a WA solution. 
Depends on https://github.com/openvinotoolkit/openvino/pull/34175 

E-200274

## Checklist:


src/cpp/src/whisper/pipeline.cpp
```diff
@@ -86,6 +86,10 @@ class WhisperPipeline::WhisperPipelineStatefulImpl : public WhisperPipeline::Whi
             // NB: only batch_size == 1 is supported now for NPU
             reshape_to_static_encoder(encoder_model, 1, m_feature_extractor.feature_size);
             compiled_model = core.compile_model(encoder_model, "NPU", properties_copy);
+
+            // WA: Push eos_token to NPUW to stop infer() if KVcache is full
+            auto eos_token = m_generation_config.eos_token_id == -1 ? m_generation_config.eos_token_id : m_tokenizer.get_eos_token_id();
```
as-suvorov: Please double check ternary expression. Looks like intention was oposite:

```suggestion
            auto eos_token = m_generation_config.eos_token_id == -1 ? m_tokenizer.get_eos_token_id(): m_generation_config.eos_token_id;
```

eshiryae: Thanks, fixed!

PR #3451, opened 2026-03-06
[SD] Dynamic Tree Search Sampler for draft generation and main validation

GuoliangShiIntel:
## Description
The EAGLE tree search sampler implements speculative decoding via a tree-structured draft, where each draft round:

1. Runs the draft model for `tree_depth` steps, expanding a tree of candidate token sequences.
2. Selects the top `num_assistant_tokens` nodes from the tree.
3. Passes the selected tokens to the target model in a single batched forward pass.
4. Validates the candidates and accepts as many as possible.

There is an example:
image

CVS-182483

## Checklist:


src/cpp/src/generation_config.cpp
```diff
@@ -158,6 +158,9 @@ void GenerationConfig::update_generation_config(const ov::AnyMap& properties) {
     // CDPruner
     read_anymap_param(properties, "pruning_ratio", pruning_ratio);
     read_anymap_param(properties, "relevance_weight", relevance_weight);
+
+    // EAGLE tree search
+    read_anymap_param(properties, "eagle_tree_params", eagle_tree_params);
```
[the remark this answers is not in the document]
GuoliangShiIntel: Fix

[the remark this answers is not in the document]
GuoliangShiIntel: done

src/cpp/src/generation_config.cpp
```diff
@@ -251,18 +251,21 @@ size_t GenerationConfig::get_max_new_tokens(size_t prompt_length) const {
...
-    return !do_sample && !is_beam_search();
+    return !do_sample && !is_beam_search() && !is_tree_search();
 }
 
 bool GenerationConfig::is_beam_search() const {
     return num_beams > 1;
 }
 
+bool GenerationConfig::is_tree_search() const {
+    return eagle_tree_params.tree_depth > 0;
```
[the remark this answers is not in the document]
GuoliangShiIntel: ASSERT implement at `GenerationConfig::validate()` function, no need here

src/cpp/src/generation_config.cpp
```diff
@@ -346,6 +352,23 @@ void GenerationConfig::validate() const {
...
+            eagle_tree_params.total_tokens > 1,
+            "'total_tokens' must be > 1 when tree search is enabled (root + at least one draft node), but got ",
+            eagle_tree_params.total_tokens);
+        OPENVINO_ASSERT(eagle_tree_params.total_tokens > eagle_tree_params.tree_depth,
+                        "'total_tokens' (",
+                        eagle_tree_params.total_tokens,
+                        ") must be greater than 'tree_depth' (",
+                        eagle_tree_params.tree_depth,
+                        ") to allow at least one branching node");
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: No change needed here.

src/cpp/src/sequence_group.hpp
```diff
@@ -125,6 +138,10 @@ class Sequence {
         return m_status == SequenceStatus::WAITING;
     }
 
+    bool is_suspended() const {
+        return m_status == SequenceStatus::SUSPENDED;
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: Will be used by the pipeline

src/python/py_generation_config.cpp
```diff
@@ -412,6 +412,25 @@ void init_generation_config(py::module_& m) {
...
+                       "How deep to look ahead in the EAGLE tree")
+        .def_readwrite("total_tokens",
+                       &GenerationConfig::EagleParams::total_tokens,
+                       "Number of nodes selected from the built EAGLE tree after reranking, "
+                       "used as the final set of draft token candidates submitted to the target model for verification")
+        .def("__repr__", [](const GenerationConfig::EagleParams& self) {
+            return "EagleParams(branching_factor=" + std::to_string(self.branching_factor) +
+                   ", tree_depth=" + std::to_string(self.tree_depth) +
+                   ", total_tokens=" + std::to_string(self.total_tokens) + ")";
+        });
```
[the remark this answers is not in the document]
GuoliangShiIntel: Fixed

src/python/py_generation_config.cpp
```diff
@@ -444,6 +463,7 @@ void init_generation_config(py::module_& m) {
         .def_readwrite("assistant_confidence_threshold", &GenerationConfig::assistant_confidence_threshold)
         .def_readwrite("num_assistant_tokens", &GenerationConfig::num_assistant_tokens)
         .def_readwrite("max_ngram_size", &GenerationConfig::max_ngram_size)
+        .def_readwrite("eagle_tree_params", &GenerationConfig::eagle_tree_params, "EAGLE tree parameters for speculative decoding")
```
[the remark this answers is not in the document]
GuoliangShiIntel: Fixed

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,400 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+        return a.score > b.score;
+    };
+    std::priority_queue<Node, std::vector<Node>, decltype(cmp_min)> min_heap(cmp_min);
+
+    for (const auto& [id, node] : m_nodes) {
+        if (id == 0u)
+            continue;
+        if (min_heap.size() < static_cast<size_t>(m_max_candidate_nodes)) {
+            min_heap.push(node.data);
+        } else if (node.data.score > min_heap.top().score) {
```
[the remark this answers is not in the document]
GuoliangShiIntel: Fixed

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +488,376 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+    OPENVINO_ASSERT(m_sequence_group->num_running_seqs() == 1,
+                    "TreeSearcher must be initialized with exactly one running sequence");
+    // m_original_grouped_id is set authoritatively in tree_reset() at the start of each draft round.
+}
+
+void Sampler::TreeSearcher::tree_reset() {
+    m_candidate_graph.emplace(-1,
+                              0.0f,
+                              m_parameters.eagle_tree_params.total_tokens - 1,
+                              m_parameters.eagle_tree_params.tree_depth);
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +488,395 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
```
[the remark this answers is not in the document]
GuoliangShiIntel: done

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,400 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+    const size_t branching_factor = m_parameters.eagle_tree_params.branching_factor;
+    const size_t kept = std::min(branching_factor, candidates.size());
+    std::partial_sort(candidates.begin(),
+                      candidates.begin() + kept,
+                      candidates.end(),
+                      [](const CandidateBeam& a, const CandidateBeam& b) {
+                          return a.score > b.score;
+                      });
+    candidates.resize(kept);
+    return candidates;
```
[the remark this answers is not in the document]
GuoliangShiIntel: wrong comment

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,400 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+    // Roll back tokens appended during drafting, then apply the final candidate tokens
+    OPENVINO_ASSERT(sequence->get_generated_len() >= m_pre_draft_generated_len,
+                    "finalize_tree: generated_len (", sequence->get_generated_len(),
+                    ") is less than pre-draft length (", m_pre_draft_generated_len,
+                    "); sequence state is inconsistent");
+    sequence->remove_last_tokens(sequence->get_generated_len() - m_pre_draft_generated_len);
+    for (size_t t = 1; t < final_nodes.size(); ++t) {
+        sequence->append_token(final_nodes[t].token_id, 0.0f);
+        logit_processor.register_new_generated_token(final_nodes[t].token_id);
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +488,395 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+
+    OPENVINO_ASSERT(final_nodes.size() > static_cast<size_t>(m_parameters.eagle_tree_params.tree_depth),
+                    "finalize_tree: fewer candidate nodes than tree_depth; tree parameters may be inconsistent");
+    const size_t extra_processed_tokens = final_nodes.size() - static_cast<size_t>(m_parameters.eagle_tree_params.tree_depth) - 1;
+    m_sequence_group->update_processed_tokens_num(m_sequence_group->get_num_processed_tokens() + extra_processed_tokens);
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
```diff
@@ -681,22 +1051,151 @@ align_all_sequence_len(SequenceGroup::Ptr& sequence_group,
...
 
+// Validate tree results from the target model using retrieve_indices and logits
+// Returns the number of valid tokens, and truncates sequence if mismatch is found
+size_t Sampler::validate_tree_candidates(Sequence::Ptr& sequence,
+                                         const ov::Tensor& sequence_group_logits,
+                                         LogitProcessor& logit_processor,
+                                         size_t num_tokens_to_validate) {
+    auto& eagle_metadata = sequence->get_eagle_metadata();
+    auto retrieve_indices = eagle_metadata.retrieve_indices;
+
```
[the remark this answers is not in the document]
GuoliangShiIntel: no change need

src/cpp/src/sampling/sampler.cpp
```diff
@@ -681,6 +1049,135 @@ align_all_sequence_len(SequenceGroup::Ptr& sequence_group,
...
+    if (retrieve_indices.empty()) {
+        return 0;
+    }
+
+    // reconstruct the candidate lists from retrieve_indices
+    auto generated_len = sequence->get_generated_len();
+    auto generated_ids = sequence->get_generated_ids();
+
```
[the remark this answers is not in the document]
GuoliangShiIntel: no change need

src/cpp/src/sampling/sampler.cpp
```diff
@@ -681,22 +1073,163 @@ align_all_sequence_len(SequenceGroup::Ptr& sequence_group,
...
+    }
+
+    // remove tokens from sequence
+    sequence->remove_last_tokens(num_tokens_to_validate);
+
+    // append validated tokens back to sequence
+    for (size_t i = 1; i < validated_steps; ++i) {
+        sequence->append_token(candidate_path[i], 0.0f);
+        logit_processor.register_new_generated_token(candidate_path[i]);
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: no change need

src/cpp/src/sampling/sampler.cpp
```diff
@@ -921,8 +1467,49 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
...
+            TreeSearcher* tree_searcher;
+            {
+                std::lock_guard<std::mutex> lock(m_tree_search_info_mutex);
+                if (m_tree_search_info.find(request_id) == m_tree_search_info.end()) {
+                    m_tree_search_info.emplace(
+                        request_id,
+                        TreeSearcher(sequence_group, m_d2t_mapping ? m_d2t_mapping->get_tensor_view() : ov::Tensor()));
+                }
+                tree_searcher = &m_tree_search_info.at(request_id);
+            }
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sequence_group.hpp
```diff
@@ -760,6 +785,10 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
         return m_sampling_params;
     }
 
+    void set_sampling_parameters(const ov::genai::GenerationConfig& params) {
+        m_sampling_params = params;
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: Will be used in pipeline, so keep it

src/cpp/src/sampling/sampler.cpp
```diff
@@ -921,6 +1512,48 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
...
+        uint64_t request_id = sequence_group->get_request_id();
+
+        if (is_validation_mode_enabled && num_generated_tokens_to_validate > 0) {
+            // Target-model validation pass: validate the draft tree candidates.
+            std::vector<Sequence::Ptr> running_sequences = sequence_group->get_running_sequences();
+            OPENVINO_ASSERT(running_sequences.size() == 1);
+            size_t valid_tokens = verify_draft_tree(running_sequences[0],
+                                                    sequence_group_logits,
+                                                    logit_processor,
+                                                    num_generated_tokens_to_validate);
```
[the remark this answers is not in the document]
GuoliangShiIntel: Eagle pipeline required, will add in the following PR

src/cpp/src/generation_config.cpp
```diff
@@ -346,6 +360,23 @@ void GenerationConfig::validate() const {
         // OPENVINO_ASSERT(length_penalty == 1.0f, "'length_penalty' is set to ", length_penalty, " (default is 1.0f), which is supported only by beam search sampling");
     }
 
+    // tree search (EAGLE)
+    if (is_tree_search()) {
+        OPENVINO_ASSERT(eagle_tree_params.branching_factor > 0,
+                        "'branching_factor' must be > 0 when tree search is enabled, but got ",
+                        eagle_tree_params.branching_factor);
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
```diff
@@ -681,22 +1109,185 @@ align_all_sequence_len(SequenceGroup::Ptr& sequence_group,
...
+    OPENVINO_ASSERT(num_tokens_to_validate > 0,
+                    "verify_draft_tree: num_tokens_to_validate must be > 0, got ",
+                    num_tokens_to_validate);
+
+    auto& eagle_metadata = sequence->get_eagle_metadata();
+    auto retrieve_indices = eagle_metadata.retrieve_indices;
+
+    if (retrieve_indices.empty()) {
+        return 0;
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,436 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+    // m_original_grouped_id is set authoritatively in tree_reset() at the start of each draft round.
+}
+
+void Sampler::TreeSearcher::tree_reset() {
+    const size_t total_tokens = m_parameters.eagle_tree_params.total_tokens;
+    OPENVINO_ASSERT(total_tokens > 1,
+                    "eagle_tree_params.total_tokens must be greater than 1, got ", total_tokens);
+    const int max_non_root_candidate_nodes = static_cast<int>(total_tokens - 1);
+    m_candidate_graph.emplace(-1, 0.0f, max_non_root_candidate_nodes, m_parameters.eagle_tree_params.tree_depth);
+
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -621,8 +621,14 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
  *        copy as initial value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both backends.
  * @param max_ngram_size is maximum ngram to use when looking for matches in the prompt.
  *
+ * EAGLE tree search parameters:
+ * @param eagle_tree_params.branching_factor number of top-k candidates expanded at each tree node (branching factor).
+ * @param eagle_tree_params.tree_depth lookahead depth of the EAGLE tree; the draft model runs `tree_depth` iterations.
+ * @param eagle_tree_params.total_tokens number of nodes selected from the built EAGLE tree after reranking,
+ *        used as the final set of draft token candidates submitted to the target model for verification.
```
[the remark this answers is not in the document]
GuoliangShiIntel: No change needs

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -621,8 +621,14 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
  *        copy as initial value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both backends.
  * @param max_ngram_size is maximum ngram to use when looking for matches in the prompt.
  *
+ * EAGLE tree search parameters:
+ * @param eagle_tree_params.branching_factor number of top-k candidates expanded at each tree node (branching factor).
+ * @param eagle_tree_params.tree_depth lookahead depth of the EAGLE tree; the draft model runs `tree_depth` iterations.
+ * @param eagle_tree_params.total_tokens number of nodes selected from the built EAGLE tree after reranking,
```
songbell: we have decided to reuse num_speculative_tokens instead of total_tokens to keep interface clean, cc @wangleis

GuoliangShiIntel: I have updated it

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -621,8 +621,14 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
  *        copy as initial value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both backends.
  * @param max_ngram_size is maximum ngram to use when looking for matches in the prompt.
  *
+ * EAGLE tree search parameters:
+ * @param eagle_tree_params.branching_factor number of top-k candidates expanded at each tree node (branching factor).
+ * @param eagle_tree_params.tree_depth lookahead depth of the EAGLE tree; the draft model runs `tree_depth` iterations.
```
songbell: should be tree_depth + 1 iterations?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -621,8 +621,14 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
...
+ * @param eagle_tree_params.num_speculative_tokens number of draft (non-root) tokens from the EAGLE tree submitted
+ *        to the target model for verification. Total tree nodes = `num_speculative_tokens + 1` (including root).
```
songbell: we should have aligned to reuse the num_assistant_token for this param, and reuse its default value as well. @wangleis , please advice if not.

songbell:
```suggestion
 *        to the target model for verification. Total tree nodes for validation = `num_speculative_tokens + 1` (including root).
```

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -675,6 +681,19 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
     size_t num_assistant_tokens = 0;
     size_t max_ngram_size = 0;
 
+    /**
+     * @brief Parameters for EAGLE3 speculative decoding tree search
+     */
+    struct EagleParams {
```
songbell: Maybe use TreeParams? to less bounded to Eagle

src/cpp/src/generation_config.cpp
```diff
@@ -251,18 +262,21 @@ size_t GenerationConfig::get_max_new_tokens(size_t prompt_length) const {
...
-    return !do_sample && !is_beam_search();
+    return !do_sample && !is_beam_search() && !is_tree_search();
 }
 
 bool GenerationConfig::is_beam_search() const {
     return num_beams > 1;
 }
 
+bool GenerationConfig::is_tree_search() const {
+    return eagle_tree_params.tree_depth > 0;
```
songbell: to define the tree search, should have the branching factor as well
```suggestion
    return return eagle_tree_params.tree_depth > 0 && eagle_tree_params.branching_factor > 1;
```

src/cpp/src/generation_config.cpp
```diff
@@ -346,6 +360,29 @@ void GenerationConfig::validate() const {
...
 
+    // tree search (EAGLE)
+    if (is_tree_search()) {
+        OPENVINO_ASSERT(!do_sample,
+                        "Tree search (EAGLE) is incompatible with do_sample=true; "
+                        "set eagle_tree_params.tree_depth=0 or do_sample=false");
+        OPENVINO_ASSERT(num_beams == 1,
+                        "Tree search (EAGLE) is incompatible with beam search; "
+                        "set eagle_tree_params.tree_depth=0 or num_beams=1");
+        OPENVINO_ASSERT(eagle_tree_params.branching_factor > 0,
```
songbell:
```suggestion
        OPENVINO_ASSERT(eagle_tree_params.branching_factor > 1,
```
I think = 1 does not need the tree sample and tree validation right?

src/cpp/src/sequence_group.hpp
```diff
@@ -30,6 +34,14 @@ enum class SequenceGroupType {
 
 using TokenIds = std::vector<int64_t>;
 using LogProbs = std::vector<float>;
+
+struct EagleMetaData {
```
songbell:
```suggestion
struct TreeMetaData {
```

src/cpp/src/sampling/sampler.hpp
```diff
@@ -89,6 +150,9 @@ class Sampler {
     std::map<uint64_t, GroupBeamSearcher> m_beam_search_info;
     std::mutex m_beam_search_info_mutex;
 
+    // request ID => tree search tracking information
+    std::map<uint64_t, std::shared_ptr<TreeSearcher>> m_tree_search_info;
+    std::mutex m_tree_search_info_mutex;
```
songbell: can reuse with m_beam_search_info_mutex? like merge to m_search_info_mutex.

songbell:
btw, I'm trying to merge this to my local impl, will update if changes needed then

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: Let's use this case as an example to clarify this question:
for my perspective, the lookahead depth of this EAGLE tree is 3, and the draft model runs 3 times.
image

src/cpp/src/generation_config.cpp
GuoliangShiIntel: that is right, but i keep 1 for top 1 solution. it also works well. So that we can cover top1 and topk solution with totally same logic.

src/cpp/src/sequence_group.hpp
```diff
@@ -20,7 +20,11 @@ enum class SequenceStatus {
     RUNNING = 0,
     FINISHED = 1,
     OUT_OF_MEMORY = 2,
-    WAITING = 3
+    WAITING = 3,
+    // The sequence is temporarily parked during EAGLE tree drafting: kept alive in the
+    // sequence group so finalize_tree() can locate and restore it, but excluded from
+    // get_running_sequences() to prevent it from being scheduled for inference.
+    SUSPENDED = 4
```
songbell: mark: this part I'll elaborate if I have better solution

src/cpp/src/sampling/sampler.hpp
```diff
@@ -70,7 +75,55 @@ struct SequenceGroupSamplingInfo {
     }
 };
 
+// Tree of draft token candidates for EAGLE speculative decoding.
+// Stores token IDs and cumulative scores without Sequence references.
+class EagleCandidateGraph {
+public:
+    struct Node {
+        uint64_t id = 0;
```
songbell: see if we can optimize
1) duplicated Node and InternalNode used, seems confusing
2) avoid id, seems we can avoid using it.  see below to check if you have better optimization
https://github.com/songbell/openvino.genai/blob/bell/clean_up_code/src/cpp/src/sampling/sampler.hpp#L183

src/cpp/src/sampling/sampler.cpp
```diff
@@ -921,6 +1511,48 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
             // stop sequence by max_new_tokens
             beam_searcher->finalize(sg_sampling_info.sampler_output);
         }
+    } else if (sampling_params.is_tree_search()) {
+        uint64_t request_id = sequence_group->get_request_id();
+
+        if (is_validation_mode_enabled && num_generated_tokens_to_validate > 0) {
```
songbell, line 1627: how can main model sample the first token?

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,435 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+
+    // Roll back draft tokens, then re-apply the selected tree nodes (skipping root at index 0).
+    OPENVINO_ASSERT(sequence->get_generated_len() >= m_pre_draft_generated_len,
+                    "finalize_tree: generated_len (", sequence->get_generated_len(),
+                    ") is less than pre-draft length (", m_pre_draft_generated_len,
+                    "); sequence state is inconsistent");
+    sequence->remove_last_tokens(sequence->get_generated_len() - m_pre_draft_generated_len);
+    for (size_t t = 1; t < final_nodes.size(); ++t)
+        sequence->append_token(final_nodes[t].token_id, 0.0f);
+
```
songbell, line 870: need to update information to logic processor as well, otherwise, some other checking in pipeline will report failure

songbell:
some common ideas from me: 
1) better not to have eagle specific naming in other components, for example, sampling code, we can make it more generic
2) better avoid "draft" naming in common part, but this is not a must

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: I discussed it with you. For main fist token uses greedy sampler at pipeline (change sampler config and restore back : ))

songbell: where to restore it? can you share your caller?
in addition, in CB pipeline, we can have several requests batched, while some in prefill some in generation, how to control the branch with only 1 sampler config?

GuoliangShiIntel: Here is [code](https://github.com/GuoliangShiIntel/openvino.genai/blob/sgl/eagle_3_topk_npu_support_v3/src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp#L653-L662). It is not a good solution, any suggestion here?

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: Done

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: agree, done

src/cpp/src/generation_config.cpp
GuoliangShiIntel: I’ll keep it unchanged, because I think the top‑1 solution should also go through the tree search. A branching_factor of 1 still means there is only one path, but it’s still a specific tree.

src/cpp/src/sequence_group.hpp
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.hpp
GuoliangShiIntel: make sense, done

src/cpp/src/sampling/sampler.hpp
GuoliangShiIntel: good proposal, i have updated it.

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: done

GuoliangShiIntel:
@songbell I’ve addressed almost all of the comments, and for the ones I didn’t change, I left explanations. Since we made some major updates—for example, merging certain node structures—I haven’t fully verified that everything works correctly at the pipeline level yet. I’ll perform that verification in the NPU pipeline. Thanks for the valuable feedback.

GuoliangShiIntel:
> @songbell I’ve addressed almost all of the comments, and for the ones I didn’t change, I left explanations. Since we made some major updates—for example, merging certain node structures—I haven’t fully verified that everything works correctly at the pipeline level yet. I’ll perform that verification in the NPU pipeline. Thanks for the valuable feedback.

@songbell FYI, it works well with NPU pipeline

src/cpp/src/sampling/sampler.cpp
songbell: not pushed or ?  something like:
https://github.com/songbell/openvino.genai/blob/bell/clean_up_code/src/cpp/src/sampling/sampler.cpp#L739

GuoliangShiIntel: please check here: https://github.com/GuoliangShiIntel/openvino.genai/blob/sgl/eagle_tree_search_sampler/src/cpp/src/sampling/sampler.cpp#L1532

src/cpp/src/sequence_group.hpp
```diff
@@ -787,6 +812,10 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
         return m_sampling_params;
     }
 
+    void set_sampling_parameters(const ov::genai::GenerationConfig& params) {
+        m_sampling_params = params;
+    }
+
```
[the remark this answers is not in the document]
GuoliangShiIntel: will be used by pipeline

src/cpp/src/sampling/sampler.cpp
songbell: not only length, but also register the tokens: logit_processor.register_new_generated_token
other wise it will break some common logic check

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,432 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+    // extra_processed_tokens: tree nodes beyond the deepest path (= total - depth - 1 for root).
+    // These are processed simultaneously by the verifier and must be counted now.
+    OPENVINO_ASSERT(final_nodes.size() > static_cast<size_t>(m_parameters.tree_params.tree_depth),
+                    "finalize_tree: final_nodes.size() (",
+                    final_nodes.size(),
+                    ") must exceed tree_depth (",
+                    m_parameters.tree_params.tree_depth,
+                    ")");
+    const size_t extra_processed_tokens =
+        final_nodes.size() - static_cast<size_t>(m_parameters.tree_params.tree_depth) - 1;
```
songbell:
```suggestion
        final_nodes.size() - static_cast<size_t>(m_parameters.tree_params.tree_depth);
```
my original impl needs to -1 because I will let draft run depth +1 times, but now this impl don't need to do this adjustment.

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: Got it — but let’s double check to be sure.

Actually, I’ve already added this logic in the verify_draft_tree function:
<https://github.com/GuoliangShiIntel/openvino.genai/blob/sgl/eagle_tree_search_sampler/src/cpp/src/sampling/sampler.cpp#L1253>.

From my understanding, only the tokens that are accepted should call register_new_generated_token, correct?

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: I think the logic here is right, let‘s use this case as an example (branching_factor = 2, tree_depth = 4, num_speculative_tokens = 7):

final_nodes.size() = 8 (root + 7 candidates)
The linear‑path verifier processes 4 draft tokens
The tree verifier processes 7 draft tokens
Therefore: extra = 8 - 4 - 1 = 3 (3 additional tokens were processed)

So, i think the "-1" is to remove root token. and for your previous logic, there is no "-1" https://github.com/songbell/openvino.genai/blob/bell/clean_up_code/src/cpp/src/sampling/sampler.cpp#L745-L746

src/cpp/src/sampling/sampler.cpp
songbell: yes, the link you provided above is for main , for draft model, we also have to update the correct status

GuoliangShiIntel: Done, thanks

src/cpp/src/sampling/sampler.cpp
songbell: it does not matter where or how we put the adjusting, the logic is target to ensure we have correct num_processed_tokens been set for draft, because we are using a "fake" draft sequence to record all the selected draft tokens:
branching_factor = 2, tree_depth = 4, num_speculative_tokens = 7, suppose at current step we have previous processed num as 100, then in this iter:
1) final_nodes.size() = 8 (root + 7 candidates)
2) in finalize api, seq_group->get_num_processed_token() only come to 103 (another 1 is after the sampling finished), but actually we are trying to attaching 8 tokens to this draft sequence group, that's why we need to adjust the num_processed_tokens
3) after draft finished, the num_processed_token for draft should reach 100 + 8 = 108
in your code, currently this number will be 103 + (8 - 4 -1) + 1 = 107, this may introduce other assertions in pipeline, because in other logic, we all assume the num_processed_tokens is the marker of where we are at the sequence

GuoliangShiIntel: After discussing this offline, I realized I had a misunderstanding about `num_processed_token`. I’ve updated the logic accordingly. Thanks.

src/cpp/src/sequence_group.hpp
songbell: I revisit this part today, I think it's not a good idea to add a new status here though.
we added this only for draft candidate sequence-reconstruct, previously for fast-draft or top-1, we only have 1 running sequence for a request, now we have multiple beams in draft forwarding, however, it's ok to use whatever sequence in the group during the tree finalization. We can agree on using the first sequence in the queue, what do you think? I have verified locally with a few code changes in the pipeline,  CB can run well, can you have a try?
image

src/cpp/src/sampling/sampler.cpp
songbell: > Here is [code](https://github.com/GuoliangShiIntel/openvino.genai/blob/sgl/eagle_3_topk_npu_support_v3/src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp#L653-L662). It is not a good solution, any suggestion here?
as 
maybe it's ok for non-cb, but in cb, the solution is not applicable, reason as I state above, so an easy impl which can fit for both cases, would be change the main model sampling to greedy (not tree sampling, main is not tree sampling anyway), then use a branch to control when main in validation mode, whether we should do tree validation or normal validation, or only do a prefill sampling, what do you think?

src/cpp/src/sequence_group.hpp
```diff
@@ -30,6 +34,14 @@ enum class SequenceGroupType {
...
 using TokenIds = std::vector<int64_t>;
 using LogProbs = std::vector<float>;
+
+struct TreeMetaData {
+    std::vector<std::vector<uint8_t>> tree_mask;
+    std::vector<std::vector<int64_t>> retrieve_indices;
+    std::vector<int64_t> tree_position_ids;
+    std::vector<int64_t> validated_indices;
+};
+
```
songbell, line 40: nit: 1) for indices , is int64_t is a must? if the member is int64_t, maybe the indice in naming is not a good one
      2) if a must, in line 35, there is a alias as TokenIds, able to re-use?
and for the place of declaration, since the meta data will need to be used in SD pipeline, maybe in utils.hpp is better? not sure...

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,434 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+            else if (logit > min_heap.top().first) {
+                min_heap.pop();
+                min_heap.emplace(logit, i);
+            }
+        }
+
+        // Pass 2: compute log-sum-exp for softmax normalization.
+        float sum_exp = 0.0f;
+        for (size_t i = 0; i < vocab_size; ++i)
+            sum_exp += std::exp(logit_row[i] - max_logit);
```
songbell: I have a finding here, can we leverage the online softmax idea of flash attention to avoid another round of loop here? by adding compensation to the sum_exp during the loop.  we can integrate heap construct, max finding and exp_sum into 1 loop. worth a try?

GuoliangShiIntel: done, thanks

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: Done. The current logic for target model infer; we can treat all with validation mode.
For prefill, tokens_to_validate == 0, 
For validation, tokens_to_validate > 0,
which we can use to decide between greedy and tree-based validation.

src/cpp/src/sequence_group.hpp
GuoliangShiIntel: I’ve removed the SUSPEND status. However, there is an issue in the NPU pipeline: during draft iteration, all running sequences are required for the next inference. To address this, I reused the WAITING status to replace SUSPEND.

src/cpp/src/sequence_group.hpp
GuoliangShiIntel: I have not change code in this commit.

maybe uint8_t is enough? and tree_mask also change to bool?

GuoliangShiIntel: done

songbell:
LGTM, please make CI clean

src/cpp/src/sampling/sampler.cpp
```diff
@@ -921,6 +1515,66 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
...
+            // Target-model validation pass: validate the candidate tree against target logits.
+            std::vector<Sequence::Ptr> running_sequences = sequence_group->get_running_sequences();
+            OPENVINO_ASSERT(running_sequences.size() == 1);
+            size_t valid_tokens = verify_draft_tree(running_sequences[0],
+                                                    sequence_group_logits,
+                                                    logit_processor,
+                                                    num_generated_tokens_to_validate);
+            assisting_pipeline_info.max_removed_tokens_per_request =
+                std::max(assisting_pipeline_info.max_removed_tokens_per_request,
+                         num_generated_tokens_to_validate + 1 - valid_tokens);
```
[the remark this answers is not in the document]
GuoliangShiIntel: false positive

songbell (approved):

songbell:
@rkazants Hi Roman, will you take a look at this?

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,436 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+                    "TreeSearcher must be initialized with exactly one running sequence");
+    // m_original_grouped_id is set authoritatively in tree_reset() at the start of each draft round.
+}
+
+void Sampler::TreeSearcher::tree_reset() {
+    const size_t num_tree_nodes = m_parameters.num_assistant_tokens;
+    OPENVINO_ASSERT(num_tree_nodes > 0,
+                    "num_assistant_tokens must be greater than 0 for tree search, got ", num_tree_nodes);
+    m_candidate_graph.emplace(-1, 0.0f, static_cast<int>(num_tree_nodes), static_cast<int>(m_parameters.tree_params.tree_depth));
+
```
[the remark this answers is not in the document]
GuoliangShiIntel: Typically, `num_assistant_tokens / tree_depth` should not be significantly larger than INT_MAX.

Wovchena:
@sgonorov, please review

sgonorov (approved):

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -617,12 +617,18 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
  * @param assistant_confidence_threshold the lower token probability of candidate to be validated by main model in case of dynamic strategy candidates number update.
           NOTE: `assistant_confidence_threshold` is supported only by ContinuousBatching backend for Speculative Decode.
  * @param num_assistant_tokens the defined candidates number to be generated by draft model/prompt lookup in case of static strategy candidates number update.
+ *        For tree search this serves as the number of candidate (non-root) tokens submitted to the target model for verification;
```
Wovchena:
```suggestion
 *        For tree search this serves as the overall number of candidate (non-root) tokens submitted to the target model for verification;
```

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -617,12 +617,18 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
...
+ *        total tree nodes = `num_assistant_tokens + 1` (including root).
  *        NOTE: ContinuousBatching backend for Speculative Decode uses `num_assistant_tokens` as is. Stateful backend for Speculative Decode uses `num_assistant_tokens`'s
  *        copy as initial value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both backends.
  * @param max_ngram_size is maximum ngram to use when looking for matches in the prompt.
  *
+ * Tree search parameters:
+ * @param tree_params.branching_factor number of top-k candidates selected per tree node and kept globally per tree layer.
+ * @param tree_params.tree_depth lookahead depth of the candidate tree; the draft model runs `tree_depth` iterations.
```
Wovchena: Explain what happens if `branching_factor ** tree_depth != num_assistant_tokens`

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,436 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
         }
     }
 }
+/* tree generator logic */
+
+// ------------------------------------------------------------------
+// CandidateGraph
+// ------------------------------------------------------------------
+
```
Wovchena:
```suggestion
```
That's what `CandidateGraph::` already says

src/cpp/src/sampling/sampler.cpp
```diff
@@ -491,6 +489,436 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
...
+std::vector<CandidateGraph::NodePtr> CandidateGraph::select_candidate_nodes() const {
+    if (m_max_candidate_nodes <= 0)
+        return {m_root};
+
+    // Strict total order: better node has higher score, then shallower layer, then earlier allocation address.
+    const auto is_better = [](const NodePtr& a, const NodePtr& b) {
+        if (a->score != b->score)
+            return a->score > b->score;
+        if (a->tree_layer != b->tree_layer)
+            return a->tree_layer < b->tree_layer;
```
Wovchena, line 527: 1. Does it ever trigger? It's unlikely two `score`s match.
2. Why is shallower tree better end not vice versa?

src/cpp/src/sampling/sampler.cpp
```diff
@@ -681,22 +1109,188 @@ align_all_sequence_len(SequenceGroup::Ptr& sequence_group,
...
+        logit_processor.apply(logit_vector);
+        bonus_token = _greedy_sample(logit_vector, 0);
+    }
+
+    sequence->append_token(bonus_token->m_index, bonus_token->m_log_prob);
+    logit_processor.register_new_generated_token(bonus_token->m_index);
+
+    // Update EAGLE metadata with the accepted path for the next draft round.
+    TreeMetaData updated_metadata;
+    updated_metadata.tree_position_ids = tree_metadata.tree_position_ids;
```
Wovchena, line 1372: That's the only occurance of `tree_position_ids`. Remove the member

src/cpp/src/sampling/sampler.hpp
```diff
@@ -70,7 +75,51 @@ struct SequenceGroupSamplingInfo {
...
 };
 
+// Tree of draft token candidates for speculative decoding.
+// Stores token IDs and cumulative scores without Sequence references.
+class CandidateGraph {
+public:
+    struct Node {
+        int64_t token_id = -1;
+        float score = -std::numeric_limits<float>::infinity();
+        int tree_layer = 0;
```
Wovchena:
```suggestion
        size_t tree_layer = 0;
```

likholat (changes requested):
Please update the docs https://github.com/openvinotoolkit/openvino.genai/tree/master/site

tests/cpp/eagle_tree_search.cpp
```diff
@@ -0,0 +1,531 @@
+// Copyright (C) 2026 Intel Corporation
```
likholat, line 1: I see tests for the tree logic itself, but do we also have a test that runs the full generation flow with tree_params enabled?

GuoliangShiIntel: I have updated tests with full generation flow with `tree_params` https://github.com/openvinotoolkit/openvino.genai/pull/3451/changes/ee59b40244727542353290da623c1d45a065dea8

GuoliangShiIntel:
> Please update the docs https://github.com/openvinotoolkit/openvino.genai/tree/master/site

Could you please provide more details about what docs should i update?

likholat:
> > Please update the docs https://github.com/openvinotoolkit/openvino.genai/tree/master/site
> 
> Could you please provide more details about what docs should i update?

I think we should add the docs here: https://github.com/openvinotoolkit/openvino.genai/tree/master/site/docs/concepts

@Wovchena, could you please take a look and confirm that this would be the right place for the EAGLE docs?

Wovchena:
It should be https://openvinotoolkit.github.io/openvino.genai/docs/concepts/optimization-techniques/speculative-decoding/ CVS-175015

GuoliangShiIntel:
> It should be https://openvinotoolkit.github.io/openvino.genai/docs/concepts/optimization-techniques/speculative-decoding/ [CVS-175015](https://jira.devtools.intel.com/browse/CVS-175015)

@Wovchena @likholat If this document is meant for speculative decoding, would it be better to add it in the pipeline PR, or to submit it as a separate PR after everything is ready, instead of including it in this PR?

tests/cpp/eagle_tree_search.cpp
likholat: Thanks for extending the test coverage, really appreciate that.

However, this is not exactly what I had in mind. I was referring more to an end-to-end test that exercises the full pipeline, similar to how it is done here:
https://github.com/openvinotoolkit/openvino.genai/blob/1030cba83e80fa642a5b24fd70f935ce4acd7704/tests/python_tests/test_continuous_batching.py#L668

GuoliangShiIntel: > Thanks for extending the test coverage, really appreciate that.
> 
> However, this is not exactly what I had in mind. I was referring more to an end-to-end test that exercises the full pipeline, similar to how it is done here:
> 
> https://github.com/openvinotoolkit/openvino.genai/blob/1030cba83e80fa642a5b24fd70f935ce4acd7704/tests/python_tests/test_continuous_batching.py#L668

Got it. However, the pipeline has not been merged yet. This PR is the first step toward supporting the Eagle3 TOPK implementation. After this PR is merged, we will review and refine the pipeline code, and potentially add more end-to-end pipeline tests in the pipeline PR.

GuoliangShiIntel: It seems there may be some misunderstanding about the Eagle3 plan:

**Stage 1:** Support Eagle3 with a Top‑1–only pipeline (Merged already)
The pipeline you currently see in GenAI is only for the Top‑1 scenario.

**Stage 2:** Support Eagle3 Top‑K.
Since Top‑K requires a new sampling method—the tree sampler—this is the motivation for the current PR.
After this PR is merged, we will open the actual Top‑K pipeline for review.

likholat: Got it! Thanks a lot for the clarification!

likholat:
> > It should be https://openvinotoolkit.github.io/openvino.genai/docs/concepts/optimization-techniques/speculative-decoding/ [CVS-175015](https://jira.devtools.intel.com/browse/CVS-175015)
> 
> @Wovchena @likholat If this document is meant for speculative decoding, would it be better to add it in the pipeline PR, or to submit it as a separate PR after everything is ready, instead of including it in this PR?

Got it now. I think we can postpone the documentation support and add it later with the pipeline PR

likholat (approved):

GuoliangShiIntel:
**Regarding the commit changes:**

In fact, the two recent reviews only added a single additional [commit](https://github.com/openvinotoolkit/openvino.genai/pull/3451/commits/ee59b40244727542353290da623c1d45a065dea8) on top of the existing code, mainly for expanded test coverage. The reason so many commit changes are shown is that during rebasing, the latest commit was placed after all previous changes.

I’ll be more careful going forward and rebase the changes to the front of the commit history to avoid confusion during review. Sorry for the inconvenience.

GuoliangShiIntel:
@Wovchena Do you have any additional comments on this PR? Also, are there any other reviewers we’re waiting on, or any remaining steps before it can move forward?

sbalandi (commented):
Hi @GuoliangShiIntel ! Could you, please, update tools: wwb and llm_bench according to these changes ? I mean add support of tree_depth and branching_factor to [TextEvaluator](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/text_evaluator.py#L209) for WWB and to [text_generation.py](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/task/text_generation.py#L264) for llm_bench and add tests. It could be done in another PR

GuoliangShiIntel:
> Hi @GuoliangShiIntel ! Could you, please, update tools: wwb and llm_bench according to these changes ? I mean add support of tree_depth and branching_factor to [TextEvaluator](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/text_evaluator.py#L209) for WWB and to [text_generation.py](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/task/text_generation.py#L264) for llm_bench and add tests. It could be done in another PR

Hi @sbalandi Thanks for the comment, sure, we will update it when the pipeline be supported. Here is the Jira to track:
https://jira.devtools.intel.com/browse/CVS-185476

sbalandi:
> > Hi @GuoliangShiIntel ! Could you, please, update tools: wwb and llm_bench according to these changes ? I mean add support of tree_depth and branching_factor to [TextEvaluator](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/text_evaluator.py#L209) for WWB and to [text_generation.py](https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/task/text_generation.py#L264) for llm_bench and add tests. It could be done in another PR
> 
> Hi @sbalandi Thanks for the comment, sure, we will update it when the pipeline be supported. Here is the Jira to track: https://jira.devtools.intel.com/browse/CVS-185476

great, thank you !

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -675,6 +681,16 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
...
 
+    /**
+     * @brief Parameters for speculative decoding tree search
+     */
+    struct TreeParams {
+        /// Number of top-k candidates expanded at each tree node (branching factor)
+        size_t branching_factor = 1;
+        /// Lookahead depth of the candidate tree; the draft model runs `tree_depth` iterations
+        size_t tree_depth = 0;
+    } tree_params;
```
Wovchena: `TreeParams` relates to assisting generation parameters which don't have a dedicated structure. Move `TreeParams`'s members to `GenerationConfig` and delete `TreeParams`

Wovchena (changes requested):

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: In this PR, yes, `tree_position_ids` will not be used.

However, for the pipeline implementation, we need this information to quickly determine which layer each token belongs to. Otherwise, we would have to recompute this information from the `tree_mask`, which is both time‑consuming and not elegant.

So, i have not removed this member in this PR.

GuoliangShiIntel: Here is NPU pipeline example to use this info. [Code link ](https://github.com/GuoliangShiIntel/openvino.genai/blob/sgl/eagle_3_topk_npu_support_v3/src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp#L244)

I think GPU has the same situation right? @songbell pls more comment, thanks

src/cpp/src/sampling/sampler.hpp
GuoliangShiIntel: Done

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: Done

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: > Does it ever trigger? It's unlikely two scores match.

I have not found it in a real case, exactly, It's unlikely two scores same, but it is not 100% ensure.
I don’t think this will impact pipeline efficiency, so I haven’t removed it.

> Why is shallower tree better end not vice versa?

Here is the [eagle source code](https://github.com/SafeAILab/EAGLE/blob/main/eagle/model/cnets.py#L762-L764). 
In cnets.py, the final selection is:
```
  top_scores = torch.topk(scores_list, total_tokens, dim=-1)
  top_scores_index = top_scores.indices
  top_scores_index = torch.sort(top_scores_index).values
```
`scores_list` is built by appending depth-by-depth, so shallower nodes always have smaller indices. When `torch.topk` breaks ties, it naturally picks smaller indices first — meaning shallower nodes win.

This is also the correct behavior: a deeper node only gets accepted if every ancestor on its path is accepted first.

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: The total Draft model generate tokens number should be:
`total_draft_tokens = (branching_factor * branching_factor) * (tree_depth - 1) + branching_factor`

`num_assistant_tokens` represents the total number of draft tokens to be validated.

We always require `total_draft_tokens >= num_assistant_tokens` as a prerequisite, and we will also enforce a strict check for this condition in the pipeline.

take this case as an example:
```
total_draft_tokens = (2*2)*(3-1)+2 = 10
num_assistant_tokens = 7
```
we select the top 7 draft token and plus with root token to target model validation.
So we needs manage eagle metadata to present those info.
image

GuoliangShiIntel:
@Wovchena Thanks for the solid and detailed comments. I’ve addressed all of them, and for the items where no changes were made, I’ve added comments explaining the rationale. Please take a look when you have a chance.

All change exist in this commit https://github.com/openvinotoolkit/openvino.genai/pull/3451/commits/47a584fa45eab68ea4502dcc903a0af4941a83de

likholat:
@GuoliangShiIntel  Could you please confirm if I understand correctly that PR #3330 continues this PR and adds pipeline support?

GuoliangShiIntel:
> @GuoliangShiIntel Could you please confirm if I understand correctly that PR #3330 continues this PR and adds pipeline support?

Hi @likholat As i have mentioned above, we have two stages for eagle3 feature:

**Top - 1:** LLM has already merged. The PR #3330 is to support VLM
**Top - K:** NPU pipeline #3396, PR for GPU pipeline have not been created

peterchen-intel:
GPU/CPU EAGLE3 TopK Pipeline: https://github.com/openvinotoolkit/openvino.genai/pull/3755

src/cpp/include/openvino/genai/generation_config.hpp
Wovchena: Add the explanation to docstrings

GuoliangShiIntel: Update in this commit: https://github.com/openvinotoolkit/openvino.genai/pull/3451/changes/f4f44ebfd1b7cbde27f232a29bfd2092806c0b37

src/cpp/src/sequence_group.hpp
```diff
@@ -56,6 +64,7 @@ class Sequence {
     size_t m_hidden_size;
     std::vector<ov::Tensor> m_position_ids_list;
     int64_t m_rope_delta;
+    TreeMetaData m_tree_metadata;
 
     // Embeddings hash calculation params
     static constexpr size_t m_embeddings_hash_max_num_values = 10; // max number of values used for embeddings hash calculation
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/sequence_group.hpp
```diff
@@ -152,6 +161,14 @@ class Sequence {
         return m_hidden_state;
     }
 
+    void set_tree_metadata(const TreeMetaData& metadata) {
+        m_tree_metadata = metadata;
+    }
+
+    const TreeMetaData& get_tree_metadata() const {
+        return m_tree_metadata;
+    }
```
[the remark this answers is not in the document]
GuoliangShiIntel: Done

src/cpp/src/generation_config.cpp
```diff
@@ -346,6 +357,29 @@ void GenerationConfig::validate() const {
...
+        OPENVINO_ASSERT(
+            num_assistant_tokens > 0,
+            "'num_assistant_tokens' must be > 0 when tree search is enabled, but got ",
+            num_assistant_tokens);
+        OPENVINO_ASSERT(num_assistant_tokens >= tree_depth,
+                        "'num_assistant_tokens' (",
+                        num_assistant_tokens,
+                        ") must be >= 'tree_depth' (",
+                        tree_depth,
+                        ") to allow at least one node per draft layer");
```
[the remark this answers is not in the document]
GuoliangShiIntel: No change. This check should be handled on the pipeline side.

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -617,12 +617,21 @@ operator|(const StructuredOutputConfig::StructuralTag& lhs,
...
  *        NOTE: ContinuousBatching backend for Speculative Decode uses `num_assistant_tokens` as is. Stateful backend for Speculative Decode uses `num_assistant_tokens`'s
  *        copy as initial value and adjusts it based on recent number of accepted tokens. If `num_assistant_tokens` is not set, it defaults to `5` for both backends.
  * @param max_ngram_size is maximum ngram to use when looking for matches in the prompt.
  *
+ * Tree search parameters:
+ * @param branching_factor number of top-k candidates selected per tree node and kept globally per tree layer.
+ * @param tree_depth lookahead depth of the candidate tree; the draft model runs `tree_depth` iterations.
+ *        NOTE: The total number of draft tokens produced by the tree is
+ *        `total_draft_tokens = branching_factor^2 * (tree_depth - 1) + branching_factor`.
+ *        `total_draft_tokens >= num_assistant_tokens` must hold. The top `num_assistant_tokens` candidates (by score) are selected from the tree for verification.
```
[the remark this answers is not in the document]
GuoliangShiIntel: No change. Since pipeline has not been supported. Will add it with pipeline PR

src/cpp/src/sampling/sampler.cpp
songbell: yes, this will be used for position ids offset in the pipeline

Wovchena (approved):

src/cpp/src/sampling/sampler.cpp
```diff
@@ -1015,6 +1606,66 @@ SequenceGroupSamplingInfo Sampler::sample_from_sequence_group(SequenceGroup::Ptr
...
+            // Target-model validation pass: validate the candidate tree against target logits.
+            std::vector<Sequence::Ptr> running_sequences = sequence_group->get_running_sequences();
+            OPENVINO_ASSERT(running_sequences.size() == 1);
+            size_t valid_tokens = verify_draft_tree(running_sequences[0],
+                                                    sequence_group_logits,
+                                                    logit_processor,
+                                                    num_generated_tokens_to_validate);
+            assisting_pipeline_info.max_removed_tokens_per_request =
+                std::max(assisting_pipeline_info.max_removed_tokens_per_request,
+                         num_generated_tokens_to_validate + 1 - valid_tokens);
```
songbell, line 1637: the stop_tokens is not handled, it may cause unable to stop main generation if stop ids are hit.

likholat:
@GuoliangShiIntel could you please resolve the conflicts?

src/cpp/src/sampling/sampler.cpp
GuoliangShiIntel: Good cache, i have fixed it by https://github.com/openvinotoolkit/openvino.genai/pull/3451/changes/d165d11f2cf2cccd59ebaba690b58cc6af0b97e7

GuoliangShiIntel:
> @GuoliangShiIntel could you please resolve the conflicts?

done

src/cpp/src/sequence_group.hpp
```diff
@@ -787,6 +805,10 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
         return m_sampling_params;
     }
 
+    void set_sampling_parameters(const ov::genai::GenerationConfig& params) {
+        m_sampling_params = params;
+    }
+
```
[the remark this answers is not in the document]
GuoliangShiIntel: will be used by pipeline, so keep it

src/cpp/src/sequence_group.hpp
```diff
@@ -787,6 +805,10 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
         return m_sampling_params;
     }
 
+    void set_sampling_parameters(const ov::genai::GenerationConfig& params) {
+        m_sampling_params = params;
+    }
```
songbell: seems I did not call this API in cb pipeline. can you double confirm if this is a must api? @GuoliangShiIntel

GuoliangShiIntel: good cache, i double check and yes, it can be removed.

previous, for prefill stage we use the greedy_sampler and it change in the pipeline. Now, everything has moved into sampler logic. it can be removed safely. Thanks

GuoliangShiIntel:
**CI Status:**

We got two regression report:

1. [Linux (Ubuntu 22.04, Python 3.11) / Python (Cacheopt E2E (Part 1)) Tests (wheel) (pull_request)](https://github.com/openvinotoolkit/openvino.genai/actions/runs/25091348053/job/73538731874?pr=3451)
same regression shows in other PR [report](https://github.com/openvinotoolkit/openvino.genai/actions/runs/25042581373/job/73358894790?pr=3772)

2. [Manylinux 2_28 / Python (Cacheopt E2E (Part 1)) Tests (wheel) (pull_request)](https://github.com/openvinotoolkit/openvino.genai/actions/runs/25091348069/job/73538743990?pr=3451)
same regression shows in other PR [report](https://github.com/openvinotoolkit/openvino.genai/actions/runs/25042581249/job/73358527799?pr=3772)

I don’t think this regression is caused by this PR. However, it would be helpful to know whether there is an existing Jira issue tracking it. Also, regarding the merge progress, should we wait for the fix PR to be merged and CI to be fully green, or can we move this PR forward in parallel?

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -675,6 +684,10 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
     size_t num_assistant_tokens = 0;
     size_t max_ngram_size = 0;
 
+    // Tree search parameters
+    size_t branching_factor = 1;
+    size_t tree_depth = 0;
```
Wovchena, line 691: Hide the new params for now. So if the full feature isn't merged by the release, there are no dangling params. Not sure how to create the internal temporary alias though

Wovchena (changes requested):

src/cpp/include/openvino/genai/generation_config.hpp
GuoliangShiIntel: Thanks for the suggestion. I understand the concern about exposing unfinished features at release time, but I believe these parameters fall into a different category than “dangling”.

`branching_factor` and `tree_depth` sit at the same abstraction level as the existing `num_assistant_tokens`. They are pipeline-level knobs for speculative decoding, not high level, user facing `LLMPipeline` options. With this PR, the tree search sampler is fully implemented end-to-end and functionally complete. Hiding these parameters would actually risk breaking functional completeness, and our tests might no longer be able to cover the relevant paths correctly.

From a C++ implementation perspective, hiding them would also be awkward. The `ov::Property` system does not provide a clean way to create a temporary alias or internal only property without introducing a separate internal configuration structure. That would be a non-trivial refactor and would likely introduce more churn than the risk we are trying to avoid.

My suggestion is to keep the parameters public as they are, but defer adding user facing documentation and samples until the pipeline PR lands. We have already aligned that the documentation will be added as part of the pipeline PR. This way, there is no API churn, no dangling surface, and the feature remains discoverable for advanced developers who may need it prior to the release.

Wovchena: `src/cpp/include/openvino/genai/generation_config.hpp` is part of the public API, so the abstraction level does not change the concern here.

The scope of this PR is limited to sampling, while these parameters only make sense as part of the full feature and have no real standalone value. Exposing them now expands the public API surface before the feature is complete.

Since there is currently no way to avoid that API impact in this PR, I don’t think we should merge it before code freeze. Keeping this blocked until after code freeze.

Wovchena (approved):

PR #3465, opened 2026-03-09
[GPU] Route VLM per-model properties via utils::get_model_properties

sungeunk:
## Summary
- Add `utils::get_model_properties(properties, model_role)` that resolves
a sub-model's effective config by merging two priority layers
(global → `PER_MODEL_PROPERTIES[model_role]`). The `PER_MODEL_PROPERTIES`
meta key is stripped from the result. Meta keys and model roles are
matched exactly (case-sensitive).
- Wire the helper into both VLM-aware `ContinuousBatchingPipeline`
constructors so callers can already route `cache_dir`, `NUM_STREAMS`,
and other properties to the vision encoder per model on the current
public API — no breaking changes.

CVS-182861

## Checklist:
- [x] This PR follows [GenAI Contributing guidelines](../CONTRIBUTING.md).
  Added five unit tests in `tests/cpp/utils.cpp` for `utils::get_model_properties()`:
  globals-only passthrough, `PER_MODEL_PROPERTIES` meta-key exclusion,
  per-model overriding global, per-model missing role falls back to
  globals, and disjoint-key merging across both layers.
- [x] This PR fully addresses the ticket (CVS-182861).
  No doc update required: this is an internal behavior fix; the public
  constructor signatures of `ContinuousBatchingPipeline` are unchanged.

---------
Target Machine: PTLH
Ran it four times in a row after removing the cache files.

### master
```
| Task                                      | R1.LAT   | R2.LAT   | R3.LAT   | R4.LAT   |
| ContinuousBatchingPipeline constructor #1 | 8.123 s  | 4.586 s  | 3.283 s  | 3.270 s  |
| │  └─ Language model read                 | 0.177 s  | 0.124 s  | 0.123 s  | 0.127 s  |
| │  └─ *Tokenizer*                         | 0.896 s  | 0.628 s  | 0.633 s  | 0.650 s  |
| │  └─ InputsEmbedder creation             | 2.247 s  | 2.167 s  | 2.026 s  | 1.986 s  |
| │    └─ VisionEncoder                     | 0.366 s  | 0.311 s  | 0.235 s  | 0.241 s  |
| │    └─ EmbeddingsModel                   | 0.173 s  | 0.182 s  | 0.174 s  | 0.175 s  |
| │    └─ *Tokenizer*                       | 0.646 s  | 0.618 s  | 0.645 s  | 0.618 s  |
| │    └─ *VisionEmbeddingsMerger*          | 0.987 s  | 0.964 s  | 0.897 s  | 0.872 s  |  no cache
| │    └─ VisionEmbeddingsPos               | 0.016 s  | 0.032 s  | 0.015 s  | 0.014 s  |  no cache
| │  └─ Language model                      | 4.612 s  | 1.606 s  | 0.447 s  | 0.447 s  |  cache
```

### master + This PR
* Tokenizer: 650 ms -> 55 ms, 618 ms -> 47 ms
* VisionEmbeddingsMerger: 872 ms -> 408 ms
```
| Task                                      | R1.LAT   | R2.LAT   | R3.LAT   | R4.LAT   |
| ContinuousBatchingPipeline constructor #1 | 8.317 s  | 4.436 s  | 2.143 s  | 2.191 s  |
| │  └─ Language model read                 | 0.178 s  | 0.126 s  | 0.125 s  | 0.134 s  |
| │  └─ *Tokenizer*                         | 0.305 s  | 0.054 s  | 0.055 s  | 0.055 s  |
| │  └─ InputsEmbedder creation             | 2.966 s  | 2.519 s  | 1.477 s  | 1.482 s  |
| │    └─ VisionEncoder                     | 0.359 s  | 0.375 s  | 0.157 s  | 0.172 s  |
| │    └─ EmbeddingsModel                   | 0.595 s  | 0.881 s  | 0.182 s  | 0.185 s  |
| │    └─ *Tokenizer*                       | 0.049 s  | 0.045 s  | 0.048 s  | 0.047 s  |
| │    └─ *VisionEmbeddingsMerger*          | 1.277 s  | 0.520 s  | 0.437 s  | 0.408 s  |  cache
| │    └─ VisionEmbeddingsPos               | 0.020 s  | 0.035 s  | 0.004 s  | 0.004 s  |  cache
| │  └─ Language model                      | 4.664 s  | 1.680 s  | 0.428 s  | 0.461 s  |  cache
```

### DIFF
--- a/src/cpp/src/continuous_batching/pipeline.cpp
+++ b/src/cpp/src/continuous_batching/pipeline.cpp
@@ -60,8 +60,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        auto vision_encoder_properties_for_embedder = utils::inherit_cache_properties(vision_encoder_properties, properties);
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties_for_embedder);
+        embedder = std::make_shared<InputsEmbedder>(models_path, device,
+            utils::inherit_cache_properties(vision_encoder_properties, properties));
     }
 
     utils::print_scheduler_config_info(scheduler_config);
@@ -112,8 +112,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(const std::shared_ptr<ov:
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        auto vision_encoder_properties_for_embedder = utils::inherit_cache_properties(vision_encoder_properties, properties);
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties_for_embedder);
+        embedder = std::make_shared<InputsEmbedder>(models_path, device,
+            utils::inherit_cache_properties(vision_encoder_properties, properties));
     }
 
     utils::print_scheduler_config_info(scheduler_config);
--- a/src/cpp/src/continuous_batching/pipeline.cpp
+++ b/src/cpp/src/continuous_batching/pipeline.cpp
@@ -60,7 +60,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties);
+        auto vision_encoder_properties_for_embedder = utils::inherit_cache_properties(vision_encoder_properties, properties);
+        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties_for_embedder);
     }
 
     utils::print_scheduler_config_info(scheduler_config);
@@ -111,7 +112,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(const std::shared_ptr<ov:
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties);
+        auto vision_encoder_properties_for_embedder = utils::inherit_cache_properties(vision_encoder_properties, properties);
+        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties_for_embedder);
     }
 
     utils::print_scheduler_config_info(scheduler_config);
--- a/src/cpp/src/continuous_batching/pipeline.cpp
+++ b/src/cpp/src/continuous_batching/pipeline.cpp
@@ -60,8 +60,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device,
-            utils::inherit_cache_properties(vision_encoder_properties, properties));
+        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings", device);
+        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_props);
     }
 
     utils::print_scheduler_config_info(scheduler_config);
@@ -112,8 +112,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(const std::shared_ptr<ov:
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device,
-            utils::inherit_cache_properties(vision_encoder_properties, properties));
+        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings", device);
+        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_props);
     }
 
     utils::print_scheduler_config_info(scheduler_config);
--- a/src/cpp/src/continuous_batching/pipeline.cpp
+++ b/src/cpp/src/continuous_batching/pipeline.cpp
@@ -60,7 +60,7 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings", device);
+        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings");
         embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_props);
     }
 
@@ -112,7 +112,7 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(const std::shared_ptr<ov:
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings", device);
+        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings");
         embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_props);
     }
 
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -4,7 +4,6 @@
 #include "utils.hpp"
 
 #include <algorithm>
-#include <array>
 #include <cctype>
 #include <variant>
 #include <fstream>
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -3,10 +3,12 @@
 
 #include "utils.hpp"
 
+#include <array>
 #include <variant>
 #include <fstream>
 #include <memory>
 
+#include "openvino/runtime/properties.hpp"
 #include "openvino/op/add.hpp"
 #include "openvino/op/divide.hpp"
 #include "openvino/op/gather.hpp"
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -378,7 +378,6 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
 } // namespace
 
 const std::string PER_MODEL_PROPERTIES = "MODEL_PROPERTIES";
-const std::string DEVICE_PROPERTIES    = "DEVICE_PROPERTIES";
 
 namespace {
 std::string to_lower(const std::string& s) {
@@ -408,7 +407,7 @@ ov::AnyMap get_model_properties(const ov::AnyMap& properties, const std::string&
     // Meta key exclusion is case-insensitive (e.g. "per_model_properties"
     // is also treated as a meta key).
     const auto per_model_lowered = to_lower(PER_MODEL_PROPERTIES);
-    const auto device_props_lowered = to_lower(DEVICE_PROPERTIES);
+    const auto device_props_lowered = to_lower(ov::device::properties.name());
     for (const auto& kv : properties) {
         const auto key_lowered = to_lower(kv.first);
         if (key_lowered == per_model_lowered || key_lowered == device_props_lowered) {
@@ -417,9 +416,9 @@ ov::AnyMap get_model_properties(const ov::AnyMap& properties, const std::string&
         result.insert_or_assign(kv.first, kv.second);
     }
 
-    // Layer 2: DEVICE_PROPERTIES[device] overrides globals for this device.
+    // Layer 2: ov::device::properties[device] overrides globals for this device.
     // Both the meta key and the device name are matched case-insensitively.
-    auto dev_it = find_case_insensitive(properties, DEVICE_PROPERTIES);
+    auto dev_it = find_case_insensitive(properties, ov::device::properties.name());
     if (dev_it != properties.end()) {
         const auto& by_device = dev_it->second.as<ov::AnyMap>();
         auto role_dev_it = find_case_insensitive(by_device, device);
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -4,7 +4,6 @@
 #include "utils.hpp"
 
 #include <algorithm>
-#include <cctype>
 #include <variant>
 #include <fstream>
 #include <memory>
--- a/src/cpp/src/utils.cpp
+++ b/src/cpp/src/utils.cpp
@@ -379,7 +379,6 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
 
 const std::string PER_MODEL_PROPERTIES = "MODEL_PROPERTIES";
 
-namespace {
 std::string to_lower(const std::string& s) {
     std::string result(s.size(), '\0');
     std::transform(s.begin(), s.end(), result.begin(),
@@ -398,7 +397,6 @@ ov::AnyMap::const_iterator find_case_insensitive(const ov::AnyMap& map, const st
     }
     return map.end();
 }
-} // namespace
 
 ov::AnyMap get_model_properties(const ov::AnyMap& properties, const std::string& model_role, const std::string& device) {
     ov::AnyMap result;
@@ -421,9 +419,9 @@ ov::AnyMap get_model_properties(const ov::AnyMap& properties, const std::string&
     auto dev_it = find_case_insensitive(properties, ov::device::properties.name());
     if (dev_it != properties.end()) {
         const auto& by_device = dev_it->second.as<ov::AnyMap>();
-        auto role_dev_it = find_case_insensitive(by_device, device);
-        if (role_dev_it != by_device.end()) {
-            for (const auto& kv : role_dev_it->second.as<ov::AnyMap>()) {
+        auto device_it = find_case_insensitive(by_device, device);
+        if (device_it != by_device.end()) {
+            for (const auto& kv : device_it->second.as<ov::AnyMap>()) {
                 result.insert_or_assign(kv.first, kv.second);
             }
         }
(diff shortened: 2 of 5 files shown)


src/cpp/src/tokenizer/tokenizer_impl.cpp
```diff
@@ -424,7 +424,7 @@ void Tokenizer::TokenizerImpl::setup_tokenizer(const std::pair<std::shared_ptr<o
 
         // Initialize tokenizer's cache to save time later.
         // TODO CVS-150630: Empty strings sporadically can fail, therefore use nonempty string for warmup.
-        encode("non empty string");
+        // encode("non empty string");
```
as-suvorov: Tokenizer warm-up times has been optimized in https://github.com/openvinotoolkit/openvino_tokenizers/pull/597

```suggestion
        encode("non empty string");
```

src/cpp/src/tokenizer/tokenizer_impl.cpp
```diff
@@ -448,7 +448,7 @@ void Tokenizer::TokenizerImpl::setup_tokenizer(const std::pair<std::shared_ptr<o
         if (m_eos_token_id != -1 && m_eos_token.empty())
             m_eos_token = decode(std::vector{m_eos_token_id}, {ov::genai::skip_special_tokens(false)});
         // Initialize detokenizer's cache to save time later.
-        decode({1, 33, 199, 42, 42});
+        // decode({1, 33, 199, 42, 42});
```
as-suvorov:
```suggestion
        decode({1, 33, 199, 42, 42});
```

as-suvorov (changes requested):

src/cpp/src/tokenizer/tokenizer_impl.cpp
sungeunk: @as-suvorov  encode() still uses 100 ms in model loading. Target loading time is under 1.5 s.

as-suvorov: There are also targets for TTFT for other models which will be affected by this change

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -59,7 +59,14 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties);
+        auto vision_encoder_properties_for_embedder = vision_encoder_properties;
```
as-suvorov: Can `cache_dir` be set directly to `vision_encoder_properties` by user code?

src/cpp/src/tokenizer/tokenizer_impl.cpp
as-suvorov: The preferred way to address tokenizers warmup is to run async encode in ctor. With this approach tokenizer warmup will happen in parallel to other models compilation. Also we need to ensure tokenizer is first model initialized in pipeline (already the case as I understand).

src/cpp/src/continuous_batching/pipeline.cpp
sungeunk: @as-suvorov sure. This logic will work when 'vision_encoder_properties' is empty.

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -59,7 +59,14 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
...
+        if (vision_encoder_properties_for_embedder.empty()) {
+            const auto cache_dir_name = ov::cache_dir.name();
+            if (properties.find(cache_dir_name) != properties.end()) {
+                vision_encoder_properties_for_embedder[cache_dir_name] = properties.at(cache_dir_name);
```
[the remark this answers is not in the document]
sungeunk: The intention is that if the user does not set cache_dir, it should still work as-is.

src/cpp/src/continuous_batching/pipeline.cpp
as-suvorov: @sungeunk `vision_encoder_properties` are split on purpose, from my pov implicitly copying properties can be confusing for the user. If user can set `cache_dir` to `vision_encoder_properties` in the current API, I propose to not copy properties.

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -59,7 +59,14 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties);
+        auto vision_encoder_properties_for_embedder = vision_encoder_properties;
+        if (vision_encoder_properties_for_embedder.empty()) {
```
popovaan: Why is this property set only when `vision_encoder_properties_for_embedder.empty()`?
Shouldn't we check that "cache_dir" was not set instead?

src/cpp/src/continuous_batching/pipeline.cpp
sungeunk: @as-suvorov  No path to set vision_encoder_properties to set the properties in ContinuousBatchingPipeline.
https://github.com/openvinotoolkit/openvino.genai/blob/dbe751505d67874980e34302b2e22dbd0ff7eaa2/src/cpp/src/visual_language/pipeline.cpp#L709
Could the GenAI team consider how to resolve this issue?

as-suvorov: @sungeunk I see the problem now, thanks. 
To summarize:
The requirement is to set `vision_encoder_properties` (`cache_dir` for this particular case). There is an API to set it with `ContinuousBatchingPipeline`. The problem that there is seem to be no API to set it with `VLMPipeline` ctor.
@yatarkan Is there a way to set `vision_encoder_properties` with `VLMPipeline` ctor?

sungeunk: @yatarkan any update? If you are running into this issue in another task, please let me know.

yatarkan: Currently as I can see in `VLMPipeline` there is no separation between different types of properties maps like in `ContinuousBatchingPipeline`, so there is no way to setup `vision_encoder_properties` explicitly from user perspective with `VLMPipeline` ctor.
However, in stateful `VLMPipeline` there is extraction logic for `embedder_properties` (which is then passed to `InputsEmbedder`) and it seems `cache_dir` will be propagated for SDPA use case.
The problem is that `VLMContinuousBatchingAdapter` does not use it.
I believe such properties extraction logic should be unified, aligned and reused in `VLMPipeline`, `VLMContinuousBatchingAdapter` and `ContinuousBatchingPipeline` classes.

sungeunk: @yatarkan Could you let me know the current status of the work?  If there is already a PR, I would appreciate it if you could share it.

yatarkan: @sungeunk I am not working on that, currently I have other more prioritized tasks.
You can implement properties alignment and propagation to `ContinuousBatchingPipeline` within this PR.

src/cpp/src/continuous_batching/pipeline.cpp
sungeunk: Good catch — you are right. The previous `.empty()` check was too coarse: as soon as any unrelated key (e.g. `NUM_STREAMS`) was added to `vision_encoder_properties`, inheritance of `cache_dir` was silently disabled.

The latest revision replaces that logic with `utils::inherit_cache_properties()`, which checks **per key** whether the property is already set in the sub map, exactly as you suggested. It also covers `cache_mode` and
`cache_encryption_callbacks` alongside `cache_dir` for consistency, and preserves explicit per-sub-model overrides.

sgonorov (approved):

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -111,7 +112,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(const std::shared_ptr<ov:
 
     std::shared_ptr<InputsEmbedder> embedder;
     if (std::filesystem::exists(models_path / "openvino_text_embeddings_model.xml")) {
-        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_encoder_properties);
+        auto vision_props = utils::get_model_properties(properties_without_draft_model, "vision_embeddings", device);
```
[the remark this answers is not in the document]
sungeunk: In this constructor, 'vision_encoder_properties' is empty. And this arg will be removed after implement https://github.com/dkalinowski/openvino.genai/blob/vlm-design/design.md.

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -60,7 +60,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline( const std::filesystem::p
...
```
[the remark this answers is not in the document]
sungeunk: In this constructor, 'vision_encoder_properties' is empty. And this arg will be removed after implement https://github.com/dkalinowski/openvino.genai/blob/vlm-design/design.md.

src/cpp/src/utils.hpp
```diff
@@ -142,6 +142,28 @@ ov::Core& singleton_core();
...
+///        2. DEVICE_PROPERTIES[device]
+///        3. PER_MODEL_PROPERTIES[model_role]
+/// @param properties The main properties map. Not modified.
+/// @param model_role Sub-model role (e.g. "vision_embeddings"). Matched
+///        case-insensitively against PER_MODEL_PROPERTIES keys.
+/// @param device Device for the sub-model (e.g. "GPU"). Matched
+///        case-insensitively against DEVICE_PROPERTIES keys.
+/// @return A new ov::AnyMap with the merged result. The input map is left
+///         untouched so callers may continue using the meta keys.
+ov::AnyMap get_model_properties(const ov::AnyMap& properties, const std::string& model_role, const std::string& device);
```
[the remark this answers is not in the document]
sungeunk: description updated

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,75 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
 
 } // namespace
 
+const std::string PER_MODEL_PROPERTIES = "PER_MODEL_PROPERTIES";
```
Wovchena:
```suggestion
const std::string PER_MODEL_PROPERTIES = "MODEL_PROPERTIES";
```
@dkalinowski agreed to rename but didn't update https://github.com/dkalinowski/openvino.genai/blob/vlm-design/design.md

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,75 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
+const std::string DEVICE_PROPERTIES    = "DEVICE_PROPERTIES";
```
Wovchena: `DEVICE_PROPERTIES` is OpenVINO's feature and doesn't require special handling. Existing example: https://github.com/openvinotoolkit/openvino.genai/blob/1030cba83e80fa642a5b24fd70f935ce4acd7704/tests/python_tests/test_vlm_pipeline.py#L194

Wovchena (changes requested):

tests/cpp/utils.cpp
```diff
@@ -19,3 +21,131 @@ TEST(TestIsContainer, test_is_container) {
...
+
+TEST(TestGetModelProperties, per_model_missing_role_leaves_lower_layers) {
+    ov::AnyMap main_props = {
+        {"NUM_STREAMS", std::string("2")},
+        {PER_MODEL_PROPERTIES, ov::AnyMap{
+            {"text_embeddings", ov::AnyMap{{"NUM_STREAMS", std::string("8")}}},
+        }},
+    };
+    auto result = get_model_properties(main_props, "vision_embeddings", "GPU");
+    EXPECT_EQ(result.at("NUM_STREAMS").as<std::string>(), "2");
```
dkalinowski, line 99:
```suggestion
    ASSERT_EQ(result.count("NUM_STREAMS"), 1);
    EXPECT_EQ(result.at("NUM_STREAMS").as<std::string>(), "2");
```
prefer assertion over unhandled exception in case the tests starts failing

src/cpp/src/utils.cpp
sungeunk: I'll update it.

tests/cpp/utils.cpp
sungeunk: I'll update it.

src/cpp/src/utils.cpp
Wovchena: Remove `DEVICE_PROPERTIES` so OpenVINO handles it

src/cpp/src/utils.cpp
dkalinowski: @Wovchena good catch, I updated the design.md just now

dkalinowski (approved):

src/cpp/src/continuous_batching/pipeline.cpp
```diff
@@ -111,7 +112,8 @@ ContinuousBatchingPipeline::ContinuousBatchingPipeline(const std::shared_ptr<ov:
...
+        embedder = std::make_shared<InputsEmbedder>(models_path, device, vision_props);
```
dkalinowski, line 116: In the final implementation we need to pass all properties down the InputsEmbedder constructor and extract separate text/vision/merger properties there

src/cpp/src/utils.cpp
sungeunk: Oh mis-understand. I'll remove it.

src/python/py_generation_config.cpp
```diff
@@ -447,7 +447,7 @@ void init_generation_config(py::module_& m) {
         .def_readwrite("include_stop_str_in_output", &GenerationConfig::include_stop_str_in_output)
         .def_readwrite("stop_token_ids", &GenerationConfig::stop_token_ids)
         .def_readwrite("structured_output_config", &GenerationConfig::structured_output_config)
-        .def_readwrite("parsers", &GenerationConfig::parsers, py::keep_alive<1, 2>())
+        .def_readwrite("parsers", &GenerationConfig::parsers)
```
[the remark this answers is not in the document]
sungeunk: restored

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,74 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
+    }
+
+    // Layer 2: ov::device::properties[device] overrides globals for this device.
+    // Both the meta key and the device name are matched case-insensitively.
+    auto dev_it = find_case_insensitive(properties, ov::device::properties.name());
+    if (dev_it != properties.end()) {
+        const auto& by_device = dev_it->second.as<ov::AnyMap>();
+        auto role_dev_it = find_case_insensitive(by_device, device);
+        if (role_dev_it != by_device.end()) {
+            for (const auto& kv : role_dev_it->second.as<ov::AnyMap>()) {
```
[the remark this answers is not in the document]
sungeunk: updated

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,74 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
 
 } // namespace
 
+const std::string PER_MODEL_PROPERTIES = "MODEL_PROPERTIES";
```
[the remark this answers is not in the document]
sungeunk: PER_MODEL_PROPERTIES is in ov::genai::utils namespace scope.

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,72 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
+        }
+    }
+    return map.end();
+}
+
+ov::AnyMap get_model_properties(const ov::AnyMap& properties, const std::string& model_role, const std::string& device) {
+    ov::AnyMap result;
+
+    // Layer 1 (lowest priority): global properties, excluding meta keys.
+    // Meta key exclusion is case-insensitive (e.g. "per_model_properties"
```
Wovchena:
```suggestion
    // Meta key exclusion is case-insensitive (e.g. "model_properties"
```

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,72 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
+    // is also treated as a meta key).
```
Wovchena: Why do you need case-insensitive key exclusion? Stay consistent with OpenVINO

src/cpp/src/utils.cpp
```diff
@@ -373,6 +377,72 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
+    if (pm_it != properties.end()) {
+        const auto& per_model = pm_it->second.as<ov::AnyMap>();
+        auto role_it = find_case_insensitive(per_model, model_role);
+        if (role_it != per_model.end()) {
+            for (const auto& kv : role_it->second.as<ov::AnyMap>()) {
+                result.insert_or_assign(kv.first, kv.second);
+            }
+        }
+    }
+
```
Wovchena:
```suggestion
ov::AnyMap get_model_properties(ov::AnyMap& properties, const std::string& model_role) {
    ov::AnyMap result;
    auto model_properties = properties.find(PER_MODEL_PROPERTIES);
    if (model_properties != properties.end()) {
        const ov::AnyMap& model_map = model_properties->second.as<ov::AnyMap>();
        auto role = model_map.find(model_role);
        if (role != model_map.end()) {
            result = role->second.as<ov::AnyMap>();
        }
    }
    result.merge(properties);
    result.erase(PER_MODEL_PROPERTIES);
```

Wovchena (changes requested):

src/cpp/src/utils.cpp
sungeunk: updated it

src/cpp/src/utils.cpp
sungeunk: no special reason. updated code comments.

src/cpp/src/utils.cpp
sungeunk: applied suggestion

Wovchena (approved):

src/cpp/src/utils.cpp
```diff
@@ -373,6 +376,24 @@ bool is_gguf_model(const std::filesystem::path& file_path) {
...
+    auto model_properties = properties.find(PER_MODEL_PROPERTIES);
+    if (model_properties != properties.end()) {
+        const ov::AnyMap& model_map = model_properties->second.as<ov::AnyMap>();
+        auto role = model_map.find(model_role);
+        if (role != model_map.end()) {
+            result = role->second.as<ov::AnyMap>();
+        }
+    }
+    result.merge(properties);
+    result.erase(PER_MODEL_PROPERTIES);
```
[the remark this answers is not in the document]
Wovchena: My version had a copy. Fix it in my or Copilos's way

sungeunk:
@as-suvorov With the help of several colleagues, I have updated the redesigned PR.
Please review it and merge if there are no issues.

as-suvorov (approved):

PR #3482, opened 2026-03-12
Sswierze long json with chattemplate

sswierze:
## Description

CVS-###

Fixes #(issue)

## Checklist:


tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -642,10 +656,10 @@ def create_evaluator(base_model, args):
 
     try:
         EvaluatorCLS = EVALUATOR_REGISTRY[task]
-        prompts = load_prompts(args)
 
         if task == "text":
-            tokenizer = load_tokenizer(args) if not args.llamacpp else None
+            tokenizer = load_tokenizer(args)
```
sbalandi: why condition is deleted ?

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -655,15 +669,14 @@ def create_evaluator(base_model, args):
                 gen_answer_fn = None
 
             use_chat_template = (
-                tokenizer is not None and tokenizer.chat_template is not None and not args.omit_chat_template
+                tokenizer is not None and tokenizer.chat_template is not None and not args.omit_chat_template and not is_json_dataset(args.dataset)
```
sbalandi: we have --omit-chat-template option, user can pass in json non-message message

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -14,13 +15,32 @@
...
+)
 
 # Configure logging
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
 
+def _log_prompts_summary(prompts, source):
+    if not prompts:
+        logger.info("Prompts summary (%s): none", source)
```
sbalandi: let's raise error in such case

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -14,13 +15,32 @@
...
+        return
+    prompt_list = prompts.get("prompts", []) if isinstance(prompts, dict) else []
+    logger.info("Prompts summary (%s): count=%d", source, len(prompt_list))
+    for idx, prompt in enumerate(prompt_list):
+        prompt_str = str(prompt).replace("\r", " ").replace("\n", " ")
+        preview = prompt_str[:200]
+        last_chars = prompt_str[-200:] if len(prompt_str) > 200 else ""
+        logger.info("Prompt %d: %s ... (last 200 chars: %s)", idx, preview, last_chars)
```
sbalandi: Function will print 400 chars of n prompts, it's a lot of information. I would remove it , user can find prompts in the files or let's print it if --verbose is set

tools/who_what_benchmark/whowhatbench/prompts/messages_short.jsonl
sbalandi, line 1: it's not valid json format, let's make it jsonl

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -230,3 +236,34 @@ def default_gen_answer(model, tokenizer, prompt, max_new_tokens, crop_question,
         df["prompt_length_type"] = 'long' if self.long_prompt else 'short'
 
         return df
+
+    def _log_greedy_generation_params(self, gen_answer_fn):
+        if self._logged_generation_params:
+            return
+
+        self._logged_generation_params = True
```
sbalandi: if you would like to print param once, move it to constructor

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -230,3 +236,34 @@ def default_gen_answer(model, tokenizer, prompt, max_new_tokens, crop_question,
...
+        elif gen_fn_name == "llamacpp_gen_text":
+            backend = "llamacpp"
+        else:
+            backend = "transformers"
+
+        params = {
+            "backend": backend,
+            "do_sample": False,
+            "max_new_tokens": self.max_new_tokens,
+            "use_chat_template": bool(self.use_chat_template),
```
sbalandi: for optimum it can be also apply_chat_template

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -230,3 +236,34 @@ def default_gen_answer(model, tokenizer, prompt, max_new_tokens, crop_question,
...
+            return
+
+        self._logged_generation_params = True
+        gen_fn_name = getattr(gen_answer_fn, "__name__", "")
+        if gen_fn_name == "genai_gen_text":
+            backend = "genai"
+        elif gen_fn_name == "llamacpp_gen_text":
+            backend = "llamacpp"
+        else:
+            backend = "transformers"
```
sbalandi: backend can be also optimum-intel

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -230,3 +236,34 @@ def default_gen_answer(model, tokenizer, prompt, max_new_tokens, crop_question,
...
+
+        params = {
+            "backend": backend,
+            "do_sample": False,
+            "max_new_tokens": self.max_new_tokens,
+            "use_chat_template": bool(self.use_chat_template),
+            "ignored_sampling_params": ["temperature", "top_p", "top_k"],
```
sbalandi: why do you need ignored_sampling_params ? do_sample: Flase equal to that

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -259,3 +259,142 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+    logger.info(f"Loaded {len(items)} records from JSON dataset")
+    return items
+
+
+def get_chat_template_for_model(model_path, tokenizer=None):
+    """Get the appropriate chat template for known models."""
+    import re
+    
+    # Extract model name from path
+    model_name_lower = model_path.lower() if model_path else ""
```
sbalandi, line 331: let's define it by model_type , model_type can be taken from AutoConfig , the path can be absolutely any

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -259,3 +259,142 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+    if chat_template_source:
+        if chat_template_source.startswith("http://") or chat_template_source.startswith("https://"):
+            import requests
+            logger.info(f"Loading chat template from URL: {chat_template_source}")
+            chat_template = requests.get(chat_template_source).text
+        else:
+            from transformers import AutoTokenizer
+            logger.info(f"Loading chat template from model: {chat_template_source}")
+            custom_tokenizer = AutoTokenizer.from_pretrained(chat_template_source)
+            chat_template = custom_tokenizer.get_chat_template()
```
sbalandi, line 372: chat_template source will be path or None, this branch looks unnecessary

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -259,3 +259,142 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+        "gpt-oss-20b": "https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_gpt_oss.jinja",
+    }
+    
+    # Find matching model
+    for key, template_source in chat_template_map.items():
+        if key in model_name_lower:
+            logger.info(f"Matched model '{key}' with chat template source: {template_source}")
+            return template_source
+    
+    logger.info(f"No hardcoded chat template found for model: {model_path}, using tokenizer")
```
sbalandi: I would avoid to log that, it's better to avoid logging too much. If we found a predefined template, then okay, but if we didn't find it, there's no need to write about it.

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -259,3 +259,142 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+    if chat_template is None:
+        logger.error("Chat template is missing; cannot render prompts for JSON dataset")
+        raise ValueError("Chat template is missing; cannot render prompts for JSON dataset.")
+    
+    for item in records:
+        messages = item["messages"]
+        tools = item["tools"]
+        
+        try:
+            jinja_env = jinja2.Environment()
```
sbalandi: why not apply_chat_template() ?
prompt = tokenizer.apply_chat_template(
    messages,
    tools=tools,
    add_generation_prompt=True,
    enable_thinking=False
    reasoning_effort="low",
    tokenize=False
)

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -259,3 +259,142 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+    for key, template_source in chat_template_map.items():
+        if key in model_name_lower:
+            logger.info(f"Matched model '{key}' with chat template source: {template_source}")
+            return template_source
+    
+    logger.info(f"No hardcoded chat template found for model: {model_path}, using tokenizer")
+    return None
+
+
+def render_messages_dataset_to_prompts(records, tokenizer, model_path=None):
```
sbalandi, line 349: wwb will have the same prompts for optimum and for genai here, but genAI also have own way to manage message and tools https://github.com/openvinotoolkit/openvino.genai/blob/master/samples/python/text_generation/README.md#12-tool-calling-with-structural-tags-sample-structural_tags_generation 
I would suggest to move creation of prompts from messages to text_evalutor.py. It can be orginized as gen_answer_fn.
Also I would suggest add separate flag for message mode. Because for optimum/transformers we also need to omit chat template in that case , for genAI message must be pass to pipeline via chat history

tools/who_what_benchmark/whowhatbench/registry.py
```diff
@@ -35,15 +36,27 @@ def score(self, model_or_data, **kwargs):
...
+ 
     @abstractmethod
     def get_generation_fn(self):
         raise NotImplementedError("generation_fn should be returned")
 
 
 class BaseEvaluator(Evaluator):
+    @staticmethod
+    def _escape_for_csv(df):
+        """Escape newlines and tabs in dataframe string values for clean CSV output."""
```
sbalandi: it will change anwer and prompts.
if prompts/answer is big or have specific structure let's save it to separate files and add to .csv lint to that files , example find  https://github.com/sbalandi/openvino.genai/blob/bc08e33ec9f331d47a4bb38977ef52e380a6f813/tools/who_what_benchmark/whowhatbench/chat_text_evaluator.py#L214

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -230,3 +236,34 @@ def default_gen_answer(model, tokenizer, prompt, max_new_tokens, crop_question,
         df["prompt_length_type"] = 'long' if self.long_prompt else 'short'
 
         return df
+
+    def _log_greedy_generation_params(self, gen_answer_fn):
```
sbalandi: I would suggest to save info to the csv , not print

sbalandi (changes requested):

tools/who_what_benchmark/whowhatbench/text_evaluator.py
sbalandi, line 1: please, add test for that case

tools/who_what_benchmark/whowhatbench/utils.py
sswierze: I had a case when model id was used as chat_template input. atm it is not used but I would like to be able to use this kind of input

PR #3486, opened 2026-03-12
[llm_bench] Avoid run pipeline on optimum if genai was required

sbalandi:
## Description
Currently if GenAI pipeline is not created successfully, llm_bench just logging it and falls back to optimum-intel pipeline. This might lead to incorrect metrics reporting and other misleading information about execution. Let's raise error in such cases.

[CVS-175415](https://jira.devtools.intel.com/browse/CVS-175415)

Fixes #(issue)

## Checklist:

### DIFF
--- a/tools/llm_bench/requirements.txt
+++ b/tools/llm_bench/requirements.txt
@@ -14,7 +14,7 @@ diffusers>=0.22.0,<=0.37.0
 optimum-intel[nncf]>=1.25.0,<=1.27.0
 packaging>=20.0,<=26.0
 psutil
-timm<=1.0.25
+timm<=1.0.26
 tiktoken>=0.3.0,<=0.12.0
 librosa==0.11.0 # For Whisper
 matplotlib>=3.3.4,<=3.10.8


tests/python_tests/utils/atomic_download.py
```diff
@@ -20,7 +20,7 @@ def is_complete(self) -> bool:
         return self.final_path.exists()
 
     def execute(self, download_fn: Callable[[Path], None]) -> None:
-        if self.is_complete():
+        if self.is_complete() and "Qwen3-Embedding" not in str(self.final_path):
```
yatarkan: Should we move "Qwen3-Embedding" check to helper and reuse?

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -122,12 +122,13 @@ def create_text_gen_model(model_path, device, memory_data_collector, **kwargs):
...
-            if model_class != UseCaseTextGen.ov_cls and "mpt" not in use_case.model_types and "chatglm" not in use_case.model_types:
-                log.warning("OpenVINO GenAI based benchmarking is not available for required model type. Will be switched to default benchmarking")
-            else:
-                log.info("Selected OpenVINO GenAI for benchmarking")
-                return create_genai_text_gen_model(model_path, device, ov_config, memory_data_collector, **kwargs)
+        if kwargs.get("genai", True):
+            if not is_genai_available(log_msg=True):
+                raise RuntimeError("OpenVINO GenAI based benchmarking is required, but not available.")
+            if (model_class != UseCaseTextGen.ov_cls and "mpt" not in use_case.model_types and "chatglm" not in use_case.model_types):
+                raise RuntimeError("OpenVINO GenAI based benchmarking is not available for required model type.")
```
yatarkan, line 133: Does it make sense to print required model type (here and in similar errors) for easier debugging?

tests/python_tests/utils/atomic_download.py
sbalandi: This is a debug check. The test for embedding with this model fails. I assumed it's due to a wrong conversion parameters. CI doesn't re-convert a model if it exists. So I hard-coded model name to force re-convertion, but that didn't help, so I'm continuing to investigate.

tools/llm_bench/llm_bench_utils/ov_utils.py
sbalandi: good idea, I will add model type

sbalandi: just remembered that model_type is always output as general information in the log, for example:
```
python tools/llm_bench/benchmark.py -m ./models/TinyLlama-1.1B-Chat-v1.0_int8/ -n 1
[ INFO ] ==SUCCESS FOUND==: use_case: text_gen, model_type: llama, model_Name: TinyLlama-1.1B-Chat-v1.0_int8
```

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -122,12 +122,17 @@ def create_text_gen_model(model_path, device, memory_data_collector, **kwargs):
...
-            else:
-                log.info("Selected OpenVINO GenAI for benchmarking")
-                return create_genai_text_gen_model(model_path, device, ov_config, memory_data_collector, **kwargs)
+        if kwargs.get("genai", True):
+            if not is_genai_available(log_msg=True):
+                raise RuntimeError("OpenVINO GenAI based benchmarking is required, but not available.")
+            if (
+                model_class != UseCaseTextGen.ov_cls
+                and "mpt" not in use_case.model_types
+                and "chatglm" not in use_case.model_types
```
as-suvorov, line 131: Could you please explain why this check needed?

tests/python_tests/samples/conftest.py
```diff
@@ -332,7 +332,10 @@ def download_to_temp(temp_path: Path) -> None:
     else:
         converted_models_dir = get_ov_cache_converted_models_dir()
         model_cache = converted_models_dir / model_id
-        model_path = model_cache / model_name
+        if "alias" in model:
+            model_path = model_cache / model["alias"]
```
as-suvorov: I think we should solve this either on atomic download or on CI side.
@sgonorov Could you please take a look at the issue from atomic download side.

tools/llm_bench/llm_bench_utils/ov_utils.py
sbalandi: to avoid run models which are not supported by GenAI with GenAI

PR #3555, opened 2026-03-24
[Image generation pipeline] Add `m_is_blob` flag to handle infer request when import the blob to NPU 

JohnLeFeng:
## Description

Before this change, users cannot run the image generation pipeline with NPU Unet blob when guidence scale > 1, as pipeline cannot make sure the correct batch size of UNet. Therefore, we only can support guidence scale < 0.5. However, with flag, `m_is_blob`, we can update infer request based on batch of hidden state size.

[CVS-174803](https://jira.devtools.intel.com/browse/CVS-174803)

Fixes #(issue)

## Checklist:

### DIFF
--- a/.github/skills/genai-model-checker/SKILL.md
+++ b/.github/skills/genai-model-checker/SKILL.md
@@ -1,0 +1,92 @@
+---
+name: genai-model-checker
+description: "Validate a newly supported optimum-intel model with OpenVINO GenAI. Use when: checking new model support, verifying model export to OpenVINO IR, running GenAI inference test with llm_bench, benchmarking model accuracy with who-what-benchmark."
+argument-hint: "model_id and task (e.g. tencent/HY-MT1.5-1.8B text-generation-with-past)"
+---
+
+# GenAI Model Checker
+
+Validates that a HuggingFace model exported via optimum-intel works correctly with OpenVINO GenAI pipelines and passes accuracy benchmarks.
+
+## When to Use
+
+- A new model was added to optimum-intel and needs GenAI validation
+- Verify a HuggingFace model exports to OpenVINO IR and runs inference
+- Check model accuracy after conversion using who-what-benchmark
+
+## Inputs
+
+The user must provide:
+
+- **model_id**: HuggingFace model identifier (e.g. `tencent/HY-MT1.5-1.8B`)
+- **task**: optimum-cli export task. Supported values:
+  - `text-generation-with-past`
+  - `image-text-to-text`
+  - `text-to-image`
+  - `image-to-image`
+  - `feature-extraction`
+  - `text-classification`
+  - `text-to-video`
+  - `automatic-speech-recognition`
+
+## Prerequisites
+
+Activate the Python virtual environment before running any commands.
+
+1. **Locate the virtual environment** — check for common directories at the repository root: `.venv/`, `venv/`, `env/`. Use `list_dir` to find it. If none is found, ask the user for its location.
+2. **Activate** based on the current platform:
+   - **Linux/macOS**: `source <venv_path>/bin/activate`
+   - **Windows (cmd)**: `<venv_path>\Scripts\activate.bat`
+   - **Windows (PowerShell)**: `<venv_path>\Scripts\Activate.ps1`
+
+## Procedure
+
+### Step 1: Run check_model.py
+
+Run the checker script from the repository root:
+
+```
+python3 .github/skills/genai-model-checker/scripts/check_model.py \
+    --model-id <model_id> \
+    --task <export_task> \
+    --work-dir /tmp/genai-model-check
+```
+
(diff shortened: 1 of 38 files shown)


src/cpp/src/image_generation/models/unet_inference_static_bs1.hpp
```diff
@@ -72,6 +72,19 @@ class UNet2DConditionModel::UNetInferenceStaticBS1 : public UNet2DConditionModel
...
+                      << "The batch size of encoder hidden states does not match the batch size of native, therefore, "
+                      << "update native batch size to align batch size of encoder hidden states."
+                      << std::endl;
+            auto compiled_model = m_requests[0].get_compiled_model();
+            m_native_batch_size = encoder_hidden_states_bs;
+            
+            m_requests.resize(m_native_batch_size);
+            for (int i = 0; i < m_native_batch_size; i++) {
+                m_requests[i] = compiled_model.create_infer_request();
+            }
```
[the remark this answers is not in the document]
JohnLeFeng: I think we can use comment in here.

src/cpp/src/image_generation/models/unet_inference_static_bs1.hpp
```diff
@@ -167,22 +183,22 @@ class UNet2DConditionModel::UNetInferenceStaticBS1 : public UNet2DConditionModel
...
+
         m_native_batch_size = compiled_model.input("sample").get_shape()[0];
         m_requests.resize(m_native_batch_size);
 
         ov::genai::utils::print_compiled_model_properties(compiled_model, "UNet 2D Condition batch-1 model");
 
-        for (int i = 0; i < m_native_batch_size; i++) {
+        for (size_t i = 0; i < m_native_batch_size; i++) {
             m_requests[i] = compiled_model.create_infer_request();
         }
```
as-suvorov, line 210: As I understand with the proposed approach we set requests twice: once for `import_model` and once for `set_hidden_states`. Can we set requests only once?

JohnLeFeng: Yes, it will set infer requests twice. Actually, I also would like to initiate infer request once, however, it will introduce a new private parameter for `compile_model` and it only be used in this case. Therefore, I think re-creating the infer requests at `set_hidden_states` is better option. How do you think?

JohnLeFeng: or do you think we can simplify it at `import_model` function as it must be batch size `1` model?

JohnLeFeng:
> Could you please add a python test. Please note that CI has CPU device only, so CPU fallback for NPU device is required.

Thanks and added test for blob model.

src/cpp/src/image_generation/models/unet_inference_static_bs1.hpp
```diff
@@ -27,6 +28,9 @@ class UNet2DConditionModel::UNetInferenceStaticBS1 : public UNet2DConditionModel
         auto clone = std::make_shared<UNetInferenceStaticBS1>();
         clone->m_native_batch_size = m_native_batch_size;
         clone->m_requests.reserve(m_requests.size());
+        clone->m_is_blob = m_is_blob;
+        clone->m_adapter_controller = m_adapter_controller;
+        clone->m_last_adapter_config = m_last_adapter_config;
```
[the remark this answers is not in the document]
JohnLeFeng: Updated this, but I think we need to discuss about whether to have a follow up PR for keep adapter when use clone(). Both [`unet_inference_static_bs1.hpp`](https://github.com/openvinotoolkit/openvino.genai/pull/3555/files/06cc6724e34918ee75ea6eb2c384cd9bdc7ef4dc#diff-6520883bece9761476a6bbb4a25dab2709daeb3c4600aab3d2b51aae713bac07) and [`unet_inference_dynamic.hpp`](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/image_generation/models/unet_inference_dynamic.hpp) are not cloning the adapter.

as-suvorov:
@RyanMetcalfeInt8 We tested different approaches and single `compile_model` showed best perf. We created ticket to investigate `read` + `compile` vs `compile` behavior: 172563

JohnLeFeng:
Hi team, could you please re-trigger Ci? I think the issue can be fixed by #3580

as-suvorov:
> Hi team, could you please re-trigger Ci? I think the issue can be fixed by #3580

Could you please update branch to master, this should solve CI fail

JohnLeFeng:
> > Hi team, could you please re-trigger Ci? I think the issue can be fixed by #3580
> 
> Could you please update branch to master, this should solve CI fail

Updated.

as-suvorov:
build_jenkins

as-suvorov:
build_jenkins

PR #3564, opened 2026-03-24
[llm_bench] [wwb] Update torch related deps

wkobielx:
After recent releases:
- https://pypi.org/project/torch/#history
- https://pypi.org/project/torchaudio/#history
- https://pypi.org/project/torchvision/#history

We are facing issues in validation when installing llm_bench/wwb together with custom optimum-intel, where torch is not as restricted: `"torch>=2.1",`

```
09:08:37  torchvision 0.25.0+cpu requires torch==2.10.0, but you have torch 2.11.0+cpu which is incompatible.
09:08:37  torchaudio 2.10.0+cpu requires torch==2.10.0, but you have torch 2.11.0+cpu which is incompatible.
```

### DIFF
--- a/tools/llm_bench/requirements.txt
+++ b/tools/llm_bench/requirements.txt
@@ -7,7 +7,7 @@ openvino-tokenizers
 openvino_genai
 # Image processing for VLMs
 pillow==12.1.1
-torch>=2.1.0,<=2.10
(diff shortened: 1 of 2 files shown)


wkobielx:
Probably need to port this to releases/2026/1 branch before RC2

PR #3565, opened 2026-03-24
[llm_bench] [wwb] Update torch related deps - port to releases/2026/1

wkobielx:
After recent releases:

- https://pypi.org/project/torch/#history
- https://pypi.org/project/torchaudio/#history
- https://pypi.org/project/torchvision/#history

We are facing issues in validation when installing llm_bench/wwb together with custom optimum-intel, where torch is not as restricted: "torch>=2.1",
```
09:08:37  torchvision 0.25.0+cpu requires torch==2.10.0, but you have torch 2.11.0+cpu which is incompatible.
09:08:37  torchaudio 2.10.0+cpu requires torch==2.10.0, but you have torch 2.11.0+cpu which is incompatible.
```

Related PR to master: https://github.com/openvinotoolkit/openvino.genai/pull/3564

### DIFF
--- a/tools/llm_bench/requirements.txt
+++ b/tools/llm_bench/requirements.txt
@@ -7,7 +7,7 @@ openvino-tokenizers
 openvino_genai
 # Image processing for VLMs
 pillow==12.1.1
-torch>=2.1.0,<=2.10
+torch>=2.1.0,<=2.11
 transformers[sentencepiece]>=4.40.0,<5.4.0
 diffusers>=0.22.0,<=0.37.0
 # optimum is in dependency list of optimum-intel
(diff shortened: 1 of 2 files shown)


as-suvorov:
@wkobielx In order to merge to release branch technical committee approval is required.
Also issue you facing in validation is not clear. Optimum-intel has [`torch>=2.1`](https://github.com/huggingface/optimum-intel/blob/main/setup.py#L30C6-L30C11) which inline with current release dependencies `torch>=2.1.0,<=2.10`

wkobielx:
@as-suvorov okay, leave it for now, I will sync internally and try to fix it on pipeline side

PR #3572, opened 2026-03-26
Apply clang-format and cmake-format to all C, C++ and CMake files

goyaladitya05:
This PR includes a one-time bulk reformat of all C, C++ and CMake files to bring the codebase in line with the existing [.clang-format](vscode-webview://0d65538tdp3k8kv7gkoh92mugr1tdiv2hqjl3h3tcqgsh5qgmbtn/.clang-format) config (Google style, 4-space indent, 120 column limit) and cmake-format defaults.

The .clang-format config was introduced in #312 and applied at that time, but has never been enforced since. This commit re-applies it to all files that have drifted.

A follow-up PR #3573 will add `clang-format` and `cmake-format` pre-commit hooks to enforce formatting going forward.

## Checklist:
- I have made corresponding changes to the documentation. NA

### DIFF
--- a/.cmake-format.yaml
+++ b/.cmake-format.yaml
@@ -1,0 +1,6 @@
+format:
+  line_width: 120
+  tab_size: 2
+
+markup:
+  first_comment_is_literal: true


goyaladitya05:
@Wovchena Since there are a lot of modified files, I want to confirm, is this the right approach? Is there a good time to merge this to minimize conflicts with other PRs? I'll add the clang-format to pre-commit hooks in a follow up PR (#3573), wanted to keep this strictly for one time formatting.

I think this could also be split directory wise, to make it smaller, and would reduce merge conflicts.

I'm following the pre-commit introductions in #2980, but one at a time.

Wovchena:
@sgonorov, @as-suvorov, please advice

sgonorov:
clang-format is needed, so i think it's an okay to migrate all of the codebase, the release has passed just recently so it's the best time. Also you can introduce cmake format (as it's closely related)
```yaml
  - repo: https://github.com/cheshirekow/cmake-format-precommit
    rev: v0.6.13
    hooks:
      - id: cmake-format
        types: [cmake]
```

.cmake-format.yaml
goyaladitya05, line 1: Used these settings for the cmake-format.

`first_comment_is_literal: true` is used to ensure the copyright statement does not change.
Without it, cmake-format was reflowing the two lines into one.

```
# Copyright (C) 2018-2026 Intel Corporation
```

to 

```
# Copyright (C) 2018-2026 Intel Corporation SPDX-License-Identifier: Apache-2.0
```

as-suvorov:
@sgonorov We decided not to lint the whole python codebase as we have incremental linter. From my pov we should be aligned here. Either introduce incremental c++ linter, or lint the whole codebase for python as well.
I don't mind linting the whole codebase.

The linting should also introduce pre-commit and CI checks in the same PR.
I also have concerns regarding reviewing 18k changed lines. We can split PR into several smaller ones and manually review them. Or linting should be done by GenAI team member.

sgonorov:
@as-suvorov Sadly i haven't found an incremental C++ formatter. This is not linting, clang-tidy is a linter, and it works fine, but it will also review the whole file, not just touched parts. So it's the only way i think. Splitting is possible but do we need it? It will take much more time to do.

goyaladitya05:
> @as-suvorov Sadly i haven't found an incremental C++ formatter. This is not linting, clang-tidy is a linter, and it works fine, but it will also review the whole file, not just touched parts. So it's the only way i think. Splitting is possible but do we need it? It will take much more time to do.
  
  @sgonorov Actually, `git clang-format --staged` formats only the staged diff (changed lines), not the whole file. I verified this locally. It's the C++ equivalent of darker for Python. Could you please check it out?

goyaladitya05:
I'm searching for an incremental formatter. But this one time formatting would need to be done anyways. I'm willing to do this directory/subdirectory wise in seperate PRs, although I think the current approach would be better.

I understand if you prefer to handle it internally.

as-suvorov:
Another concern is non incremental formatting would lead to conflicts in PRs. As I remember it was one of the reasons we have not formatted all python files.
@goyaladitya05 please give us some time to re-iterate this

PR #3573, opened 2026-03-26
[GHA] Add incremental clang-format pre-commit hook for C/C++

goyaladitya05:
This PR adds a clang-format pre-commit hook based on `git-clang-format`, which parses the git diff and passes `--lines start:end` to clang-format, formatting only the touched lines rather than whole files. This was released in clang-format 23.1.0, and was added to LLVM in llvm/llvm-project#188813.

The `--extensions` allowlist on the hook entry is required because `git-clang-format` otherwise forwards every language clang-format supports, so with the codebase's `Cpp`-only .clang-format, a PR touching .js/.json files alongside C++ made clang-format error and the hook abort. Restricting it to C/C++ extensions keeps clang-format out of JS/JSON entirely (JS is formatted by Prettier in src/js; pre-commit only validates JSON via check-json). Per-language `DisableFormat` sections in .clang-format would also work, but make it multi-document YAML, which the existing check-yaml hook rejects.

Validated this on my fork: a misformatted change fails with the diff scoped to touched files only (goyaladitya05/openvino.genai#193), the same change correctly formatted passes (goyaladitya05/openvino.genai#194), and a .json + .cpp change in one PR is handled cleanly (goyaladitya05/openvino.genai#195). Pre-existing drift in untouched files (e.g. chat_history.hpp, imwrite.cpp) is never reported.

## Checklist:

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -8,7 +8,7 @@
 7. Contributions with use of AI must comply with [OpenVINO's AI Usage Policy](https://github.com/openvinotoolkit/openvino/blob/c4f4325c57977c684184e758449d1f8825ebbfd7/AI_USAGE_POLICY.md).
 
 # Code Quality
-This repository uses [pre-commit](https://pre-commit.com/) hooks to catch common issues early (JSON/YAML syntax errors, merge conflict markers, line endings, etc.).
+This repository uses [pre-commit](https://pre-commit.com/) hooks to catch common issues early (JSON/YAML syntax errors, merge conflict markers, line endings, C/C++ formatting on changed lines, etc.).
 
 To set up locally:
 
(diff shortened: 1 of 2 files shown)


goyaladitya05:
Closing the PR since this is being handled by the genai team.

goyaladitya05:
Hi @Wovchena, @as-suvorov and @sgonorov,
could we reopen this discussion from #3572? Previously, the blocker was that no incremental C++ formatter existed. However, it's been introduced in clang-format 23.1.0. `git-clang-format` natively supports pre-commit's `--from-ref`/`--to-ref` (llvm/llvm-project#188813), so it formats only the lines a PR touches, sort of the cpp equivalent of what darker does for Python.

There's no mention of this in the release notes, as I recently learned that for LLVM, the modifications needs to be done by the author themselves, and I didn't know it when I made the PR (I introduced this feature in LLVM).

I've updated this branch to introduce it, and mentioned more details in the PR desc.

PR #3600, opened 2026-03-29
fix: replace no-op OPENVINO_ASSERT with OPENVINO_THROW for unsupported type branches

VinKarwal:
image
As seen in this screenshot, OPENVINO_ASSERT uses condition to run, in case of the code where i have replaced OPENVINO_ASSERT with OPENVINO_THROW, the string literal is used as a condition rather than a message, and as we know that any non-empty string as a condition is true in C++, the block never executes and hence replaced with OPENVINO_THROW in these error branches which will throw the exception nonetheless.


VinKarwal:
For the pipeline branches checked the existing codebase patterns and `OPENVINO_ASSERT(false, ...)` is consistently used, hence aligned with that convention. 

Also accepted GGUF tensor name/type suggestion for better debugging.

as-suvorov (approved):

as-suvorov:
build_jenkins

as-suvorov:
build_jenkins

PR #3362, opened 2026-02-20
[Video Generation] Add LoRA adapter support for Text2VideoPipeline (Supersedes #3309)

goyaladitya05:
This PR implements LoRA adapter support for `Text2VideoPipeline`.

## Changes

C++ (LTXVideoTransformer3DModel)
- Removed the blocking assertion in compile() and initialized AdapterController following the FluxTransformer2DModel pattern
- Implemented set_adapters() with a safety check to prevent segfaults on uninitialized controllers
- Added `std::optional<AdapterConfig> adapters` to VideoGenerationConfig

Pipeline (ltx_pipeline.hpp)
- Added m_transformer->set_adapters() before the denoising loop. Adapters were previously extracted from properties but never applied.
- Fixed rebuild_models() to forward m_compile_properties so adapters survive model reshapes

Python
- Exposed LTXVideoTransformer3DModel.set_adapters() via pybind11 and updated the stub files

Samples & docs
- Added C++ and Python samples
- Updated both READMEs with LoRA usage instructions

Tests
- Added TestLoRAVideoGeneration covering constructor, generate(), and set_adapters(). All tests use an empty AdapterConfig() and run without real LoRA files

## Verification
without LoRA
564786724-f3189e62-e088-4aea-8de0-50048622272e

with LoRA 
Adapter:[(ltx_pixel_pytorch_lora_weights.safetensors)](https://huggingface.co/svjack/ltx_video_pixel_early_lora)
564786695-3a0c0714-b14f-4881-99bd-be69f5861bc2

These samples demonstrate that the LoRA adapter is successfully loaded and applied during the denoising process in the Text2VideoPipeline.

Environment:
Model: LTX-Video
Alpha: 3.0
Prompt: "In the style of Pixel, the video shifts to a majestic castle under a starry sky."

## Checklist:
- [x] This PR follows [GenAI Contributing guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing). 
- [x] Tests have been updated or added to cover the new code. 
    * Verified `tests/python_tests/test_video_generation.py` passes locally.
- [x] This PR fully addresses the ticket. 
- [x] I have made corresponding changes to the documentation.

Closes #3306
Supersedes #3309

PR #3605, opened 2026-03-30
Add LoRA adapter support for text encoders in SD3 and Flux pipelines

goyaladitya05:
## Description
This PR resolves the `// TODO: Add LoRA Adapter support for text encoders` in stable_diffusion_3_pipeline.hpp.

- SD3: CLIP1, CLIP2, and T5-XXL text encoders now receive LoRA weights during set_lora_adapters, with per-encoder prefix filtering (lora_te1, lora_te2, lora_te3).

- Flux: T5 encoder (text_encoder_2) is now included in set_lora_adapters with lora_te2 prefix filtering. CLIP (text_encoder) prefix is corrected from lora_te to lora_te1, matching the kohya sd-scripts convention used by community Flux LoRAs (same as SDXL). 
e.g https://huggingface.co/obann001/Saraptah_Flux_LoRA_v1 , https://huggingface.co/Shakker-Labs/FLUX.1-dev-LoRA-Dark-Fantasy

- T5EncoderModel: Refactored compile to use AdapterController (dynamic LoRA) instead of static fusing, and added set_adapters, following the same pattern as CLIPTextModel.

## Checklist:
- I have made corresponding changes to the documentation. (Did not find any seperate docs for this case.)


goyaladitya05:
@likholat I've been working on this for a while, and wanted to attach a sample similar to how it's done in #3362. I tried looking for LoRA adapters, but couldn't find any suitable ones. Could you please share the link of an adapter I could use for sample generation and comparison?

src/cpp/src/image_generation/stable_diffusion_3_pipeline.hpp
```diff
@@ -141,24 +141,28 @@ class StableDiffusion3Pipeline : public DiffusionPipeline {
         using utils::read_json_param;
 
         set_scheduler(Scheduler::from_config(root_dir / "scheduler/scheduler_config.json"));
+        auto updated_properties = update_adapters_in_properties(properties, &StableDiffusion3Pipeline::derived_adapters);
         const std::string text_encoder = data["text_encoder"][1].get<std::string>();
         if (text_encoder == "CLIPTextModelWithProjection") {
             m_clip_text_encoder_1 =
```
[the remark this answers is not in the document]
goyaladitya05: updated

src/cpp/src/image_generation/models/t5_encoder_model.cpp
```diff
@@ -72,7 +72,12 @@ T5EncoderModel& T5EncoderModel::reshape(int batch_size, int max_sequence_length)
 
 T5EncoderModel& T5EncoderModel::compile(const std::string& device, const ov::AnyMap& properties) {
     OPENVINO_ASSERT(m_model, "Model has been already compiled. Cannot re-compile already compiled model");
-    ov::CompiledModel compiled_model = utils::singleton_core().compile_model(m_model, device, *extract_adapters_from_properties(properties));
+    std::optional<AdapterConfig> adapters;
+    auto filtered_properties = extract_adapters_from_properties(properties, &adapters);
+    if (adapters) {
```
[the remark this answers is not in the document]
goyaladitya05: no, T5EncoderModel is used in both SD3 (lora_te3) and Flux (lora_te2), so there's no single correct default prefix.

goyaladitya05:
cc @sgonorov @likholat

src/cpp/src/image_generation/stable_diffusion_3_pipeline.hpp
```diff
@@ -783,6 +792,19 @@ class StableDiffusion3Pipeline : public DiffusionPipeline {
         }
     }
 
+    utils::SharedOptional<const ov::AnyMap> properties_for_text_encoder(
+            const ov::AnyMap& properties, const std::string& tensor_name_prefix) {
+        return update_adapters_in_properties(properties,
```
sgonorov: Maybe it can be move to some utils? I see it's being copied across code.

tests/python_tests/test_image_generation.py
```diff
@@ -192,6 +194,82 @@ def callback(step, num_steps, latent):
         assert len(callback_calls) > 0
 
 
+class TestLoRAFluxImageGeneration:
```
sgonorov, line 197: Adapter config is always empty in these tests - maybe create some mock one? Otherwise it's not testing much.

tests/python_tests/conftest.py
```diff
@@ -76,6 +76,7 @@ def pytest_configure(config: pytest.Config):
     "tiny-random-latent-consistency": "echarlaix/tiny-random-latent-consistency",
     "tiny-random-flux": "optimum-intel-internal-testing/tiny-random-flux",
     "tiny-random-sdxl": "echarlaix/tiny-random-stable-diffusion-xl",
+    "stable-diffusion-3-tiny-random": "optimum-intel-internal-testing/stable-diffusion-3-tiny-random",
```
sgonorov:
```suggestion
    "tiny-random-stable-diffusion-3": "optimum-intel-internal-testing/stable-diffusion-3-tiny-random",
```
Feels better for consistency.

sgonorov (changes requested):

likholat:
Hi @goyaladitya05,

> @likholat I've been working on this for a while, and wanted to attach a sample similar to how it's done in https://github.com/openvinotoolkit/openvino.genai/pull/3362. I tried looking for LoRA adapters, but couldn't find any suitable ones. Could you please share the link of an adapter I could use for sample generation and comparison?

I couldn't find such a LoRA adapter either, so it seems we don't need to support this case.

src/cpp/src/image_generation/stable_diffusion_3_pipeline.hpp
goyaladitya05: done

goyaladitya05:
> I couldn't find such a LoRA adapter either, so it seems we don't need to support this case.

Hi @likholat , I'd like to clarify why text encoder LoRA support matters here.

Text encoder LoRA support is already implemented in the codebase for SDXL. This PR completes the same pattern for SD3 and Flux.

Please correct me if I'm wrong here. Another reason why this matter would be that many LoRA adapters include text encoder weights alongside the transformer/UNet weights. If the pipeline silently ignores the text encoder LoRA weights, users will experience text encoder misalignment, the text encoding space drifts from what the LoRA was trained on, leading to degraded prompt adherence and style consistency. This is a known and documented issue (e.g. [https://www.reddit.com/r/StableDiffusion/comments/1hrno1j/global_text_encoder_misalign](https://www.reddit.com/r/StableDiffusion/comments/1hrno1j/global_text_encoder_misalignment_potential/)

Since LoRA is supported for SD3 and Flux, then this should also be supported as this is a part of that.

tests/python_tests/test_image_generation.py
```diff
@@ -192,6 +195,125 @@ def callback(step, num_steps, latent):
...
+    @pytest.mark.parametrize("image_generation_model", [FLUX_MODEL_ID], indirect=True)
+    def test_lora_adapters_constructor(self, image_generation_model):
+        """Test that a real LoRA adapter can be passed to the Flux pipeline constructor"""
+        lora_path = download_hf_file("goyaladitya05/openvino-genai-test-files", "flux_tiny_dummy_lora.safetensors")
+        adapter = ov_genai.Adapter(lora_path)
+        adapter_config = ov_genai.AdapterConfig(adapter)
+        pipe = ov_genai.Text2ImagePipeline(image_generation_model, "CPU", adapters=adapter_config)
+        assert pipe is not None
```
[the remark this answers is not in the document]
goyaladitya05: added lora_te* tensors to all three dummy LoRA hf adapters

tests/python_tests/utils/hugging_face.py
```diff
@@ -365,23 +365,28 @@ def convert_to_temp(temp_path: Path) -> None:
     )
 
 
-def download_gguf_model(
-    gguf_model_id: str,
-    gguf_filename: str,
-):
-    gguf_dir_name = sanitize_model_id(gguf_model_id)
+def download_hf_file(
```
goyaladitya05, line 368: renamed this to avoid confusion, as its a generic hf file downloader (ref #3718)

tests/python_tests/test_image_generation.py
goyaladitya05: updated.

goyaladitya05:
after more research, I agree. While my previous analysis holds true, but text encoders LoRA weights are rare, and I couldn't find a popular LoRA to show the comparison. I'm going to close this PR, thanks.

> > I couldn't find such a LoRA adapter either, so it seems we don't need to support this case.
> 
> Hi @likholat , I'd like to clarify why text encoder LoRA support matters here.
> 
> Text encoder LoRA support is already implemented in the codebase for SDXL. This PR completes the same pattern for SD3 and Flux.
> 
> Please correct me if I'm wrong here. Another reason why this matter would be that many LoRA adapters include text encoder weights alongside the transformer/UNet weights. If the pipeline silently ignores the text encoder LoRA weights, users will experience text encoder misalignment, the text encoding space drifts from what the LoRA was trained on, leading to degraded prompt adherence and style consistency. This is a known and documented issue (e.g. [https://www.reddit.com/r/StableDiffusion/comments/1hrno1j/global_text_encoder_misalign](https://www.reddit.com/r/StableDiffusion/comments/1hrno1j/global_text_encoder_misalignment_potential/)
> 
> Since LoRA is supported for SD3 and Flux, then this should also be supported as this is a part of that.

PR #3630, opened 2026-03-31
[Bug Fix] LTX-Video vae_scale_factor formula [Run checks]

Yash-Vijay29:
## Description

Fix VAE scale factor computation to correctly account for patch-based spatial compression.

Previously, vae_scale_factor did not include patch_size, which led to incorrect spatial scaling when mapping between latent and pixel space. This change updates the computation to:

scale = patch_size * 2^(sum(spatio_temporal_scaling))

This aligns the scale factor with how spatial compression is applied in the VAE (patch embedding + hierarchical downsampling).

Note:

This change reflects spatial scaling only, which is sufficient for the current text-to-video pipeline.
Temporal scaling is handled separately via patch_size_t and is not included in this function.
The definition may need to be extended in the future to explicitly represent both spatial and temporal scaling for image-to-video or bidirectional pipelines.

Fixes #3629

## Checklist:


sgonorov (approved):
Looks like at autoencoder_kl_ltx_video.cpp and ltx_pipeline.hpp we're doing the same but using std::pow variant. Maybe it's a good idea to make all of them the same way, also extract an utility to avoid duplication. But it's a minor thing.

Yash-Vijay29:
> Looks like at autoencoder_kl_ltx_video.cpp and ltx_pipeline.hpp we're doing the same but using std::pow variant. Maybe it's a good idea to make all of them the same way, also extract an utility to avoid duplication. But it's a minor thing.

Okay i ill finish that!

Yash-Vijay29:
@sgonorov @likholat  hey added a helper function and did deduplication.  Does it seem fine?

src/cpp/src/video_generation/compression_utils.hpp
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2025-2026 Intel Corporation
```
likholat:
```suggestion
// Copyright (C) 2026 Intel Corporation
```

likholat (approved):

likholat:
@Yash-Vijay29 please update the branch

Yash-Vijay29:
@likholat  done! run checks please

Yash-Vijay29:
@likholat  Is it normal for it to fail to build for ubuntu 22 python 3.10, 3.12 and 3.13 but work for 3.11-

Yash-Vijay29:
@likholat All tests passed last time, branch went out of date so updated, please run checks.

Yash-Vijay29:
Hey @likholat  does this formula fix seem fine? or is something incorrect still?

PR #3648, opened 2026-04-03
perf: Add T5 prompt embedding caching to LTX pipeline

Ashitpatel001:
This PR implements a stateful cache for the T5 text encoder within the `LTXPipeline` to accelerate iterative video generation.

During workflows where users keep the text prompt constant but iterate on other parameters (like the random seed or `num_inference_steps`), the pipeline previously recomputed the exact same text embeddings on every call. This PR introduces an LRU-style caching mechanism inside `compute_hidden_states` to eliminate this redundant computation.

**Key Implementation Details:**
* **Cache Evaluation:** The pipeline now strictly checks `positive_prompt`, `negative_prompt`, `do_classifier_free_guidance`, and `max_sequence_length`. 

* **Zero-Cost Hits:** On a cache hit, the T5 inference is bypassed entirely, and the telemetry metric `m_perf_metrics.encoder_inference_duration["text_encoder"]` is explicitly set to `0`.

* **Memory Safety:** On a cache miss, the resulting tensors are deep-copied into state variables using `copy_to()`. This strict memory isolation prevents downstream tensor mutations from corrupting the saved cache.

Resolves #3647

### Checklist:

### DIFF
--- a/src/cpp/src/video_generation/ltx_pipeline.hpp
+++ b/src/cpp/src/video_generation/ltx_pipeline.hpp
@@ -310,6 +310,14 @@ class Text2VideoPipeline::LTXPipeline {
     std::string m_vae_device;
     ov::AnyMap m_compile_properties;
 
+    // Prompt Caching State
+    std::string m_cached_positive_prompt;
+    std::string m_cached_negative_prompt;
+    bool m_cached_do_cfg = false;
+    size_t m_cached_max_sequence_length = 0;
+    ov::Tensor m_cached_prompt_embeds;
+    ov::Tensor m_cached_attention_mask;
+
     ov::Tensor prepare_latents(const ov::genai::VideoGenerationConfig& generation_config,
                                size_t num_channels_latents,
                                size_t transformer_spatial_patch_size,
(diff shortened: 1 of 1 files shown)


as-suvorov (changes requested):
Please apply original `pull_request_template.md`

Ashitpatel001:
Thanks @as-suvorov  for catching that. I accidentally overwrote the default template while formatting the initial description. I have now updated the main PR body to include the standard checklist.

Regarding the unchecked boxes:

Documentation: I left this unchecked because this is strictly an internal engine optimization. The public API surface remains exactly the same, so user facing documentation does not require updating.

Tests: The existing test suite will verify that the pipeline still produces correct outputs without regression. However, I left this box unchecked for now to ask: would you like me to add a dedicated Python or C++ unit test specifically to assert the cache hit and miss execution paths? I am fully prepared to write and push those tests if you want them in the suite.

Ashitpatel001:
@as-suvorov , @likholat  Just following up on this PR. I updated the description with the standard template last week.

Could you let me know if you would like me to add those explicit cache hit unit tests, or if we are good to proceed with reviewing the core caching architecture? I am fully ready to execute whichever direction you prefer.

PR #3655, opened 2026-04-07
Sswierze long json with chattemple agent

sswierze:
## Description

Latest commit adds a full WWB text-agent evaluation feature, including a new text_agent_evaluator, JSON message prompt datasets, and integration updates across CLI/registry/model-loading/metrics/tests.

CVS-181518

Fixes #(issue)

## Checklist:


tools/who_what_benchmark/tests/test_cli_text.py
```diff
@@ -360,3 +360,52 @@ def test_text_chat_model(model_id, tmp_path):
...
+        gt_path,
+        "--dataset",
+        dataset_path,
+        "--model-type",
+        "text-agent",
+        "--num-samples",
+        "1",
+        "--device",
+        "CPU",
+        "--hf",
```
sbalandi, line 405: could you please add tests also for GenAI and optimum ?

tools/who_what_benchmark/whowhatbench/prompts/messages_short.jsonl
```diff
@@ -0,0 +1,8 @@
+{"messages":[{"role":"user","content":"Find all the roots of a quadratic equation given coefficients a = 3, b = -11, and c = -4."}],"model":"ovms-model","max_completion_tokens":2048,"store":false,"temperature":0.0,"tool_choice":"auto","tools":[{"type":"function","function":{"name":"solve_quadratic","description":"Solve a quadratic equation given coefficients a, b, and c. If optional 'root_type' is 'real', the function will only return real roots. If not specified, function may return complex roots. Note that the provided function is in Python 3 syntax.","parameters":{"type":"object","properties":{"a":{"type":"integer","description":"The coefficient of the squared term in the quadratic equation."},"b":{"type":"integer","description":"The coefficient of the linear term in the quadratic equation."},"c":{"type":"integer","description":"The constant term in the quadratic equation."},"root_type":{"type":"string","description":"The type of roots to return: 'real' for real roots, 'all' for both real and complex roots. Default value is 'real'."}},"required":["a","b","c"]}}}]}
```
sbalandi, line 1: do we need "model", "store" here in default data ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,521 @@
+# Copyright (C) 2023-2026 Intel Corporation
```
sbalandi, line 1:
```suggestion
# Copyright (C) 2026 Intel Corporation
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,521 @@
...
+            kwargs["enable_thinking"] = False
+        if "reasoning_effort" in chat_template:
+            kwargs["reasoning_effort"] = "low"
+
+    def _generate_non_genai(self, model, tokenizer, record: Dict[str, Any]) -> str:
+        messages = record.get("messages")
+        if not isinstance(messages, list):
+            raise ValueError("Each JSON record must contain a 'messages' list")
+        if len(messages) == 0:
+            raise ValueError("Each JSON record must contain a non-empty 'messages' list")
```
sbalandi, line 373: Could you please move it to _generate_data ? it is the same code as in _generate_genai

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,521 @@
...
+                try:
+                    tokenizer_chat_template = tokenizer.get_chat_template()
+                except Exception:
+                    tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
+            else:
+                tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
+            self._apply_thinking_controls_for_template(chat_kwargs, tokenizer_chat_template)
+            inputs = tokenizer.apply_chat_template(messages, **chat_kwargs).to(device)
+
+        if "input_ids" not in inputs or inputs["input_ids"].shape[-1] == 0:
```
sbalandi, line 406: Can this really happen when the message is not empty?

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -307,3 +307,55 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+        logger.info(f"JSON dataset found at local path: {dataset_path}")
+        return dataset_path
+
+    from importlib.resources import files
+    resource_path = files('whowhatbench.prompts').joinpath(dataset_path)
+    if resource_path.is_file():
+        logger.info(f"JSON dataset found in whowhatbench.prompts resources: {resource_path}")
+        return str(resource_path)
+
+    logger.error(f"JSON dataset file '{dataset_path}' not found in local filesystem or whowhatbench.prompts resources")
```
sbalandi, line 330: Why are messages needed here if they appear as exceptions ?

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -784,6 +809,7 @@ def create_evaluator(base_model, args):
                 seed=args.seed,
             )
         elif task == "text-to-video":
+            prompts = load_prompts(args)
```
sbalandi, line 812: please, remove it here, for text-to-video is expected more complex data

sbalandi (commented):
Thanks for changing the approach and preparing the code for the separate agent pipeline

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -19,13 +19,41 @@
...
+        raw_prompt_list = prompts.get("prompts")
+        if raw_prompt_list is None:
+            raise ValueError(f"'prompts' key missing in prompts source: {source}")
+        if not isinstance(raw_prompt_list, list):
+            raise ValueError(f"'prompts' field must be a list in prompts source: {source}")
+        if len(raw_prompt_list) == 0:
+            raise ValueError(f"No prompts found in 'prompts' list from source: {source}")
+        prompt_list = raw_prompt_list
+    logger.info("Prompts summary (%s): count=%d", source, len(prompt_list))
+    if verbose:
```
sbalandi: Do you use this ? I would remove as we know source or leave 50 chars, 25 from start and 25 from the end.

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -392,24 +422,39 @@ def check_args(args):
         raise ValueError(f"Speaker embedding file does not exist: {args.speaker_embeddings}")
 
 
-def load_prompts(args):
+def load_prompts(args, tokenizer=None):
```
sbalandi: tokenizer is not used here, please, remove

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -392,24 +422,39 @@ def check_args(args):
...
+    if not separator:
         name = None
-    data = load_dataset(path=path, name=name, split=split)
+    data = load_dataset(path=path, name=name, split=dataset_split)
 
     res = data[args.dataset_field]
-    res = {"prompts": list(res)}
-    return res
+    prompts = {"prompts": list(res)}
+    _log_prompts_summary(prompts, f"dataset:{path}:{dataset_split}", verbose=args.verbose)
```
sbalandi:
```suggestion
   logger.info("Prompts source path %s name %s split %s: count=%d", path, name, dataset_split, len(res))
```

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -230,6 +234,12 @@ def load_text_hf_pipeline(model_id, device, **kwargs):
     if kwargs.get("adapters") is not None:
         model = apply_peft_adapters(model, kwargs["adapters"], kwargs.get("alphas", None))
 
+    if is_cpu and getattr(getattr(model, "config", None), "model_type", None) == "gpt_oss":
```
sbalandi: please, remove, changes for gpt_oss exists above

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -189,27 +192,28 @@ def load_text_llamacpp_pipeline(model_dir):
...
-            trust_remote_code = True
-        # MoE weights of gpt_oss models post-trained with MXFP4 quantization
-        # they are dequantized with torch.bfloat16 https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/integrations/mxfp4.py#L104
-        # forcing the model to any other type will cause a type mismatch error
-        # https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L117
-        if getattr(config, "model_type", None) == "gpt_oss":
-            model_kwargs["torch_dtype"] = "auto"
-
-    if not torch.cuda.is_available or device.lower() == "cpu":
+    if is_cpu:
```
sbalandi: please, left unchanged, changes for gpt oss already in master

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -392,24 +422,39 @@ def check_args(args):
...
+
+
+def load_agent_dataset(args):
+    if args.dataset is None:
+        dataset_name = "messages_long.jsonl" if args.long_prompt else "messages_short.jsonl"
+        resolved_path = resolve_json_dataset_path(dataset_name)
+    else:
+        if not is_json_dataset(args.dataset):
+            raise ValueError("text-agent model type requires --dataset to be a .json/.jsonl file")
+        resolved_path = resolve_json_dataset_path(args.dataset)
```
sbalandi: pelase, merge `is_json_dataset` and `resolve_json_dataset_path`, add `check dataset.lower().endswith((".json", ".jsonl"))` to resolve_json_dataset_path and remove `is_json_dataset`

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+                )
+            prompts.append(prompt_preview)
+            answers.append(gen_answer_fn(model, self.tokenizer, record))
+
+        prompts_result = prompts
+        answers_result = answers
+        if result_dir:
+            prompts_result, answers_result = self._save_prompts_and_answers(prompts_result, answers_result, result_dir)
+
+        return pd.DataFrame({"prompts": prompts_result, "answers": answers_result})
```
sbalandi:
```suggestion
        if result_dir:
            prompts_result, answers_result = self._save_prompts_and_answers(prompts, answers, result_dir)

        return pd.DataFrame({"prompts": prompts, "answers": answers})
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+
+    @staticmethod
+    def _message_has_content(msg: Dict[str, Any]) -> bool:
+        if not isinstance(msg, dict):
+            return False
+        content = msg.get("content")
+        if isinstance(content, str):
+            return bool(content.strip())
+        if isinstance(content, list):
+            return len(content) > 0
```
sbalandi: Can be elemet of the list be None or empty string ? should we check that ?
What else could content be besides a string or a list ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+        chat_template_map = {
+            "qwen3coder": "https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_qwen3coder_instruct.jinja",
+            "gpt_oss": "https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_gpt_oss.jinja",
+            "gpt-oss-120b": "https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_gpt_oss.jinja",
+            "gpt-oss-20b": "https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_gpt_oss.jinja",
+        }
+
+        config = getattr(model, "config", None)
+        model_type = getattr(config, "model_type", None)
+        if model_type in chat_template_map:
```
sbalandi: if model_type is none it will cause error:
`TypeError: 'in <string>' requires string as left operand, not NoneType`

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+            kwargs["enable_thinking"] = False
+        if "reasoning_effort" in chat_template:
+            kwargs["reasoning_effort"] = "low"
+
+    def _generate_non_genai(self, model, tokenizer, record: Dict[str, Any]) -> str:
+        messages = record["messages"]
+
+        tools = record.get("tools")
+        device = getattr(model, "device", "cpu")
+        chat_template = self._load_chat_template(model, tokenizer)
```
sbalandi: move loading chat_template outside the loop, we can load it once for the model

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+            template_source = chat_template_map[model_type]
+            logger.info(
+                "Matched model_type '%s' with chat template source: %s",
+                model_type,
+                template_source,
+            )
+            return template_source
+
+        candidate_names = []
+        for value in (
```
sbalandi: tokenizer could be from another model, don't use it 
model_dir is one for the model, don't need to collect list

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+                import requests
+
+                logger.info("Loading chat template from URL: %s", chat_template_source)
+                response = requests.get(chat_template_source, timeout=30)
+                response.raise_for_status()
+                return response.text
+
+            from transformers import AutoTokenizer
+
+            logger.info("Loading chat template from model: %s", chat_template_source)
```
sbalandi, line 315: When is it used?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+        else:
+            chat_kwargs = {
+                "tokenize": True,
+                "add_generation_prompt": True,
+                "return_tensors": "pt",
+                "return_dict": True,
+            }
+            if tools is not None:
+                chat_kwargs["tools"] = tools
+            tokenizer_chat_template = None
```
sbalandi: We already took chat_template on lines 381, why isn't it used?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+            if hasattr(tokenizer, "get_chat_template"):
+                try:
+                    tokenizer_chat_template = tokenizer.get_chat_template()
+                except Exception:
+                    tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
+            else:
+                tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
```
sbalandi:
```suggestion
       try:
           tokenizer_chat_template = tokenizer.get_chat_template()
       except Exception:
           tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+            self._apply_thinking_controls_for_template(chat_kwargs, tokenizer_chat_template)
```
sbalandi: we will retun from _apply_thinking_controls_for_template if `self.omit_chat_template or not isinstance(chat_template, str)`
but we will go to this line if `chat_template is None or self.omit_chat_template`, so looks like code from _apply_thinking_controls_for_template will not applied

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+        self.empty_adapters = empty_adapters
+        self.num_assistant_tokens = num_assistant_tokens
+        self.assistant_confidence_threshold = assistant_confidence_threshold
+        self.is_genai_backend = is_genai_backend
+        self.omit_chat_template = omit_chat_template
+        self.gt_dir = os.path.dirname(gt_data or "")
+        self.target_dir = None
+
+        if base_model:
+            result_dir = os.path.join(self.gt_dir, "reference") if self.gt_dir else None
```
sbalandi: results should be saved for reference
```suggestion
            result_dir = os.path.join(self.gt_dir, "reference")
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+                raise ValueError(
+                    f"Prompt preview is empty for record index {idx}; "
+                    "check the messages content in your dataset"
+                )
+            prompts.append(prompt_preview)
+            answers.append(gen_answer_fn(model, self.tokenizer, record))
+
+        prompts_result = prompts
+        answers_result = answers
+        if result_dir:
```
sbalandi, line 519: let's save results anyway, for reference to "reference" dir, for target to output_dir or "target"

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+                model_type,
+                template_source,
+            )
+            return template_source
+
+        candidate_names = []
+        for value in (
+            getattr(config, "_name_or_path", None),
+            getattr(tokenizer, "name_or_path", None),
+            getattr(model, "model_dir", None),
```
sbalandi: to use model_dir it's need to add model_dir as attribute of class here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/model_loaders.py#L69 , but I'm not sure if we need this,  as we can take if from config , so add to class or remove it

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
sbalandi, line 1: You add 4 files with prompts, but only use 2. please, add cmd options for the rest or store them in the validation repo

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+
+        return None
+
+    def _read_text_data(self, data: pd.DataFrame) -> pd.DataFrame:
+        text_data = {"answers": [], "prompts": []}
+
+        fallback_prompt_dirs = [
+            os.path.join(self.gt_dir, "reference", "prompts") if self.gt_dir else None,
+            os.path.join(self.target_dir, "prompts") if self.target_dir else None,
+        ]
```
sbalandi: This method is not deterministic, there may not be a "reference" folder, user can compare two different target runs. I would use just path from .csv file.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,522 @@
...
+        if "reasoning_effort" in chat_template:
+            kwargs["reasoning_effort"] = "low"
+
+    def _generate_non_genai(self, model, tokenizer, record: Dict[str, Any]) -> str:
+        messages = record["messages"]
+
+        tools = record.get("tools")
+        device = getattr(model, "device", "cpu")
+        chat_template = self._load_chat_template(model, tokenizer)
+        if chat_template is not None and not self.omit_chat_template:
```
sbalandi: --omit-chat-template isn't what's usually meant, chat_template will be used in any case. Please, find out whether the branch with omit-chat-template true is necessary in practice. If yes, it should be called something else: internal-chat-template, not-compile-chat-template

tools/who_what_benchmark/tests/test_cli_text.py
```diff
@@ -360,3 +362,125 @@ def test_text_chat_model(model_id, tmp_path):
...
+
+
+@pytest.mark.parametrize(
+    ("dataset_name", "as_jsonl"),
+    [
+        ("messages_genai.json", False),
+        ("messages_genai.jsonl", True),
+    ],
+)
+def test_text_agent_dataset_json_genai(dataset_name, as_jsonl, tmp_path):
```
sbalandi: please, add also test with --test-data

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -392,24 +422,39 @@ def check_args(args):
...
+
+    data = read_json_dataset(resolved_path)
+    if args.num_samples is not None:
+        data = data[: args.num_samples]
```
sbalandi: data is cut by num_samples here and in evaluator, let's left it only in evaluator

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -392,24 +422,39 @@ def check_args(args):
...
+    logger.info("Agent dataset summary (json:%s): count=%d", resolved_path, len(data))
```
sbalandi: resolved_path is printed twice here and in read_json_dataset, please, remove it here

tools/who_what_benchmark/tests/test_cli_text.py
sswierze: done

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
sswierze: all datasets are going to be used in our validation pipelines.

tools/who_what_benchmark/tests/test_cli_text.py
sswierze: I need more information about

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: I've remove usaded of omit atm

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: The loaded chat_template is used right away: if it is available, we render the prompt with Jinja in _generate_non_genai -> _render_messages_to_prompt, and similarly in GenAI at L473.
The fallback branch only runs when no template is available, where tokenizer/model internal apply_chat_template is used at L406.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: This can happen even when the message is not empty, depending on how the tokenizer processes the input.

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: Leaved  _log_prompts_summary as it could be helpful in case of very long prompts

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -189,27 +192,22 @@ def load_text_llamacpp_pipeline(model_dir):
...
-            config = AutoConfig.from_pretrained(model_id)
-        except Exception:
-            config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
-            trust_remote_code = True
-        # MoE weights of gpt_oss models post-trained with MXFP4 quantization
-        # they are dequantized with torch.bfloat16 https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/integrations/mxfp4.py#L104
-        # forcing the model to any other type will cause a type mismatch error
-        # https://github.com/huggingface/transformers/blob/v4.57.6/src/transformers/models/gpt_oss/modeling_gpt_oss.py#L117
-        if getattr(config, "model_type", None) == "gpt_oss":
-            model_kwargs["torch_dtype"] = "auto"
```
sbalandi: please, keep code for gpt_oss , and it should work not only for cpu

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
sbalandi: As far as I know, there is a separate good-structured repository with datasets for wwb for validation. I think if file is not directly use in the tool, it should be moved there.

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -19,13 +19,40 @@
...
+    if isinstance(prompts, dict):
+        raw_prompt_list = prompts.get("prompts")
+        if raw_prompt_list is None:
+            raise ValueError(f"'prompts' key missing in prompts source: {source}")
+        if not isinstance(raw_prompt_list, list):
+            raise ValueError(f"'prompts' field must be a list in prompts source: {source}")
+        if len(raw_prompt_list) == 0:
+            raise ValueError(f"No prompts found in 'prompts' list from source: {source}")
+        prompt_list = raw_prompt_list
+    logger.info("Prompts summary (%s): count=%d", source, len(prompt_list))
```
sbalandi: prompts was created in load_prompts() as dict with "prompts" as key and list as value , we don't control whether it's empty or not , so if this function is really needed , then I would left only this check 
```suggestion
    if len(raw_prompt_list) == 0:
        raise ValueError(f"No prompts found in 'prompts' list from source: {source}")
    prompt_list = raw_prompt_list
    logger.info("Prompts summary (%s): count=%d", source, len(prompt_list))
```
and left logger.info in one place here or in load_prompts

tools/who_what_benchmark/whowhatbench/wwb.py
sbalandi: How does this help ? The prompts will be saved to files after generation, and you also know the dataset you're passing

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
```diff
@@ -179,15 +179,31 @@ def __init__(self, model_id) -> None:
...
             model_kwargs["tokenizer_kwargs"] = {"pad_token": pad_token}
 
-        self.model = SentenceTransformer(model_id, **model_kwargs)
+        try:
+            self.model = SentenceTransformer(
+                model_id,
+                **model_kwargs,
+            )
+        except TypeError:
+            # sentence-transformers<3.0 does not accept tokenizer_kwargs
```
sbalandi: adding tokenizer kwargs is under 
`if "tokenizer_kwargs" in inspect.signature(SentenceTransformer.__init__).parameters:`
Why did you add this code? What failed did you observe ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sbalandi: I mean custom_tokenizer = AutoTokenizer.from_pretrained(chat_template_source, trust_remote_code=True) , all sources are http(s)

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sbalandi: if chat template is not avalible tokenizer can't apply it with apply_chat_template , because there is nothing to apply

sbalandi: similar code already exists in _load_chat_template , no need to do it again
```
            try:
                tokenizer_chat_template = tokenizer.get_chat_template()
            except Exception:
                tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
```

so, if there is no template is available, then tokenizer_chat_template will None here, what will return apply_chat_template if tokenizer_chat_template is None ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sbalandi: if result_dir is None, then please define it with some default value

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+@pytest.mark.parametrize(
+    ("dataset_name", "as_jsonl"),
+    [
+        ("messages_genai.json", False),
+        ("messages_genai.jsonl", True),
+    ],
+)
+def test_text_agent_dataset_json_genai(dataset_name, as_jsonl, tmp_path):
```
sbalandi: looks like only this one may be left

sswierze: I simplified the test coverage by merging the backend-specific

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
sswierze: I do not have information about this. As far as I know all scenarios may be used atm.

tools/who_what_benchmark/whowhatbench/model_loaders.py
sswierze: DONE

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
sswierze: Lets leave this as it is atm, next step I will try to adjust those datasets  this validation repo which is not know for me  for now.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: DONE

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: DONE

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: Yes, updated to check for non-empty text within list items; None content is already handled by the final fallback.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: Good catch, fixed with a None guard before the in check.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: DONE

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: added attribute

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: Fixed — removed the redundant tokenizer template fetch and the now-unused _apply_thinking_controls_for_template helper.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: _apply_thinking_controls_for_template was removed entirely. Thinking controls are now applied directly in _render_messages_to_prompt where the template string is available to inspect for "enable_thinking" and "reasoning_effort".

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: Updated to "target" subdirectory — consistent with wwb naming (reference for baseline, target for predictions).

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: Moved chat template loading outside the loop — now loads once per _generate_data call and passes it to each generation function. Reduces redundant model lookups during batch processing.

tools/who_what_benchmark/whowhatbench/whowhat_metrics.py
sswierze: removed those changes

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: Done

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: removed

tools/who_what_benchmark/whowhatbench/utils.py
```diff
@@ -329,3 +324,53 @@ def parquet_generate_tables(self, files, *args, **kwargs):
...
+        pass
+
+    # Fall back to JSONL (line-delimited JSON)
+    items = []
+    for line in content.splitlines():
+        line = line.strip()
+        if line:
+            items.append(json.loads(line))
+    logger.info(f"Loaded {len(items)} records from JSON dataset")
+    return items
```
sbalandi:
```suggestion
    items = []
    if Path(dataset_path).suffix == 'json':
        items = json.loads(content)
    else:   
        for line in content.splitlines():
            line = line.strip()
            if line:
                items.append(json.loads(line))
    logger.info(f"Loaded {len(items)} records from JSON dataset")
    return items
```

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -42,7 +42,32 @@
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
-PYTORCH_MODEL_DTYPE_KWARG = {"torch_dtype": torch.float32}
+if Version(__version__) >= Version("4.56"):
+    PYTORCH_MODEL_DTYPE_KWARG = {"dtype": "auto"}
+else:
+    PYTORCH_MODEL_DTYPE_KWARG = {"torch_dtype": torch.float32}
```
sbalandi: it was set torch.float32 in older transformers inside

```suggestion
     PYTORCH_MODEL_DTYPE_KWARG = {"torch_dtype": torch.float32}
```

dtype will be changed to auto in pr https://github.com/openvinotoolkit/openvino.genai/pull/3793

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -42,7 +42,32 @@
...
+
+
+def _normalize_hf_device_map(device: str) -> str:
+    normalized = str(device).strip().lower()
+    # Transformers device_map does not recognize "gpu".
+    if normalized == "gpu":
+        return "cuda:0"
+    return normalized
```
sbalandi: move these changes into a separate commit, please

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -189,27 +215,42 @@ def load_text_llamacpp_pipeline(model_dir):
...
         try:
             config = AutoConfig.from_pretrained(model_id)
         except Exception:
-            config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
             trust_remote_code = True
+            try:
+                config = AutoConfig.from_pretrained(model_id, trust_remote_code=True)
+            except Exception as e:
+                # Some very new model types are unknown to the installed Transformers.
+                # Keep loading path alive and let model.from_pretrained with trust_remote_code decide.
```
sbalandi: to work with new model it's need to install new transformers, please, remove it

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+
+        # Prompt is already rendered with chat template; do not re-apply template in GenAI.
+        kwargs["apply_chat_template"] = False
+        res = model.generate(prompt, **kwargs)
+
+        if hasattr(res, "texts") and len(res.texts) > 0:
+            return res.texts[0]
+        return str(res)
+
+    def _generate_data(self, model, gen_answer_fn=None, result_dir=None, output_dir=None):
```
sbalandi:
```suggestion
    def _generate_data(self, model, gen_answer_fn=None, result_dir=None):
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+        if gen_answer_fn is None:
+            gen_answer_fn = self._generate_genai if self.is_genai_backend else self._generate_non_genai
+
+        if output_dir is not None:
+            result_dir = output_dir
```
sbalandi:
```suggestion
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+            try:
+                tokenizer_chat_template = tokenizer.get_chat_template()
+            except Exception:
+                tokenizer_chat_template = getattr(tokenizer, "chat_template", None)
+
+            logger.info("Loading chat template from model: %s", chat_template_source)
+            custom_tokenizer = AutoTokenizer.from_pretrained(chat_template_source, trust_remote_code=True)
+            if hasattr(custom_tokenizer, "get_chat_template"):
+                return custom_tokenizer.get_chat_template()
+            return custom_tokenizer.chat_template
```
sbalandi: remove this code, tt will never be executed
```suggestion
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+        for msg in messages:
+            if not isinstance(msg, dict):
+                continue
+            content = msg.get("content", "")
+            if isinstance(content, str) and content.strip():
+                return content
+            if isinstance(content, list) and len(content) > 0:
+                return json.dumps(content, ensure_ascii=False)
+
+        return ""
```
sbalandi, line 231: the prompt is an only user messages or the entire context (both user and assistant)?

Please, remove fallback. Why not leave the prompt empty if that's what it is?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+
+        return self._apply_tokenizer_chat_template_to_prompt(tokenizer, messages, tools)
+
+    def _generate_non_genai(self, model, tokenizer, record: Dict[str, Any], chat_template: Optional[str] = None) -> str:
+        messages = record["messages"]
+
+        tools = record.get("tools")
+        device = getattr(model, "device", "cpu")
+        if chat_template is None:
+            chat_template = self._load_chat_template(model, tokenizer)
```
sbalandi: please, remove
```suggestion
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+        temperature = record.get("temperature")
+        if temperature is not None:
+            kwargs["temperature"] = float(temperature)
+
+        if self.empty_adapters:
+            kwargs["adapters"] = openvino_genai.AdapterConfig()
+
+        tools = record.get("tools")
+        if chat_template is None:
+            chat_template = self._load_chat_template(model, _tokenizer)
```
sbalandi: please, remove it
_generate_genai is internal function, which called only in another internal function. remove default value for chat_template to avoid setting it to None and remove this strings

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+                    f"Prompt preview is empty for record index {idx}; "
+                    "check the messages content in your dataset"
+                )
+            prompts.append(prompt_preview)
+            # Pass chat_template if the generation function accepts it
+            try:
+                answers.append(gen_answer_fn(model, self.tokenizer, record, chat_template=chat_template))
+            except TypeError:
+                # Fallback for custom gen_answer_fn that don't accept chat_template parameter
+                answers.append(gen_answer_fn(model, self.tokenizer, record))
```
sbalandi: First, the chat template is loaded using a function _load_chat_template, then it's passed to the function gen_answer_fn. And if something fails, gen_answer_fn will be called without the chat template. But the chat template will be loaded into gen_answer_fn through the same function _load_chat_template. It doesn't look very useful. 

```suggestion
        answers.append(gen_answer_fn(model, self.tokenizer, record, chat_template=chat_template))
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+        if is_awq:
+            with patch_awq_for_inference(is_awq):
+                tokens = model.generate(**inputs, **generation_kwargs)
+        else:
+            tokens = model.generate(**inputs, **generation_kwargs)
+
+        answer_tokens = tokens[:, inputs["input_ids"].shape[-1]:]
+        return tokenizer.batch_decode(answer_tokens, skip_special_tokens=True)[0]
+
+    def _generate_genai(self, model, _tokenizer, record: Dict[str, Any], chat_template: Optional[str] = None) -> str:
```
sbalandi:
```suggestion
    def _generate_genai(self, model, _tokenizer, record: Dict[str, Any], chat_template: Optional[str]) -> str:
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,514 @@
...
+            except Exception as render_exc:
+                logger.warning(
+                    "Template rendering failed for current record in %s path; falling back to tokenizer.apply_chat_template. Error: %s",
+                    backend_name,
+                    render_exc,
+                )
+
+        return self._apply_tokenizer_chat_template_to_prompt(tokenizer, messages, tools)
+
+    def _generate_non_genai(self, model, tokenizer, record: Dict[str, Any], chat_template: Optional[str] = None) -> str:
```
sbalandi:
```suggestion
    def _generate_non_genai(self, model, tokenizer, record: Dict[str, Any], chat_template: Optional[str]) -> str:
```

sbalandi (changes requested):

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,184 @@
...
+    model_local = snapshot_download(model_id)
+    tokenizer = AutoTokenizer.from_pretrained(model_local)
+    base_model = OVModelForCausalLM.from_pretrained(model_local)
+    base_model.save_pretrained(temp_path)
+    tokenizer.save_pretrained(temp_path)
+    export_tokenizer(tokenizer, temp_path)
+
+
+agent_model_id = "optimum-intel-internal-testing/tiny-random-Phi3ForCausalLM"
+agent_model_path = convert_text_model(agent_model_id, "tiny_phi3_agent", _convert_base)
```
sbalandi: add the model to https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/tests/conftest.py#L16 and convert_model(model_id) from conftest.py in tests

sbalandi: This is used for text models because there are two types of conversion(full and quantized), but in the case of an agent, we can go through a common mechanism

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,184 @@
...
+        "--target-data",
+        target_data_path,
+        "--model-type",
+        "text-agent",
+        "--num-samples",
+        "1",
+        "--device",
+        "CPU",
+        *(["--genai"] if use_genai else []),
+    ])
```
sbalandi:
```suggestion
    output_with_target_data = run_wwb([
        "--gt-data",
        gt_path,
        "--target-data",
        target_data_path,
        "--model-type",
        "text-agent",
        "--num-samples",
        "1",
        "--device",
        "CPU",
    ])
```

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,184 @@
...
+        "--model-type",
+        "text-agent",
+        "--num-samples",
+        "1",
+        "--device",
+        "CPU",
+        "--output",
+        output_dir,
+    ])
+    assert "Metrics for model" in output_with_models
```
sbalandi: please, add check of similarity also

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,184 @@
...
+        gt_path,
+        "--model-type",
+        "text-agent",
+        "--num-samples",
+        "1",
+        "--device",
+        "CPU",
+        "--output",
+        output_dir,
+    ])
```
sbalandi:
```suggestion
    output_with_models = run_wwb([
        "--target-model",
        agent_model_path,
        "--gt-data",
        gt_path,
        "--model-type",
        "text-agent",
        "--num-samples",
        "1",
        "--device",
        "CPU",
        "--output",
        output_dir,
        --genai
    ])
```

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,184 @@
...
+        gt_path,
+        "--dataset",
+        dataset_path,
+        "--model-type",
+        "text-agent",
+        "--num-samples",
+        "1",
+        "--device",
+        "CPU",
+    ])
```
sbalandi: All tests can be merged in one and left one with 4 wwb run (hf, optimum, genai, target-data)
All assert can be moved to that tests. And also, please. add check similarity and target artifacts.

```suggestion
    run_wwb([
        "--base-model",
        [hf-model],
        "--gt-data",
        gt_path,
        "--dataset",
        dataset_path,
        "--model-type",
        "text-agent",
        "--num-samples",
        "1",
        "--device",
        "CPU",
        --hf
    ])
    
    opt_output = run_wwb([
        "--target-model",
        agent_model_path,
        "--gt-data",
        gt_path,
        "--dataset",
        dataset_path,
        "--model-type",
        "text-agent",
        "--num-samples",
        "1",
        "--device",
        "CPU",
    ])
```

sbalandi: and also would be good to have one test with one case from real messages(prompts folder).

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
sbalandi: please, add cmd options for all files then
it can be smth like --agent-dataset and variants [base, small, medium, large]

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: I would like to  leave possibility to use chat templates in other ways than https (like model id - "openai/gpt-oss-20b")

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: fixed

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -893,7 +910,14 @@ def load_model(
...
-    if model_type == "text" or model_type == "text-chat":
+    if model_type == "text" or model_type == "text-chat" or model_type == "text-agent":
+        if model_type == "text-agent" and use_genai and _is_visual_text_model(model_id):
+            logger.info(
+                "text-agent requested for multimodal model '%s'; using VLMPipeline via visual-text loader",
+                model_id,
+            )
+            kwargs["model_type"] = kwargs.get("model_type", "visual-text")
+            return load_visual_text_model(model_id, device, ov_options, use_hf, use_genai, **kwargs)
         return load_text_model(model_id, device, ov_options, use_hf, use_genai, use_llamacpp, **kwargs)
```
sbalandi, line 915: please, don't mix pipelines, we need a clear understanding of what we are comparing

if you really need to test vlm models with agentic scenario, let's add one mode model-type visual-text-agent, register it to register_evaluator https://github.com/sswierze/openvino.genai/blob/ad593f5d16e2f974a591380bd4a39a6537660c29/tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py#L21 and load model as VLM for all backend

```suggestion
    if model_type == "text" or model_type == "text-chat":
        return load_text_model(model_id, device, ov_options, use_hf, use_genai, use_llamacpp, **kwargs)
    elif model_type == "text-agent":
        if _is_visual_text_model(model_id):
            logger.info(
                "text-agent requested for multimodal model '%s'; using VLM Pipeline via visual-text loader",
                model_id,
            )
            kwargs["model_type"] = kwargs.get("model_type", "visual-text")
            return load_visual_text_model(model_id, device, ov_options, use_hf, use_genai, **kwargs)
        else:
            return load_text_model(model_id, device, ov_options, use_hf, use_genai, use_llamacpp, **kwargs)
```

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -204,18 +220,19 @@ def load_text_hf_pipeline(model_id, device, **kwargs):
...
+
         with mock_AwqQuantizer_validate_environment(is_awq), mock_torch_cuda_is_available(is_gptq or is_awq):
             model = AutoModelForCausalLM.from_pretrained(model_id, trust_remote_code=trust_remote_code, device_map="cpu", **model_kwargs)
         if is_awq:
             model.is_awq = is_awq
     else:
         try:
             model = AutoModelForCausalLM.from_pretrained(
-                model_id, trust_remote_code=False, device_map=device.lower(), **model_kwargs
+                model_id, trust_remote_code=False, device_map=device, **model_kwargs
```
sbalandi: please, restore .lower , we have capital letters as the default input , which applicable for genai ,

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,531 @@
...
+        if tokenizer is None:
+            return None
+
+        if hasattr(tokenizer, "get_chat_template"):
+            try:
+                return tokenizer.get_chat_template()
+            except Exception:
+                return getattr(tokenizer, "chat_template", None)
+
+        return getattr(tokenizer, "chat_template", None)
```
sbalandi, line 302:
```suggestion
        try:
            return tokenizer.get_chat_template()
        except Exception:
            return getattr(tokenizer, "chat_template", None)
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,531 @@
...
+            logger.info("Loading chat template from model: %s", chat_template_source)
+            custom_tokenizer = AutoTokenizer.from_pretrained(chat_template_source, trust_remote_code=True)
+            custom_template = self._get_tokenizer_chat_template(custom_tokenizer)
+            if custom_template is not None:
+                return custom_template
+
+        if tokenizer is not None:
+            try:
+                chat_template = self._get_tokenizer_chat_template(tokenizer)
+                if chat_template is None:
```
sbalandi, line 324: Here exception will be raised if chat_template is None, but if no tokenizer found, then function return None as chat_template. Should exception will be also raised in such case on str 331 ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,531 @@
...
+        except Exception as template_exc:
+            if tools is None:
+                raise
+            logger.warning(
+                "tokenizer.apply_chat_template failed with tools in %s path; "
+                "retrying without tools. Error: %s",
+                backend_name,
+                template_exc,
+            )
+            return self._build_prompt_text_without_tools(tokenizer, messages, chat_template, backend_name)
```
sbalandi: looks like, you execute code of _build_prompt_text_without_tools twice(with and without tools), but the first time you decompose it into code, I would suggest doing it like this

```suggestion
    def _build_templated_prompt(self, tokenizer, messages, chat_template, backend_name: str, tools: = None) -> str:
        if chat_template is not None:
            try:
                return self._render_messages_to_prompt(messages, tools, chat_template)
            except Exception as render_exc:
                logger.warning(
                    "Template rendering failed for current record in %s path; "
                    "falling back to tokenizer.apply_chat_template. Error: %s",
                    backend_name,
                    render_exc,
                )

        return self._apply_tokenizer_chat_template_to_prompt(tokenizer, messages, None)

    def _build_prompt_text(self, tokenizer, messages, tools, chat_template, backend_name: str) -> str:
        try:
            return _build_prompt_text(tokenizer, messages, chat_template, backend_name, tools)
        except Exception as render_exc:
            logger.warning(
                "Retrying prompt rendering without tools in %s path due to incompatible tools format",
                backend_name,
             )

        return _build_prompt_text(tokenizer, messages, chat_template, backend_name, tools=None)
```

sbalandi:
One more comments:
I test PR with gpt-oss-20b mode and got the similarity with model-type text-agent:
 hf vs optimum INT8 - 0.844735
 hf vs genai INT8 - 0.832675
, but with model-type text I got similarity for  hf vs optimum INT8 - 0.990431

0.84 looks quite low, is it expected ?

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -425,21 +453,45 @@ def check_args(args):
...
+    elif args.agent_dataset is not None:
+        dataset_name = AGENT_DATASET_VARIANTS[args.agent_dataset]
+        resolved_path = resolve_json_dataset_path(dataset_name)
+        logger.info(
+            "Text-agent dataset selected from --agent-dataset=%s: %s",
+            args.agent_dataset,
+            resolved_path,
+        )
+    else:
+        dataset_name = "messages_20k.jsonl" if args.long_prompt else "messages_500.jsonl"
```
sbalandi, line 486: why, if no --long_prompt is set, dataset will be messages_20k, but not messages_5k ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: I’d keep it as is. chat_template is None is not an error by itself here, because we still have a fallback via tokenizer.apply_chat_template(). The real error case is when both are unavailable: no resolved chat template and no usable tokenizer. That is already raised later during prompt rendering, so I don’t think we should fail earlier in _load_chat_template().

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: By default (without --long-prompt) we use messages_500.jsonl as the baseline set.
messages_20k.jsonl is only selected for the legacy long-context path (--long-prompt) and this was agreed with the team.
messages_5k.jsonl and messages_100k.jsonl are extension presets and should be used only when explicitly requested via --agent-dataset (not by default).

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -43,6 +43,20 @@
 logger = logging.getLogger(__name__)
 
 
+def _is_visual_text_model(model_id: str) -> bool:
```
sbalandi, line 46: it is not used, please, remove

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,133 @@
...
+    dataset_path = tmp_path / dataset_name
+    gt_path = tmp_path / f"gt_{dataset_name}.csv"
+    optimum_output_dir = tmp_path / "optimum_output"
+    genai_output_dir = tmp_path / "genai_output"
+    optimum_target_data_path = optimum_output_dir / "target.csv"
+    expected_prompt = _create_messages_dataset(dataset_path, as_jsonl=as_jsonl, use_real_prompt=use_real_prompt)
+
+    hf_output = run_wwb([
+        "--base-model",
+        "Qwen/Qwen2-0.5B",
```
sbalandi, line 60: you run evaluation with "Qwen/Qwen2-0.5B" and compare with "optimum-intel-internal-testing/tiny-random-Phi3ForCausalLM", this should not work right, please, use one model

tools/who_what_benchmark/tests/test_cli_text_agent.py
```diff
@@ -0,0 +1,133 @@
...
+
+import pandas as pd
+import pytest
+
+from conftest import convert_model, run_wwb
+from test_cli_image import get_similarity
+
+
+agent_model_id = "optimum-intel-internal-testing/tiny-random-Phi3ForCausalLM"
+agent_model_path = convert_model(agent_model_id)
```
sbalandi, line 16: you add model here and to conftest, please, leave it in one place, you can parametrize test as here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/tests/test_cli_vlm.py#L183 and convert inside test, tests will use cached variant

tools/who_what_benchmark/whowhatbench/wwb.py
sbalandi: so if it is not long prompt it will be 500, but if long, than it will be 20, looks illogical, we have --short-prompt, I suggest to use 20 by it in that case
```suggestion
        dataset_name = "messages_20k.jsonl" if args.short-prompt else "messages_500.jsonl"
```

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,524 @@
...
+            if chat_template_source.startswith(("http://", "https://")):
+                import requests
+
+                logger.info("Loading chat template from URL: %s", chat_template_source)
+                response = requests.get(chat_template_source, timeout=30)
+                response.raise_for_status()
+                return response.text
+
+            logger.info("Loading chat template from model: %s", chat_template_source)
+            custom_tokenizer = AutoTokenizer.from_pretrained(chat_template_source, trust_remote_code=True)
```
sbalandi, line 316: you wrote, that
"I would like to leave possibility to use chat templates in other ways than https (like model id - "openai/gpt-oss-20b")"
but we will be here only if model type in the chat_template_map , but no records like openai/gpt-oss-20b in the chat_template_map 

when this code block will be called ?

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sbalandi: if chat_template is None is not an error by itself , then, please, remove raising 

actually if no tokenizer found it's a problem, generation can't be run without tokenizer, but I think we will fail later

tools/who_what_benchmark/whowhatbench/prompts/messages_100k.jsonl
avasenin-14: I agree with Sofia, the current wheel size is ~150KB, adding 1.5MBs of prompt files to it doesn't seems reasonable. I'd suggest to either upload them as custom dataset to HF and then use with load_dataset, as it's already done with media for other images; or upload it to a share, if you only plan to use them in validation.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
```diff
@@ -0,0 +1,524 @@
...
+
+    @staticmethod
+    def _strftime_now(format_string: str):
+        from datetime import datetime
+
+        return datetime.now().strftime(format_string)
+
+    def _get_chat_template_for_model(self, model, tokenizer=None):
+        chat_template_map = {
+            "qwen3coder": "https://raw.githubusercontent.com/openvinotoolkit/model_server/refs/heads/main/extras/chat_template_examples/chat_template_qwen3coder_instruct.jinja",
```
avasenin-14, line 260: Please either add this templates to this repo, or add --chat-template CLI option to provide them as a file. My main concerns are:
1. These templates can be changed/renamed/removed in the upstream repo independetly, we won't know about it and that will break the WWB.
2. Additional dependency on network access to GitHub with all the downsides (running a local model need network connection now, need potenital proxies for github, github shouldn't be down).

avasenin-14 (changes requested):

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: changed

sswierze:
> One more comments: I test PR with gpt-oss-20b mode and got the similarity with model-type text-agent: hf vs optimum INT8 - 0.844735 hf vs genai INT8 - 0.832675 , but with model-type text I got similarity for hf vs optimum INT8 - 0.990431
> 
> 0.84 looks quite low, is it expected ?

Thanks for raising this .Yes, this can be expected: text-agent usually has lower similarity than text because it uses a different flow (messages/chat template/tools), so the scores are not directly comparable.

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: done

tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -43,6 +43,20 @@
...
+    """Best-effort check whether model config describes a multimodal VLM."""
```
sbalandi, line 47: function is still here, but is not used, please, remove

tools/who_what_benchmark/whowhatbench/wwb.py
sbalandi: @sswierze the changes is not in PR, please, check all comments and apply it

sbalandi (changes requested):

tools/who_what_benchmark/whowhatbench/wwb.py
sswierze: change applied in https://github.com/openvinotoolkit/openvino.genai/pull/4371

tools/who_what_benchmark/whowhatbench/model_loaders.py
sswierze: removed in https://github.com/openvinotoolkit/openvino.genai/pull/4371

tools/who_what_benchmark/tests/test_cli_text_agent.py
sswierze: one model used, change in https://github.com/openvinotoolkit/openvino.genai/pull/4371

tools/who_what_benchmark/tests/test_cli_text_agent.py
sswierze: I hope that it is done in https://github.com/openvinotoolkit/openvino.genai/pull/4371

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: finally, in https://github.com/openvinotoolkit/openvino.genai/pull/4371 I have decided to remove chat template from the code, our pipeline will provide chat template as new wwb argument --chat-template-source.

I am going to place chat template and prompts templates in frameworks.ai.openvino.llm.prompts repository.

sswierze: finally, in https://github.com/openvinotoolkit/openvino.genai/pull/4371 I have decided to remove chat template from the code, our pipeline will provide chat template as new wwb argument --chat-template-source.

I am going to place chat template and prompts templates in frameworks.ai.openvino.llm.prompts repository.

tools/who_what_benchmark/whowhatbench/text_agent_evaluator.py
sswierze: I've decided to move this PR to another branch and PR to avoid problem with current ad-hoc executions: https://github.com/openvinotoolkit/openvino.genai/pull/4371

I've removed prompts templates from genai repository, it will be passed by --dataset parameter .

tools/who_what_benchmark/whowhatbench/model_loaders.py
sswierze: it was removed in : https://github.com/openvinotoolkit/openvino.genai/pull/4371

PR #3672, opened 2026-04-09
Enable LTO on GenAI Obj

apaniukov:
## Description
Current optimization applies to openvino_genai, the final shared-library target. But that target does not compile those .cpp files itself. It only links together object files that were already produced by openvino_genai_obj.

Result: libopenvino_genai.so went from 12,274,864 to 9,660,872 bytes, a reduction of 2,613,992 bytes, about 21.3% on local compilation.

## Checklist:


as-suvorov (approved):

src/cpp/CMakeLists.txt
```diff
@@ -221,7 +221,23 @@ if(MSVC)
...
+        target_compile_options(${TARGET_NAME_OBJ} PRIVATE "/arch:AVX2")
+        message(STATUS "Enable SIMD compilation for ${TARGET_NAME_OBJ}")
+    else()
+        message(STATUS "Disable SIMD compilation for ${TARGET_NAME_OBJ}")
+    endif()
+endif()
+
+set_target_properties(${TARGET_NAME_OBJ} PROPERTIES
+    POSITION_INDEPENDENT_CODE ON
+    INTERPROCEDURAL_OPTIMIZATION_RELEASE ${ENABLE_LTO})
```
as-suvorov, line 226: @artanokhov Do you know if we have `ENABLE_LTO` set for production/nighlty builds?

PR #3677, opened 2026-04-10
LoRA support for ContinuousBatchingPipeline via add_request

dkalinowski:
## Description
Apply LoRA during CB pipeline initialization. Currently it is only done in `generate` method, so inference with lora via `add_request` is completely ignored. With this change, it is possible to setup LoRA once in initialization, and then have it applied when using `add_request`.

+ test

CVS-182580

## Checklist:

### DIFF
--- a/src/cpp/src/continuous_batching/pipeline_impl.cpp
+++ b/src/cpp/src/continuous_batching/pipeline_impl.cpp
@@ -222,6 +222,10 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::initialize_pipeline(
     m_sampler = std::make_shared<Sampler>(m_tokenizer, sampler_num_threads);
     m_sampler->set_seed(m_generation_config.rng_seed);
 
+    // Apply initial LoRA adapters to the infer request so that
+    // add_request + step() workflow uses the adapters specified at construction.
+    set_adapters(m_generation_config.adapters);
+
     // If eos_token_id was not provided, take value
     if (m_generation_config.eos_token_id == -1)
         m_generation_config.set_eos_token_id(m_tokenizer.get_eos_token_id());
--- a/tests/python_tests/test_vlm_pipeline.py
+++ b/tests/python_tests/test_vlm_pipeline.py
@@ -759,19 +759,25 @@ def test_vlm_cb_lora_generate_vs_add_request(cat_tensor: openvino.Tensor, tmp_pa
     gc_no_lora = get_greedy()
     gc_no_lora.max_new_tokens = DEFAULT_MAX_NEW_TOKENS
     res_no_lora = pipe_no_lora.generate(
-        [PROMPTS[0]], images=[[cat_tensor]], generation_config=[gc_no_lora],
+        [PROMPTS[0]],
+        images=[[cat_tensor]],
+        generation_config=[gc_no_lora],
     )
 
     # Path 1: generate() - calls set_adapters() internally (known-good)
     pipe_generate = ContinuousBatchingPipeline(
-        models_path, scheduler_config, "CPU",
+        models_path,
+        scheduler_config,
+        "CPU",
         properties={"adapters": adapter_config},
     )
     gen_config_generate = get_greedy()
     gen_config_generate.max_new_tokens = DEFAULT_MAX_NEW_TOKENS
     gen_config_generate.adapters = adapter_config
     res_generate = pipe_generate.generate(
-        [PROMPTS[0]], images=[[cat_tensor]], generation_config=[gen_config_generate],
+        [PROMPTS[0]],
+        images=[[cat_tensor]],
+        generation_config=[gen_config_generate],
     )
 
     assert res_generate[0].texts[0] != res_no_lora[0].texts[0], (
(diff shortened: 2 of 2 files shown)


src/cpp/src/continuous_batching/pipeline_impl.cpp
```diff
@@ -222,6 +222,10 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::initialize_pipeline(
     m_sampler = std::make_shared<Sampler>(m_tokenizer, sampler_num_threads);
     m_sampler->set_seed(m_generation_config.rng_seed);
 
+    // Apply initial LoRA adapters to the infer request so that
+    // add_request + step() workflow uses the adapters specified at construction.
+    set_adapters(m_generation_config.adapters);
```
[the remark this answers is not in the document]
dkalinowski: Inside set_adapters there is a check if controller is initialized, I dont think we need another if statement here

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -711,6 +711,116 @@ def test_vlm_continuous_batching_generate_vs_add_request_for_gemma(
...
+    outputs = handle.read_all()
+
+    # Both LoRA paths must produce identical token sequences
+    for out_idx, output in enumerate(outputs):
+        text = tokenizer.decode(output.generated_ids)
+        assert text == res_generate[0].texts[out_idx], (
+            f"add_request()+step() output differs from generate() - "
+            f"LoRA likely not applied during add_request()+step() workflow. "
+            f"Got '{text}', expected '{res_generate[0].texts[out_idx]}'"
+        )
```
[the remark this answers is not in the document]
dkalinowski: The outputs will vary alot, I dont think we need token-level matching

Wovchena (approved):

dkalinowski:
Can anyone add it to merge queue? @Wovchena @likholat

likholat:
@dkalinowski Could you please provide a short script to reproduce the original issue?

> Apply LoRA during CB pipeline initialization. Currently it is only done in generate method, so inference with lora via add_request is completely ignored. With this change, it is possible to setup LoRA once in initialization, and then have it applied when using add_request.

Could this be resolved by calling `set_adapters` instead?

likholat (commented):
@yatarkan @Wovchena @dkalinowski I’m not sure about the suggested solution, because it looks like it could cause `set_adapters` to be called twice: once during pipeline initialization and then again during the `generate`
Please correct me if I'm wrong

likholat:
@yatarkan please take a look

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -711,6 +711,116 @@ def test_vlm_continuous_batching_generate_vs_add_request_for_gemma(
             )
 
 
+LORA_ADAPTER_REPO_ID = "likholat/tiny-random-qwen2vl-lora"
+
+
+def _create_nonzero_lora_adapter(base_adapter_path, tmp_dir):
```
yatarkan, line 717: Please add typings (`Path` or `str`)

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -711,6 +711,116 @@ def test_vlm_continuous_batching_generate_vs_add_request_for_gemma(
...
+    """
+    Regression test: LoRA adapters specified at CB pipeline construction
+    must be applied during add_request()+step() workflow, not only during generate().
+    Compares generate() vs add_request()+step() outputs - they must be identical.
+    Uses a modified adapter with non-zero B weights to ensure LoRA has observable effect.
+    """
+    from openvino_genai import Adapter, AdapterConfig
+
+    models_path = _get_ov_model("optimum-intel-internal-testing/tiny-random-qwen2vl")
+    lora_path = snapshot_download(LORA_ADAPTER_REPO_ID)
```
yatarkan, line 748: Let's move it to pytest params

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -711,6 +711,116 @@ def test_vlm_continuous_batching_generate_vs_add_request_for_gemma(
...
+    for key in tensors:
+        if "lora_B" in key:
+            tensors[key] = rng.standard_normal(tensors[key].shape).astype(tensors[key].dtype) * 0.1
+
+    out_path = Path(tmp_dir) / "adapter_model.safetensors"
+    save_file(tensors, str(out_path))
+    return out_path
+
+
+def test_vlm_cb_lora_generate_vs_add_request(cat_tensor: openvino.Tensor, tmp_path):
```
yatarkan, line 738: Please add typings

dkalinowski:
> @yatarkan @Wovchena @dkalinowski I’m not sure about the suggested solution, because it looks like it could cause `set_adapters` to be called twice: once during pipeline initialization and then again during the `generate` Please correct me if I'm wrong

I don't think it will be called twice, let me explain:
- existing applications that already use LoRA + ContinuousBatchingPipeline + generate, dont apply adapters at pipeline constructors, simply because its not supported now. For such scenario, set_adapters is no-op. Only set_adapters during ::generate() call will take effect
- existing applications that dont use LoRA at all via ContinuousBatchingPipeline + generate - will have both set_adapters basically no-op. No overhead

The only affected scenario, is when new applications will initialize some AdapterConfig at initialization time (because it will be possible now after my change) and during ::generate() call as well. Intentionally. When they pass it explicitly in pipeline constructor. And also in GenerationConfig's param of generate call. But even then there is `compare_configs` logic that quickly deduces that adapters are equal and it is no-op.

Here are my measurements, first set_adapter call (during pipeline initialization):
```
AdapterControllerImpl::apply took 155.956 ms
AdapterControllerImpl::compare_configs took 0.012185 ms
ContinuousBatchingImpl::set_adapters took 156.008 ms
```

Second set_adapter call, with the same adapter config:
```
AdapterControllerImpl::apply took 0.005179 ms
AdapterControllerImpl::compare_configs took 0.003834 ms
ContinuousBatchingImpl::set_adapters took 0.059887 ms <--------------------------
```

I assume the overhead is negligible.
Also if someone changes adapter config via generate method, the operation will take longer, but it needs to - because user requested different LoRA.


Plus, finally, this change supports LoRA + ContinuousBatchingPipeline + **add_request** path, which is currently forgotted as of now.

@likholat @Wovchena @yatarkan

likholat:
@dkalinowski Could you please verify whether the current solution (master branch, without your changes) works for your case with a LoRA adapter in MODE_STATIC?

dkalinowski:
> @dkalinowski Could you please verify whether the current solution (master branch, without your changes) works for your case with a LoRA adapter in MODE_STATIC?

Yes, MODE_STATIC works in my case. Thank you!
I think we can close this PR since there is feasible workaround.

PR #3718, opened 2026-04-16
Rename download_gguf_model to download_hf_model in test utilities

goyaladitya05:
## Description
`download_gguf_model` in tests/python_tests/utils/hugging_face.py is a generic file downloader that uses hf_hub_download, it works for any file type, not just GGUF. Renamed it to download_hf_file to reflect its purpose and enable reuse for downloading other file types (e.g. LoRA .safetensors adapters) in upcoming PRs (e.g #3605).

Updated all calls in test_gguf_lora.py, test_gguf_reader.py, and samples/test_tools_llm_benchmark.py.

## Checklist:
- I have made corresponding changes to the documentation. NA

### DIFF
--- a/tests/python_tests/samples/test_tools_llm_benchmark.py
+++ b/tests/python_tests/samples/test_tools_llm_benchmark.py
@@ -8,7 +8,7 @@ from pathlib import Path
 
 from test_utils import run_sample
 from data.models import GGUF_MODEL_LIST
-from utils.hugging_face import download_hf_model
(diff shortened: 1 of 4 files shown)


as-suvorov:
https://github.com/openvinotoolkit/openvino.genai/pull/3605 is not very big so renaming can be done there. Let's move renaming to https://github.com/openvinotoolkit/openvino.genai/pull/3605

PR #3721, opened 2026-04-16
Properly handle per request seed for multinomial sampling

### LINKED ISSUE #3584: [Bug] LLMPipeline.generate doesn't properly sample
**Bug:**

Running the same prompts through `LLMPipeline.generate` produces the same results in separate runs. I'm uncertain of the cause but I believe it relates to the seed being ignored. Given the answers are often the exact same during a run, I believe it could also be that the sampler is choosing the most likely option. 

Adjusting `top_p`, `top_k`, and `temperature` will still produce different results. So will `max_new_tokens` as long as the adjustment has an effect on the output (eg, going from 10 => 500 or 500 => 20 will have an effect but going from 500 => 1000 won't). 

The issue is occurring when using either my GPU or CPU. 

**Expected Results:**

I would expect that the same prompt would produce different results each time it is run. While it does produce different runs with the same pipe, the results are identical when run again. 

I have an example script I've been using below. The model used is [Echo9Zulu/Qwen3-4B-Instruct-2507-int4_asym-awq-ov](https://huggingface.co/Echo9Zulu/Qwen3-4B-Instruct-2507-int4_asym-awq-ov); however, every model I've tested has the same issue. The only change between the two items is the seed. Even still, you'll notice it produced the exact same ten responses. 

Example script:

```python
import openvino_genai as ov
pipe = ov.LLMPipeline("qwen3-instruct", "GPU")
config = pipe.get_generation_config()
config.top_p = 0.95
config.top_k = 25
config.temperature = 2.0
config.rng_seed = 1658945 # Ignored silently
config.max_new_tokens = 100
pipe.set_generation_config(config)
prompt = "Generate a random number between 100,000 and 100,000,000,000."
for i in range(0,10):
    print(pipe.generate(prompt))

7829346512 🎲
*(This is a random number between 100,000 and 100,000,000,000.)* ✅
8764329015
This is a randomly generated number between 100,000 and 100,000,000,000.
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**5,328,647,912**
Let me know if you'd like a different range or have another request! 😊 numbers!
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**4,829,156,378**
Let me know if you'd like a different range or more digits! (Note: This is randomly selected for your request — numbers up to 100 billion with 11 or 12 digits.)
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**7,482,316,953**
Let me know if you’d like a different range or format! 🎯
9876543210
This is a random number between 100,000 and 100,000,000,000.
9742863452
Sure! Here's a random number between 100,000 and 100,000,000,000:
**4,723,851,960**
Let me know if you'd like another or a different range! 🍀
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**4,783,261,590**
Let me know if you want it within a specific range or format! 🎱
97,643,218,590
```

```python
import openvino_genai as ov
pipe = ov.LLMPipeline("qwen3-instruct", "GPU")
config = pipe.get_generation_config()
config.top_p = 0.95
config.top_k = 25
config.temperature = 2.0
config.rng_seed = 494189156 # Ignored silently
config.max_new_tokens = 100
pipe.set_generation_config(config)
prompt = "Generate a random number between 100,000 and 100,000,000,000."
for i in range(0,10):
    print(pipe.generate(prompt))

7829346512 🎲
*(This is a random number between 100,000 and 100,000,000,000.)* ✅
8764329015
This is a randomly generated number between 100,000 and 100,000,000,000.
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**5,328,647,912**
Let me know if you'd like a different range or have another request! 😊 numbers!
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**4,829,156,378**
Let me know if you'd like a different range or more digits! (Note: This is randomly selected for your request — numbers up to 100 billion with 11 or 12 digits.)
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**7,482,316,953**
Let me know if you’d like a different range or format! 🎯
9876543210
This is a random number between 100,000 and 100,000,000,000.
9742863452
Sure! Here's a random number between 100,000 and 100,000,000,000:
**4,723,851,960**
Let me know if you'd like another or a different range! 🍀
Sure! Here's a randomly generated number between 100,000 and 100,000,000,000:
**4,783,261,590**
Let me know if you want it within a specific range or format! 🎱
97,643,218,590
```

However, if I run it with `optimum.intel`, I don't have the same issue:

```python
from optimum.intel.openvino import OVModelForCausalLM
from transformers import AutoProcessor
model_id = "qwen3-instruct"
model = OVModelForCausalLM.from_pretrained(model_id, device="GPU")
processor = AutoProcessor.from_pretrained(model_id, use_fast=False)
prompt = "Generate a random number between 100,000 and 100,000,000,000."
input = processor(prompt, return_tensors="pt")
for i in range(0,10):
    output = model.generate(**input)
    print(processor.batch_decode(output, skip_special_tokens=True, clean_up_tokenization_spaces=True))

["Generate a random number between 100,000 and 100,000,000,000. The number should be 11 digits long. Here's a randomly generated 11-digit number"]
['Generate a random number between 100,000 and 100,000,000,000. A random number between 100,000 and 100,00']
['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number. A random number between 100,000']
['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number, and it should not contain any repeated digits. Additionally, ensure']
['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number.\n\nWe are to generate a random whole number between 10']
['Generate a random number between 100,000 and 100,000,000,000. A random number between 100,000 and 100,00']
['Generate a random number between 100,000 and 100,000,000,000. We are to generate a random number between 100,000 and 10']
['Generate a random number between 100,000 and 100,000,000,000. To generate a random number between 100,000 and 100,']['Generate a random number between 100,000 and 100,000,000,000. To generate a random number between 100,000 and 100,']['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number. The number should be between 100,00']
```

```python
from optimum.intel.openvino import OVModelForCausalLM
from transformers import AutoProcessor
from transformers import AutoProcessor
model_id = "qwen3-instruct"
model = OVModelForCausalLM.from_pretrained(model_id, device="GPU")
processor = AutoProcessor.from_pretrained(model_id, use_fast=False)
prompt = "Generate a random number between 100,000 and 100,000,000,000."
input = processor(prompt, return_tensors="pt")
for i in range(0,10):
    output = model.generate(**input)
    print(processor.batch_decode(output, skip_special_tokens=True, clean_up_tokenization_spaces=True))

['Generate a random number between 100,000 and 100,000,000,000. To generate a random number between 100,000 and 100,']['Generate a random number between 100,000 and 100,000,000,000. We are generating a random number between 100,000 and
100']
['Generate a random number between 100,000 and 100,000,000,000. To generate a random number between 100,000 and 100,']['Generate a random number between 100,000 and 100,000,000,000. We are generating a random number between 100,000 and
100']
['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number. The number should not be a multiple of 10.']
['Generate a random number between 100,000 and 100,000,000,000. To generate a random number between 100,000 and 100,']['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number.\n\nWe are to generate a random whole number between 10']
["Generate a random number between 100,000 and 100,000,000,000. The number must be a multiple of 100,000. Here's a random"]
['Generate a random number between 100,000 and 100,000,000,000. The number should be a whole number. A random number between 100,000']
['Generate a random number between 100,000 and 100,000,000,000. A random number between 100,000 and 100,00']
```
**Environment:**

OS: Ubuntu 24.04, Linux Kernel 6.17.0-19-generic
CPU: AMD 5600x
GPU: Sparkle Intel Arc B60 Pro 24GB

Python:
* `openvino`: 2026.0.0
* `openvino-genai`: 2026.0.0.0

System packages:
* `libze-dev`: 1.27.0.1-24.04-ppa2
* `libze-intel-gpu-dev`: 26.05.37020.3-1-24.04-ppa3
* `libze-intel-gpu1`: 26.05.37020.3-1-24.04-ppa3
* `libze1`: 1.27.0.1-24.04-ppa2
* `intel-opencl-icd`: 26.05.37020.3-1-24.04-ppa3

I've also tested with `openvino-genai` 2025.4.1.0 but still experience the same issue.

mzegla:
Accept per request seed for multinomial sampling instead of initializing default seed on pipeline initialization (continuous batching). 
Addressing: https://github.com/openvinotoolkit/openvino.genai/issues/3584


apaniukov (approved):

samples/cpp/text_generation/multinomial_causal_lm.cpp
```diff
@@ -1,11 +1,16 @@
...
+#include <iostream>
+#include <limits>
+#include <stdexcept>
+#include <string>
 #include "openvino/genai/llm_pipeline.hpp"
 
 int main(int argc, char* argv[]) try {
-    if (3 != argc) {
-        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> '<PROMPT>'");
+    if (argc < 3 || argc > 4) {
```
as-suvorov: We do not expose other config properties as arguments. Let's revert seed argument and keep samples simple

dtrawins: The expectation is that with multinomial algorithm, responses should differ slightly each time. I guess @mzegla  is trying to showcase the difference that way but the correct behavior should be to have random default seed.
If we have fixed static seed, the point of multinomial is not met.
I propose to drop this parameter from sample but set random seed by default.

mzegla: Makes sense to me. @as-suvorov , @apaniukov - what do you think?

as-suvorov: @dtrawins  Do you mean to set random seed to [`GenerationConfig`](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/generation_config.hpp#L667)?

dtrawins: @as-suvorov  yes, the seed should be random by default. So it would be the best to set random value in the constructor by default. That would achieve correct behavior of multinomial algorithm and give option to get deterministic responses with a static seed value in the user request.
Potentially that might impact unit tests but this is needed. Open question is if that should be done in the same PR or separate. I think it could be one as it states in title "Properly handle per request seed for multinomial sampling" ;-)

as-suvorov: I don't remember if current static seed was set on purpose. I don't see why static seed is not correct multinomial behavior. Please check how transformers and probably vllm handle this.
cc: @Wovchena

dtrawins: There is no default seed in the Hugging Face transformers library. If you don’t explicitly set a seed, generation uses the current global RNG state (from PyTorch / NumPy / Python), which typically results in non‑deterministic outputs across runs.

mzegla: llama.cpp and vllm use non-deterministic generation by default and getting deterministic sequence requires setting the seed explicitly
How shall we proceed?

as-suvorov: It's fine by me. @Wovchena @apaniukov Do you have any objections?

Wovchena: Is transformers and llama.cpp alignment the only reason? A counter reason would be not to suddenly change GenAI's default people could already get used to.

@likholat, can you describe the behavior of random generators for image generation?

tests/python_tests/test_llm_pipeline.py
```diff
@@ -1142,3 +1143,71 @@ def test_llm_pipeline_add_extension(
...
+                )
+            )
+        )
+        for seed in rng_seeds
+    ]
+
+    assert len(set(text_tuples)) > 1, (
+        f"generate() with different rng_seeds {rng_seeds} must produce at least one distinct output, "
+        f"but all produced: {text_tuples[0]!r}"
+    )
```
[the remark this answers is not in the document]
mzegla: We already check 4 different seeds and expect at least one to be different. Issue with that seems really unlikely, so I would go with current implementation for now.

src/cpp/src/sampling/sampler.cpp
```diff
@@ -1130,13 +1138,16 @@ void Sampler::create_logit_processor(uint64_t request_id, const GenerationConfig
...
+        std::forward_as_tuple(request_id),
+        std::forward_as_tuple(sampling_params.rng_seed, std::move(lp)));
 }
 
 void Sampler::clear_request_info(uint64_t request_id) {
     m_beam_search_info.erase(request_id);
-    m_logit_processors.erase(request_id);
-    m_stop_strings.erase(request_id);
+    m_request_contexts.erase(request_id);
 }
```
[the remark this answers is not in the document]
mzegla: `clear_request_info` is used on the pipeline level - after or before actual sampling. The parallel sampling and clearing is not the expected flow

samples/cpp/text_generation/multinomial_causal_lm.cpp
mzegla: From my perspective it's more of a misconfiguration that is now being fixed. 
I would say that general expectation when using random sampling is to get different outputs for exactly the same prompt by default (https://github.com/openvinotoolkit/openvino.genai/issues/3584) and deterministic behavior only on-demand when setting specific seed.

mzegla: Pushed few commits with default random seed changed. Let me know what the final decision is. 
If there are objections I will revert it and only keep the fix for correct seed usage.

dtrawins (approved):

pavel-esir (approved):

PR #3749, opened 2026-04-23
[whisper]: Add `language` for whisper results

as-suvorov:
### Public API design:
I analyzed `transformers` and `openai` implementations. They implement public API in a bit different ways.
```python
# transformers
from transformers import pipeline
pipe = pipeline("automatic-speech-recognition", model="openai/whisper-tiny")
result = pipe(sample, return_language=True)
# {'text': '...', 'chunks': [{'language': None, 'text': '...'}]}

# openai
import whisper
model = whisper.load_model("tiny")
result = model.transcribe(sample["array"])

# {
#     "text": " Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.",
#     "segments": [
#         {
#             "id": 0,
#             "seek": 0,
#             "start": 0.0,
#             "end": 6.0,
#             "text": " Mr. Quilter is the apostle of the middle classes and we are glad to welcome his gospel.",
#             "tokens": [50364, 2221, 13, 2326, 388, 391, 307, 264, 50244, 295, 264, 2808, 5359, 293, 321, 366, 5404, 281, 2928, 702, 14943, 13, 50664],
#             "temperature": 0.0,
#             "avg_logprob": -0.2779924472173055,
#             "compression_ratio": 1.0875,
#             "no_speech_prob": 0.0022630461025983095,
#         }
#     ],
#     "language": "en",
# }
```

Transformers implementation detects language for each chunk, when openai implementation detects language for the whole sample.
Transformers language detection implementation is currently broken: https://github.com/huggingface/transformers/blob/07e3831133a37487ee6faf57aed965f39c962a43/tests/pipelines/test_pipelines_automatic_speech_recognition.py#L413

#### Motivation for `GenAI` per sample language detection API:
* Current `GenAI` implementation prepares `sot` tokens once for the whole sample: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/whisper/whisper.cpp#L314.
* According to copilot Whisper was designed and trained for per sample language detection (not verified, but matches to openai implementation)
* Implementation of per-chunk language detection would require additional encoder-decoder inference at the start of each chunk.

### Implementation details:
Proposed implementation always return `language` (e.g. `en`, `de`) with `WhisperDecodedResults`.
In case of user provided `WhisperGenerationConfig.language` the result field is just set as `WhisperDecodedResults.language = WhisperGenerationConfig.language`.
In case of `.en` models `WhisperDecodedResults.language` is hardcoded as `en`.
Otherwise language is auto detected by pipeline.

CVS-184551

## Checklist:


src/cpp/src/whisper/pipeline_static.cpp
```diff
@@ -245,30 +245,48 @@ int64_t detect_language(ov::Tensor& encoder_hidden_state,
...
     }
 
     int64_t language_token_id = 0;
+    std::string language;
     if (config.language.has_value()) {
-        std::string language = *config.language;
+        language = *config.language;
         if (config.lang_to_id.count(language)) {
             language_token_id = config.lang_to_id.at(language);
         }
```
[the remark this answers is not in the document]
as-suvorov: Checked on config validation

src/cpp/src/whisper/models/decoder.cpp
```diff
@@ -35,7 +35,18 @@ std::pair<int64_t, float> WhisperDecoder::detect_language(const ov::Tensor& enco
     auto output_tensor = wait();
     const auto infer_ms = ov::genai::PerfMetrics::get_microsec(std::chrono::steady_clock::now() - infer_start);
 
-    int64_t output_token = ov::genai::utils::argmax(output_tensor, 0);
+    auto logits_data = output_tensor.data<float>();
```
eshiryae, line 39: Why `ov::genai::utils::argmax()` won't work here as it was before?

src/cpp/src/whisper/whisper.cpp
```diff
@@ -222,23 +223,41 @@ ov::Tensor encode(ov::InferRequest& request,
     return request.get_tensor("last_hidden_state");
 }
 
-std::vector<int64_t> prepare_sot_tokens(ov::Tensor& encoder_hidden_state,
-                                        std::shared_ptr<ov::genai::WhisperDecoder> decoder,
-                                        const ov::genai::WhisperGenerationConfig& config,
-                                        ov::genai::RawPerfMetrics& raw_metrics) {
+std::string find_language_by_token_id(const std::map<std::string, int64_t>& lang_to_id, int64_t token_id) {
```
eshiryae: `find_language_by_token_id()` and `to_unescaped_language()` can be moved to whisper utils

tests/python_tests/test_whisper_pipeline_static.py
```diff
@@ -266,3 +271,21 @@ def test_static_whisper_stateful_word_timestamps(model_descr, sample_from_datase
 
     compare_results_with_assert(expected, actual_out)
     compare_word_timestamps_results_with_assert(expected, actual_out)
+
```
eshiryae, line 287: Should we add `test_language_detection_en` also for NPU pipeline?

src/cpp/src/whisper/models/decoder.cpp
as-suvorov: This change aligns npu and stateful implementations. Instead of iterating through all vocab we just pick language token ids. This also ensures returned detected language is actually a language token.

src/cpp/src/whisper/whisper.cpp
as-suvorov: Moved

tests/python_tests/test_whisper_pipeline_static.py
as-suvorov: added

eshiryae (approved):
LGTM

src/cpp/src/whisper/whisper_utils.cpp
```diff
@@ -57,6 +59,50 @@ int64_t argmax(const ov::Tensor& logits, const size_t batch_idx) {
...
+            return language;
+        }
+    }
+
+    OPENVINO_THROW("Language token id ", token_id, " not found in lang_to_id map.");
+}
+
+std::string to_unescaped_language(const std::string& language) {
+    // "<|en|>" -> "en"
+    return std::regex_replace(language, std::regex("[|<>]+"), "");
```
sgonorov: I think it's easier and faster to just erase_if instead of a whole regex.

src/cpp/src/whisper/models/decoder.cpp
```diff
@@ -35,7 +35,18 @@ std::pair<int64_t, float> WhisperDecoder::detect_language(const ov::Tensor& enco
...
+
+    int64_t output_token;
```
sgonorov:
```suggestion
    int64_t output_token = 0;
```

sgonorov (approved):
I think it also a good idea to update C and JS API with a new language field.  Just not sure is it in scope.

src/cpp/src/whisper/whisper_utils.cpp
as-suvorov: `erase_if` is c++ 20 feature, switched to erase + remove_if

src/cpp/src/whisper/models/decoder.cpp
as-suvorov: used -1 as initilizer

PR #3773, opened 2026-04-28
[EXPERIMENT][WIP] llm: always fall back to SDPA for hybrid models (Qwen3.6-35B-A3B)

mlukasze:
## Description

Fix SDPA fallback for hybrid models (e.g. Qwen3.6-35B-A3B with GatedDeltaNet + full-attention layers) when `ATTENTION_BACKEND=PA` is set.

**Root cause:** `explicitly_requires_paged_attention()` conflated `ATTENTION_BACKEND=PA` (a user preference) with hard PA requirements (scheduler_config / speculative-decoding / prompt_lookup). As a result the hybrid-model SDPA fallback in all three LLMPipeline constructors was dead code — the outer branch always warned and continued with PA, leading to a crash:
```
RuntimeError: Model references undeclared parameters: beam_idx
```

**Fix:** Introduce `strictly_requires_paged_attention()` that checks only scheduler_config, draft_model, and prompt_lookup — not ATTENTION_BACKEND. Update all three LLMPipeline constructors to:
1. Use `OPENVINO_ASSERT(!strictly_requires_paged_attention(...))` for the hybrid-model guard so a truly required PA on a hybrid model fails fast with a clear error (C++ Core Guidelines E.12).
2. Use `strictly_requires_paged_attention()` for the CB-adapter selection so `ATTENTION_BACKEND=PA` alone falls through to the optional try/catch CB path, not the strict always-crash path.

`explicitly_requires_paged_attention()` is preserved (delegates to the new function after the ATTENTION_BACKEND check) so existing callers are unaffected.

## Checklist:

### DIFF
--- a/src/cpp/src/llm/pipeline.cpp
+++ b/src/cpp/src/llm/pipeline.cpp
@@ -212,19 +212,18 @@ ov::genai::LLMPipeline::LLMPipeline(
     // PA backend does not support linear attention states (conv/SSM caches).
     if (attention_backend == PA_BACKEND && !is_npu_requested
         && utils::has_linear_attention_states(model)) {
-        if (utils::explicitly_requires_paged_attention(user_properties)) {
-            GENAI_WARN("PA backend does not support models with linear attention states. The model may work incorrectly.");
-        } else {
-            // Always fall back to SDPA for hybrid models (linear + full attention).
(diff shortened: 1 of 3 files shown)


mlukasze:
will be fully recreated

PR #3803, opened 2026-05-05
Kczarnec malloc trim in memory monitor cooldown

krzyczar:
### Add portable `malloc_trim` (`memory_monitor.py`)

Introduces a portable heap-trimming helper that releases free memory back to the OS from the **main (benchmark) process** during each cooldown phase.

**New additions:**
- `_load_libc()` — resolves the C standard library at runtime without hard-coding `libc.so.6`:
  - Linux/glibc: uses `ctypes.CDLL(None)` (already mapped into the process)
  - Linux/musl / other: falls back to `ctypes.util.find_library("c")`
  - Windows: skipped (handled separately)
- `malloc_trim() -> bool` — cross-platform memory release:
  - **Linux/glibc**: calls `malloc_trim(0)` to return free heap pages to the OS
  - **Windows**: iterates all process heaps via `GetProcessHeaps` + `HeapCompact`
  - **macOS / musl / unsupported**: returns `False` silently, no crash
- `activate_cooldown()`: add a call to `malloc_trim()` + log output. The call runs unconditionally (outside `if self.mmh:`) so it always acts on the main process heap regardless of monitoring mode.

**Why it matters:** glibc retains freed memory in its internal pool by default. Calling `malloc_trim(0)` after a model unload or between benchmark iterations forces the OS to reclaim those pages, giving cleaner baseline readings in subsequent measurements.

---

### Remove idle mode (`memory_monitor.py`, `benchmark.py`)

The idle/cooldown-only monitoring mode (`-mc 5`) has been removed. It was the least-used mode and added unnecessary branching in the background worker.

**Removed:**
- `MonitorLevel.IDLE = 3` enum value
- `MonitorMode.PROCESS_IDLE` enum member
- `MonitorMode.is_idle` property
- `MemoryMarkerMonitor.idle_loop()` method
- `is_idle` parameter from `MemoryMarkerHandler.__init__()`, `pargs` tuple, and `background_worker()` signature
- `if is_idle: mmm.idle_loop(…) else:` branch in `background_worker` — now unconditionally calls `mmm.loop()`

**Valid mode range updated:** `0–5` → `0–4`

**`benchmark.py` help text updated** to reflect the current four modes:
```
0 = off (default)
1 = thread, warm-up only
2 = thread, all iterations
3 = separate process, warm-up only
4 = separate process, all iterations
```

## Checklist:

### DIFF
--- a/tools/llm_bench/llm_bench_utils/memory_monitor.py
+++ b/tools/llm_bench/llm_bench_utils/memory_monitor.py
@@ -31,10 +31,56 @@ import logging as log
 import traceback
 import json
 import sys
+import ctypes
+import ctypes.util
 
 
 # CUSTOM FIX TO AVOID ISSUE: RuntimeError: main thread is not in main loop.
 matplotlib.use("Agg")
+
+# ── portable malloc_trim ──────────────────────────────────────────────────────
+def _load_libc():
+    """Load the C standard library portably (Linux / macOS). Returns None on Windows."""
+    if sys.platform == "win32":
+        return None
+    lib = ctypes.CDLL(None)          # fast path: already mapped into the process
+    if not hasattr(lib, "malloc_trim"):
+        name = ctypes.util.find_library("c")
+        lib = ctypes.CDLL(name) if name else None
+    return lib
+
+_libc = _load_libc()
+
+def malloc_trim() -> bool:
+    """
+    Release free heap memory back to the OS in the current (main) process.
+    - Linux / glibc : malloc_trim(0) — returns free heap pages to the OS
+    - Windows       : HeapCompact()  — coalesces / decommits all process heaps
+    Returns True if memory was trimmed, False otherwise.
+    Safe to call at any time — only touches already-freed memory.
+    """
+    if sys.platform == "win32":
+        try:
+            kernel32 = ctypes.windll.kernel32
+            count = kernel32.GetProcessHeaps(0, None)
+            HeapsArray = ctypes.c_void_p * count
+            heaps = HeapsArray()
+            kernel32.GetProcessHeaps(count, heaps)
+            trimmed = False
+            for heap in heaps:
+                if kernel32.HeapCompact(heap, 0):
+                    trimmed = True
+            return trimmed
+        except OSError:
+            return False
+
+    if _libc is None or not hasattr(_libc, "malloc_trim"):
+        return False
+    _libc.malloc_trim.restype  = ctypes.c_int
+    _libc.malloc_trim.argtypes = [ctypes.c_size_t]
+    return bool(_libc.malloc_trim(0))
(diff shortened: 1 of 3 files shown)


sbalandi (approved):

sbalandi:
Could you please, add description of modes 3, 4 to README https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/README.md#8-memory-consumption-mode ? And also extend info about cooldown ?
here or in another PR

sbalandi:
build_jenkins

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -31,17 +31,93 @@
...
+
+
+def drop_caches():
+    os.sync()
+    cache_path = "/proc/sys/vm/drop_caches"
+    try:
+        with open(cache_path, "w") as f:
+            f.write("3")
+    except PermissionError:
+        os.stsem("sudo echo 3 > /proc/sys/vm/drop_caches")
```
avasenin-14:
```suggestion
        os.system("sudo echo 3 > /proc/sys/vm/drop_caches")
```

avasenin-14: Hi @krzyczar , do you think it makes sense to avoid calling 'sudo' from a python code? 
From my point of view it is a potential security vulnerability and a possible hang on a password prompt. Also a side note that sudo might not be present on all linux systems.
I'd suggest writing a message to the user to re-run the script with sudo permissions, if that's possible?
Though I might be wrong and I can may miss something here.

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -31,17 +31,93 @@
...
+
+_libc = _load_libc()
+
+
+def drop_caches():
+    os.sync()
+    cache_path = "/proc/sys/vm/drop_caches"
+    try:
+        with open(cache_path, "w") as f:
+            f.write("3")
```
avasenin-14, line 66: Do I understand correctly that 3 will drop the caches system-wide?
https://docs.kernel.org/admin-guide/sysctl/vm.html#drop-caches
If I'm not overthinking things, won't it slow down the next run significantly, considering OS may need to load all the stuff back to memory? Or is it the intended behaviour?

avasenin-14 (changes requested):

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: it is intended

avasenin-14 (approved):

sbalandi:
build_jenkins

PR #3841, opened 2026-05-11
PA LoRA fix

likholat:
## Description

[CVS-186315](https://jira.devtools.intel.com/browse/CVS-186315)

## Checklist:

### DIFF
--- a/src/cpp/src/continuous_batching/pipeline.cpp
+++ b/src/cpp/src/continuous_batching/pipeline.cpp
@@ -20,6 +20,8 @@
 #include "visual_language/inputs_embedder.hpp"
 #include "visual_language/vision_properties.hpp"
 #include "json_utils.hpp"
+#include "lora/helper.hpp"
+
 
 using namespace ov::genai;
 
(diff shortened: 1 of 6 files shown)


src/cpp/src/continuous_batching/pipeline.cpp
as-suvorov, line 52: `vision_encoder_properties` is not used now. It should have been addressed in properties routing PR I believe. We need to remove it from arguments, not sure if it should be done in this PR though.

as-suvorov (commented):
As I understand `ATTENTION_BACKEND` added to sample for testing purposes? Can we add python test instead?

likholat:
> As I understand `ATTENTION_BACKEND` added to sample for testing purposes? Can we add python test instead?

done

src/cpp/src/continuous_batching/pipeline.cpp
likholat: I believe this should be addressed in a separate PR

as-suvorov (approved):

yatarkan (approved):

PR #3846, opened 2026-05-11
Fix inference error for `gemma-4-31B-it`

p-wysocki:
### Details

The `has_token_type_ids()` is an incorrect classifier for attaching `token_type_ids`. The model mentioned in the ticket, `gemma-4-31B-it`, is not an MoE model, but in its config it sets `"use_bidirectional_attention": "vision"`, which uses `token_type_ids`. 

Proof:
Small models do not use `token_type_ids`: `"use_bidirectional_attention": null,`
Gemma4 E2B: [config.json · google/gemma-4-E2B-it at main](https://huggingface.co/google/gemma-4-E2B-it/blob/main/config.json#L137)
Gemma4 E4B: [config.json · google/gemma-4-E4B-it at main](https://huggingface.co/google/gemma-4-E4B-it/blob/main/config.json#L144)
 
But a bigger model, which is NOT an MoE model, has `"use_bidirectional_attention": "vision",`:
Gemma4 31B: [config.json · google/gemma-4-31B-it at main](https://huggingface.co/google/gemma-4-31B-it/blob/main/config.json#L123)
 
Same as the MoE Gemma:
[config.json · google/gemma-4-26B-A4B-it at main](https://huggingface.co/google/gemma-4-26B-A4B-it/blob/main/config.json#L93)

`token_type_ids` is required for bidirectional mask creation. This fix removes the error mentioned in the ticket in my local setup, however there are still accuracy issues when using INT4 model. FP16 works fine for me.

### Ticket

CVS-186483

### DIFF
--- a/src/cpp/src/visual_language/gemma4/classes.cpp
+++ b/src/cpp/src/visual_language/gemma4/classes.cpp
@@ -312,7 +312,7 @@ std::pair<ov::Tensor, ov::Tensor> InputsEmbedderGemma4::get_inputs_embeds_with_t
 }
 
 bool InputsEmbedderGemma4::has_token_type_ids() const {
-    return m_vlm_config.enable_moe_block;
+    return m_vlm_config.use_bidirectional_attention;
 }
 
 ov::Tensor InputsEmbedderGemma4::get_token_type_ids(const ov::Tensor& input_ids) {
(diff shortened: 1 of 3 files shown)


as-suvorov:
Thanks for the PR. To be fixed in: https://github.com/openvinotoolkit/openvino.genai/pull/3844

PR #3860, opened 2026-05-14
Expose VLM request-level metrics via GenerationHandle

junruizh2021:
Add VLMPerfMetrics support to GenerationHandle so that continuous batching callers can retrieve per-request VLM performance metrics (prepare_embeddings_duration, image_slice_count, etc.) after generation completes.

Key changes:
- GenerationHandleImpl: add m_vlm_perf_metrics field, constructor overload, get/set_vlm_perf_metrics(), get_perf_metrics()
- GenerationStream: add set/get_perf_metrics() for passing metrics from SequenceGroup to handles
- SequenceGroup: accumulate per-step inference metrics and expose get_perf_metrics()/update_perf_metrics()
- VLMPerfMetrics: add image_slice_count field and getter
- VLMRawPerfMetrics: add image_slice_counts vector
- pipeline_base.cpp: track image slice counts and embeddings duration in add_request overloads, attach metrics to handles
- pipeline.cpp: track image slice counts in SDPA VLM pipeline
- pipeline_impl.cpp: propagate perf metrics to streams on generation completion and request cleanup
- Python bindings: expose get_image_slice_count() and image_slice_counts

## Description

CVS-186710

Fixes #(issue)

## Checklist:


src/cpp/src/continuous_batching/pipeline_impl.cpp
```diff
@@ -470,6 +470,10 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::step() {
         report_tokens_timer.end();
     }
 
+    for (const auto& request : m_requests) {
+        request->update_perf_metrics(MicroSeconds(m_pipeline_metrics.inference_duration), m_batch_size);
+    }
```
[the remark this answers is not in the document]
junruizh2021: Good catch. Fixed in the updated commit — now only requests that are both `is_scheduled()` and `is_running()` get their metrics updated, and the batch size passed is `1` (per-request token count) instead of the total `m_batch_size`:

```cpp
for (const auto& request : m_requests) {
    if (request->is_scheduled() && request->is_running()) {
        request->update_perf_metrics(MicroSeconds(m_pipeline_metrics.inference_duration), 1);
    }
}
```

This ensures requests that didn't produce a token in this step are skipped, and each request's `num_generated_tokens` accumulates correctly.

src/cpp/src/sequence_group.hpp
```diff
@@ -802,6 +805,30 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
...
+    PerfMetrics get_perf_metrics() {
+        auto& raw_metrics = m_perf_metrics.raw_metrics;
+        if (raw_metrics.m_inference_durations.empty()) {
+            raw_metrics.m_inference_durations = {{MicroSeconds(0.0f)}};
+        }
+        m_perf_metrics.num_input_tokens = get_prompt_len();
+        m_perf_metrics.m_evaluated = false;
+        m_perf_metrics.evaluate_statistics(m_start_time);
+        return m_perf_metrics;
+    }
```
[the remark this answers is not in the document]
junruizh2021: Fixed. `get_perf_metrics()` now records `generate_durations` as `now - m_start_time` before calling `evaluate_statistics()`, so `get_generate_duration()` returns a valid per-request wall-clock duration:

```cpp
if (raw_metrics.generate_durations.empty()) {
    auto now = std::chrono::steady_clock::now();
    auto total_duration = std::chrono::duration_cast<MicroSeconds>(now - m_start_time);
    raw_metrics.generate_durations.push_back(total_duration);
}
```

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -20,6 +20,19 @@ std::unordered_map<std::string, ov::Tensor> deep_copy_tensors_map(
...
+        return 1;
+    }
+    return image.slices_shape[0] * image.slices_shape[1];
+}
+
+void update_image_slice_counts(ov::genai::VLMPerfMetrics& metrics, const std::vector<ov::genai::EncodedImage>& images) {
+    for (const auto& image : images) {
+        metrics.vlm_raw_metrics.image_slice_counts.emplace_back(get_image_slice_count(image));
+    }
+}
```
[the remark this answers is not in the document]
junruizh2021: Agreed. Extracted both functions as inline helpers into src/cpp/src/visual_language/vlm_utils.hpp under ov::genai::vlm_utils namespace, and updated all call sites in pipeline_base.cpp and pipeline.cpp to use vlm_utils::update_image_slice_counts(). No more duplication.

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -27,6 +27,19 @@
...
+        return 1;
+    }
+    return image.slices_shape[0] * image.slices_shape[1];
+}
+
+void update_image_slice_counts(VLMPerfMetrics& metrics, const std::vector<EncodedImage>& images) {
+    for (const auto& image : images) {
+        metrics.vlm_raw_metrics.image_slice_counts.emplace_back(get_image_slice_count(image));
+    }
+}
```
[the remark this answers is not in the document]
junruizh2021: Same as above — consolidated into vlm_utils.hpp. Both files now use `vlm_utils::update_image_slice_counts()`.

src/cpp/include/openvino/genai/visual_language/perf_metrics.hpp
```diff
@@ -12,17 +12,22 @@ namespace ov::genai {
...
+    /** @brief Total number of image slices produced for the request */
+    size_t image_slice_count = 0;
 
     MeanStdPair get_prepare_embeddings_duration();
+    size_t get_image_slice_count();
 
     VLMPerfMetrics() = default;
 
-    VLMPerfMetrics(PerfMetrics& perf_metrics) : PerfMetrics(perf_metrics), prepare_embeddings_duration(){};
+    VLMPerfMetrics(PerfMetrics& perf_metrics) : PerfMetrics(perf_metrics), prepare_embeddings_duration(), image_slice_count(0){};
```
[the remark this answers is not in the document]
junruizh2021: Fixed. Changed to `explicit VLMPerfMetrics(const PerfMetrics& perf_metrics)`.

src/python/py_vlm_pipeline.cpp
```diff
@@ -198,11 +198,13 @@ void init_vlm_pipeline(py::module_& m) {
...
             return common_utils::get_ms(rw, &ov::genai::VLMRawPerfMetrics::prepare_embeddings_durations);
-        });
+        })
+        .def_readonly("image_slice_counts", &ov::genai::VLMRawPerfMetrics::image_slice_counts);
 
     py::class_<ov::genai::VLMPerfMetrics, ov::genai::PerfMetrics>(m, "VLMPerfMetrics", perf_metrics_docstring)
         .def(py::init<>())
         .def("get_prepare_embeddings_duration", &ov::genai::VLMPerfMetrics::get_prepare_embeddings_duration)
+        .def("get_image_slice_count", &ov::genai::VLMPerfMetrics::get_image_slice_count)
         .def_readonly("vlm_raw_metrics", &ov::genai::VLMPerfMetrics::vlm_raw_metrics);
```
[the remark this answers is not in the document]
junruizh2021: Added. Extended `test_perf_metrics` with assertions for both `perf_metrics.get_image_slice_count()` and `vlm_raw_metrics.image_slice_counts`.

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -119,6 +128,10 @@ GenerationHandleImpl {
     GenerationOutputs read();
     // Reads all generated tokens for all sequences
     std::vector<GenerationOutput> read_all();
+
+    PerfMetrics get_perf_metrics();
+    VLMPerfMetrics get_vlm_perf_metrics();
+    void set_vlm_perf_metrics(VLMPerfMetrics vlm_perf_metrics);
```
yatarkan: Update python bindings for GenerationHandle

src/cpp/include/openvino/genai/visual_language/perf_metrics.hpp
```diff
@@ -12,17 +12,22 @@ namespace ov::genai {
...
     std::vector<MicroSeconds> prepare_embeddings_durations;
+    /** @brief Number of image slices produced for each input image */
+    std::vector<size_t> image_slice_counts;
 };
 
 struct OPENVINO_GENAI_EXPORTS VLMPerfMetrics : public PerfMetrics {
     /** @brief Mean and standard deviation of preparation of embeddings in milliseconds */
     MeanStdPair prepare_embeddings_duration;
+    /** @brief Total number of image slices produced for the request */
+    size_t image_slice_count = 0;
```
yatarkan: `image_slice_counts` and `image_slice_count` are a bit confusing - not clear which is per image and which is total per request. Let's rename and update getter as well.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -678,19 +677,27 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::add_request(uint64_t re
...
+        vlm_utils::update_image_slice_counts(metrics, encoded_images);
 
         const auto [unified_prompt, image_sequence, video_sequence] = m_inputs_embedder->normalize_prompt(prompt, 0, encoded_images);
         if (m_inputs_embedder->has_token_type_ids()) {
             std::tie(inputs, token_type_ids) = m_inputs_embedder->get_inputs_embeds_with_token_type_ids(unified_prompt, encoded_images, metrics, true, image_sequence);
         } else {
             inputs = m_inputs_embedder->get_inputs_embeds(unified_prompt, encoded_images, metrics, true, image_sequence);
         }
-        return add_request(request_id, inputs, sampling_params, token_type_ids, prompt_ids, m_inputs_embedder->get_lm_extra_inputs());
+        lm_extra_inputs = deep_copy_tensors_map(m_inputs_embedder->get_lm_extra_inputs());
```
yatarkan: Why deep copy is introduced here?

src/cpp/src/continuous_batching/pipeline_impl.cpp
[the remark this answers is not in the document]
yatarkan: How about chunked prefill case, will it work?

src/cpp/src/visual_language/vlm_utils.hpp
```diff
@@ -13,10 +13,25 @@
...
 #include "openvino/core/except.hpp"
 #include "openvino/genai/tokenizer.hpp"
+#include "openvino/genai/visual_language/perf_metrics.hpp"
 #include "openvino/runtime/tensor.hpp"
+#include "visual_language/vision_encoder.hpp"
 
 namespace ov::genai::vlm_utils {
 
+inline size_t get_image_slice_count(const EncodedImage& image) {
+    if (image.slices_shape.size() < 2) {
```
yatarkan, line 25: Returning 1 should be documented in docstring for this function.
`slices_shape` seems to be used and populated only for minicpm model.

.gitignore
```diff
@@ -47,3 +47,7 @@ _codeql_detected_source_root
 
 # agentic model enabler
 .model_enabler
+PR_DESCRIPTION.md
+test_vlm_metrics
+test_vlm_metrics.cpp
+kernel.errors.txt
```
yatarkan: This should not be added to gitignore

tests/python_tests/test_vlm_pipeline.py
yatarkan, line 1: Currently tests are extended for stateful VLM pipeline, but continuous batching pipeline with `add_request` and newly introduced `get_vlm_perf_metrics` is not tested.

yatarkan (changes requested):

yatarkan:
There is merge conflict, fyi.
And please double check new Copilot's comments.

src/cpp/include/openvino/genai/generation_handle.hpp
junruizh2021: Agreed. I’ll expose both get_perf_metrics() and get_vlm_perf_metrics() on the Python GenerationHandle binding and update the Python type stubs accordingly. I’ll also cover the new APIs in a continuous-batching Python test.

src/cpp/include/openvino/genai/visual_language/perf_metrics.hpp
junruizh2021: Agreed. The current singular/plural distinction does not clearly communicate the aggregation level. I’ll rename the raw per-image field to per_image_slice_counts, and the request-level aggregate to total_image_slice_count, with a corresponding get_total_image_slice_count() getter.

src/cpp/src/continuous_batching/pipeline_base.cpp
junruizh2021: The deep copy was introduced because get_lm_extra_inputs() returns tensors owned by the mutable inputs embedder, while the new code moved the downstream add_request() call outside m_embeddings_mutex. A shallow map copy would still share tensor storage that may be reused by another request. However, this adds an unnecessary and potentially expensive copy. I’ll preserve the original lifetime/locking behavior by invoking the internal add_request() while holding the mutex, then attach the collected metrics to the returned handle after leaving the critical section.

junruizh2021:
https://github.com/openvinotoolkit/openvino.genai/pull/3860#discussion_r3537023217

Good point. The current is_scheduled() && is_running() check is insufficient for chunked prefill: a request may be scheduled for a prefill chunk without producing a generated token. Recording one token for every such step would overcount generated tokens and distort TTFT/TPOT. I’ll derive the per-request generated-token count from the scheduler/sampler output and update token metrics only when that count is non-zero. Prefill inference time should be tracked separately without appending a generated-token timestamp or batch-size entry.

src/cpp/src/visual_language/vlm_utils.hpp
junruizh2021: Correct. slices_shape is currently populated only for models that explicitly split an image, such as MiniCPM. For other models, an encoded image is treated as one unsliced image unit, which is why the fallback is 1. I’ll document this behavior explicitly and rename the getter to get_total_image_slice_count() as discussed above.

.gitignore
junruizh2021: Removed these entries from the PR.

tests/python_tests/test_vlm_pipeline.py
junruizh2021: Agreed. The existing assertions only validate metrics returned by the stateful VLM pipeline. I’ll add a continuous-batching test that submits a VLM request through add_request(), waits for completion, retrieves metrics from the returned GenerationHandle, and validates both the base and VLM-specific metrics.

src/cpp/src/generation_handle.cpp
```diff
@@ -77,3 +77,22 @@ std::vector<GenerationOutput> GenerationHandleImpl::read_all() {
...
+VLMPerfMetrics GenerationHandleImpl::get_vlm_perf_metrics() {
+    auto base_metrics = get_perf_metrics();
+    VLMPerfMetrics metrics(base_metrics);
+    if (m_vlm_perf_metrics.has_value()) {
+        metrics.vlm_raw_metrics = m_vlm_perf_metrics->vlm_raw_metrics;
+    }
+    metrics.m_evaluated = false;
+    metrics.evaluate_statistics();
+    return metrics;
+}
```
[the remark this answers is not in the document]
junruizh2021: Good catch. m_vlm_perf_metrics contains the request-setup metrics collected by InputsEmbedder, while get_perf_metrics() returns the generation-stage metrics from GenerationStream. The current implementation only preserves the VLM-specific raw fields and therefore drops tokenization and chat-template durations. I’ll merge the request-setup raw durations into the generation metrics before reevaluating the statistics, and add assertions for these fields in the continuous-batching test.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -674,19 +677,27 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::add_request(uint64_t re
     std::optional<ov::Tensor> token_type_ids;
     // FIXME prompt_ids is not populated for VLM prompt lookup with add_request API
     std::optional<ov::Tensor> prompt_ids;
+    std::unordered_map<std::string, ov::Tensor> lm_extra_inputs;
+    const auto start_get_inputs_embeds = std::chrono::steady_clock::now();
     {
         std::lock_guard<std::mutex> lock(m_embeddings_mutex);
         m_inputs_embedder->set_apply_chat_template_status(sampling_params.apply_chat_template);
```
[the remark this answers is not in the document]
junruizh2021: Agreed. The metric is intended to measure embeddings preparation, not queueing time on m_embeddings_mutex. I’ll move both timestamps inside the critical section so the duration starts after the lock has been acquired and ends immediately after the embeddings and extra inputs have been prepared.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -716,16 +727,24 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::add_request(
...
```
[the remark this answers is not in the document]
junruizh2021: Agreed. This overload has the same issue. I’ll apply the same change here so both image-only and image/video request paths measure embeddings preparation only after acquiring m_embeddings_mutex.

src/cpp/src/sequence_group.hpp
```diff
@@ -858,6 +861,36 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
...
+        auto& raw_metrics = m_perf_metrics.raw_metrics;
+        if (raw_metrics.m_inference_durations.empty()) {
+            raw_metrics.m_inference_durations = {{MicroSeconds(0.0f)}};
+        }
+        raw_metrics.m_inference_durations[0] += inference_duration;
+        if (batch_size > 0) {
+            raw_metrics.m_token_infer_durations.emplace_back(inference_duration);
+            raw_metrics.m_new_token_times.emplace_back(std::chrono::steady_clock::now());
+            raw_metrics.m_batch_sizes.emplace_back(batch_size);
+        }
```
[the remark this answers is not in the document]
yatarkan: Please check

src/cpp/include/openvino/genai/generation_handle.hpp
yatarkan: 1. New public API setter `set_vlm_perf_metrics()` is added, but I can't see where it is used. Also there is no such perf metrics setter w/o vlm specific (`set_perf_metrics()`)
2. There is no python bindings for this setter `set_vlm_perf_metrics`

Do we really need it?

src/python/py_vlm_pipeline.cpp
```diff
@@ -198,11 +198,16 @@ void init_vlm_pipeline(py::module_& m) {
...
+        })
+        .def_readonly("per_image_slice_counts", &ov::genai::VLMRawPerfMetrics::per_image_slice_counts);
 
     py::class_<ov::genai::VLMPerfMetrics, ov::genai::PerfMetrics>(m, "VLMPerfMetrics", perf_metrics_docstring)
         .def(py::init<>())
         .def("get_prepare_embeddings_duration", &ov::genai::VLMPerfMetrics::get_prepare_embeddings_duration)
+        .def("get_total_image_slice_count", &ov::genai::VLMPerfMetrics::get_total_image_slice_count,
+             R"(Returns the total number of image slices processed for the request.
+
+An input image without explicit slicing metadata counts as one slice.)")
```
yatarkan, line 267: Did you generate `py_openvino_genai.pyi` file with `pybind11-stubgen` ?
It seems `py_openvino_genai.pyi` and `py_vlm_pipeline.cpp` are not aligned (docstrings). If so CI will fail.

tests/python_tests/test_vlm_pipeline.py
```diff
@@ -849,6 +849,18 @@ def test_vlm_continuous_batching_generate_vs_add_request(
...
+
+        assert perf_metrics.get_num_generated_tokens() > 0
+        assert vlm_perf_metrics.get_num_generated_tokens() == perf_metrics.get_num_generated_tokens()
+        assert vlm_perf_metrics.get_tokenization_duration().mean > 0
+        assert len(vlm_perf_metrics.vlm_raw_metrics.prepare_embeddings_durations) == 1
+        assert vlm_perf_metrics.get_prepare_embeddings_duration().mean > 0
+        assert len(vlm_perf_metrics.vlm_raw_metrics.per_image_slice_counts) == len(images)
+        assert vlm_perf_metrics.get_total_image_slice_count() == sum(
+            vlm_perf_metrics.vlm_raw_metrics.per_image_slice_counts
+        )
```
yatarkan, line 890: Does it make sense to also check/compare perf_metrics & vlm_perf_metrics between handle (`add_request()` API) and CB `generate()` API?

yatarkan (changes requested):

src/cpp/src/sequence_group.hpp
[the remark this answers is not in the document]
junruizh2021: Fixed in `9b9f4aa8`. `update_perf_metrics()` now divides the step inference duration by the per-request generated-token count and records one duration entry for each generated token. This keeps `m_token_infer_durations` consistent with its per-token semantics and preserves the correct weighting when `batch_size > 1`.

I also added assertions verifying that the number of token inference duration entries and the sum of recorded batch sizes match the number of generated tokens. Both greedy and beam-search cases pass (`2 passed`).

src/cpp/include/openvino/genai/generation_handle.hpp
junruizh2021: Good point. The setter was only used as internal plumbing by the VLM `add_request()` paths. In `9b9f4aa8`, I removed the public `set_vlm_perf_metrics()` API and the VLM-specific `GenerationHandleImpl` constructor. The VLM setup metrics are now passed through the internal request path and stored in the internal `GenerationStream`.

Therefore, no Python setter binding is needed; only `get_perf_metrics()` and `get_vlm_perf_metrics()` remain public.

src/python/py_vlm_pipeline.cpp
junruizh2021: Yes. I regenerated `py_openvino_genai.pyi` with the project-pinned `pybind11-stubgen==2.5.5` after aligning the pybind docstring. The generated stubs now exactly match the committed `.pyi` files after applying the repository cleanup scripts, and the Python extension build passes.

tests/python_tests/test_vlm_pipeline.py
junruizh2021: Yes. I extended `test_vlm_continuous_batching_generate_vs_add_request` to compare the metrics returned by the handle with the metrics returned by the CB `generate()` API. The test compares deterministic fields such as input/generated token counts, embedding metric entry counts, per-image slice counts, and total image slice count; timing values are checked for validity rather than exact equality because the requests run independently.

Both greedy and beam-search cases pass (`2 passed`).

src/cpp/src/continuous_batching/pipeline_base.hpp
```diff
@@ -85,7 +85,8 @@ class ContinuousBatchingPipeline::IContinuousBatchingPipeline {
                                          const GenerationConfig& sampling_params,
                                          std::optional<ov::Tensor> token_type_ids = std::nullopt,
                                          std::optional<ov::Tensor> prompt_ids = std::nullopt,
-                                         std::optional<std::unordered_map<std::string, ov::Tensor>> lm_extra_inputs = std::nullopt) = 0;
+                                         std::optional<std::unordered_map<std::string, ov::Tensor>> lm_extra_inputs = std::nullopt,
+                                         std::optional<VLMPerfMetrics> vlm_perf_metrics = std::nullopt) = 0;
```
yatarkan: I am quite concerned about propagating `std::optional<VLMPerfMetrics> vlm_perf_metrics` through all `add_request()` overrides in `ContinuousBatchingImpl`, `PromptLookupImpl`, `SpeculativeDecodingImpl`, `Eagle3DecodingImpl`.
Why not to store VLM metrics in `GenerationStream` (as they are stored now) and set them on the stream after getting the handle? For example:
```cpp
GenerationHandle handle = add_request(request_id, inputs, sampling_params, token_type_ids, prompt_ids, lm_extra_inputs);
handle->get_generation_stream()->set_vlm_perf_metrics(std::move(metrics));
```

src/cpp/src/visual_language/pipeline.cpp
```diff
@@ -458,7 +460,7 @@ class VLMPipeline::VLMPipelineImpl : public VLMPipelineBase{
         }
 
         auto generate_end_time = std::chrono::steady_clock::now();
-        decoded.perf_metrics = encoded_result.perf_metrics;
+        static_cast<PerfMetrics&>(decoded.perf_metrics) = encoded_result.perf_metrics;
```
yatarkan: Can we use explicit constructor instead of static cast?
```cpp
decoded.perf_metrics = VLMPerfMetrics(encoded_result.perf_metrics);
```

src/cpp/src/sequence_group.hpp
```diff
@@ -392,6 +393,8 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
     size_t m_output_seq_len = 0;
 
     size_t m_num_streamed_tokens = 0, m_stream_window_size = 0;
+    TimePoint m_start_time = std::chrono::steady_clock::now();
```
yatarkan, line 433: `m_start_time` is set when SequenceGroup is created, not when it starts generating. If request is in the queue this will affect `raw_metrics.generate_durations`

yatarkan (changes requested):
1. Please also check out Copilot's comments:
    - https://github.com/openvinotoolkit/openvino.genai/pull/3860#discussion_r3577338869
    - https://github.com/openvinotoolkit/openvino.genai/pull/3860#discussion_r3577338905
2. [Lint job](https://github.com/openvinotoolkit/openvino.genai/actions/runs/29317928927/job/87063087789?pr=3860) in CI is failing

src/cpp/src/continuous_batching/pipeline_base.hpp
junruizh2021: Addressed. I removed the std::optional<VLMPerfMetrics> parameter from the encoded-input add_request() interface and from the ContinuousBatching, PromptLookup, SpeculativeDecoding, and Eagle3 overrides.
The VLM overload now calls the regular add_request() first and then stores the request-setup metrics in the returned handle's GenerationStream. I added a small setter on GenerationHandleImpl for this instead of exposing the internal stream directly. This keeps the VLM-specific data out of the common add_request() call chain.

src/cpp/src/visual_language/pipeline.cpp
junruizh2021: Addressed. I replaced the base-class static_cast assignment with:
`decoded.perf_metrics = VLMPerfMetrics(encoded_result.perf_metrics);`
I updated both occurrences in pipeline.cpp.

src/cpp/src/sequence_group.hpp
junruizh2021: Yes, `m_start_time` is captured when the `SequenceGroup` is created, so the current `generate_duration` includes time spent waiting in the scheduler queue.
I have left this unchanged for now because the same timestamp is also used as the TTFT baseline, where including queueing time represents the request's end-to-end latency. If `generate_duration` is expected to measure active scheduled generation only, I think we should introduce a separate timestamp set when the request is first scheduled rather than change `m_start_time`, since changing it would also alter the TTFT semantics. Could you please confirm which meaning is expected here?

src/cpp/src/continuous_batching/pipeline_impl.cpp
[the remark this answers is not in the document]
junruizh2021: Thanks for checking this. The full batched inference duration is intentionally attributed to every scheduled request because it represents the wall-clock latency experienced by each request during that step. Dividing the duration among the requests would make their individual inference latency appear artificially lower.
The per-request generated-token count is used only to derive that request’s per-token inference duration. A scheduled request with zero generated tokens may still be processing prompt tokens, for example during chunked prefill, so its inference duration should still be accumulated while no token-level timing sample is added.
Therefore, I have kept the current attribution unchanged.

src/cpp/src/sequence_group.hpp
```diff
@@ -858,6 +861,36 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
...
+        raw_metrics.m_inference_durations[0] += inference_duration;
+        if (batch_size > 0) {
+            raw_metrics.m_token_infer_durations.emplace_back(inference_duration);
+            raw_metrics.m_new_token_times.emplace_back(std::chrono::steady_clock::now());
+            raw_metrics.m_batch_sizes.emplace_back(batch_size);
+        }
+    }
+
+    PerfMetrics get_perf_metrics() {
+        auto& raw_metrics = m_perf_metrics.raw_metrics;
```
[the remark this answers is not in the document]
junruizh2021: m_new_token_times is intentionally recorded once per generation step rather than once per generated token. Each timestamp is paired with the corresponding entry in m_batch_sizes, and PerfMetrics::evaluate_statistics() uses that batch size to expand the interval into per-token TPOT samples.
Appending batch_size identical timestamps would break the one-to-one alignment between m_new_token_times and m_batch_sizes, or produce zero-duration intervals if both vectors were expanded. The current step-level timestamp plus token-count representation is therefore intentional, so I have left it unchanged.

junruizh2021:
> 1. Please also check out Copilot's comments:
>    
>    * [#3860 (comment)](https://github.com/openvinotoolkit/openvino.genai/pull/3860#discussion_r3577338869)
>    * [#3860 (comment)](https://github.com/openvinotoolkit/openvino.genai/pull/3860#discussion_r3577338905)
> 2. [Lint job](https://github.com/openvinotoolkit/openvino.genai/actions/runs/29317928927/job/87063087789?pr=3860) in CI is failing

Q1 Answer: I provided more information in these two threads.

Q2 Answer: The lint job failed because darker detected a formatting issue in tests/python_tests/test_vlm_pipeline.py. The assertion was split across multiple lines, while the configured formatter expected it to fit on a single line.
I reformatted the assertion and verified the fix with:
`pre-commit run --files tests/python_tests/test_vlm_pipeline.py`
All pre-commit hooks, including darker, now pass.

src/cpp/src/sequence_group.hpp
yatarkan: Ok, let's keep it as is now if there is the same pattern for TTFT

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -119,6 +122,10 @@ GenerationHandleImpl {
...
```
yatarkan: I believe `set_vlm_perf_metrics` should not be public and exposed to user - it is used only internally in `pipeline_base.cpp`.

yatarkan (changes requested):

src/cpp/src/generation_handle.cpp
```diff
@@ -77,3 +77,38 @@ std::vector<GenerationOutput> GenerationHandleImpl::read_all() {
...
+
+void GenerationHandleImpl::set_vlm_perf_metrics(VLMPerfMetrics vlm_perf_metrics) {
+    m_generation_stream->set_vlm_perf_metrics(std::move(vlm_perf_metrics));
+}
+
+VLMPerfMetrics GenerationHandleImpl::get_vlm_perf_metrics() {
+    auto base_metrics = get_perf_metrics();
+    VLMPerfMetrics metrics(base_metrics);
+    const auto request_setup_metrics = m_generation_stream->get_vlm_perf_metrics();
```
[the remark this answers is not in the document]
junruizh2021: Fixed in the latest update. I added an explicit `VLMPerfMetrics(PerfMetrics&&)` constructor and now construct the result with `std::move(base_metrics)`. This removes the second copy of the potentially large per-token vectors; the copy returned from the stored stream metrics remains necessary.

src/cpp/src/generation_stream.hpp
```diff
@@ -2,10 +2,13 @@
...
 #pragma once
-#include <mutex>
 #include <atomic>
+#include <mutex>
+#include <optional>
+
 #include "openvino/genai/continuous_batching_pipeline.hpp"
 #include "openvino/genai/generation_handle.hpp"
+#include "openvino/genai/visual_language/perf_metrics.hpp"
 #include "synchronized_queue.hpp"
```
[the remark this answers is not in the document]
junruizh2021: Fixed. `generation_stream.hpp` now explicitly includes `openvino/core/except.hpp`, so its use of `OPENVINO_ASSERT` no longer depends on transitive includes.

src/cpp/include/openvino/genai/generation_handle.hpp
junruizh2021: Fixed. `set_vlm_perf_metrics` is no longer part of the public `GenerationHandleImpl` API. `pipeline_base.cpp` now uses an internal friend accessor to attach the metrics directly to the underlying `GenerationStream`, while the user-facing handle only exposes the metric getters.

src/cpp/src/sequence_group.hpp
```diff
@@ -858,6 +861,39 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
...
+        }
+        raw_metrics.m_inference_durations[0] += inference_duration;
+        if (batch_size > 0) {
+            const auto per_token_inference_duration = inference_duration / batch_size;
+            raw_metrics.m_token_infer_durations.insert(raw_metrics.m_token_infer_durations.end(),
+                                                       batch_size,
+                                                       per_token_inference_duration);
+            raw_metrics.m_new_token_times.emplace_back(std::chrono::steady_clock::now());
+            raw_metrics.m_batch_sizes.emplace_back(batch_size);
+        }
```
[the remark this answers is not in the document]
junruizh2021: Fixed. The pipeline now captures a single `step_end_time` before updating the scheduled requests and passes it, together with the step sampling duration, into `update_perf_metrics(...)`. Each token-producing request records that shared timestamp and its sampling-duration sample consistently. I also added test assertions that sampling durations are populated and aligned with `m_batch_sizes`.

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -4,11 +4,14 @@
 #pragma once
 
 #include <memory>
+#include <optional>
 #include <unordered_map>
+#include <utility>
```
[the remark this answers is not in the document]
junruizh2021: Updated. I removed the unused `<optional>` include. `<utility>` is still required in this header because the `GenerationHandleImpl` constructor uses `std::move(generation_stream)`, so I kept that include.

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -119,6 +128,9 @@ GenerationHandleImpl {
     GenerationOutputs read();
     // Reads all generated tokens for all sequences
     std::vector<GenerationOutput> read_all();
+
+    PerfMetrics get_perf_metrics();
+    VLMPerfMetrics get_vlm_perf_metrics();
 };
```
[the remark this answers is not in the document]
junruizh2021: Fixed in `16a7740f`. Both `GenerationHandleImpl::get_perf_metrics()` and `get_vlm_perf_metrics()` are now `const` in their declarations and definitions. The Python binding target also builds successfully with the const-qualified method pointers.

src/cpp/src/generation_handle.cpp
```diff
@@ -77,3 +77,34 @@ std::vector<GenerationOutput> GenerationHandleImpl::read_all() {
...
+            request_setup_raw_metrics.chat_template_durations.end());
+        result_raw_metrics.detokenization_durations.insert(
+            result_raw_metrics.detokenization_durations.end(),
+            request_setup_raw_metrics.detokenization_durations.begin(),
+            request_setup_raw_metrics.detokenization_durations.end());
+    }
+    metrics.m_evaluated = false;
+    metrics.evaluate_statistics();
+    return metrics;
+}
```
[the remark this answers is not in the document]
junruizh2021: Fixed in `16a7740f`. `get_vlm_perf_metrics()` now asserts that request-setup VLM metrics are present before constructing the result, with a clear error for non-VLM requests. I also removed the previous empty-optional path, so the method can no longer return a misleading VLM metrics object without VLM data.

src/cpp/src/visual_language/perf_metrics.cpp
```diff
@@ -31,6 +40,11 @@ VLMPerfMetrics VLMPerfMetrics::operator+(const VLMPerfMetrics& right) const {
     result_prepare_embeddings_durations.insert(result_prepare_embeddings_durations.end(),
                                                 right_prepare_embeddings_durations.begin(),
                                                 right_prepare_embeddings_durations.end());
+    auto& result_per_image_slice_counts = result.vlm_raw_metrics.per_image_slice_counts;
+    const auto& right_per_image_slice_counts = right.vlm_raw_metrics.per_image_slice_counts;
```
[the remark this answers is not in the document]
junruizh2021: Fixed in `1be45dfd`. `VLMPerfMetrics::operator+` now constructs the result with `std::move(base_result)`, reusing the rvalue constructor and avoiding the extra copy of the base metrics and their per-token vectors. Pre-commit checks, `git diff --check`, and the `py_openvino_genai` build all pass.

src/cpp/src/continuous_batching/pipeline_impl.cpp
```diff
@@ -649,6 +664,7 @@ ContinuousBatchingPipeline::ContinuousBatchingImpl::generate(const std::vector<o
         perf_metrics.evaluate_statistics(start_time);
 
         result.perf_metrics = perf_metrics;
+        request->get_generation_stream()->set_perf_metrics(perf_metrics);
         results.push_back(std::move(result));
```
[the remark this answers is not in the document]
junruizh2021: Fixed in `dd137172`. I removed the redundant `set_perf_metrics(perf_metrics)` call from `generate()`. The stream now retains the request-level metrics stored by `_free_non_running_requests()`, while the aggregate metrics remain only in `result.perf_metrics`, avoiding both the overwrite and the extra copy. Pre-commit checks, `git diff --check`, and the `py_openvino_genai` build all pass.

src/cpp/src/continuous_batching/pipeline_impl.cpp
```diff
@@ -669,6 +684,7 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::_free_non_running_reque
     while (requests_iterator != m_requests.end()) {
         const auto& request = *requests_iterator;
         if(request->has_finished() || request->handle_stopped() || request->handle_cancelled()) {
+            request->get_generation_stream()->set_perf_metrics(request->get_perf_metrics());
             for (const auto& sequence: request->get_sequences()) {
```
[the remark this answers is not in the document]
junruizh2021: Fixed in `3ee4f184`. `_free_non_running_requests()` now sets `perf_metrics.load_time` from the pipeline's `m_load_time_ms` before moving the metrics into `GenerationStream`, so the value also propagates through `get_vlm_perf_metrics()`.

I also removed the duplicate `m_load_time_ms` member from `ContinuousBatchingImpl`, which was shadowing the initialized base-class field and would otherwise keep the handle value at `0`. The existing VLM handle test now asserts that load time is positive and identical for `get_perf_metrics()` and `get_vlm_perf_metrics()`.

The `py_openvino_genai` build, targeted pytest (`2 passed`), pre-commit checks, and `git diff --check` all pass.

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -5,13 +5,19 @@
...
 #include "openvino/genai/generation_config.hpp"
 #include "openvino/genai/visibility.hpp"
 #include "openvino/genai/perf_metrics.hpp"
+#include "openvino/genai/visual_language/perf_metrics.hpp"
 
 namespace ov::genai {
 
+namespace detail {
+class GenerationHandleImplAccess;
+}
```
yatarkan: My concerns regarding `detail` namespace and `GenerationHandleImplAccess`:
1. New `detail` namespace is in a public header and not hidden so a user can still overwrite perf metrics by calling `datail::GenerationHandleImplAccess::set_vlm_perf_metrics(...)`.
2. `detail` namespace leaks private `GenerationStream` dependency into public header.
3. We do not have such patter in GenAI codebase.

Instead, `GenerationHandleImpl` can have `ContinuousBatchingPipeline` as a friend class so in `pipeline_base.cpp` in `add_request(...)` method we can access handle's generation stream and set VLM perf metrics:
```cpp
handle->m_generation_stream->set_vlm_perf_metrics(metrics);
```

yatarkan (changes requested):

src/cpp/include/openvino/genai/generation_handle.hpp
junruizh2021: Addressed in [`b849c45b`](https://github.com/openvinotoolkit/openvino.genai/commit/b849c45b2b7e139a33b0c98b537493c7ecc9c5cd). I removed `detail::GenerationHandleImplAccess`, forward-declared `ContinuousBatchingPipeline`, and made it a friend of `GenerationHandleImpl`. The VLM `add_request(...)` paths now set the metrics directly on `handle->m_generation_stream`, preserving `std::move(metrics)` to avoid an extra copy.

`git diff --check` and the full `py_openvino_genai` build pass.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -791,8 +798,18 @@ GenerationHandle ContinuousBatchingPipeline::IContinuousBatchingPipeline::add_re
...
+        const auto end_get_inputs_embeds = std::chrono::steady_clock::now();
+        metrics.vlm_raw_metrics.prepare_embeddings_durations.emplace_back(
+            PerfMetrics::get_microsec(end_get_inputs_embeds - start_get_inputs_embeds));
+        handle = add_request(request_id,
+                             inputs,
+                             sampling_params,
+                             token_type_ids,
+                             prompt_ids,
+                             m_inputs_embedder->get_lm_extra_inputs());
+        handle->m_generation_stream->set_vlm_perf_metrics(std::move(metrics));
```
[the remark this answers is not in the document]
junruizh2021: Thanks for checking. This access is valid C++: a nested class has the same access rights as any other member of its enclosing class. Since `GenerationHandleImpl` declares `ContinuousBatchingPipeline` as a friend, `ContinuousBatchingPipeline::IContinuousBatchingPipeline` can access `m_generation_stream`.

I verified this with the project build after the change: `pipeline_base.cpp` compiled successfully and the full `py_openvino_genai` target linked successfully. No additional setter or accessor is needed.

src/cpp/src/continuous_batching/pipeline_base.cpp
```diff
@@ -821,16 +838,29 @@ ContinuousBatchingPipeline::IContinuousBatchingPipeline::add_request(
...
```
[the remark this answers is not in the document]
junruizh2021: This overload uses the same valid nested-class access described above. Both direct `handle->m_generation_stream` accesses compiled successfully, and the full `py_openvino_genai` build completed and linked without errors. Therefore no API or friendship change is required.

yatarkan:
build_jenkins

yatarkan (approved):

PR #3867, opened 2026-05-14
Fix Gemma3 chat template formatting

yatarkan:
## Description
Currently `chat_template_fallback_map` includes basic Gemma3 chat template, which is the same and applicable for several Gemma3 models - `gemma3-1b-it` (LLM), `gemma3-4b-it` (VLM) etc. So original model chat template is replaced with the fallback chat template w/o trim filter (`| trim`) to be aligned with HF/optimum-intel chat template processing. But this processing alignment is needed only for Gemma3 **VLM** models (`gemma3-4b-it`) and not for **LLM** models (`gemma3-1b-it`), so such general replacement is affecting LLM models accuracy, especially with long prompt input.

This PR fixes accuracy for Gemma3 **LLM** model keeping the same accuracy for **VLM** model.

WWB accuracy results GenAI vs Optimum (SDPA, CPU, `--long-prompt`):
|        | `gemma3-1b-it` (LLM) | `gemma3-4b-it` (VLM) |
|--------|----------------------|----------------------|
| master | 0.9365119            | 0.98293966           |
| PR     | 0.99999046           | 0.98293966           |

Also this PR adds safe guard to fix corner-case issue with `text_config.use_bidirectional_attention` in `config.json` (introduced in #3844) when Gemma3 model is exported with `transformers==5.0.0` and `text_config.use_bidirectional_attention` is boolean (`false`).

CVS-181746

## Checklist:
- [x] Tests have been updated or added to cover the new code - **N/A**.
- [x] I have made corresponding changes to the documentation - **N/A**.


src/cpp/src/visual_language/vlm_config.cpp
```diff
@@ -82,7 +82,14 @@ VLMConfig::VLMConfig(const std::filesystem::path& json_path) {
...
     read_json_param(parsed, "text_config.hidden_size_per_layer_input", hidden_size_per_layer_input);
-    read_json_param(parsed, "text_config.use_bidirectional_attention", use_bidirectional_attention);
+    // For gemma3 `text_config.use_bidirectional_attention` can be absent or boolean
+    if (
+        parsed.contains("text_config")
+        && parsed.at("text_config").contains("use_bidirectional_attention")
+        && parsed.at("text_config").at("use_bidirectional_attention").is_string()
+    ) {
+        read_json_param(parsed, "text_config.use_bidirectional_attention", use_bidirectional_attention);
+    }
```
yatarkan, line 92: Sequential key access with `.at()` and checks with `.contains()` might be not efficient.

Other possible options:
1. Check for Gemma4 model type:
```cpp
if (model_type == VLMModelType::GEMMA4) {
    read_json_param(parsed, "text_config.hidden_size_per_layer_input", hidden_size_per_layer_input);
    // For gemma3 `text_config.use_bidirectional_attention` can be absent or boolean
    read_json_param(parsed, "text_config.use_bidirectional_attention", use_bidirectional_attention);
}
```

2. Set with one-liner using `.value()` getter with default value and nested key path with `_json_pointer` built-in from `nlohmann::json` instead of `read_json_param` from utils:
```cpp
use_bidirectional_attention = parsed.value("/text_config/use_bidirectional_attention"_json_pointer, "");
```

@as-suvorov @Wovchena which do you prefer?

Wovchena: I like the fancy one-liner

as-suvorov: How `_json_pointer` handles `{"use_bidirectional_attention": false}` case?
From docs:
```
The function can throw the following exceptions:
Throws [type_error.302](https://json.nlohmann.me/home/exceptions/#jsonexceptiontype_error302) 
if default_value does not match the type of the value at ptr
```

yatarkan: @as-suvorov you are right, it will also throw. `_json_pointer` would also need `.contains()` and type check:
```cpp
if (parsed.contains("/text_config/use_bidirectional_attention"_json_pointer) 
    && parsed["/text_config/use_bidirectional_attention"_json_pointer].is_string()
) {
    read_json_param(parsed, "text_config.use_bidirectional_attention", use_bidirectional_attention);
}
```
Probably we can keep the current implementation.

as-suvorov (approved):

PR #3868, opened 2026-05-15
[VLM] QWEN-3-Omni support

sgonorov:
## Description                                                                                                                                                                                                           
  Adds `Qwen3-Omni` VLM pipeline (Dense and MoE) with audio input/output support:
  - New `qwen3_omni/` classes, audio encoder, and speech pipeline                                                                                                                                                          
  - Audio streaming via `AudioStreamerBase` + `audio_utils`                                                                                                                                                                
  - Python/C++ samples (`qwen3_omni_chat`, `audio_to_text`) and pytest coverage                                                                                                                                            
  - Wires audio config/token IDs into `VLMConfig`, extends VLM and CB pipelines with audio plumbing    
                                                      
  ## Checklist:                                                                                                                                                                                                            
  - [x] Tests have been updated or added to cover the new code. <!-- tests/python_tests/test_qwen3_omni.py -->                                                                                                             

### DIFF
--- a/samples/cpp/omni/CMakeLists.txt
+++ b/samples/cpp/omni/CMakeLists.txt
@@ -1,0 +1,48 @@
+# Copyright (C) 2023-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+if (MSVC)
+  set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+endif()
+
+find_package(OpenVINOGenAI REQUIRED
+    PATHS
+        "${CMAKE_BINARY_DIR}"  # Reuse the package from the build.
+        ${OpenVINO_DIR}  # GenAI may be installed alogside OpenVINO.
+    NO_CMAKE_FIND_ROOT_PATH
+)
+
+file(DOWNLOAD
+    https://raw.githubusercontent.com/nothings/stb/f75e8d1cad7d90d72ef7a4661f1b994ef78b4e31/stb_image.h
+    ${CMAKE_BINARY_DIR}/stb_image.h
+    EXPECTED_HASH MD5=27932e6fb3a2f26aee2fc33f2cb4e696)
+
+if(POLICY CMP0135)
+    cmake_policy(SET CMP0135 NEW)
+endif()
+
+if(POLICY CMP0169)
+    cmake_policy(SET CMP0169 OLD)
+endif()
+
+include(FetchContent)
+
+if(NOT TARGET dr_libs)
+    FetchContent_Declare(dr_libs
+        URL https://github.com/mackron/dr_libs/archive/da35f9d6c7374a95353fd1df1d394d44ab66cf01.tar.gz
+        URL_HASH SHA256=2704d347f480ca1bc92233fb01747e4550cc8031735b6ea62ca9990ebb8851ae)
+    FetchContent_MakeAvailable(dr_libs)
+endif()
+
+add_executable(omni_chat omni_chat.cpp ../automatic_speech_recognition/audio_utils.cpp ../visual_language_chat/load_image.cpp)
+target_include_directories(omni_chat PRIVATE "${CMAKE_BINARY_DIR}" "../visual_language_chat" "../automatic_speech_recognition" "$<BUILD_INTERFACE:${dr_libs_SOURCE_DIR}>")
+target_link_libraries(omni_chat PRIVATE openvino::genai)
+
+set_target_properties(omni_chat PROPERTIES
+    # Ensure out of box LC_RPATH on macOS with SIP
+    INSTALL_RPATH_USE_LINK_PATH ON)
+
+install(TARGETS omni_chat
+        RUNTIME DESTINATION samples_bin/
+        COMPONENT samples_bin
+        EXCLUDE_FROM_ALL)
--- a/samples/cpp/omni/README.md
+++ b/samples/cpp/omni/README.md
@@ -1,0 +1,97 @@
+# Qwen3-Omni Chat Sample
+
+Interactive multimodal chat with Qwen3-Omni models supporting text, image, and speech input/output.
+
+## Description
+
+Demonstrates `ov::genai::OmniPipeline` for end-to-end multimodal conversations. Qwen3-Omni models accept text, images, and audio inputs, generating text and optionally synthesized speech responses.
+
+**Key features:**
+- Multi-turn conversations with image context
+- Optional speech synthesis (talker mode)
+- Separate generation configs for text decoding and speech output
+- Interactive CLI interface
+
+## Prerequisites
+
+Prepare image files (JPG, PNG) and audio files (16kHz mono WAV) for testing.
+
+## Usage
+
+```bash
+omni_chat <MODEL_DIR> <IMAGE_FILE_OR_DIR> <AUDIO_FILE>
+```
+
+**Parameters:**
+- `<MODEL_DIR>` — Path to exported Qwen3-Omni OpenVINO model directory
+- `<IMAGE_FILE_OR_DIR>` — Path to input image(s) for visual context
+- `<AUDIO_FILE>` — Path to input audio file (16kHz mono WAV)
+
+**Example:**
+
+```bash
+./build/samples/cpp/omni/omni_chat /models/qwen3-omni-ov ./coco.jpg ./audio.wav
+```
+
+**Interactive usage:**
+1. Images are loaded once at startup and available to all turns
+2. Type questions and press Enter
+3. Model responds with streaming text; speech output indicated when enabled
+4. Continue the conversation across multiple turns
+5. Press Ctrl+D to exit
+
+## Model Support
+
+Compatible with Qwen3-Omni models exported to OpenVINO format.
+
+Export using [Optimum Intel](https://github.com/huggingface/optimum-intel) CLI or Python API.
+
+## Configuration
+
+### Text Generation
+
+`GenerationConfig` controls text decoding (the "thinker" phase):
+
+```cpp
+ov::genai::GenerationConfig text_config;
+text_config.max_new_tokens = 256;
+```
+
+### Speech Synthesis
+
+`OmniTalkerSpeechConfig` controls talker and speech output:
+
+```cpp
+ov::genai::OmniTalkerSpeechConfig talker_speech_config(models_path);
+talker_speech_config.return_audio = true;  // Enable speech synthesis
+talker_speech_config.speaker = "Cherry";   // Select voice (optional)
+```
+
+**Available voices** vary by checkpoint. MoE models typically expose: `"Ethan"`, `"Chelsie"`, `"Aiden"`, `"Cherry"`. Check `talker_config.speaker_id` in the model's `config.json` for the full list. Leaving `speaker` empty selects the default voice.
+
+**Speech output:** 24kHz mono PCM samples returned in `OmniDecodedResults.speech_outputs`.
+
+## GPU Inference
+
+**Important:** Qwen3-Omni speech synthesis requires FP32 precision on GPU. FP16 causes numerical drift → codec token corruption → distorted audio. The pipeline enforces `INFERENCE_PRECISION_HINT=f32` automatically for GPU devices. Change the device argument in `OmniPipeline` constructor:
+
+```cpp
+ov::genai::OmniPipeline pipe(models_path, "GPU");
+```
+
+**MoE models** (30B-A3B) may exceed available GPU memory. Use CPU for MoE inference.
+
+## Speaker API Features
+
+The main sample demonstrates speaker enumeration and voice blending at startup:
+
+1. **List speakers** — shows all named voices in the model
+2. **Voice blending** — when 2+ speakers are available, creates a 50/50 blend and uses it for generation
+
+**Custom blending:** Retrieve speaker embeddings via `pipe.get_talker()->get_speaker_embedding(name)`, mix with custom ratios (e.g., 0.75A + 0.25B), and pass via `OmniTalkerSpeechConfig::speaker`.
+
+## See Also
+
+- [Python Qwen3-Omni chat sample](../../python/omni/)
+- [OmniPipeline C++ API reference](../../../src/cpp/include/openvino/genai/omni/pipeline.hpp)
+- [Qwen3-Omni model documentation](https://github.com/QwenLM/Qwen3-Omni)
--- a/samples/cpp/omni/omni_chat.cpp
+++ b/samples/cpp/omni/omni_chat.cpp
@@ -1,0 +1,140 @@
+// Copyright (C) 2023-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include <filesystem>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "audio_utils.hpp"
+#include "load_image.hpp"
+#include "openvino/genai/generation_config.hpp"
+#include "openvino/genai/omni/pipeline.hpp"
+#include "openvino/genai/omni/talker_speech_config.hpp"
+
+ov::genai::StreamingStatus print_subword(std::string&& subword) {
+    std::cout << subword << std::flush;
+    return ov::genai::StreamingStatus::RUNNING;
+}
+
+int main(int argc, char* argv[]) try {
+    if (argc != 4) {
+        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> <IMAGE_FILE_OR_DIR> <AUDIO_FILE>");
+    }
+
+    const std::filesystem::path models_path = argv[1];
+
+    // Two configs: text_config drives the thinker text decode, talker_speech_config drives
+    // the talker + speech output. Speech output is hardcoded on here to show the multimodal
+    // path. Set talker_speech_config.return_audio = false to get text-only responses.
+    ov::genai::GenerationConfig text_config;
+    text_config.max_new_tokens = 256;
+
+    ov::genai::OmniTalkerSpeechConfig talker_speech_config(models_path);
+    talker_speech_config.return_audio = true;
+    // Leaving talker_speech_config.speaker empty selects the model's default voice. Available
+    // voices vary by checkpoint.
+
+    std::vector<ov::Tensor> rgbs = utils::load_images(argv[2]);
+
+    ov::Tensor audio_tensor = utils::audio::read_wav_as_tensor(argv[3]);
+    std::vector<ov::Tensor> audios = {std::move(audio_tensor)};
+
+    // Compose OmniPipeline from separate VLM and Talker components.
+    // This allows independent device selection (e.g., VLM on GPU, Talker on CPU),
+    // sharing a VLM base across pipelines, or injecting custom TalkerBase implementations.
+    // For simpler use cases: ov::genai::OmniPipeline pipe(models_path, "CPU");
+
+    auto vlm_base = ov::genai::VLMPipeline::create_base(models_path, "CPU", {});
+    auto talker = std::shared_ptr<ov::genai::Talker>(new ov::genai::Talker(models_path, "CPU"));
+    ov::genai::OmniPipeline pipe(vlm_base, talker);
+
+    std::cout << "OmniPipeline composed successfully.\n";
+
+    // Speaker API demo: list available speakers and demonstrate voice blending.
+    std::vector<std::string> speakers = talker->list_speakers();
+    std::cout << "\n=== Available Speakers ===\n";
+    if (speakers.empty()) {
+        std::cout << "No named speakers (model has single default voice)\n";
+    } else {
+        std::cout << "Found " << speakers.size() << " speakers:\n";
+        for (const auto& name : speakers) {
+            std::cout << "  - " << name << "\n";
+        }
+        std::cout << "\n";
+
+        if (speakers.size() >= 2) {
+            std::cout << "=== Voice Blending Demo ===\n";
+            std::cout << "Blending " << speakers[0] << " + " << speakers[1] << " (50/50 mix)\n";
+            ov::Tensor emb1 = talker->get_speaker_embedding(speakers[0]);
+            ov::Tensor emb2 = talker->get_speaker_embedding(speakers[1]);
+            ov::Tensor blended(emb1.get_element_type(), emb1.get_shape());
+            const float* data1 = emb1.data<float>();
+            const float* data2 = emb2.data<float>();
+            float* blended_data = blended.data<float>();
+            for (size_t i = 0; i < emb1.get_size(); ++i) {
+                blended_data[i] = 0.5f * data1[i] + 0.5f * data2[i];
+            }
+            talker_speech_config.speaker = blended;
+            std::cout << "Using blended voice for generation.\n\n";
+        }
+    }
+
+    ov::genai::ChatHistory history;
+    std::vector<ov::Tensor> videos;  // Empty: sample uses images and audio only
+
+    std::string prompt;
+    std::cout << "question:\n";
+    std::getline(std::cin, prompt);
+
+    history.push_back({{"role", "user"}, {"content", std::move(prompt)}});
+    ov::genai::OmniDecodedResults decoded_results = pipe.generate(history,
+                                                                   rgbs,
+                                                                   videos,
+                                                                   audios,
+                                                                   text_config,
+                                                                   talker_speech_config,
+                                                                   print_subword);
+    history.push_back({{"role", "assistant"}, {"content", std::move(decoded_results.texts[0])}});
+
+    if (!decoded_results.speech_result.speech_outputs.empty()) {
+        std::cout << "\n[Speech output: " << decoded_results.speech_result.speech_outputs[0].get_size() << " samples at 24kHz]"
+                  << std::endl;
+    }
+
+    std::cout << "\n----------\n"
+                 "question:\n";
+    while (std::getline(std::cin, prompt)) {
+        history.push_back({{"role", "user"}, {"content", std::move(prompt)}});
+        // New images and audio can be passed at each turn; here we rely on the info from turn 1.
+        std::vector<ov::Tensor> images, turn_audios;
+        decoded_results = pipe.generate(history,
+                                        images,
+                                        videos,
+                                        turn_audios,
+                                        text_config,
+                                        talker_speech_config,
+                                        print_subword);
+        history.push_back({{"role", "assistant"}, {"content", std::move(decoded_results.texts[0])}});
+
+        if (!decoded_results.speech_result.speech_outputs.empty()) {
+            std::cout << "\n[Speech output: " << decoded_results.speech_result.speech_outputs[0].get_size() << " samples at 24kHz]"
+                      << std::endl;
+        }
+        std::cout << "\n----------\n"
+                     "question:\n";
+    }
+    return 0;
+} catch (const std::exception& error) {
+    try {
+        std::cerr << error.what() << '\n';
+    } catch (const std::ios_base::failure&) {
+    }
+    return EXIT_FAILURE;
+} catch (...) {
+    try {
+        std::cerr << "Non-exception object thrown\n";
+    } catch (const std::ios_base::failure&) {
+    }
+    return EXIT_FAILURE;
+}
--- a/samples/python/omni/qwen3_omni_chat.py
+++ b/samples/python/omni/qwen3_omni_chat.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+# Copyright (C) 2023-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Qwen3-Omni multimodal chat sample.
+
+Demonstrates text + image + audio -> text + speech output using the ChatHistory API.
+
+Usage:
+    python qwen3_omni_chat.py <MODEL_DIR> <IMAGE_FILE_OR_DIR> [--audio AUDIO_WAV]
+"""
+
+import argparse
+from pathlib import Path
+
+import librosa
+import numpy as np
+import openvino_genai
+import soundfile as sf
+from openvino import Tensor
+from PIL import Image
+
+
+def streamer(subword: str) -> openvino_genai.StreamingStatus:
+    """Stream text tokens to stdout."""
+    print(subword, end="", flush=True)
+    return openvino_genai.StreamingStatus.RUNNING
+
+
+def read_image(path: str) -> Tensor:
+    pic = Image.open(path).convert("RGB")
+    image_data = np.array(pic)
+    return Tensor(image_data)
+
+
+def read_images(path: str) -> list[Tensor]:
+    entry = Path(path)
+    if entry.is_dir():
+        return [read_image(str(file)) for file in sorted(entry.iterdir())]
+    return [read_image(path)]
+
+
+def load_audio(audio_path: str, target_sr: int = 16000) -> Tensor:
+    """Load audio from WAV file and convert to float32 mono tensor at target_sr."""
+    audio_data, sample_rate = sf.read(audio_path, dtype="float32")
+
+    if audio_data.ndim > 1:
+        audio_data = audio_data.mean(axis=1)
+
+    if sample_rate != target_sr:
+        audio_data = librosa.resample(audio_data, orig_sr=sample_rate, target_sr=target_sr)
+
+    return Tensor(audio_data.astype(np.float32))
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Qwen3-Omni multimodal chat")
+    parser.add_argument("model_dir", help="Path to the OpenVINO model directory")
+    parser.add_argument("image_dir", help="Image file or directory with images")
+    parser.add_argument("--audio", help="Path to input audio WAV file (optional)")
+    args = parser.parse_args()
+
+    rgbs = read_images(args.image_dir)
+
+    pipe = openvino_genai.OmniPipeline(args.model_dir, "CPU")
+
+    # Two configs: text_config drives the thinker text decode; talker_speech_config drives the
+    # talker + speech output. Speech output is hardcoded on here to show the multimodal path.
+    # Set talker_speech_config.return_audio = False to get text-only responses.
+    text_config = openvino_genai.GenerationConfig()
+    text_config.max_new_tokens = 256
+
+    talker_speech_config = openvino_genai.OmniTalkerSpeechConfig(args.model_dir)
+    talker_speech_config.return_audio = True
+    # Leaving speaker empty selects the model's default voice. Available voices vary by checkpoint
+    # (e.g. MoE exposes "Ethan", "Chelsie", "Aiden", "Cherry"); the full list is in
+    # talker_config.speaker_id of the model's config.json.
+
+    videos = []
+    audios = [load_audio(args.audio)] if args.audio else []
+
+    history = openvino_genai.ChatHistory()
+    prompt = input("question:\n")
+    history.append({"role": "user", "content": prompt})
+    decoded_results = pipe.generate(
+        history,
+        images=rgbs,
+        videos=videos,
+        audios=audios,
+        text_config=text_config,
+        talker_speech_config=talker_speech_config,
+        streamer=streamer,
+    )
+    history.append({"role": "assistant", "content": decoded_results.texts[0]})
+
+    if decoded_results.speech_result.speech_outputs:
+        print(f"\n[Speech output: {decoded_results.speech_result.speech_outputs[0].get_size()} samples at 24kHz]")
+
+    while True:
+        try:
+            prompt = input("\n----------\nquestion:\n")
+        except EOFError:
+            break
+
+        history.append({"role": "user", "content": prompt})
+        # New images can be passed at each turn; here we rely on the info from turn 1.
+        images = []
+        decoded_results = pipe.generate(
+            history,
+            images=images,
+            videos=videos,
+            audios=audios,
+            text_config=text_config,
+            talker_speech_config=talker_speech_config,
+            streamer=streamer,
+        )
+        history.append({"role": "assistant", "content": decoded_results.texts[0]})
+
+        if decoded_results.speech_result.speech_outputs:
+            print(f"\n[Speech output: {decoded_results.speech_result.speech_outputs[0].get_size()} samples at 24kHz]")
+
+
+if __name__ == "__main__":
+    main()
--- a/src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
+++ b/src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
@@ -384,4 +384,10 @@ OPENVINO_GENAI_EXPORTS std::pair<std::string, ov::Any> videos_metadata_batches(
     const std::vector<std::vector<VideoMetadata>>& videos_metadata_batches
 );
 
+/// @brief Factory for audios_batches AnyMap entry.
+/// Audios must be encoded before text tokenization for Qwen3-Omni's interleaved layout.
+OPENVINO_GENAI_EXPORTS std::pair<std::string, ov::Any> audios_batches(
+    const std::vector<std::vector<ov::Tensor>>& audios_batches
+);
+
 }
--- a/src/cpp/include/openvino/genai/generation_handle.hpp
+++ b/src/cpp/include/openvino/genai/generation_handle.hpp
@@ -50,6 +50,14 @@ struct EncodedGenerationResult {
     // To get metrics, it should be cast to corresponding class for extended perf metrics from pipeline
     // Cast to SDPerModelsPerfMetrics for SpeculativeDecoding
     std::shared_ptr<ExtendedPerfMetrics> extended_perf_metrics;
+
+    // Accumulated intermediate hidden states per sequence.
+    // Outer vector: per return sequence. Inner vector: one tensor per generation step.
+    std::vector<std::vector<ov::Tensor>> m_intermediate_hidden_states;
+
+    // Full prompt + generated token IDs (needed for talker input construction).
+    // Outer vector: per return sequence, aligned with m_intermediate_hidden_states.
+    std::vector<std::vector<int64_t>> m_full_token_ids;
 };
 
 struct GenerationResult {
@@ -82,6 +90,7 @@ struct GenerationOutput {
     std::vector<float> generated_log_probs;
     float score = 0;
     GenerationFinishReason finish_reason = GenerationFinishReason::NONE;
+    std::vector<ov::Tensor> intermediate_hidden_states;
 };
 
 using GenerationOutputs = std::unordered_map<uint64_t, GenerationOutput>;
--- a/src/cpp/include/openvino/genai/omni/decoded_results.hpp
+++ b/src/cpp/include/openvino/genai/omni/decoded_results.hpp
@@ -1,0 +1,23 @@
+// Copyright (C) 2023-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "openvino/genai/omni/talker.hpp"
+#include "openvino/genai/visual_language/pipeline.hpp"
+
+namespace ov::genai {
+
+/**
+ * @brief Omni-specific decoded results including speech outputs.
+ *
+ * Extends VLMDecodedResults with a TalkerResults that holds speech waveforms and perf metrics.
+ */
+class OPENVINO_GENAI_EXPORTS OmniDecodedResults : public VLMDecodedResults {
+public:
+    /// Talker-side result: speech waveforms + perf metrics.
+    /// `speech_result.speech_outputs` is empty when speech generation was not requested.
+    TalkerResults speech_result;
+};
+
+}  // namespace ov::genai
--- a/src/cpp/include/openvino/genai/omni/pipeline.hpp
+++ b/src/cpp/include/openvino/genai/omni/pipeline.hpp
@@ -1,0 +1,199 @@
+// Copyright (C) 2024-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <filesystem>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "openvino/core/any.hpp"  // util::EnableIfAllStringAny
+#include "openvino/genai/chat_history.hpp"
+#include "openvino/genai/generation_config.hpp"
+#include "openvino/genai/llm_pipeline.hpp"  // StreamerVariant
+#include "openvino/genai/omni/decoded_results.hpp"
+#include "openvino/genai/omni/speech_streamer_base.hpp"
+#include "openvino/genai/omni/talker.hpp"
+#include "openvino/genai/omni/talker_speech_config.hpp"
+#include "openvino/genai/streamer_base.hpp"
+#include "openvino/genai/visibility.hpp"
+#include "openvino/genai/visual_language/pipeline.hpp"
+#include "openvino/runtime/tensor.hpp"
+
+namespace ov::genai {
+
+/**
+ * @brief Public Qwen3-Omni pipeline supporting text + speech outputs.
+ *
+ * OmniPipeline composes a VLM pipeline (text generation with hidden-state collection)
+ * with a Qwen3-Omni speech pipeline (Talker + CodePredictor + Code2Wav). Each call to
+ * `generate` takes two configs:
+ *
+ *   - `GenerationConfig text_config`     — drives the thinker text decode
+ *     (max_new_tokens, do_sample, temperature, top_p, top_k, repetition_penalty,
+ *      num_beams, eos_token_id, ...).
+ *   - `OmniTalkerSpeechConfig talker_speech_config` — drives the talker + speech output
+ *     (return_audio, speaker (name or embedding tensor), audio_chunk_frames, max_new_tokens,
+ *      rng_seed, talker_x / cp_x sampling overrides). Speech is gated per-call by
+ *      `talker_speech_config.return_audio`.
+ *
+ * Cross-config validation (e.g. return_audio incompatible with beam search) runs inside
+ * generate() before any work begins.
+ *
+ * Constructors:
+ *   - Path-based: loads VLM and speech models from a single models_path.
+ *   - Pure-DI: takes pre-built VLM and talker components for independent device choices
+ *              or a custom TalkerBase subclass.
+ *
+ * Both ctors enforce that the loaded model is Qwen3-Omni capable (model_type == QWEN3_OMNI
+ * and enable_audio_output) — non-Omni models throw at construction time.
+ */
+class OPENVINO_GENAI_EXPORTS OmniPipeline {
+public:
+    /// @brief Sugar ctor: loads the VLM and the default Qwen3-Omni talker from a single
+    /// `models_path`, both compiled to `device` with the same `properties`. Equivalent to
+    /// constructing a `VLMPipeline::create_base(models_path, device, properties)` plus a
+    /// `Talker(models_path, device, properties)` and passing them to the DI ctor.
+    OmniPipeline(const std::filesystem::path& models_path,
+                 const std::string& device,
+                 const ov::AnyMap& properties = {});
+
+    /// @brief Pure-DI ctor: takes pre-built VLM and talker components. Use when you want
+    /// independent device/property choices for the two stages (e.g. VLM on GPU, talker on
+    /// CPU), or to inject a custom `TalkerBase` subclass.
+    /// @param vlm Backing VLM pipeline base. Shared ownership allows independent lifetime management.
+    /// @param talker Backing speech generator. The default impl is `Talker`.
+    OmniPipeline(const std::shared_ptr<VLMPipeline::VLMBackend>& vlm, const std::shared_ptr<TalkerBase>& talker);
+
+    ~OmniPipeline();
+
+    /// @brief Generate text + (optionally) speech from a flat prompt.
+    /// @param prompt The user prompt.
+    /// @param images Image tensors to be prepended to the prompt.
+    /// @param videos Video tensors to be prepended to the prompt.
+    /// @param audios Audio tensors to be prepended to the prompt.
+    /// @param text_config Thinker text-decode config.
+    /// @param talker_speech_config Talker + speech-output config.
+    /// @param streamer Optional streamer for text tokens.
+    /// @param speech_streamer Optional streamer for audio chunks.
+    /// @return OmniDecodedResults with `speech_result.speech_outputs` populated when
+    ///         `talker_speech_config.return_audio` is true.
+    OmniDecodedResults generate(const std::string& prompt,
+                                const std::vector<ov::Tensor>& images,
+                                const std::vector<ov::Tensor>& videos,
+                                const std::vector<ov::Tensor>& audios,
+                                const GenerationConfig& text_config,
+                                const OmniTalkerSpeechConfig& talker_speech_config,
+                                const StreamerVariant& streamer = std::monostate{},
+                                const OmniSpeechStreamerVariant& speech_streamer = std::monostate{});
+
+    /// @brief Images-only flat-prompt overload (no videos / no audios).
+    OmniDecodedResults generate(const std::string& prompt,
+                                const std::vector<ov::Tensor>& images,
+                                const GenerationConfig& text_config,
+                                const OmniTalkerSpeechConfig& talker_speech_config,
+                                const StreamerVariant& streamer = std::monostate{},
+                                const OmniSpeechStreamerVariant& speech_streamer = std::monostate{});
+
+    /// @brief Images + videos flat-prompt overload (no audios).
+    OmniDecodedResults generate(const std::string& prompt,
+                                const std::vector<ov::Tensor>& images,
+                                const std::vector<ov::Tensor>& videos,
+                                const GenerationConfig& text_config,
+                                const OmniTalkerSpeechConfig& talker_speech_config,
+                                const StreamerVariant& streamer = std::monostate{},
+                                const OmniSpeechStreamerVariant& speech_streamer = std::monostate{});
+
+    /// @brief Single-image flat-prompt overload.
+    OmniDecodedResults generate(const std::string& prompt,
+                                const ov::Tensor& image,
+                                const GenerationConfig& text_config,
+                                const OmniTalkerSpeechConfig& talker_speech_config,
+                                const StreamerVariant& streamer = std::monostate{},
+                                const OmniSpeechStreamerVariant& speech_streamer = std::monostate{});
+
+    /// @brief Property-bag overload for flat-prompt generation. Recognized keys:
+    ///        `images`, `videos`, `audios`, `text_config`, `talker_speech_config`, `streamer`,
+    ///        `speech_streamer`, plus any field of `GenerationConfig` or `OmniTalkerSpeechConfig`.
+    /// @note Leftover keys are forwarded to BOTH configs' `update_generation_config`. Shared
+    ///       field names (e.g. `max_new_tokens`, `rng_seed`) end up on both — this is the
+    ///       intended broadcast behavior. Pass typed configs explicitly to set different values.
+    OmniDecodedResults generate(const std::string& prompt, const ov::AnyMap& config_map);
+
+    /// @brief Variadic-properties flat-prompt overload.
+    /// Example:
+    ///   `pipe.generate("text", ov::genai::images(rgbs), ov::genai::audios(audios),
+    ///                   ov::genai::text_config(g), ov::genai::talker_speech_config(t));`
+    template <typename... Properties>
+    util::EnableIfAllStringAny<OmniDecodedResults, Properties...> generate(const std::string& prompt,
+                                                                           Properties&&... properties) {
+        return generate(prompt, AnyMap{std::forward<Properties>(properties)...});
+    }
+
+    /// @brief Generate text + (optionally) speech from a chat history.
+    OmniDecodedResults generate(const ChatHistory& history,
+                                const std::vector<ov::Tensor>& images,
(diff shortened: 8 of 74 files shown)


src/cpp/src/whisper/feature_extractor.hpp
```diff
@@ -57,11 +59,10 @@ class WhisperFeatureExtractor {
     WhisperFeatures extract(const std::vector<float>& raw_speech);
 
 private:
-    std::vector<float> sin_vals;
-    std::vector<float> cos_vals;
-    std::vector<float> mel_filter;
+    // Deferred construction: config values are loaded from JSON inside the constructor body,
+    // so the extractor cannot be built in the member initializer list.
+    std::optional<audio_utils::MelSpectrogramExtractor> m_extractor;
```
as-suvorov: Please reuse `WhisperFeatureExtractor` instead of re-implementing it

as-suvorov (changes requested):

src/cpp/src/whisper/feature_extractor.hpp
sgonorov: I've just extracted common logic and made it separate. In my opinion it's a better approach than have multiple constructors with different behaviours and also different outputs. Or outputs which will require some post-processing. So it's not reimplementing, it's just a bit of refactorin so Whisper and Qwen3Omni could use common stuff but keep their unique behaviours by themselves, without mixing.

as-suvorov: > I've just extracted common logic and made it separate. In my opinion it's a better approach than have multiple constructors with different behaviours and also different outputs.

You introduced second constructor `audio_utils::MelSpectrogramExtractor`.

> So it's not reimplementing, it's just a bit of refactorin

Implementation is different now, so it's re-implementation from my pov.

> their unique behaviours by themselves, without mixing.

From my understanding Qwen3Omni uses WhisperFeatureExtractor: https://huggingface.co/Qwen/Qwen3-Omni-30B-A3B-Instruct/blob/main/preprocessor_config.json#L3
`"feature_extractor_type": "WhisperFeatureExtractor",`.
I would prefer to keep single ctor which is `WhisperFeatureExtractor`. Could you please explain what unique behavior feature extractor for Qwen3Omni has comparing to Whisper?

sgonorov: Okay, moved everything back to WhisperFeatureExtractor 👍

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -13,15 +13,15 @@
 namespace ov::genai {
 
 enum class GenerationStatus {
-    RUNNING = 0, // Default status for ongoing generation
-    FINISHED = 1, // Status set when generation has been finished
```
Wovchena, line 17: Knowing that something is default is useful. Keep it. I agree that
> // Status of generation

is pointless but removing it increases the risk of introducing merge conflicts and makes it harder to review. Remove the comments in a separate PR

src/cpp/src/whisper/feature_extractor.hpp
as-suvorov: Not moved back

src/cpp/src/whisper/feature_extractor.cpp
```diff
@@ -1,425 +1,17 @@
...
 #include <openvino/core/except.hpp>
-#include <openvino/openvino.hpp>
-#include <string>
 #include <thread>
-#include <vector>
 
 #include "json_utils.hpp"
 #include "openvino/genai/visibility.hpp"
-
-namespace {
```
as-suvorov, line 25: Please keep existing feature extraction implementation. Update in place if changes required,

as-suvorov (changes requested):

src/cpp/include/openvino/genai/omni/pipeline.hpp
Wovchena, line 1: Why did you remove `omni` folder

src/cpp/include/openvino/genai/omni_pipeline.hpp
```diff
@@ -0,0 +1,106 @@
...
+        "Please, use generate() with ChatHistory argument.")
+    void start_chat(const std::string& system_message = "");
+
+    /// @brief Deactivate chat mode on the underlying VLM pipeline.
+    /// @deprecated start_chat() / finish_chat() API is deprecated and will be removed in the next major release.
+    /// Please, use generate() with ChatHistory argument.
+    OPENVINO_DEPRECATED(
+        "start_chat() / finish_chat() API is deprecated and will be removed in the next major release. "
+        "Please, use generate() with ChatHistory argument.")
+    void finish_chat();
```
Wovchena:
```suggestion
```
Don't introduce deprecated methods

src/cpp/src/omni/pipeline.cpp
```diff
@@ -0,0 +1,53 @@
...
+#include "omni/pipeline_impl.hpp"
+
+namespace ov::genai {
+
+OmniPipeline::OmniPipeline(const std::filesystem::path& models_path,
+                           const std::string& device,
+                           const ov::AnyMap& properties)
+    : m_pimpl(std::make_unique<OmniPipelineImpl>(models_path, device, properties)) {}
+
+OmniPipeline::OmniPipeline(const std::shared_ptr<VLMPipeline::VLMPipelineBase>& vlm,
```
Wovchena: Why not inherit `VLMPipeline` from `VLMPipelineBase`?

src/cpp/include/openvino/genai/visual_language/pipeline_base.hpp
```diff
@@ -0,0 +1,166 @@
...
+#include "openvino/genai/visibility.hpp"
+#include "openvino/genai/visual_language/pipeline.hpp"
+#include "openvino/runtime/tensor.hpp"
+
+namespace ov {
+namespace genai {
+
+// Forward-declared so the public header doesn't drag in the internal vlm_config.hpp.
+// The underlying type is fixed to `int` to keep the forward declaration valid.
+enum class VLMModelType : int;
```
Wovchena: If `VLMModelType` is defined in internal header, how will people use it?

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,91 @@
...
+    const std::filesystem::path models_path = argv[1];
+
+    // Speech output is hardcoded here to show the multimodal path.
+    // Set return_audio = false to get text-only responses.
+    ov::genai::OmniSpeechGenerationConfig speech_config(models_path);
+    speech_config.max_new_tokens = 256;
+    speech_config.return_audio = true;
+    // Leaving speaker empty selects the model's default voice. Available voices vary by checkpoint
+    // (e.g. MoE exposes "Ethan", "Chelsie", "Aiden", "Cherry"); the full list is in
+    // talker_config.speaker_id of the model's config.json.
```
Wovchena: Is speaker represented with embedding internally? In that case we should expose it as a parameter to allow mixing different speakers as one

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/pipeline.hpp
sgonorov: At first i thought it's better for consistency like llm, whisper tts pipelines, but yeah, I'll restore it back.

src/cpp/include/openvino/genai/omni_pipeline.hpp
sgonorov: Removed

samples/cpp/omni/omni_chat.cpp
sgonorov: Yes, we can do it, but not sure it should go in this PR.

src/cpp/include/openvino/genai/visual_language/pipeline_base.hpp
sgonorov: Yes, removed it for now.

src/cpp/src/omni/pipeline.cpp
sgonorov: If VLMPipeline inherited VLMPipelineBase, omni-specific functions would be public on every VLM user, which I wanted to avoid.

src/cpp/include/openvino/genai/generation_handle.hpp
sgonorov: Fixed it

src/cpp/src/whisper/feature_extractor.cpp
sgonorov: Reverted 👍

src/cpp/src/whisper/feature_extractor.hpp
sgonorov: Now moved fully

samples/cpp/omni/omni_chat.cpp
Wovchena, line 1: Create a separate omni folder and add README

samples/cpp/visual_language_chat/qwen3_omni_chat.cpp
```diff
@@ -0,0 +1,97 @@
...
+#include "openvino/genai/omni/talker_speech_config.hpp"
+
+ov::genai::StreamingStatus print_subword(std::string&& subword) {
+    std::cout << subword << std::flush;
+    return ov::genai::StreamingStatus::RUNNING;
+}
+
+int main(int argc, char* argv[]) try {
+    if (argc != 3) {
+        throw std::runtime_error(std::string{"Usage "} + argv[0] + " <MODEL_DIR> <IMAGE_FILE_OR_DIR>");
```
Wovchena: Add audio input

samples/cpp/visual_language_chat/qwen3_omni_chat.cpp
```diff
@@ -0,0 +1,97 @@
...
+    ov::genai::ChatHistory history;
+
+    std::string prompt;
+    std::cout << "question:\n";
+    std::getline(std::cin, prompt);
+
+    history.push_back({{"role", "user"}, {"content", std::move(prompt)}});
+    ov::genai::OmniDecodedResults decoded_results = pipe.generate(history,
+                                                                   rgbs,
+                                                                   /*videos=*/{},
```
Wovchena: Create an empty var so the comments aren't needed

src/cpp/src/omni/pipeline.cpp
```diff
@@ -0,0 +1,275 @@
...
+            out.leftover.emplace(key, value);
+        }
+    }
+    return out;
+}
+
+}  // namespace
+
+OmniDecodedResults OmniPipeline::generate(const std::string& prompt, const ov::AnyMap& config_map) {
+    auto p = extract_omni_props(config_map);
```
Wovchena: Avoid single letter var names. It's hard to search such name ocurances

src/cpp/src/omni/pipeline_impl.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+}  // namespace
+
+OmniPipeline::OmniPipelineImpl::OmniPipelineImpl(const std::shared_ptr<VLMPipeline::VLMPipelineBase>& vlm,
+                                                  const std::shared_ptr<TalkerBase>& talker) {
+    OPENVINO_ASSERT(vlm != nullptr, "OmniPipeline: VLM pointer is null");
+    OPENVINO_ASSERT(talker != nullptr, "OmniPipeline: talker pointer is null");
+    m_vlm = vlm;
+    m_talker = talker;
+    assert_omni_capable();
```
Wovchena:
```suggestion
                                                  const std::shared_ptr<TalkerBase>& talker) :
        m_vlm{vlm}, m_talker{talker} {
    OPENVINO_ASSERT(m_vlm, "OmniPipeline: VLM pointer is null");
    OPENVINO_ASSERT(m_talker, "OmniPipeline: talker pointer is null");
```

Wovchena (changes requested):

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
```diff
@@ -0,0 +1,1061 @@
...
+    auto build_result = [&](ov::Tensor waveform) -> TalkerResult {
+        TalkerResult result;
+        if (waveform && waveform.get_size() > 0) {
+            result.perf_metrics.num_generated_samples = waveform.get_size();
+            result.speech_outputs.push_back(std::move(waveform));
+        }
+        const auto duration =
+            ov::genai::PerfMetrics::get_microsec(std::chrono::steady_clock::now() - speech_start_time);
+        result.perf_metrics.raw_metrics.generate_durations.emplace_back(duration);
+        result.perf_metrics.evaluate_statistics(speech_start_time);
```
Wovchena: Only `speech_outputs` and `raw_metrics` are filled. Should a separate type be introduced instead of reusing `SpeechGenerationPerfMetrics`?

src/cpp/src/omni/talker.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "visual_language/vlm_config.hpp"
+
+namespace ov::genai {
+
+namespace {
+
+/// @brief Flatten per-step hidden states into per-token hidden states.
+/// Each step tensor has shape [num_tokens, 1, hidden_size]. Multi-token tensors
+/// (from prefill) are split into individual [1, 1, hidden_size] slices.
+std::vector<ov::Tensor> flatten_hidden_states(const std::vector<ov::Tensor>& per_step_hs) {
```
Wovchena:
```suggestion
std::vector<ov::Tensor> split_hidden_states(const std::vector<ov::Tensor>& per_step_hs) {
```

src/cpp/src/omni/talker.cpp
```diff
@@ -0,0 +1,118 @@
...
+        const auto shape = step_hs.get_shape();
+        const size_t num_tokens = shape[0];
+        const size_t hidden_size = shape.back();
+        const size_t elem_size = step_hs.get_element_type().size();
+        const auto* src = static_cast<const uint8_t*>(step_hs.data());
+        const size_t token_bytes = hidden_size * elem_size;
+        const size_t stride = (shape.size() == 3) ? shape[1] * hidden_size * elem_size : token_bytes;
+        for (size_t t = 0; t < num_tokens; t++) {
+            ov::Tensor token_hs(step_hs.get_element_type(), {1, 1, hidden_size});
+            std::memcpy(token_hs.data(), src + t * stride, token_bytes);
```
Wovchena: `ov::Tensor` has `Coordinate` based constructor which creates a view instead of copy. Use it

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * Public-API thin wrapper around the internal `Qwen3OmniSpeechPipeline`. Construct
+ * directly when you want to inject your own VLMPipelineBase but keep the default speech
+ * backend:
+ *
+ *     auto vlm = ov::genai::VLMPipeline::create_base(model_dir, "GPU", props);
+ *     auto talker = std::make_shared<ov::genai::Qwen3OmniTalker>(model_dir, "CPU", props);
+ *     ov::genai::OmniPipeline pipe(vlm, talker);
+ */
+class OPENVINO_GENAI_EXPORTS Qwen3OmniTalker : public TalkerBase {
```
Wovchena:
```suggestion
class OPENVINO_GENAI_EXPORTS Talker : public TalkerBase {
```

Other pipelines don't provide model specific definitions neither should new API

Wovchena (changes requested):

samples/cpp/omni/omni_chat.cpp
Wovchena: We discussed so it's just a reminder. API should be designed now, the implementation can be in another PR

samples/python/visual_language_chat/audio_to_text.py
Wovchena, line 1: Fuse this sample with image based one

samples/cpp/visual_language_chat/qwen3_omni_chat.cpp
```diff
@@ -0,0 +1,97 @@
...
+    if (!decoded_results.speech_outputs.empty()) {
+        std::cout << "\n[Speech output: " << decoded_results.speech_outputs[0].get_size() << " samples at 24kHz]"
+                  << std::endl;
+    }
+
+    std::cout << "\n----------\n"
+                 "question:\n";
+    while (std::getline(std::cin, prompt)) {
+        history.push_back({{"role", "user"}, {"content", std::move(prompt)}});
+        // New images can be passed at each turn; here we reuse the initial one only on turn 1.
```
Wovchena:
```suggestion
        // New images can be passed at each turn; here we rely on the info from turn 1.
```
Same for Python

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,236 @@
...
+namespace ov::genai {
+
+/**
+ * @brief Public Qwen3-Omni pipeline supporting text + speech output.
+ *
+ * OmniPipeline composes a VLM pipeline (text generation with hidden-state collection)
+ * with a Qwen3-Omni speech pipeline (Talker + CodePredictor + Code2Wav). Each call to
+ * `generate` takes two configs:
+ *
+ *   - `GenerationConfig text_config`     — drives the thinker text decode
```
Wovchena, line 33:
```suggestion
 *   - `GenerationConfig text_config` — drives the thinker text decode
```

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,236 @@
...
+#include "openvino/genai/streamer_base.hpp"
+#include "openvino/genai/visibility.hpp"
+#include "openvino/genai/visual_language/pipeline.hpp"
+#include "openvino/genai/visual_language/pipeline_base.hpp"
+#include "openvino/runtime/tensor.hpp"
+
+namespace ov::genai {
+
+/**
+ * @brief Public Qwen3-Omni pipeline supporting text + speech output.
```
Wovchena:
```suggestion
 * @brief Public Qwen3-Omni pipeline supporting text + speech outputs.
```

src/cpp/include/openvino/genai/omni/decoded_results.hpp
```diff
@@ -0,0 +1,28 @@
...
+class OPENVINO_GENAI_EXPORTS OmniDecodedResults : public VLMDecodedResults {
+public:
+    /// Speech output waveforms (one per generated result).
+    /// Empty if speech generation was not requested (return_audio=false).
+    std::vector<ov::Tensor> speech_outputs;
+
+    /// Speech-side perf metrics. Populated regardless of `return_audio`; when speech is
+    /// disabled, holds the trivially-fast no-op path's timing. Use alongside the inherited
+    /// `perf_metrics` (text-side) to break down end-to-end Omni latency.
+    SpeechGenerationPerfMetrics speech_perf_metrics;
```
Wovchena: Reuse `TalkerResult`

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,236 @@
...
+    ///            or build fresh via `VLMPipeline::create_base(...)`.
+    /// @param talker Backing speech generator. The default impl is `Qwen3OmniTalker`.
+    OmniPipeline(const std::shared_ptr<VLMPipeline::VLMPipelineBase>& vlm,
+                 const std::shared_ptr<TalkerBase>& talker);
+
+    ~OmniPipeline();
+    OmniPipeline(OmniPipeline&&) noexcept;
+    OmniPipeline& operator=(OmniPipeline&&) noexcept;
+    OmniPipeline(const OmniPipeline&) = delete;
+    OmniPipeline& operator=(const OmniPipeline&) = delete;
```
Wovchena:
```suggestion
```
What are they for? Same for Talker

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+    /// @brief Return the pipeline's default OmniTalkerSpeechConfig. The path-based ctor seeds it
+    /// from `<models_path>/config.json -> talker_config.speaker_id`; the DI ctor leaves it
+    /// default-constructed. Used as the default when generate() is called without an explicit
+    /// `talker_speech_config`.
+    OmniTalkerSpeechConfig get_talker_speech_config() const;
+
+    /// @brief Replace the pipeline's default OmniTalkerSpeechConfig. The new value is `validate()`-d
+    /// before being stored so a misconfigured config can never silently take effect on a later call.
+    void set_talker_speech_config(const OmniTalkerSpeechConfig& new_config);
```
Wovchena: This and above getters and setters should be moved to Talker so the only way to work with them is through `get_talker()`

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+#include <filesystem>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "openvino/core/any.hpp"
+#include "openvino/genai/chat_history.hpp"
+#include "openvino/genai/generation_config.hpp"
+#include "openvino/genai/llm_pipeline.hpp"  // util::EnableIfAllStringAny
```
Wovchena:
```suggestion
```
`EnableIfAllStringAny` is defined in OV. Probably `#include "openvino/core/any.hpp"`
Same for other places

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    template <typename... Properties>
+    util::EnableIfAllStringAny<void, Properties...> update_generation_config(Properties&&... properties) {
+        return update_generation_config(AnyMap{std::forward<Properties>(properties)...});
+    }
+
+    /// @brief Validate talker-only invariants. Cross-config rules (e.g. return_audio
+    /// vs beam search on the text_config) are checked by OmniPipelineImpl.
+    /// @throws Exception if config is invalid.
+    void validate() const;
```
Wovchena: Since it's an independent config, avoid introducing methods like it was doe for `VideoGenerationConfig`

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -0,0 +1,109 @@
...
+    /// @brief Speaker name. Empty selects the model's default speaker.
+    /// Available names are listed under `talker_config.speaker_id` in
+    /// `config.json`. Ignored when `speaker_embedding` is non-empty.
+    std::string speaker;
+
+    /// @brief Optional explicit talker speaker embedding (`[1, 1, talker_hidden_size]`, f32).
+    /// Overrides `speaker` when non-empty. Use `OmniPipeline::get_speaker_embedding`
+    /// to fetch precomputed tensors and blend them yourself (e.g. weighted sum) to
+    /// mix voices.
+    ov::Tensor speaker_embedding;
```
Wovchena: `speaker` and `speaker_embedding` should be combined by `variant`

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    /// @brief Number of codec frames accumulated before streaming each audio chunk.
+    /// Must be >= 1. Each frame is 80ms of audio at 24 kHz (1920 samples).
+    std::size_t audio_chunk_frames = 1;
+
+    /// @brief Cap on talker AR steps. Independent of `text_config.max_new_tokens`
+    /// (which caps the thinker text decode). The talker pipeline takes the min of
+    /// this value and the model's `talker_config.talker_max_new_tokens`.
```
Wovchena: Why min? A user must have an option to override the value from json

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -0,0 +1,109 @@
...
+    std::size_t max_new_tokens = std::numeric_limits<std::size_t>::max();
+
+    /// @brief RNG seed for deterministic talker + CodePredictor sampling.
```
Wovchena:
```suggestion
    /// @brief RNG seed for talker + CodePredictor sampling.
```

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -18,11 +18,87 @@ namespace ov::genai {
 class OPENVINO_GENAI_EXPORTS VLMDecodedResults : public DecodedResults{
 public:
     VLMPerfMetrics perf_metrics;
+
+    // Internal fields for the speech pipeline (not exposed to Python bindings).
```
Wovchena: The fields should be exposed to Python bindings to allow defining Talker with Python and consuming these values

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -284,25 +334,41 @@ class OPENVINO_GENAI_EXPORTS VLMPipeline {
...
+    std::shared_ptr<VLMPipelineBase> get_base() const {
+        return m_pimpl;
+    }
 
 private:
-    class VLMPipelineBase;
     class VLMPipelineImpl;
     class VLMContinuousBatchingAdapter;
-    std::unique_ptr<VLMPipelineBase> m_pimpl;
+    std::shared_ptr<VLMPipelineBase> m_pimpl;
```
Wovchena:
```suggestion

private:
class VLMPipelineImpl;
class VLMContinuousBatchingAdapter;
    std::shared_ptr<VLMPipelineImpl> m_pimpl;
```
`VLMPipeline` already implements `VLMPipelineBase`

src/cpp/src/visual_language/pipeline_base.hpp
```diff
@@ -0,0 +1,139 @@
...
+#include "openvino/genai/omni/speech_streamer_base.hpp"
+#include "openvino/genai/visibility.hpp"
+#include "openvino/genai/visual_language/pipeline.hpp"
+#include "openvino/runtime/tensor.hpp"
+
+namespace ov {
+namespace genai {
+
+/**
+ * @brief Internal Omni-aware base for `VLMPipeline` implementations.
```
Wovchena, line 17: > Internal 

So make it private

src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
```diff
@@ -348,6 +348,10 @@ class OPENVINO_GENAI_EXPORTS ContinuousBatchingPipeline {
         return generate(histories, AnyMap{std::forward<Properties>(properties)...});
     }
 
+    /// @brief Toggle hidden-states accumulation on the underlying CB pipeline. Used by
+    /// VLMPipeline's CB adapter to forward `OmniPipeline`'s hidden-states gate.
+    void set_collect_hidden_states(bool enabled);
```
Wovchena: Unconditionally return the states if the model outputs them

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -50,6 +50,14 @@ struct EncodedGenerationResult {
     // To get metrics, it should be cast to corresponding class for extended perf metrics from pipeline
     // Cast to SDPerModelsPerfMetrics for SpeculativeDecoding
     std::shared_ptr<ExtendedPerfMetrics> extended_perf_metrics;
+
+    // Accumulated hidden states per sequence (populated when return_audio is requested).
```
Wovchena:
```suggestion
    // Accumulated hidden states per sequence.
```

Wovchena (changes requested):

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
sgonorov: Yes, made a new type for this.

src/cpp/src/visual_language/pipeline_base.hpp
sgonorov: Moved it to private

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
sgonorov: Exposed

src/cpp/include/openvino/genai/omni/pipeline.hpp
sgonorov: Removed

samples/cpp/omni/audio_utils.cpp
Wovchena, line 1: `whisper_speech_recognition/audio_utils.cpp` already knows how to read wav. Reuse it

samples/cpp/omni/qwen3_omni_chat.cpp
```diff
@@ -17,11 +18,12 @@ ov::genai::StreamingStatus print_subword(std::string&& subword) {
 }
 
 int main(int argc, char* argv[]) try {
-    if (argc != 3) {
-        throw std::runtime_error(std::string{"Usage "} + argv[0] + " <MODEL_DIR> <IMAGE_FILE_OR_DIR>");
+    if (argc < 3 || argc > 4) {
+        throw std::runtime_error(std::string{"Usage "} + argv[0] + " <MODEL_DIR> <IMAGE_FILE_OR_DIR> [<AUDIO_FILE>]");
```
Wovchena: Make audio required to simplify sample implementation

samples/cpp/omni/omni_chat.cpp
Wovchena, line 1: A sample demonstrating how to compose Omni from VLM and Talker is needed. I suggest updating this sample to demonstrate composability and leave a comment that Omni can be created from a folder directly

samples/cpp/omni/README.md
```diff
@@ -0,0 +1,148 @@
...
+- Separate generation configs for text decoding and speech output
+- Interactive CLI interface
+
+## Build
+
+From the samples root:
+
+```bash
+cmake -DCMAKE_BUILD_TYPE=Release -S ./ -B ./build/
+cmake --build ./build/ --config Release -j
```
Wovchena: Don't introduce instructions which aren't present in other samples

samples/cpp/omni/README.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+**Parameters:**
+- `<MODEL_DIR>` — Path to exported Qwen3-Omni OpenVINO model directory
+- `<IMAGE_FILE_OR_DIR>` — Path to input image(s) for visual context
+
+**Example:**
+
+```bash
+./build/samples/cpp/omni/qwen3_omni_chat /models/qwen3-omni-ov ./cat.jpg
```
Wovchena: Describe how to obtain example image and audio to run the sample

samples/cpp/omni/README.md
```diff
@@ -0,0 +1,148 @@
...
+```
+
+**Interactive usage:**
+1. Images are loaded once at startup and available to all turns
```
Wovchena, line 37: And audio

samples/cpp/omni/README.md
```diff
@@ -0,0 +1,148 @@
...
+2. Type questions and press Enter
+3. Model responds with streaming text; speech output indicated when enabled
+4. Continue the conversation across multiple turns
+5. Press Ctrl+D to exit
+
+## Model Support
+
+Compatible with Qwen3-Omni models exported to OpenVINO format:
+- **Qwen3-Omni-4B-Instruct** (dense, 4B parameters)
```
Wovchena: Don't mention that model

samples/cpp/omni/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+**Custom blend ratios:**
+
+- 75% voice A + 25% voice B: `0.75 * data1[i] + 0.25 * data2[i]`
+- 25% voice A + 75% voice B: `0.25 * data1[i] + 0.75 * data2[i]`
+
+**Available speakers by model:**
+
+- **Dense (4B):** typically 1 default voice (no named speakers)
+- **MoE (30B-A3B):** `"Ethan"`, `"Chelsie"`, `"Aiden"`, `"Cherry"`
```
Wovchena: You already listed them above

samples/cpp/omni/speaker_api_demo.cpp
```diff
@@ -0,0 +1,142 @@
...
+            std::cout << "   Audio duration: " << std::fixed << std::setprecision(2) << duration_sec << " seconds\n";
+        }
+        std::cout << "\n";
+    } else {
+        std::cout << "3. Voice blending:\n";
+        std::cout << "   (Requires at least 2 speakers; model has only " << speakers.size() << ")\n\n";
+    }
+
+    std::cout << "=== Demo Complete ===\n";
+    return 0;
```
Wovchena: `EXIT_SUCCESS` to stay consistent with `EXIT_FAILURE` or don't `return` at all

samples/python/omni/qwen3_omni_chat.py
Wovchena, line 1: Move to `omni` dir

samples/cpp/omni/speaker_api_demo.cpp
Wovchena, line 1: Would it be cleaner if this sample merged with the main omni sample? Main sample could still print all speakers and blend first 2 speakers for generation

samples/cpp/omni/speaker_api_demo.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // 1. List available speakers
+    std::cout << "1. Listing available speakers:\n";
+    auto talker = pipe.get_talker();
+    std::vector<std::string> speakers = talker->list_speakers();
+
+    if (speakers.empty()) {
+        std::cout << "   No named speakers found in this model.\n";
+        std::cout << "   (Model may have a single default voice only)\n\n";
+        return 0;
+    }
```
Wovchena: People will figure out how to handle absence of speakers. Remove it to make the sample shorter.
Same for
> Show first 5 values as sample

section. It's already covered by blanding

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -55,25 +54,21 @@ class OPENVINO_GENAI_EXPORTS OmniPipeline {
...
     /// independent device/property choices for the two stages (e.g. VLM on GPU, talker on
     /// CPU), or to inject a custom `TalkerBase` subclass.
-    /// @param vlm Backing VLM. Get from `VLMPipeline::get_base()` on an existing pipeline,
-    ///            or build fresh via `VLMPipeline::create_base(...)`.
-    /// @param talker Backing speech generator. The default impl is `Qwen3OmniTalker`.
-    OmniPipeline(const std::shared_ptr<VLMPipeline::VLMPipelineBase>& vlm,
-                 const std::shared_ptr<TalkerBase>& talker);
+    /// @param vlm Backing VLM pipeline. The internal base is extracted and shared.
+    /// @param talker Backing speech generator. The default impl is `Talker`.
+    OmniPipeline(VLMPipeline& vlm, const std::shared_ptr<TalkerBase>& talker);
```
Wovchena:
```suggestion
    OmniPipeline(const shared_ptr<VLMPipelineBase>& vlm, const std::shared_ptr<TalkerBase>& talker);
```

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -55,25 +54,21 @@ class OPENVINO_GENAI_EXPORTS OmniPipeline {
...
 
     ~OmniPipeline();
     OmniPipeline(OmniPipeline&&) noexcept;
     OmniPipeline& operator=(OmniPipeline&&) noexcept;
```
Wovchena:
```suggestion
```
Why are they kept?

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,223 @@
...
+    void set_text_generation_config(const GenerationConfig& new_config);
+
+    /// @brief Return the pipeline's default OmniTalkerSpeechConfig. The path-based ctor seeds it
+    /// from `<models_path>/config.json -> talker_config.speaker_id`; the DI ctor leaves it
+    /// default-constructed. Used as the default when generate() is called without an explicit
+    /// `talker_speech_config`.
+    OmniTalkerSpeechConfig get_talker_speech_config() const;
+
+    /// @brief Replace the pipeline's default OmniTalkerSpeechConfig.
+    void set_talker_speech_config(const OmniTalkerSpeechConfig& new_config);
```
Wovchena: Remove seters and getters except for `get_vlm()` and `get_talker()`

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -0,0 +1,126 @@
...
+    /// @brief Return precomputed speaker embedding for the named speaker.
+    /// Tensor shape and dtype are backend-defined (Qwen3-Omni: `[1, 1, talker_hidden_size]`, f32).
+    /// Use to blend voices: combine two named embeddings and pass the result via
+    /// `OmniTalkerSpeechConfig::speaker` (the Tensor alternative of the variant).
+    /// @throws when the backend has no named speakers, or `name` doesn't match.
+    virtual ov::Tensor get_speaker_embedding(const std::string& name) const = 0;
+
+    /// @brief True when the talker is fully loaded and able to produce audio. False when
+    /// optional submodels are missing (e.g. the model directory ships VLM weights only).
+    virtual bool is_available() const = 0;
```
Wovchena: If submodels are missing, it should be an error. People should use `VLMPipleine` if their model has no speech generation

src/cpp/include/openvino/genai/omni/talker_perf_metrics.hpp
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Captures wall-clock time and output size for a single generate() call.
+ * Unlike SpeechGenerationPerfMetrics, this type omits LLM-oriented fields
+ * (TTFT, TPOT, tokenization durations) that are not applicable to speech codec AR.
+ */
+struct OPENVINO_GENAI_EXPORTS TalkerPerfMetrics {
+    /// Number of audio samples generated (waveform length in samples, not bytes).
+    size_t num_generated_samples = 0;
+
+    /// Total speech generation time in milliseconds (AR loop + codec decode + vocoding).
```
Wovchena: 1. What is AR?
2. Codec decode is what happens when a wav is red, isn't it?

Wovchena (changes requested):

samples/cpp/omni/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+    https://raw.githubusercontent.com/nothings/stb/f75e8d1cad7d90d72ef7a4661f1b994ef78b4e31/stb_image.h
+    ${CMAKE_BINARY_DIR}/stb_image.h
+    EXPECTED_HASH MD5=27932e6fb3a2f26aee2fc33f2cb4e696)
+
+file(DOWNLOAD
+    https://raw.githubusercontent.com/mackron/dr_libs/master/dr_wav.h
+    ${CMAKE_BINARY_DIR}/dr_wav.h
+    EXPECTED_HASH MD5=5eb20509b821ee41605e84210ed15fd5)
+
+add_executable(qwen3_omni_chat qwen3_omni_chat.cpp audio_utils.cpp ../visual_language_chat/load_image.cpp)
```
Wovchena:
```suggestion
add_executable(omni_chat qwen3_omni_chat.cpp audio_utils.cpp ../visual_language_chat/load_image.cpp)
```
Same for other occurances

src/cpp/include/openvino/genai/omni/talker_perf_metrics.hpp
sgonorov: Fixed, added some details

src/cpp/include/openvino/genai/omni/talker.hpp
sgonorov: Added assertion to the constuctor

src/cpp/include/openvino/genai/omni/pipeline.hpp
sgonorov: I thought making it movable is a nice thing. But can remove, sure.

src/cpp/include/openvino/genai/omni/pipeline.hpp
sgonorov: Fixed

samples/cpp/omni/speaker_api_demo.cpp
sgonorov: Merged in

samples/cpp/omni/README.md
sgonorov: Removed

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -0,0 +1,92 @@
...
+    /// Checkpoint default lives in `generation_config.json -> cp_temperature`.
+    std::optional<float> cp_temperature;
+
+    /// @brief CodePredictor top-k override (must be >= 1 when set).
+    /// Checkpoint default lives in `generation_config.json -> cp_top_k`.
+    std::optional<std::size_t> cp_top_k;
+
+    /// @brief CodePredictor repetition penalty override (must be > 0 when set; 1.0 = no penalty).
+    /// Checkpoint default lives in `generation_config.json -> cp_repetition_penalty`.
+    std::optional<float> cp_repetition_penalty;
```
Wovchena, line 94: I can only see `talker` prefixed entries in json: https://huggingface.co/Qwen/Qwen3-Omni-30B-A3B-Instruct/blob/main/generation_config.json. Where does `cp` come from?

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -284,25 +334,49 @@ class OPENVINO_GENAI_EXPORTS VLMPipeline {
...
+    GenerationConfig get_generation_config() const override;
 
     /// @brief Override default values for GenerationConfig
     /// @param new_config A config to override default values with.
-    void set_generation_config(const GenerationConfig& new_config);
+    void set_generation_config(const GenerationConfig& new_config) override;
+
+    /// @brief Get the model load time in milliseconds.
+    /// @return Load time in milliseconds.
+    float get_load_time_ms() const;
```
Wovchena: Why is that exposed? This metric is available as `generate()` result

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -284,25 +334,49 @@ class OPENVINO_GENAI_EXPORTS VLMPipeline {
...
+
+    /// @brief Get the attention backend used by the pipeline.
+    /// @return "PA" for Paged Attention (continuous batching), "SDPA" for scaled dot-product attention.
+    std::string get_attention_backend() const;
+
+    class VLMPipelineBase;  // internal Omni-aware base; defined in src/cpp/src/visual_language/pipeline_base.hpp.
```
Wovchena: Rename the class so it doesn't conflict with the parent of `VLMPipeline`

samples/cpp/omni/qwen3_omni_chat.cpp
sgonorov: Fixed

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
sgonorov: Yes, it's for the code predictor, it's optional, as it has hardcoded defaults. Code predictor has a separate set of parameters from talker.

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
sgonorov: Yes, leftovers, removed them

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
sgonorov: Fixed

src/cpp/src/visual_language/qwen3_omni/classes.cpp
```diff
@@ -0,0 +1,675 @@
...
+            pos_data[2 * seq_len + pos] = next_pos;
+            pos_data[3 * seq_len + pos] = next_pos;
+            next_pos++;
+        }
+    }
+
+    // Calculate rope delta from maximum position value
+    const int64_t position_ids_max_element = *std::max_element(
+        position_ids.data<int64_t>(),
+        position_ids.data<int64_t>() + position_ids.get_size()
```
byungilm: Hi, I am analyzing qwen3-omni model with the scenario of audio+image+text input and audio+text output with this PR for CVS-184584.  I faced EOS error of output. I think it is related to calculation of rope delta.
The error is fixed with the following fix.
Could you please review that this fix is reasonable?

Output text :
Before : I
After fix : I see four luxury cars from different brands, and I hear a person coughing.

source diff:
```
+    // position_ids has shape {dims, batch=1, seq_len} with rows [temporal, height, width, text].
+    // RoPE only consumes the spatial rows (config mrope_section = [24, 20, 20], i.e. 3 sections); the
+    // trailing "text" row carries large per-token sequential values inside the vision span that would
+    // pollute the global max and push the decode position far past the real sequence end. Excluding
+    // that last row mirrors HF, whose rope delta is computed from the 3D grid. Derive the spatial row
+    // count from the tensor shape so this stays correct if the row layout ever changes.
+    const size_t num_position_dims = position_ids.get_shape().at(0);
+    const size_t num_spatial_dims = num_position_dims - 1;  // drop the trailing "text" row
     const int64_t position_ids_max_element = *std::max_element(
         position_ids.data<int64_t>(),
-        position_ids.data<int64_t>() + position_ids.get_size()
+        position_ids.data<int64_t>() + num_spatial_dims * seq_len
     );
     const int64_t rope_delta = position_ids_max_element + 1 - static_cast<int64_t>(seq_len);
```

sgonorov: You're right - applied this fix

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -0,0 +1,124 @@
...
```
Wovchena: Reminder to remove `is_available()`

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/talker.hpp
sgonorov: Removed it

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
```diff
@@ -0,0 +1,1083 @@
...
+        if (!std::filesystem::exists(path)) {
+            return {};
+        }
+        auto model = utils::singleton_core().read_model(path);
+
+        // Force FP32 inference precision on GPU for talker models to match CPU behavior
+        // GPU FP16 causes numerical differences in logits → different sampled tokens → wrong speech length
+        ov::AnyMap compilation_props = properties;
+        if (device == "GPU" || device.find("GPU") == 0) {
+            compilation_props["INFERENCE_PRECISION_HINT"] = "f32";
```
byungilm, line 108: I have tested Qwen3-omnii-4B mode. And I found that this compilation of speech sub-models in f32  caused huge memory consumption. I checked that code_predictor sub-model had unrolled sampling graph and allocated all intermediate activation as fp32.
In my test, talker, code_predictor, and code2wav being only ~1,583 MB on disk, the actual device memory allocation balloons to ~8,854 MB (5.6×) after FP32 compilation, with code_predictor alone accounting for ~4.5 GiB of that
Could we have a planning on restructuring the graph into a single-step loop design?

src/cpp/include/openvino/genai/omni/pipeline.hpp
Wovchena: Reminder

src/cpp/include/openvino/genai/generation_handle.hpp
```diff
@@ -50,6 +50,13 @@ struct EncodedGenerationResult {
     // To get metrics, it should be cast to corresponding class for extended perf metrics from pipeline
     // Cast to SDPerModelsPerfMetrics for SpeculativeDecoding
     std::shared_ptr<ExtendedPerfMetrics> extended_perf_metrics;
+
+    // Accumulated intermediate hidden states per sequence.
+    // Outer vector: per return sequence. Inner vector: one tensor per generation step.
+    std::vector<std::vector<ov::Tensor>> m_intermediate_hidden_states;
+
+    // Full prompt token IDs (needed for talker input construction).
+    std::vector<int64_t> m_prompt_ids;
```
Wovchena: Extend `GenerationOutput` to enable `add_request()` API

Wovchena (changes requested):

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.hpp
```diff
@@ -0,0 +1,228 @@
...
+#include "openvino/runtime/tensor.hpp"
+#include "visual_language/vlm_config.hpp"
+
+namespace ov::genai {
+
+/// @brief Configuration for Qwen3-Omni speech generation.
+struct Qwen3OmniSpeechConfig {
+    int64_t codec_bos_id = 2149;
+    int64_t codec_eos_token_id = 2150;
+    int64_t codec_pad_id = 2148;
```
Wovchena: Set model neutral defaults and get the model defaults from json

Wovchena (changes requested):

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+    /// @brief Embed a single token via talker text embeddings.
+    /// @return Tensor [1, 1, talker_hidden_size].
+    ov::Tensor embed_talker_token(int64_t token_id);
+
+    /// @brief Get raw pointer to CodePredictor codec_embedding weights for a step/code.
+    /// @param step Step index 0..num_code_groups-2
+    /// @param code Token ID (0..cp_vocab_size-1)
+    /// @return Pointer to [cp_hidden_size] floats within m_cp_codec_embeddings.
+    const float* cp_token_weights(size_t step, int64_t code) const;
```
Wovchena:
```suggestion
```
Never called. Same for `m_cp_codec_embeddings`

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -18,11 +18,86 @@ namespace ov::genai {
...
+    // Hidden-state fields for speech generation (exposed to Python to enable custom Talker impls).
+    // Populated by the CB pipeline when `return_audio` is requested on the GenerationConfig.
+    // Consumed by Talker / TalkerBase impls to drive Qwen3-Omni speech generation.
+    // Both fields are empty (default-constructed) on the text-only path.
+    //
+    // Outer index of intermediate_hidden_states is per return sequence;
+    // inner is one tensor per generation step.
+    std::vector<std::vector<ov::Tensor>> intermediate_hidden_states;
+    // Full prompt + generated token ids for the first sequence (talker input construction).
+    std::vector<int64_t> prompt_ids;
```
Wovchena:
```suggestion
    std::vector<std::vector<int64_t>> full_token_ids;
```
Same for CB
Batch is missing

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/pipeline.hpp
```diff
@@ -0,0 +1,199 @@
...
+    /// @param speech_streamer Optional streamer for audio chunks.
+    /// @return OmniDecodedResults with `speech_result.speech_outputs` populated when
+    ///         `talker_speech_config.return_audio` is true.
+    OmniDecodedResults generate(const std::string& prompt,
+                                const std::vector<ov::Tensor>& images,
+                                const std::vector<ov::Tensor>& videos,
+                                const std::vector<ov::Tensor>& audios,
+                                const GenerationConfig& text_config,
+                                const OmniTalkerSpeechConfig& talker_speech_config,
+                                const StreamerVariant& text_streamer = std::monostate{},
```
Wovchena:
```suggestion
                                const StreamerVariant& streamer = std::monostate{},
```
Stay aligned with `streamer` property. Same for other occurances

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.hpp
sgonorov: Removed

Wovchena (changes requested):

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -284,25 +333,45 @@ class OPENVINO_GENAI_EXPORTS VLMPipeline {
...
+    /// `VLMPipeline::VLMBackend` (CB adapter or VLMPipelineImpl), which extends the
+    /// public namespace-level `ov::genai::VLMPipelineBase` and adds Omni-specific hooks.
+    static std::shared_ptr<VLMBackend> create_base(const std::filesystem::path& models_dir,
+                                                   const std::string& device,
+                                                   const ov::AnyMap& properties);
+
+    /// @brief Get the backing internal VLMBackend. Used by OmniPipeline's shared-VLM ctor
+    /// to reuse an already-loaded VLM without reloading multi-GB weights.
+    std::shared_ptr<VLMBackend> get_base() const {
+        return m_pimpl;
```
Wovchena: Remove base family methods from public API

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
```diff
@@ -284,25 +333,45 @@ class OPENVINO_GENAI_EXPORTS VLMPipeline {
...
+    GenerationConfig get_generation_config() const override;
 
     /// @brief Override default values for GenerationConfig
     /// @param new_config A config to override default values with.
-    void set_generation_config(const GenerationConfig& new_config);
+    void set_generation_config(const GenerationConfig& new_config) override;
+
+    /// @brief Get the attention backend used by the pipeline.
+    /// @return "PA" for Paged Attention (continuous batching), "SDPA" for scaled dot-product attention.
+    std::string get_attention_backend() const;
```
Wovchena: Why is that added?

src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
```diff
@@ -348,6 +348,10 @@ class OPENVINO_GENAI_EXPORTS ContinuousBatchingPipeline {
         return generate(histories, AnyMap{std::forward<Properties>(properties)...});
     }
 
+    /// @brief Toggle hidden-states accumulation on the underlying CB pipeline.
+    /// This flag controls whether they are stored in GenerationResult for caller retrieval.
+    void set_collect_hidden_states(bool enabled);
```
Wovchena: Use `GenerationConfig::return_omni_outputs` instead

Wovchena (changes requested):

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    std::vector<ov::Tensor> rgbs = utils::load_images(argv[2]);
+
+    ov::genai::RawSpeechInput pcm = utils::audio::read_wav(argv[3]);
+    ov::Tensor audio_tensor(ov::element::f32, {pcm.size()});
+    std::memcpy(audio_tensor.data<float>(), pcm.data(), pcm.size() * sizeof(float));
+    std::vector<ov::Tensor> audios = {std::move(audio_tensor)};
```
Wovchena, line 41: Add a new utils function which returns a Tensor to hide this conversion. By the way, there's a way to avoid the copy: https://github.com/openvinotoolkit/openvino.genai/blob/8981d6f848f17985979be0a9224251d181f68c56/samples/cpp/visual_language_chat/load_image.cpp#L39
move `pcm` returned from `read_wav()` into allocator defined in place

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,147 @@
...
+    std::cout << "[Loaded audio: " << audios[0].get_size() << " samples at 16kHz]\n";
```
Wovchena:
```suggestion
```

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+    // Compose OmniPipeline from separate VLM and Talker components.
+    // This allows independent device selection (e.g., VLM on GPU, Talker on CPU),
+    // sharing a VLM base across pipelines, or injecting custom TalkerBase implementations.
+    // For simpler use cases: ov::genai::OmniPipeline pipe(models_path, "CPU");
+    std::cout << "\n=== Building OmniPipeline from Components ===\n";
+    std::cout << "Creating VLM base and Talker separately...\n";
+
+    auto vlm_base = ov::genai::VLMPipeline::create_base(models_path, "CPU", {});
+    auto talker = std::shared_ptr<ov::genai::Talker>(new ov::genai::Talker(models_path, "CPU"));
```
Wovchena, line 49: > [Prefer constructing std::shared_ptr with std::make_shared here for exception safety and to avoid an extra allocation (control block + object). This is especially relevant in samples that users copy into production code.](https://github.com/openvinotoolkit/openvino.genai/pull/3868/changes#r3431671247)

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,147 @@
...
+    return ov::genai::StreamingStatus::RUNNING;
+}
+
+int main(int argc, char* argv[]) try {
+    if (argc != 4) {
+        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> <IMAGE_FILE_OR_DIR> <AUDIO_FILE>");
+    }
+
+    const std::filesystem::path models_path = argv[1];
+
```
Wovchena, line 26:
```suggestion
```
Pass `argv[1]` directly to constructors to simplify reading

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,147 @@
...
+    // Two configs: text_config drives the thinker text decode, talker_speech_config drives
+    // the talker + speech output. Speech output is hardcoded on here to show the multimodal
+    // path. Set talker_speech_config.return_audio = false to get text-only responses.
+    ov::genai::GenerationConfig text_config;
+    text_config.max_new_tokens = 256;
+
+    ov::genai::OmniTalkerSpeechConfig talker_speech_config(models_path);
+    talker_speech_config.return_audio = true;
+    // Leaving speaker empty selects the model's default voice. Available voices vary by checkpoint
```
Wovchena:
```suggestion
    // Leaving talker_speech_config.speaker empty selects the model's default voice. Available voices vary by checkpoint
```

Wovchena (changes requested):

samples/cpp/omni/omni_chat.cpp
```diff
@@ -0,0 +1,147 @@
...
+    std::vector<ov::Tensor> audios = {std::move(audio_tensor)};
+    std::cout << "[Loaded audio: " << audios[0].get_size() << " samples at 16kHz]\n";
+
+    // Compose OmniPipeline from separate VLM and Talker components.
+    // This allows independent device selection (e.g., VLM on GPU, Talker on CPU),
+    // sharing a VLM base across pipelines, or injecting custom TalkerBase implementations.
+    // For simpler use cases: ov::genai::OmniPipeline pipe(models_path, "CPU");
+    std::cout << "\n=== Building OmniPipeline from Components ===\n";
+    std::cout << "Creating VLM base and Talker separately...\n";
+
```
Wovchena:
```suggestion
```

Wovchena (changes requested):

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
sgonorov: This is fixed now. No more memory spike.

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
sgonorov: Fixed

src/cpp/include/openvino/genai/omni/pipeline.hpp
sgonorov: Fixed

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
sgonorov: Removed

src/cpp/include/openvino/genai/visual_language/pipeline.hpp
sgonorov: Leftovers, removed

src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
sgonorov: Fixed

samples/cpp/omni/omni_chat.cpp
sgonorov: Updated

samples/cpp/omni/omni_chat.cpp
sgonorov: Removed

samples/cpp/omni/omni_chat.cpp
sgonorov: I'm not sure it will simplify reading.

samples/cpp/omni/omni_chat.cpp
sgonorov: Fixed

samples/cpp/omni/omni_chat.cpp
sgonorov: Fixed

PR #3881, opened 2026-05-18
Add denoising_process C++ and Python samples

byekrang:
## Description

  Adds two new `denoising_process` samples under
  `samples/{cpp,python}/image_generation/` that visualize the
  text-to-image denoising trajectory as a video. Each sample captures
   the latent at every step from the `Text2ImagePipeline` generation
  callback, decodes it via `Text2ImagePipeline::decode(latent)`, and
  writes the per-step decoded frames as an MJPG AVI using the
  existing `save_video` helper. The last frame approximates the final
   generated image.

  ## Changes

  - **C++ sample**
  `samples/cpp/image_generation/denoising_process.cpp`: built from
  `text2image.cpp`; same hardcoded `num_inference_steps=20`,
  `width=512`, `height=512` as `text2image.cpp`; `rng_seed=42` for
  reproducibility; the existing `progress_bar` callback is retained
  for progress reporting; `pipe.decode(latent)` is invoked inside the
   callback to accumulate per-step decoded frames;
  `denoising_process.avi` is emitted via the existing `save_video`
  helper from `samples/cpp/video_generation/`.
  - **Python sample**
  `samples/python/image_generation/denoising_process.py`: mirrors the
   C++ sample's structure and parameters; reuses the existing
  `video_utils.save_video` helper from
  `samples/python/video_generation/` via a `sys.path` insert.
  - **Shared `save_video` helpers** 
  (`samples/cpp/video_generation/imwrite_video.cpp`, 
  `samples/python/video_generation/video_utils.py`): now pass 
  `cv::CAP_OPENCV_MJPEG` / `cv2.CAP_OPENCV_MJPEG` to 
  `cv::VideoWriter` explicitly, so the C++ and Python sides request 
  OpenCV's MJPEG backend explicitly rather than depending on
  whichever MJPEG backend OpenCV auto-selects per language and per
  build.
  - **README updates** in 
  `samples/{cpp,python}/image_generation/README.md` describing the 
  new sample and its output.
  - **New test** 
  `tests/python_tests/samples/test_denoising_process.py`: runs both 
  samples with `tiny-random-latent-consistency` at the sample's 
  hardcoded settings, decodes the produced `.avi` files via 
  `cv2.VideoCapture`, and asserts the decoded frame arrays are 
  exactly equal via `np.array_equal`.
  - **CI plumbing**: new `tiny_random_latent_consistency` marker in 
  `tests/python_tests/pytest.ini`; new sample-tests matrix row of the
   same name in `.github/workflows/linux.yml` and 
  `.github/workflows/windows.yml` gated on 
  `affected_components.Image_generation_samples.test`; new entry for 
  `test_denoising_process.py` under the Image generation samples
  category in `.github/labeler.yml`.

  ## Model choice for the test

  The test uses `tiny-random-latent-consistency`, following the 
  precedent set by `tests/python_tests/samples/test_taylorseer.py`, 
  which uses `tiny-random-flux` and `stable-diffusion-3-tiny-random` 
  for exact Python-vs-C++ image and video equality assertions. The 
  fixture is already declared in both 
  `tests/python_tests/conftest.py` and
  `tests/python_tests/samples/conftest.py`. The marker and CI matrix
  row mirror the existing `dreamlike_anime_1_0` and
  `LCM_Dreamshaper_v7_int8_ov` sample-test rows.

  ## Validation

  - [x] Sample built via the in-repo 
  `samples/cpp/image_generation/CMakeLists.txt`.
  - [x] Python sample standalone run emits `Step 1/20` … `Step 20/20`
   and writes `denoising_process.avi`.
  - [x] C++ sample standalone run advances `Generation step 1 / 20` …
   `20 / 20` via `indicators` and writes `denoising_process.avi`.
  - [x] Targeted pytest on `tiny-random-latent-consistency`: `1 
  passed in 204.33s (0:03:24)`, exact decoded-frame equality between 
  the Python- and C++-produced AVIs.
  - [x] Visual smoke on a real model (`LCM_Dreamshaper_v7-int8-ov`) at the finalized sample defaults (`num_inference_steps=20`, `512×512`, `rng_seed=42`): both Python and C++ samples completed; each emitted 20/20 callback/progress steps; each produced a 20-frame `(20, 512, 512, 3)` `uint8` AVI with a nonzero final frame and a plausible denoising trajectory; `np.array_equal(py_frames, cpp_frames) == True` on the decoded frames.
  - [x] `pre-commit run --from-ref origin/master --to-ref HEAD` and 
  `pre-commit run --files <all 11 changed files>` both pass.
  - [x] `git diff --check` clean.
  - [ ] Linux + Windows CI green on the new 
  `tiny_random_latent_consistency` matrix row.

  ## Notes

  - JS sample for `denoising_process` is intentionally out of scope 
  per the task spec ("two samples in total — C++ and Python"). 
  Existing `samples/js/image_generation/text2image.js` is unchanged.
  - macOS CI does not currently include an image-generation 
  sample-tests matrix row (`mac.yml` covers LLM, Whisper, Rag, Speech
   generation only). This PR preserves that convention.

  ## Checklist

  - [x] This PR follows [GenAI Contributing 
  guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=c
  ontributing-ov-file#contributing).
  - [x] This PR fully addresses the requested sample task.

### DIFF
--- a/samples/cpp/fetch_opencv.cmake
+++ b/samples/cpp/fetch_opencv.cmake
@@ -77,6 +77,36 @@ function(ov_genai_link_opencv target_name)
                 INSTALL_RPATH "@loader_path/../lib"
                 INSTALL_RPATH_USE_LINK_PATH ON)
         endif()
+
+        # Windows: deploy fetched OpenCV runtime DLLs alongside sample executables
+        # in samples_bin/. Windows has no rpath, so install/samples_bin/<exe> can
+        # only resolve its OpenCV dependencies via the loader search path. The
+        # function is invoked by multiple sample CMakeLists.txt files; guard with
+        # a GLOBAL property so the install rules are emitted exactly once.
+        if(WIN32)
+            get_property(_ov_genai_opencv_dlls_installed GLOBAL
+                PROPERTY OV_GENAI_OPENCV_RUNTIME_DLLS_INSTALLED)
+            if(NOT _ov_genai_opencv_dlls_installed)
+                foreach(component IN LISTS required_components)
+                    install(FILES "$<TARGET_FILE:opencv_${component}>"
+                            DESTINATION samples_bin/
+                            COMPONENT samples_bin
+                            EXCLUDE_FROM_ALL)
+                endforeach()
+                # FFmpeg backend plugin: prebuilt DLL OpenCV places next to
+                # opencv_videoio with no CMake target. TARGET_FILE_DIR keeps the
+                # path tied to the OpenCV build layout and config, with no
+                # version-suffix or absolute-path assumptions.
+                install(FILES
+                            "$<TARGET_FILE_DIR:opencv_videoio>/opencv_videoio_ffmpeg_64.dll"
+                        DESTINATION samples_bin/
+                        COMPONENT samples_bin
+                        EXCLUDE_FROM_ALL
+                        OPTIONAL)
+                set_property(GLOBAL PROPERTY
+                    OV_GENAI_OPENCV_RUNTIME_DLLS_INSTALLED TRUE)
+            endif()
+        endif()
     else()
         set(opencv_targets ${OpenCV_LIBS})
         if(OpenCV_INCLUDE_DIRS)
--- a/samples/cpp/fetch_opencv.cmake
+++ b/samples/cpp/fetch_opencv.cmake
@@ -78,35 +78,10 @@ function(ov_genai_link_opencv target_name)
                 INSTALL_RPATH_USE_LINK_PATH ON)
         endif()
 
-        # Windows: deploy fetched OpenCV runtime DLLs alongside sample executables
-        # in samples_bin/. Windows has no rpath, so install/samples_bin/<exe> can
-        # only resolve its OpenCV dependencies via the loader search path. The
-        # function is invoked by multiple sample CMakeLists.txt files; guard with
-        # a GLOBAL property so the install rules are emitted exactly once.
-        if(WIN32)
-            get_property(_ov_genai_opencv_dlls_installed GLOBAL
-                PROPERTY OV_GENAI_OPENCV_RUNTIME_DLLS_INSTALLED)
-            if(NOT _ov_genai_opencv_dlls_installed)
-                foreach(component IN LISTS required_components)
-                    install(FILES "$<TARGET_FILE:opencv_${component}>"
-                            DESTINATION samples_bin/
-                            COMPONENT samples_bin
-                            EXCLUDE_FROM_ALL)
-                endforeach()
-                # FFmpeg backend plugin: prebuilt DLL OpenCV places next to
-                # opencv_videoio with no CMake target. TARGET_FILE_DIR keeps the
-                # path tied to the OpenCV build layout and config, with no
-                # version-suffix or absolute-path assumptions.
-                install(FILES
-                            "$<TARGET_FILE_DIR:opencv_videoio>/opencv_videoio_ffmpeg_64.dll"
-                        DESTINATION samples_bin/
-                        COMPONENT samples_bin
-                        EXCLUDE_FROM_ALL
-                        OPTIONAL)
-                set_property(GLOBAL PROPERTY
-                    OV_GENAI_OPENCV_RUNTIME_DLLS_INSTALLED TRUE)
-            endif()
-        endif()
+        # Signal to callers that OpenCV was built from source via FetchContent,
+        # so target-scoped CMake can emit any runtime deployment rules it needs
+        # (e.g., installing the fetched DLLs next to a Windows sample EXE).
+        set_property(GLOBAL PROPERTY OV_GENAI_FETCHED_OPENCV TRUE)
     else()
         set(opencv_targets ${OpenCV_LIBS})
         if(OpenCV_INCLUDE_DIRS)
--- a/samples/cpp/image_generation/CMakeLists.txt
+++ b/samples/cpp/image_generation/CMakeLists.txt
@@ -12,6 +12,7 @@ file(DOWNLOAD https://raw.githubusercontent.com/nothings/stb/f75e8d1cad7d90d72ef
      EXPECTED_HASH MD5=27932e6fb3a2f26aee2fc33f2cb4e696)
 
 include(FetchContent)
+include(${CMAKE_CURRENT_SOURCE_DIR}/../fetch_opencv.cmake)
 
 if(POLICY CMP0135)
     cmake_policy(SET CMP0135 NEW)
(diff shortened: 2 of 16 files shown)


byekrang:
Updated the PR after the failed run and Copilot comments.

Changes in the latest push:
- Added `python/video_generation` to the Python samples install rule so installed `image_generation/denoising_process.py` can import the shared `video_utils.py` helper.
- Added `VideoWriter.isOpened()` validation in Python `video_utils.py`.
- Added `VideoCapture.isOpened()` validation in the denoising process test helper.

Local validation after the update:
- Install-rule check confirms `image_generation/denoising_process.py` and `video_generation/video_utils.py` are installed as siblings.
- Targeted pytest still passes: `1 passed in 208.47s`.
- `git diff --check` and pre-commit on the changed files passed.

The remaining OS matrix checks appear blocked pending approval after the force-push.

byekrang:
Pushed a Windows-only fix for the `tiny_random_latent_consistency` failure.

The C++ sample was exiting with `0xC0000135` (`STATUS_DLL_NOT_FOUND`) before `main()`, because the FetchContent-built OpenCV runtime DLLs were not installed beside `install/samples_bin/denoising_process.exe`.

The update installs fetched OpenCV runtime DLLs and the optional FFmpeg videoio plugin into `samples_bin/` on Windows. The block is guarded so it only applies to the FetchContent OpenCV path and does not affect Linux/macOS or system-OpenCV builds.

tests/python_tests/samples/test_denoising_process.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Must match the constants hardcoded in the denoising_process samples.
+NUM_INFERENCE_STEPS = 20
+IMAGE_HEIGHT = 512
+IMAGE_WIDTH = 512
+CHANNELS = 3
+EXPECTED_VIDEO_SHAPE = (NUM_INFERENCE_STEPS, IMAGE_HEIGHT, IMAGE_WIDTH, CHANNELS)
+
+
+def _read_video_frames(path: Path) -> np.ndarray:
```
Wovchena: `test_taylorseer.py` defines `compare_videos()`. Move `compare_videos()` to `test_utils.py` and call in both places.

samples/cpp/image_generation/CMakeLists.txt
```diff
@@ -38,6 +39,34 @@ install(TARGETS text2image
...
+    set_target_properties(denoising_process PROPERTIES
+        INSTALL_RPATH "$ORIGIN/../lib")
+elseif(APPLE)
+    set_target_properties(denoising_process PROPERTIES
+        INSTALL_RPATH "@loader_path/../lib")
+endif()
+
+set_target_properties(denoising_process PROPERTIES
+    # Ensure out of box LC_RPATH on macOS with SIP
+    INSTALL_RPATH_USE_LINK_PATH ON)
```
mryzhov: those settings will be set by ov_genai_link_opencv(), why do you repeat them?

samples/cpp/fetch_opencv.cmake
```diff
@@ -77,6 +77,36 @@ function(ov_genai_link_opencv target_name)
                 INSTALL_RPATH "@loader_path/../lib"
                 INSTALL_RPATH_USE_LINK_PATH ON)
         endif()
+
+        # Windows: deploy fetched OpenCV runtime DLLs alongside sample executables
+        # in samples_bin/. Windows has no rpath, so install/samples_bin/<exe> can
+        # only resolve its OpenCV dependencies via the loader search path. The
+        # function is invoked by multiple sample CMakeLists.txt files; guard with
+        # a GLOBAL property so the install rules are emitted exactly once.
+        if(WIN32)
```
mryzhov: That should not be part of a generic CMake function (fetch_opencv.cmake is universal for all samples now). If we need additional installation steps specific to samples, they should be handled in the target’s CMake configuration.

Another question: can we statically link sample with OpenCV?

samples/cpp/fetch_opencv.cmake
```diff
@@ -77,6 +77,36 @@ function(ov_genai_link_opencv target_name)
...
+                            DESTINATION samples_bin/
+                            COMPONENT samples_bin
+                            EXCLUDE_FROM_ALL)
+                endforeach()
+                # FFmpeg backend plugin: prebuilt DLL OpenCV places next to
+                # opencv_videoio with no CMake target. TARGET_FILE_DIR keeps the
+                # path tied to the OpenCV build layout and config, with no
+                # version-suffix or absolute-path assumptions.
+                install(FILES
+                            "$<TARGET_FILE_DIR:opencv_videoio>/opencv_videoio_ffmpeg_64.dll"
```
Wovchena: Can `ffmpeg` backend compilation be disabled instead? `cv::CAP_OPENCV_MJPEG` doesn't require `ffpmeg`

samples/cpp/fetch_opencv.cmake
Wovchena: Please, try static linking and check how it affects sample size. That way you won't need extra installation

Wovchena (changes requested):

samples/cpp/fetch_opencv.cmake
Wovchena: Never mind. It's needed for `video_to_text_chat` samples

byekrang:
Addressed the review comments in a follow-up commit.

Changes:
- Moved `compare_videos()` to `test_utils.py` and reused it from both `test_taylorseer.py` and `test_denoising_process.py`.
- Removed the duplicated `INSTALL_RPATH` block from the `denoising_process` target.
- Moved Windows OpenCV DLL deployment out of the generic `fetch_opencv.cmake` path and into the `denoising_process` target configuration.
- Kept `WITH_FFMPEG ON` unchanged; `denoising_process` installs only the OpenCV component DLLs it needs for `CAP_OPENCV_MJPEG`.

Local validation:
- pre-commit passed on the changed files,
- targeted `tiny_random_latent_consistency` pytest passed,
- `test_taylorseer.py` collects cleanly.

samples/cpp/fetch_opencv.cmake
byekrang: I investigated static linking, but did not adopt it in this PR because it would be a broader build-policy change affecting all OpenCV sample consumers, would not fully remove runtime deployment because video_to_text_chat still needs the FFmpeg videoio plugin, and would require CRT/size validation on Windows. For this PR, I kept dynamic linking and moved the Windows runtime deployment to the denoising_process target configuration so the fix stays scoped to the sample that needs it.

Wovchena (approved):

Wovchena:
build_jenkins

mryzhov (approved):

Wovchena (approved):
Ignore failed `macOS (14, Python 3.11)` runs. No action is needed from your side

byekrang:
Updated `compare_videos()` to read both captures until EOF instead of relying on `CAP_PROP_FRAME_COUNT`, so unknown/zero frame counts cannot incorrectly produce a true result.

sgonorov (approved):

Wovchena:
build_jenkins

PR #3890, opened 2026-05-20
[WWB] Create gt and output folders if they don't exist

CuriousPanCake:
## Description
Currently, wwb.py would fail to write ground truth data for evaluation if the directory provided to `--gt-data` does not exist. This may be a frustrating result to see after waiting for the model to generate reference results which may take significant amount of time for large models. The same goes for `--output` which writes resulting measurements into the provided folder and may fail if the folder doesn't exist.

Fix this by:
1) Checking if values to these parameters make sense (a file for `--gt-data` and folder a directory path to `--output`)
2) Implicitly creating directories on writing the results if they don't exist

CVS-###

## Checklist:

### DIFF
--- a/tools/who_what_benchmark/tests/test_cli_image.py
+++ b/tools/who_what_benchmark/tests/test_cli_image.py
@@ -138,6 +138,8 @@ def test_image_model_genai(model_id, model_type, tmp_path):
             "2",
             "--taylorseer-config",
             '{"disable_cache_after_step": 0}',
+            "--output",
+            tmp_path,
         ]
     )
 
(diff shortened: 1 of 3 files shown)


sbalandi (approved):

sbalandi:
@CuriousPanCake please, update also tests, some pipelines create `<gt-data parent/output>/target` folder with artifacts, and when output was not set, they were searching just `<gt-data parent>/target`. Now it's need to add `--output tmp_path` to some test_cli_image.py tests cmd

mryzhov (approved):

mryzhov (approved):
But you forgot the tests ;)

PR #3731, opened 2026-04-20
Qwen 3 asr support

dhandhalyabhavik:
## Overview

Add Qwen3-ASR (Qwen/Qwen3-ASR-0.6B and Qwen/Qwen3-ASR-1.7B) speech recognition support to the `WhisperPipeline`. The pipeline auto-detects Qwen3-ASR models from `config.json` and uses a dedicated encoder-decoder flow with chat-template prompt construction and audio pad token injection — no changes to the public API. This depends on [optimum-intel PR #1677](https://github.com/huggingface/optimum-intel/pull/1677) for OpenVINO IR export support.

## What has changed

| # | File | Changed Lines | Purpose |
|---|------|:---:|---------|
| 1 | `src/cpp/include/openvino/genai/whisper_pipeline.hpp` | +1 | Forward-declare `Qwen3ASRPipelineStatefulImpl` so the pipeline class can instantiate it |
| 2 | `src/cpp/src/whisper/config.hpp` | +15 | Add `model_type`, `audio_token_id`, `audio_start_token_id`, `audio_end_token_id` fields and `is_qwen3_asr()` helper to `WhisperConfig` |
| 3 | `src/cpp/src/whisper/config.cpp` | +14 | Parse `model_type` and Qwen3-ASR specific token IDs from the `thinker_config` section of `config.json` |
| 4 | `src/cpp/src/whisper/pipeline.cpp` | +107 | Add `Qwen3ASRPipelineStatefulImpl` class with encoder compilation, decoder loading, generate loop, and `<asr_text>` output prefix stripping; modify constructor to auto-detect model type |
| 5 | `src/cpp/src/whisper/qwen3_asr.hpp` | +32 (new) | Declare `qwen3_asr_generate()` function signature |
| 6 | `src/cpp/src/whisper/qwen3_asr.cpp` | +305 (new) | Implement Qwen3-ASR generate: mel feature extraction → encoder inference → chat-template prompt building with audio pad tokens → autoregressive decoding via existing `WhisperDecoder` and `Sampler` infrastructure |
| 7 | `samples/python/whisper_speech_recognition/benchmark_qwen3_asr.py` | +187 (new) | Benchmark script using genai built-in `PerfMetrics` to measure throughput (tok/s), TTFT, TPOT, RTF with GPU cache support |
| 8 | `tests/python_tests/test_qwen3_asr_pipeline.py` | +161 (new) | End-to-end test suite: silence, sine wave, real speech transcription, max_new_tokens, perf metrics, streamer callback |

[...]

PR #3515, opened 2026-03-18
Unified ASR Pipeline for Whisper and Paraformer-zh model

padatta:
## Summary
This PR adds comprehensive tests for both Whisper and Paraformer ASR models through a unified ASRPipeline interface.

### Python Tests (`test_asr_pipeline.py`)

- ✅ Add unified `ASRPipeline` wrapper that auto-detects model type (Whisper/Paraformer)
- ✅ Support `WhisperPipeline` when native `ASRPipeline` is not available (compatibility)
- ✅ Add `ParaformerPipeline` implementation with FBANK feature extraction
- ✅ Fix dataset loading to remove deprecated `trust_remote_code` parameter
- ✅ Mirror `test_whisper_pipeline.py` test coverage for both model types

### C++ Tests (`asr_pipeline.cpp`)
- ✅ Add multi-path fallback for model detection to support docker and dev environments
- ✅ Environment variables: `WHISPER_MODEL_DIR`, `PARAFORMER_MODEL_DIR`

### Running Tests

**Python tests:**
```bash
# With environment variables
export WHISPER_MODEL_PATH=/path/to/whisper-base-ov
export PARAFORMER_MODEL_PATH=/path/to/paraformer-zh/ov_models
pytest tests/python_tests/test_asr_pipeline.py -v

# Without env vars (uses auto-detection)
pytest tests/python_tests/test_asr_pipeline.py -v
```
**C++ tests:**
```bash
# With environment variables
export WHISPER_MODEL_DIR=/path/to/whisper-base-ov
export PARAFORMER_MODEL_DIR=/path/to/paraformer-zh/ov_models
./build/tests/cpp/openvino_genai_tests --gtest_filter="ASRPipeline*"

# Without env vars (uses fallback paths)
./build/tests/cpp/openvino_genai_tests --gtest_filter="ASRPipeline*"
```

PR #3893, opened 2026-05-21
Add ASRPipeline interface

as-suvorov:
## Description

There are requests for new automatic speech recognition models support: Qwen3-asr (https://github.com/openvinotoolkit/openvino.genai/pull/3731), Paraformer-zh (https://github.com/openvinotoolkit/openvino.genai/pull/3515).
Current automatic speech recognition implementation is model specific - [WhisperPipeline](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/whisper_pipeline.hpp).
In order to support new asr models there is a proposal for model-agnostic `ASRPipeline`. Proposed interface is a copy of `WhisperPipeline` except decoded results chunk unification, described in a section below.

**Pipeline interface** mirrors `WhisperPipeline` (same generate/config/tokenizer pattern):

```cpp
ASRPipeline pipe("model_path", "CPU");
auto result = pipe.generate(audio, ov::genai::language("<|en|>"), ov::genai::return_timestamps(true));
```

**Monolithic `ASRGenerationConfig`** — single value-type config containing all ASR model parameters. Model-specific fields are ignored by irrelevant backends.

**Unified `ASRDecodedResultChunk`** — replaces two separate WhisperPipeline structs (`WhisperDecodedResultChunk` + `WhisperWordTiming`) with one struct used for both segment-level and word-level timestamps:

```cpp
struct ASRDecodedResultChunk {
    float start_ts;
    float end_ts;
    std::string text;
    std::vector<int64_t> token_ids;
};
```

CVS-###

Fixes #(issue)

## Checklist:
- [ ] I have made corresponding changes to the documentation. - To be added in a separate PR


samples/cpp/automatic_speech_recognition/CMakeLists.txt
```diff
@@ -3,8 +3,8 @@
 
 find_package(OpenVINOGenAI REQUIRED
     PATHS
-        "${CMAKE_BINARY_DIR}"  # Reuse the package from the build.
-        ${OpenVINO_DIR}  # GenAI may be installed alogside OpenVINO.
+    "${CMAKE_BINARY_DIR}" # Reuse the package from the build.
+    ${OpenVINO_DIR} # GenAI may be installed alogside OpenVINO.
```
Wovchena:
```suggestion
        "${CMAKE_BINARY_DIR}" # Reuse the package from the build.
        ${OpenVINO_DIR} # GenAI may be installed alogside OpenVINO.
```
Both lines are PATHS

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
```diff
@@ -0,0 +1,120 @@
...
+    std::optional<std::string> hotwords = std::nullopt;
+
+    void update_generation_config(const ov::AnyMap& config_map = {});
+
+    template <typename... Properties>
+    util::EnableIfAllStringAny<void, Properties...> update_generation_config(Properties&&... properties) {
+        return update_generation_config(ov::AnyMap{std::forward<Properties>(properties)...});
+    }
+
+    void validate() const;
```
Wovchena: Inheritance conflicts with methods being non virtual. It's a good moment to avoid inheritance and methods like it was done for new `VideoGenerationConfig`

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
```diff
@@ -0,0 +1,120 @@
...
+     * Can be used to steer the model to use particular spellings or styles.
+     *
+     * Example:
+     *  auto result = pipeline.generate(raw_speech);
+     *  //  He has gone and gone for good answered Paul Icrom who...
+     *
+     *  auto result = pipeline.generate(raw_speech, ov::genai::hotwords("Polychrome"));
+     *  //  He has gone and gone for good answered Polychrome who...
+     */
+    std::optional<std::string> hotwords = std::nullopt;
```
Wovchena, line 81: Is there difference between empty `string` and `nullopt`? If not, remove `optional` for `initial_prompt` and `hotwords`

src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+    operator std::string() const {
+        std::stringstream ss;
+        ss << *this;
+        return ss.str();
+    };
+
+    operator std::vector<std::string>() const {
+        return texts;
+    };
```
Wovchena, line 48: What is implicit cast for? Adding `.texts` isn't difficult in code

src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
```diff
@@ -0,0 +1,98 @@
...
+        for (size_t i = 0; i < dr.texts.size() - 1; ++i) {
+            os << std::to_string(dr.scores[i]) << ": " << dr.texts[i] << '\n';
+        }
+        return os << std::to_string(dr.scores.back()) << ": " << dr.texts.back();
+    }
+};
+
+class OPENVINO_GENAI_EXPORTS ASRPipeline {
+    class ASRPipelineImplBase;
+    class WhisperASRPipelineAdapter;
```
Wovchena:
```suggestion
```

src/cpp/src/automatic_speech_recognition/pipeline.cpp
```diff
@@ -0,0 +1,94 @@
...
+}  // namespace
+
+namespace ov::genai {
+
+ASRPipeline::ASRPipeline(const std::filesystem::path& models_path,
+                         const std::string& device,
+                         const ov::AnyMap& properties) {
+    const ASRModelType model_type = read_model_type(models_path);
+
+    if (model_type == ASRModelType::WHISPER) {
```
Wovchena: Replace with `switch`

src/cpp/src/automatic_speech_recognition/whisper_asr_pipeline_adapter.hpp
```diff
@@ -0,0 +1,113 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "automatic_speech_recognition/pipeline_base.hpp"
+#include "openvino/genai/whisper_pipeline.hpp"
+
+namespace {
```
Wovchena:
```suggestion
```
Avoid anonymous namespace in headers

src/python/py_asr_pipeline.cpp
```diff
@@ -0,0 +1,343 @@
...
+        .def_property_readonly("texts",
+                               [](const ASRDecodedResults& dr) -> py::typing::List<py::str> {
+                                   return pyutils::handle_utf8((std::vector<std::string>)dr);
+                               })
+        .def_readonly("scores", &ASRDecodedResults::scores)
+        .def_readonly("language", &ASRDecodedResults::language)
+        .def_readonly("chunks", &ASRDecodedResults::chunks)
+        .def_readonly("words", &ASRDecodedResults::words)
+        .def_readonly("perf_metrics", &ASRDecodedResults::perf_metrics)
+        .def("__str__", [](const ASRDecodedResults& dr) -> py::str {
```
Wovchena: `__repr__` should be defined instead or additionally because it's used in interactive debugging

src/cpp/src/automatic_speech_recognition/pipeline.cpp
as-suvorov: Why?

src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
as-suvorov: Thanks for catch! Fixed

src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
as-suvorov: It's just copy of whisper and common `DecodedResults` operator. Added just to be aligned with them

src/cpp/src/automatic_speech_recognition/pipeline.cpp
Wovchena: That's more C++ style. It's also used in examples in https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#enum2-use-enumerations-to-represent-sets-of-related-named-constants
Not relevant but FYI, switch compiles into tree instead of sequential checks

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
as-suvorov: No difference on implementation side. But there is a semantic difference I prefer to keep between value not set vs value set to empty string. The same applies for image and video generation: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/include/openvino/genai/image_generation/generation_config.hpp#L84. Not sure if it's intentional though.

src/python/py_asr_pipeline.cpp
as-suvorov: Switched to  `__repr__`.

src/cpp/src/automatic_speech_recognition/whisper_asr_pipeline_adapter.hpp
as-suvorov: Fixed, thanks for catch

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
as-suvorov: Added `virtual` for `GenerationConfig` methods as discussed

samples/cpp/automatic_speech_recognition/CMakeLists.txt
as-suvorov: Fixed formatting

src/cpp/src/automatic_speech_recognition/pipeline.cpp
as-suvorov: Replaced with `switch`. We need to add it to linting rules. cc: @sgonorov

michalkulakowski:
LGTM

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
Wovchena: What is your preference based on?
Image generation distinguishes them because empty string still should be embedded while absence of string shouldn't be processed at all. `hotwords` are different because they are just inserted to the text so there's no difference between empty and absent string

src/cpp/src/automatic_speech_recognition/pipeline.cpp
Wovchena: I didn't encounter such linting rules

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
as-suvorov: `hotwords = std::nullopt` are not processed at all.
But we can check if `hotwords` is empty string and skip processing, `hotwords.has_value() -> hotwords.empty()`.
We can also remove `optional` from `task` and set to `transcribe` by default. Please tell me if you prefer to get rid of optionals where possible.

From image generation implementation it looks like `prompt_2 = std::nullopt` is set to positive string and processed anyway. And same for `negative_prompt`, it set to empty string and processed if not provided. At least for sd3.

src/cpp/src/automatic_speech_recognition/pipeline.cpp
as-suvorov: I just assumed it should exist :) Could not find it. @sgonorov please ignore request.

Wovchena (approved):

PR #3906, opened 2026-05-23
[ENH] Optimize LoRA tensor lookup by pre-filtering in constructor

andersendsa:
## Description

This change addresses a TODO comment in `src/cpp/src/lora/adapter.cpp` recommending prefix filtering during object construction.

Instead of performing string matching, `substr` operations, and a full `std::map` scan on every `LoRAWeightGetterDefault::operator()` call, we now filter the tensors in the constructor. We store matching tensors (with their stripped names and stable value pointers) in a `std::vector`, resulting in a faster allocation-free O(k) iteration over only relevant tensors on the hot path. The associated TODO comment was removed.

## Checklist:


andersendsa:
Hi @likholat  could you pls let me know if this pr needs any changes or if it is good to merge

PR #3907, opened 2026-05-25
[Video Generation] Implement Image2Video Pipeline

goyaladitya05:
## Description
This PR implementes Image2Video Pipeline for LTX-Video, i.e. image-conditioned video generation where frame 0 is anchored to a conditioning image and subsequent frames are generated from a text prompt. Also contains previously already approved commits from #3829. 

An image-to-video sample generated with this PR combined with optimum PR :

i2v_fp32 (2) (1)

- Per-token [B, S] timestep conditioning: frame-0 tokens get t=0 (clean), remaining tokens get the current denoising timestep.
- Frame-0 latent is pinned to the encoded image throughout the denoising loop.
- Auto-detects rank-1 vs rank-2 timestep shape for compatibility with both legacy and current exports.

Optimum PR: https://github.com/huggingface/optimum-intel/pull/1762

Text-to-Video sample
t2v_fp32

## Checklist:
- [x] I have made corresponding changes to the documentation.  https://goyaladitya05.github.io/openvino.genai/


src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -404,6 +409,107 @@ class Text2VideoPipeline::LTXPipeline {
...
+        return pack_latents(latents, sp_patch, tp_patch);
+    }
+
+    // Returns packed image_latent frame 0 tokens: [B, H_lat*W_lat, C].
+    // Used for Mechanism C re-lock after every scheduler step.
+    static ov::Tensor pack_image_frame0(const ov::Tensor& image_latent, size_t sp_patch, size_t tp_patch) {
+        ov::Tensor copy(image_latent.get_element_type(), image_latent.get_shape());
+        image_latent.copy_to(copy);
+        return pack_latents(copy, sp_patch, tp_patch);
+    }
```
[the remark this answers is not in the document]
goyaladitya05: dead code, has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -404,6 +409,107 @@ class Text2VideoPipeline::LTXPipeline {
...
+        std::memcpy(input_4d.data<uint8_t>(), image.data<const uint8_t>(), H * W * C);
+
+        if (static_cast<int64_t>(H) != height || static_cast<int64_t>(W) != width) {
+            input_4d = m_image_resizer->execute(input_4d, height, width);
+        }
+
+        ov::Tensor processed = m_image_processor->execute(input_4d);
+
+        ov::Tensor out(ov::element::f32, {1, C, 1, static_cast<size_t>(height), static_cast<size_t>(width)});
+        std::memcpy(out.data<float>(), processed.data<const float>(), C * static_cast<size_t>(height) * static_cast<size_t>(width) * sizeof(float));
```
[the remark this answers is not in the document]
goyaladitya05: dead code, has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -404,6 +409,107 @@ class Text2VideoPipeline::LTXPipeline {
...
+    }
+
+    // Mechanism C: overwrite frame-0 tokens in packed [B, S, D] latent with pre-packed image tokens.
+    static void relock_frame0(ov::Tensor& packed_latent, const ov::Tensor& packed_frame0) {
+        const ov::Shape& lat_shape = packed_latent.get_shape();
+        const ov::Shape& f0_shape = packed_frame0.get_shape();
+        const size_t B = lat_shape[0];
+        const size_t S = lat_shape[1];
+        const size_t D = lat_shape[2];
+        const size_t T0 = f0_shape[1];  // H_lat * W_lat
```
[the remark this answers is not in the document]
goyaladitya05: dead code, has been removed.

src/cpp/src/video_generation/generation_config_utils.cpp
```diff
@@ -11,6 +11,8 @@ void validate_generation_config(const VideoGenerationConfig& config) {
     if (config.guidance_scale <= 1.0f && config.negative_prompt != std::nullopt) {
         GENAI_WARN("Guidance scale <= 1.0 ignores negative prompt");
     }
+    OPENVINO_ASSERT(config.strength >= 0.0f && config.strength <= 1.0f,
+                    "Parameter 'strength' must be in range [0, 1] but got ", config.strength);
```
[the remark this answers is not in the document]
goyaladitya05: dead code, has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -17,6 +17,7 @@
 #include "openvino/op/divide.hpp"
 #include "openvino/op/multiply.hpp"
 
+#include "image_generation/image_processor.hpp"
 #include "image_generation/numpy_utils.hpp"
 #include "image_generation/schedulers/ischeduler.hpp"
 #include "image_generation/threaded_callback.hpp"
```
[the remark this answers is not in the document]
goyaladitya05: dead code, has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -732,6 +835,255 @@ class Text2VideoPipeline::LTXPipeline {
...
+                                                           transformer_temporal_patch_size);
+
+        const size_t video_sequence_length = m_latent_num_frames * m_latent_height * m_latent_width;
+        std::vector<float> timesteps;
+        if (merged_generation_config.strength > 0.0f) {
+            m_scheduler->set_timesteps(video_sequence_length,
+                                       merged_generation_config.num_inference_steps,
+                                       merged_generation_config.strength);
+            timesteps = m_scheduler->get_float_timesteps();
+        }
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -732,6 +835,255 @@ class Text2VideoPipeline::LTXPipeline {
...
+            const size_t B_cfg = latent_cfg.get_shape()[0];
+            const size_t S = latent_cfg.get_shape()[1];
+            ov::Tensor timestep(ov::element::f32, {B_cfg, S});
+            float* timestep_data = timestep.data<float>();
+            const float t = timesteps[inference_step];
+            for (size_t b = 0; b < B_cfg; ++b) {
+                for (size_t s = 0; s < S; ++s) {
+                    timestep_data[b * S + s] = (s < T0) ? 0.0f : t;
+                }
+            }
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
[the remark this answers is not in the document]
goyaladitya05: fixed

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -780,6 +1132,10 @@ class Text2VideoPipeline::LTXPipeline {
         m_compile_properties = properties;
         m_is_compiled = true;
         m_compiled_batch_size_multiplier = m_reshape_batch_size_multiplier;
+        if (m_has_encoder) {
+            m_image_processor = std::make_shared<ImageProcessor>("CPU");
+            m_image_resizer = std::make_shared<ImageResizer>("CPU", ov::element::u8, "NHWC", ov::op::v11::Interpolate::InterpolateMode::BICUBIC_PILLOW);
```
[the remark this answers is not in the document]
goyaladitya05: fixed

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -404,6 +410,107 @@ class Text2VideoPipeline::LTXPipeline {
...
+        if (raw_shape.size() == 4) {
+            OPENVINO_ASSERT(raw_shape[0] == 1, "Batch dimension must be 1, got ", raw_shape[0]);
+        }
+        const size_t H = raw_shape.size() == 3 ? raw_shape[0] : raw_shape[1];
+        const size_t W = raw_shape.size() == 3 ? raw_shape[1] : raw_shape[2];
+        const size_t C = raw_shape.size() == 3 ? raw_shape[2] : raw_shape[3];
+        OPENVINO_ASSERT(C == 3, "Conditioning image must have 3 channels (RGB), got ", C);
+
+        ov::Tensor input_4d(ov::element::u8, {1, H, W, C});
+        std::memcpy(input_4d.data<uint8_t>(), image.data<const uint8_t>(), H * W * C);
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -404,6 +410,107 @@ class Text2VideoPipeline::LTXPipeline {
...
+
+        for (size_t b = 0; b < B; ++b) {
+            for (size_t c = 0; c < C; ++c) {
+                // latents[b, c, 0, :, :] starts at (b*C*F + c*F) * HW
+                float* dst = lat_data + (b * C * F + c * F) * HW;
+                // image_latent[b, c, 0, :, :] starts at (b*C + c) * HW
+                const float* src = img_data + (b * C + c) * HW;
+                std::memcpy(dst, src, HW * sizeof(float));
+            }
+        }
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -732,6 +836,254 @@ class Text2VideoPipeline::LTXPipeline {
...
+        latent_shape_cfg[0] *= batch_size_multiplier;
+        ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
+
+        const size_t S = latent_shape_cfg[1];
+
+        TaylorSeerState ts_state(merged_generation_config.taylorseer_config, timesteps.size());
+
+        ov::Tensor timestep;
+        ov::Tensor noisy_residual_tensor(ov::element::f32, {});
+        for (size_t inference_step = 0; inference_step < timesteps.size(); ++inference_step) {
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -732,6 +836,254 @@ class Text2VideoPipeline::LTXPipeline {
...
+            const size_t B_cfg = latent_cfg.get_shape()[0];
+            const float t = timesteps[inference_step];
+            if (!timestep || timestep.get_shape()[0] != B_cfg) {
+                timestep = ov::Tensor(ov::element::f32, {B_cfg, S});
+            }
+            float* timestep_data = timestep.data<float>();
+            for (size_t b = 0; b < B_cfg; ++b) {
+                std::fill_n(timestep_data + b * S, T0, 0.0f);
+                std::fill_n(timestep_data + b * S + T0, S - T0, t);
+            }
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -732,6 +836,254 @@ class Text2VideoPipeline::LTXPipeline {
...
+        const size_t transformer_temporal_patch_size = transformer_config.patch_size_t;
+
+        m_latent_num_frames = (merged_generation_config.num_frames - 1) / temporal_compression_ratio + 1;
+        m_latent_height = merged_generation_config.height / spatial_compression_ratio;
+        m_latent_width = merged_generation_config.width / spatial_compression_ratio;
+
+        // Encode conditioning image → [1, C, 1, H_lat, W_lat]
+        ov::Tensor preprocessed = preprocess_conditioning_image(
+            image, merged_generation_config.height, merged_generation_config.width);
+        ov::Tensor image_latent = m_vae->encode(preprocessed, merged_generation_config.generator);
```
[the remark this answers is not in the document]
goyaladitya05: outdated, code has been removed.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -332,6 +355,71 @@ class Text2VideoPipeline::LTXPipeline {
...
+        ov::Tensor noise = config.generator->randn_tensor(shape);
+        ov::Tensor latents = pack_latents(noise, ps, ps_t);
+
+        const size_t tokens_per_frame = m_latent_height * m_latent_width;
+        const size_t D = latents.get_shape()[2];
+        for (size_t b = 0; b < config.num_videos_per_prompt; ++b) {
+            float* dst       = latents.data<float>()             + b * m_latent_num_frames * tokens_per_frame * D;
+            const float* src = image_latent_packed.data<float>() + b * tokens_per_frame * D;
+            std::memcpy(dst, src, tokens_per_frame * D * sizeof(float));
+        }
```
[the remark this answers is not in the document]
goyaladitya05: fixed

src/cpp/src/video_generation/image2video_pipeline.cpp
```diff
@@ -0,0 +1,90 @@
...
+    return m_impl->decode(latent);
+}
+
+VideoGenerationPerfMetrics Image2VideoPipeline::get_performance_metrics() {
+    return m_impl->get_performance_metrics();
+}
+
+Image2VideoPipeline Image2VideoPipeline::clone() {
+    return Image2VideoPipeline(std::make_shared<LTXPipeline>(*m_impl));
+}
```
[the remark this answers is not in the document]
goyaladitya05: fixed

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -332,6 +355,71 @@ class Text2VideoPipeline::LTXPipeline {
...
+        ov::Tensor processed = m_image_processor->execute(resized);
+
+        // Own the tensor before re-input: ImageProcessor returns an aliased buffer; reusing it
+        // as a set_input_tensor() target for another infer() causes a hang.
+        const auto& proc_shape = processed.get_shape();
+        ov::Tensor encoder_input(ov::element::f32, {proc_shape[0], proc_shape[1], 1, proc_shape[2], proc_shape[3]});
+        std::memcpy(encoder_input.data<float>(), processed.data<const float>(),
+                    proc_shape[0] * proc_shape[1] * proc_shape[2] * proc_shape[3] * sizeof(float));
+
+        ov::Tensor latent = m_vae->encode(encoder_input, config.generator);
```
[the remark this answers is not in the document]
goyaladitya05: addressed

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +621,265 @@ class Text2VideoPipeline::LTXPipeline {
...
+            const float t = timesteps[inference_step];
+            const size_t B_ts = latent_cfg.get_shape()[0];
+            ov::Tensor timestep(ov::element::f32, {B_ts, video_sequence_length});
+            float* timestep_data = timestep.data<float>();
+            for (size_t b = 0; b < B_ts; ++b) {
+                float* row = timestep_data + b * video_sequence_length;
+                for (size_t i = 0; i < video_sequence_length; ++i) {
+                    row[i] = t * (1.0f - conditioning_mask[i]);
+                }
+            }
```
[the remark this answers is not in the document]
goyaladitya05: addressed

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +639,250 @@ class Text2VideoPipeline::LTXPipeline {
...
+            static_cast<float>(temporal_compression_ratio) / frame_rate;
+        rope_interpolation_scale.data<float>()[1] = spatial_compression_ratio;
+        rope_interpolation_scale.data<float>()[2] = spatial_compression_ratio;
+        m_transformer->set_hidden_states("rope_interpolation_scale", rope_interpolation_scale);
+
+        // Frame-0 tokens are first in the packed [B, S, D] layout (pack_latents is F-outermost).
+        const size_t tokens_per_frame =
+            (m_latent_height / transformer_spatial_patch_size) *
+            (m_latent_width  / transformer_spatial_patch_size);
+        std::vector<float> conditioning_mask(video_sequence_length, 0.0f);
```
sgonorov: `conditioning_mask` seems to be unused

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +639,250 @@ class Text2VideoPipeline::LTXPipeline {
...
+        std::fill_n(conditioning_mask.begin(), tokens_per_frame, 1.0f);
+
+        ov::Shape latent_shape_cfg = latent.get_shape();
+        latent_shape_cfg[0] *= batch_size_multiplier;
+        ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
+
+        TaylorSeerState ts_state(merged_generation_config.taylorseer_config, timesteps.size());
+
+        const size_t B_ts = latent_shape_cfg[0];
+        ov::Tensor timestep(ov::element::f32, {B_ts, video_sequence_length});
```
sgonorov, line 765: As far as i remember I2V assumes rank-2 timestep and legacy rank-1 models will fail, maybe it's better to add shape check like T2V path

src/cpp/include/openvino/genai/video_generation/text2video_pipeline.hpp
```diff
@@ -178,8 +182,9 @@ class OPENVINO_GENAI_EXPORTS Text2VideoPipeline {
     ~Text2VideoPipeline();
 
 private:
-    class LTXPipeline;
-    std::unique_ptr<LTXPipeline> m_impl;
+    std::shared_ptr<LTXPipeline> m_impl;
```
sgonorov: `m_impl` changed from unique_ptr to shared_ptr so sizeof(Text2VideoPipeline) changes, may lead to ABI break as it's OPENVINO_GENAI_EXPORTS. Just check, maybe update PR description.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -443,19 +542,38 @@ class Text2VideoPipeline::LTXPipeline {
...
     }
 
     LTXPipeline(const std::filesystem::path& models_dir,
                 const std::string& device,
                 const ov::AnyMap& properties,
+                VideoPipelineType pipeline_type = VideoPipelineType::TEXT_2_VIDEO,
                 std::chrono::steady_clock::time_point start_time = std::chrono::steady_clock::now())
         : m_scheduler{cast_scheduler(Scheduler::from_config(models_dir / "scheduler/scheduler_config.json"))},
-          m_t5_text_encoder{std::make_shared<T5EncoderModel>(models_dir / "text_encoder", device, properties)},
+          m_t5_text_encoder{std::make_shared<T5EncoderModel>(models_dir / "text_encoder", device, with_cpu_fp32_default(device, properties))},
```
sgonorov: I think it changes T5 encoder behaviour on CPU to fp32. Maybe add more documentation or fill a bug.

sgonorov (changes requested):

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: added

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: removed

src/cpp/include/openvino/genai/video_generation/text2video_pipeline.hpp
goyaladitya05: I initially made that change to align with text-to-image and image-to-image pipelines, but I don't think that pattern applies here. reverted it to unique_ptr.

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: I was experimenting around that earlier, and this was a leftover bit. removed it now. After removing this block, I generated a sample to check output quality, and there wasn't any noticable difference.

goyaladitya05:
This was the output I got without forcing the transformer to fp32
i2v_fp32 (4)

goyaladitya05:
> This was the output I got without forcing the transformer to fp32 

Following this, I ran more tests.
1. The outputs were still grey/bad/noisy on genai master using Text2Video.
2. I had been using a bf16 capable CPU (Intel Sapphire Rapids). I isolate this to be an issue with how OpenVINO's CPU plugin handles bf16. 
3. I tested with a GPU using f16, and the output was okay (attached).
t2v_GPU_default

i2v_fp16

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -273,6 +275,8 @@ ov::Tensor denormalize_latents(const ov::Tensor& latents,
     return result[0];  // [B, C, F, H, W]
 }
 
+
```
likholat:
```suggestion
```

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -273,6 +275,8 @@ ov::Tensor denormalize_latents(const ov::Tensor& latents,
...
+
```
likholat:
```suggestion
```

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -332,6 +344,74 @@ class Text2VideoPipeline::LTXPipeline {
         return pack_latents(latents, transformer_spatial_patch_size, transformer_temporal_patch_size);
     }
 
+    ov::Tensor prepare_latents(const ov::Tensor& image_latent_packed,
```
likholat: We should either extend the functionality of the original prepare_latents method, or, if that’s not possible, add a docstring describing the differences between these two implementations and their use cases

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -407,7 +487,12 @@ class Text2VideoPipeline::LTXPipeline {
 public:
     VideoGenerationConfig m_generation_config;
 
+    VideoGenerationPerfMetrics get_performance_metrics() const {
+        return m_perf_metrics;
+    }
+
     LTXPipeline(const std::filesystem::path& root_dir,
```
likholat: Could we keep pipeline_type as the first parameter in all constructors to make the parameter order consistent?

```suggestion
    LTXPipeline(VideoPipelineType pipeline_type
```

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -443,19 +532,38 @@ class Text2VideoPipeline::LTXPipeline {
...
+            m_image_resizer = std::make_shared<ImageResizer>(
+                "CPU", ov::element::u8, "NHWC",
+                ov::op::v11::Interpolate::InterpolateMode::BICUBIC_PILLOW);
+            m_image_processor = std::make_shared<ImageProcessor>("CPU", true);
+        }
 
         m_load_time = Ms{std::chrono::steady_clock::now() - start_time};
     }
 
     LTXPipeline(const std::filesystem::path& models_dir,
```
likholat: same as above

likholat (changes requested):

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: updated

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: done

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: done

src/cpp/src/video_generation/image2video_pipeline.cpp
```diff
@@ -0,0 +1,90 @@
+// Copyright (C) 2025-2026 Intel Corporation
+// SPDX-License-Identifier: Apache-2.0
+
+#include "openvino/genai/video_generation/image2video_pipeline.hpp"
+#include "video_generation/ltx_pipeline.hpp"
+
+using namespace ov::genai;
+
+Image2VideoPipeline::Image2VideoPipeline(const std::filesystem::path& model_path) {
+    OPENVINO_ASSERT(std::filesystem::exists(model_path / "vae_encoder"),
```
likholat, line 13: Similar to image generation pipeline, let's check that the input model is an LTX model (for both text2video and image2video pipelines): https://github.com/openvinotoolkit/openvino.genai/blob/9a1fbf0ce882cf5e8dfc4063919028ce23696c47/src/cpp/src/image_generation/text2image_pipeline.cpp#L21

goyaladitya05:
> > This was the output I got without forcing the transformer to fp32
> 
> Following this, I ran more tests.
> 
> 1. The outputs were still grey/bad/noisy on genai master using Text2Video.
> 2. I had been using a bf16 capable CPU (Intel Sapphire Rapids). I isolate this to be an issue with how OpenVINO's CPU plugin handles bf16.
> 3. I tested with a GPU using f16, and the output was okay (attached).

Ref: https://github.com/openvinotoolkit/openvino/issues/36715

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +618,250 @@ class Text2VideoPipeline::LTXPipeline {
...
+            }
+
+            // Use TaylorSeer if enabled and caching is appropriate
+            ov::Tensor noise_pred_tensor;
+            if (ts_state.is_active() && !ts_state.should_compute(inference_step)) {
+                noise_pred_tensor = ts_state.predict(inference_step);
+            } else {
+                auto infer_start = std::chrono::steady_clock::now();
+                noise_pred_tensor = m_transformer->infer(latent_cfg, timestep);
+                auto infer_duration = ov::genai::PerfMetrics::get_microsec(std::chrono::steady_clock::now() - infer_start);
```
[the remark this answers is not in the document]
goyaladitya05: that branch cannot be reached today.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -633,22 +999,20 @@ class Text2VideoPipeline::LTXPipeline {
...
             }
-            // Match compiled model's expected batch size by repeating latent if needed
-            // (e.g., when model was compiled with CFG but current config doesn't require it)
+            // Repeat latent if model was compiled for CFG but current call has guidance_scale<=1.
             const size_t request_input_batch = m_transformer->get_request_input_batch();
             if (request_input_batch > latent_cfg.get_shape()[0]) {
                 OPENVINO_ASSERT(request_input_batch % latent_cfg.get_shape()[0] == 0,
                                 "Transformer input batch must be divisible by latent batch");
                 latent_cfg = numpy_utils::repeat(latent_cfg, request_input_batch / latent_cfg.get_shape()[0]);
             }
```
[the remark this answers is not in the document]
goyaladitya05: that branch cannot be reached today.

src/cpp/include/openvino/genai/video_generation/image2video_pipeline.hpp
```diff
@@ -0,0 +1,132 @@
+// Copyright (C) 2025-2026 Intel Corporation
```
likholat, line 1: Let's introduce `VideoPipeline` abstract base (like `DiffusionPipeline` in Image Generation)

```cpp
// Internal abstract base (similar to DiffusionPipeline)
class VideoPipeline { /* virtual interface */ };
class LTXPipeline : public VideoPipeline { /* implementation */ };

class Image2VideoPipeline {
private:
    std::shared_ptr<VideoPipeline> m_impl;
};
```

This prepares the current codebase for LTX2 support

@Wovchena do you agree with this approach?

Wovchena: pimpl is already in place. Your proposal boils down to renaming `Impl` to `VideoPipeline` and adding inheritance. I think this can be done anytime. It's usually simpler to introduce abstractions when there's a need for them. But If you already envision the need, you can introduce the abstractions now.

goyaladitya05: +1, although I think it would be better to introduce this with the LTX 2 pull request itself. would that be fine @likholat ?

src/cpp/src/video_generation/image2video_pipeline.cpp
goyaladitya05: done

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -130,6 +145,27 @@ size_t LTXVideoTransformer3DModel::get_request_input_batch() {
...
+        for (const auto& input : compiled.inputs()) {
+            if (input.get_any_name() == "timestep") {
+                return input.get_partial_shape();
+            }
+        }
+    }
+    OPENVINO_ASSERT(false,
+                    "LTXVideoTransformer3DModel: 'timestep' input not found in the model. "
+                    "The model may be corrupted or exported incorrectly.");
+}
```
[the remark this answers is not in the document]
goyaladitya05: The point after the assert is unreachable, because the assert throws and control never proceeds past it.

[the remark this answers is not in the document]
goyaladitya05: also, this is the same pattern as pipeline_static.cpp:429/:441, which end non-void functions with OPENVINO_ASSERT(false) and no trailing return.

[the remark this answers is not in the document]
goyaladitya05: The point after the assert is unreachable, because the assert throws and control never proceeds past it.
also, this is the same pattern as pipeline_static.cpp:429/:441, which end non-void functions with OPENVINO_ASSERT(false) and no trailing return.

likholat:
LGTM, just waiting for the Optimum PR to be merged

goyaladitya05:
> LGTM, just waiting for the Optimum PR to be merged

@likholat the Optimum PR has been merged.

likholat (approved):

likholat:
@sgonorov please tale a look

goyaladitya05:
@sgonorov @likholat I added 3 xfail markers in commit [29de5d5](https://github.com/openvinotoolkit/openvino.genai/pull/3907/commits/29de5d565b493fc0055b2143ad639086d8d7ef8a). These tests require the latest optimum-intel commit (the 1D to 2D timestep change). I tried updating optimum-intel to the latest commit in tests dependencies, but it introduced a chain of dependency conflicts and issues. They can be resolved once the latest version of nncf is released, so I will remove the markers then.

likholat:
build_jenkins

likholat:
> @sgonorov @likholat I added 3 xfail markers in commit [29de5d5](https://github.com/openvinotoolkit/openvino.genai/pull/3907/commits/29de5d565b493fc0055b2143ad639086d8d7ef8a). These tests require the latest optimum-intel commit (the 1D to 2D timestep change). I tried updating optimum-intel to the latest commit in tests dependencies, but it introduced a chain of dependency conflicts and issues. They can be resolved once the latest version of nncf is released, so I will remove the markers then.

@goyaladitya05 We can't merge the PR while the tests covering the new API are disabled. Could you please try using the master branch of Optimum only for the specific test cases, in the same way as was done here: https://github.com/openvinotoolkit/openvino.genai/pull/3930/changes

tests/python_tests/test_video_generation.py
```diff
@@ -433,3 +513,68 @@ def test_transformer_has_set_adapters_method(self, video_generation_model):
...
+        if model_index.exists():
+            shutil.copy(model_index, no_encoder_dir / "model_index.json")
+        with pytest.raises(RuntimeError, match="vae_encoder"):
+            ov_genai.Image2VideoPipeline(str(no_encoder_dir))
+
+    @pytest.mark.xfail(
+        reason="Test model exports rank-1 timestep. Needs optimum-intel PR #1762, "
+        "blocked by safetensors<0.8.0 vs diffusers==0.39.0",
+        raises=RuntimeError,
+    )
```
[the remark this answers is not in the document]
goyaladitya05: removed

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +632,250 @@ class Text2VideoPipeline::LTXPipeline {
...
+        latent_shape_cfg[0] *= batch_size_multiplier;
+        ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
+
+        TaylorSeerState ts_state(merged_generation_config.taylorseer_config, timesteps.size());
+
+        const size_t B_ts = latent_shape_cfg[0];
+        OPENVINO_ASSERT(m_transformer->get_timestep_partial_shape().size() == 2,
+                        "Image-to-video requires a rank-2 [B, S] timestep input. "
+                        "Re-export the model with: optimum-cli export openvino --task image-to-video");
+        ov::Tensor timestep(ov::element::f32, {B_ts, video_sequence_length});
```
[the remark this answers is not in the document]
goyaladitya05: what I really meant by that, was that it supports the Rank 1 legacy exports and new rank 2 exports for Text2Video.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -151,11 +187,17 @@ LTXVideoTransformer3DModel& LTXVideoTransformer3DModel::reshape(int64_t batch_si
 
     std::map<std::string, ov::PartialShape> name_to_shape;
 
+    const int64_t video_sequence_length = num_frames * height * width;
```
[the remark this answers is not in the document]
goyaladitya05: Not a mismatch, num_frames, height, and width are reassigned to their packed values a few lines above before multiplication. so video_sequence_length already equals (F/patch_size_t) · (H/patch_size) · (W/patch_size), the pack_latents() layout. And the hidden_states input is reshaped with the identical num_frames * height * width expression, so timestep[B, S] and hidden_states[B, S, D] share the same S by construction. No shape incompatibility.

goyaladitya05:
> @goyaladitya05 We can't merge the PR while the tests covering the new API are disabled. Could you please try using the master branch of Optimum only for the specific test cases, in the same way as was done here: https://github.com/openvinotoolkit/openvino.genai/pull/3930/changes

@likholat I understand, I have done the prescribed change.

goyaladitya05:
d556aac Fixes a crash when calling generate() twice on the same pipeline with a shrinking batch (e.g. guidance_scale=3 then guidance_scale=1), get_request_input_batch() was reading a stale batch size left over from the previous call's bound tensor, causing hidden_states and timestep to end up mismatched. 

Removed it along with the repeat logic that depended on it, since batch_size_multiplier already accounts for the compiled model's expected batch correctly. Verified with a before/after reproduction and an updated WWB text-to-video check (Optimum vs HF 0.9861, GenAI vs HF 0.9861, GenAI vs Optimum 0.9999).

goyaladitya05:
@sgonorov could you please review?

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -151,11 +187,17 @@ LTXVideoTransformer3DModel& LTXVideoTransformer3DModel::reshape(int64_t batch_si
...
     for (auto&& input : m_model->inputs()) {
         std::string input_name = input.get_any_name();
         name_to_shape[input_name] = input.get_partial_shape();
         if (input_name == "timestep") {
-            name_to_shape[input_name][0] = 1;
+            // Rank-1 [B] (legacy export) or rank-2 [B, S] (current export, per-token conditioning).
+            name_to_shape[input_name][0] = batch_size;
+            if (name_to_shape[input_name].size() >= 2) {
+                name_to_shape[input_name][1] = video_sequence_length;
+            }
```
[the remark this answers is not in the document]
goyaladitya05: same as above

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +632,250 @@ class Text2VideoPipeline::LTXPipeline {
...
+                OPENVINO_ASSERT(request_input_batch % latent_cfg.get_shape()[0] == 0,
+                                "Transformer input batch must be divisible by latent batch");
+                latent_cfg = numpy_utils::repeat(latent_cfg, request_input_batch / latent_cfg.get_shape()[0]);
+            }
+
+            const float t = timesteps[inference_step];
+            for (size_t b = 0; b < B_ts; ++b) {
+                std::fill_n(timestep_data + b * video_sequence_length + tokens_per_frame,
+                            video_sequence_length - tokens_per_frame, t);
+            }
```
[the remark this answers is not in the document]
goyaladitya05: The batch stays consistent. timestep is allocated with the same batch as latent_cfg, and batch_size_multiplier is std::max(requested, m_reshape_batch_size_multiplier, m_compiled_batch_size_multiplier), so latent_cfg's batch is always ≥ request_input_batch. Just to be safe, I've added an explicit check after the repeat in both the image-to-video and text-to-video loops, so any future change that breaks the assumption fails here instead of as a shape error inside infer().

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -502,6 +632,250 @@ class Text2VideoPipeline::LTXPipeline {
...
+
+            // Use TaylorSeer if enabled and caching is appropriate
+            ov::Tensor noise_pred_tensor;
+            if (ts_state.is_active() && !ts_state.should_compute(inference_step)) {
+                noise_pred_tensor = ts_state.predict(inference_step);
+            } else {
+                auto infer_start = std::chrono::steady_clock::now();
+                noise_pred_tensor = m_transformer->infer(latent_cfg, timestep);
```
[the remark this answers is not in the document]
goyaladitya05: same as above

samples/cpp/video_generation/CMakeLists.txt
```diff
@@ -11,6 +11,9 @@ find_package(OpenVINOGenAI REQUIRED
 file(DOWNLOAD https://raw.githubusercontent.com/nothings/stb/f75e8d1cad7d90d72ef7a4661f1b994ef78b4e31/stb_image_write.h ${CMAKE_BINARY_DIR}/stb_image_write.h
      EXPECTED_HASH MD5=845b8b43d7d941890a57a477455558ad)
 
+file(DOWNLOAD https://raw.githubusercontent.com/nothings/stb/f75e8d1cad7d90d72ef7a4661f1b994ef78b4e31/stb_image.h ${CMAKE_BINARY_DIR}/stb_image.h
+     EXPECTED_HASH MD5=27932e6fb3a2f26aee2fc33f2cb4e696)
```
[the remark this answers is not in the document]
goyaladitya05: it follows the existing pattern.

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
```diff
@@ -55,6 +57,7 @@ class OPENVINO_GENAI_EXPORTS LTXVideoTransformer3DModel {
 
     size_t get_expected_batch_size() const;
     size_t get_request_input_batch();
+    ov::PartialShape get_timestep_partial_shape();
```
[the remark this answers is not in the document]
goyaladitya05: It can't be const, the method calls m_request.get_compiled_model(), and ov::InferRequest::get_compiled_model() is non-const in the OpenVINO API (there's no const overload).

src/cpp/src/video_generation/models/autoencoder_kl_ltx_video.cpp
```diff
@@ -132,18 +210,46 @@ AutoencoderKLLTXVideo& AutoencoderKLLTXVideo::compile(const std::string& device,
...
+
+    if (m_encoder_model) {
+        ov::PartialShape input_shape = m_encoder_model->input(0).get_partial_shape();
+        OPENVINO_ASSERT(input_shape.rank().is_static() && input_shape.rank().get_length() == 5,
+            "AutoencoderKLLTXVideo encoder input must be rank 5 [B, C, F, H, W], got rank ",
+            input_shape.rank());
+        // The encoder always encodes a single conditioning frame; shape it to 1 regardless of num_frames.
+        std::map<size_t, ov::PartialShape> idx_to_shape{{0, {batch_size, input_shape[1], 1, height, width}}};
+        m_encoder_model->reshape(idx_to_shape);
+    }
```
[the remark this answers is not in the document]
goyaladitya05: this is intentional. LTX-Video image-to-video conditions on a single frame, so the encoder is reshaped to F=1 by design. it doesn't silently drop extra frames, the encoder input is reshaped to F=1, so passing F>1 fails with a shape error rather than encoding partial data.

goyaladitya05:
@as-suvorov @sgonorov could you please trigger the builds?

tests/python_tests/samples/test_image2video.py
```diff
@@ -0,0 +1,31 @@
...
+
+class TestImage2Video:
+    PROMPT = "A golden retriever in a sunlit bedroom slowly stands up and turns its head to look at the camera"
+
+    @pytest.mark.samples
+    @pytest.mark.video_generation
+    @pytest.mark.parametrize(
+        "convert_model, sample_args",
+        [
+            pytest.param("tiny-random-ltx-video", PROMPT),
```
sgonorov, line 19: `samples/conftest.py` exports this id with only `--trust-remote-code` and no optimum-master mechanism, so the rank-2 timestep export never happens and this test can't pass. I'd add a separate model id.

samples/python/video_generation/README.md
```diff
@@ -21,14 +22,22 @@ Install [../../export-requirements.txt](../../export-requirements.txt) if model
...
 
 ```sh
 optimum-cli export openvino --model Lightricks/LTX-Video --task text-to-video --weight-format fp32 ltx_video_ov/FP32
 ```
 
 > **Note:** For basic video generation without LoRA, `--weight-format int8` produces a smaller model.
 
+### For Image-to-Video
+
+The image-to-video pipeline requires a VAE encoder in addition to the standard model files. Export with `--task image-to-video`:
```
sgonorov: For diffusers the class comes from `model_index.json` `_class_name`, so `--task image-to-video` changes nothing in the LTX export. I think citing a minimum optimum-intel version works better.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -328,8 +351,74 @@ class Text2VideoPipeline::LTXPipeline {
...
+
+    ov::Tensor preprocess_and_encode_image(const ov::Tensor& image, const VideoGenerationConfig& config) {
+        OPENVINO_ASSERT(m_pipeline_type == VideoPipelineType::IMAGE_2_VIDEO,
+                        "Image2VideoPipeline requires a VAE encoder. "
+                        "Ensure 'vae_encoder' exists in the model directory.");
+
+        ov::Tensor img = image;
+        if (img.get_shape().size() == 3) {
+            const auto s = img.get_shape();
+            img.set_shape({1, s[0], s[1], s[2]});
```
sgonorov: `ov::Tensor img = image;` shares the impl, so `set_shape` promotes a caller's rank-3 tensor in place, and our own `read_image` returns rank-3. Might be worth constructing a view.

tests/python_tests/test_video_generation.py
```diff
@@ -433,3 +541,53 @@ def test_transformer_has_set_adapters_method(self, video_generation_model):
...
+class TestImage2VideoPipeline:
+    GENERATE_KWARGS = dict(height=32, width=32, num_frames=9, num_inference_steps=2)
+
+    @pytest.fixture(autouse=True)
+    def require_encoder(self, video_generation_model):
+        if not Path(video_generation_model, "vae_encoder").exists():
+            pytest.skip("vae_encoder not present in test model")
+
+    def _make_image(self, height=32, width=32):
+        return ov.Tensor(np.zeros([1, height, width, 3], dtype=np.uint8))
```
sgonorov: The conditioning image is uniform zeros and frame 0 is never compared against it, so these tests dont fail if the anchoring is deleted. I'd use a structured image.

tests/python_tests/test_video_generation.py
```diff
@@ -21,15 +28,37 @@
 
 @pytest.fixture(scope="module")
 def video_generation_model() -> str:
+    use_optimum_master = MODEL_ID in MODELS_REQUIRING_OPTIMUM_MASTER
+
     models_dir = get_ov_cache_converted_models_dir()
     model_path = Path(models_dir) / MODEL_ID / MODEL_NAME
+    # Avoid colliding with a stale pinned (rank-1 timestep) export at the same path.
+    if use_optimum_master:
+        model_path = model_path.parent / f"{model_path.name}-optimum-master"
```
sgonorov: `test_config_validate_guidance_scale_with_negative_prompt` still builds the unsuffixed path, so on a clean cache it skips forever. We could have it take the `video_generation_model` fixture.

tests/python_tests/conftest.py
```diff
@@ -84,7 +84,7 @@ def pytest_configure(config: pytest.Config):
 
 
 OPTIMUM_INTEL_MASTER = "optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@main"
-MODELS_REQUIRING_OPTIMUM_MASTER = {"tiny-random-flux.2-klein"}
+MODELS_REQUIRING_OPTIMUM_MASTER = {"tiny-random-flux.2-klein", "tiny-random-ltx-video"}
```
sgonorov: Every LTX export in CI is rank-2 now, so the rank-1 fallback at `ltx_pipeline.hpp:1001` is unreachable and will rot. Probably better to drop it or cover it.

site/docs/use-cases/video-generation/_sections/_run_model/_image2video_cpp.mdx
```diff
@@ -0,0 +1,20 @@
...
+int main(int argc, char* argv[]) {
+    const std::string models_path = argv[1];
+    const std::string image_path = argv[2], prompt = argv[3];
+
+    ov::Tensor image = utils::load_image(image_path);
+
+    ov::genai::Image2VideoPipeline pipe(models_path, "${props.device || 'CPU'}");
+    ov::Tensor video = pipe.generate(image, prompt).video;
+
+    imwrite_video("genai_video.avi", video);
```
sgonorov: There's no `imwrite_video` symbol, the header declares `save_video`. The text2video doc has the same bug, so we might as well fix both.

src/cpp/include/openvino/genai/video_generation/image2video_pipeline.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+    VideoGenerationPerfMetrics get_performance_metrics();
+
+    ~Image2VideoPipeline();
+
+private:
+    class Impl;
+    std::unique_ptr<Impl> m_impl;
+    explicit Image2VideoPipeline(std::unique_ptr<Impl> impl);
+    friend class Text2VideoPipeline;
```
sgonorov: This friend declaration looks dead, there's no conversion between teh two pipelines here or on master. I'd just drop it.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -119,15 +134,25 @@ size_t LTXVideoTransformer3DModel::get_expected_batch_size() const {
...
+    if (m_request) {
+        ov::CompiledModel compiled = m_request.get_compiled_model();
+        for (const auto& input : compiled.inputs()) {
+            if (input.get_any_name() == "timestep") {
+                return input.get_partial_shape();
+            }
+        }
     }
-    return shape[0];
+    OPENVINO_ASSERT(false,
```
sgonorov: Non-void `get_timestep_partial_shape()` has no `return` after the assert, which trips `-Wreturn-type`. This came up before and got closed without a change. We could return `{}`.

sgonorov (changes requested):

src/cpp/include/openvino/genai/video_generation/image2video_pipeline.hpp
goyaladitya05: removed

site/docs/use-cases/video-generation/_sections/_run_model/_image2video_cpp.mdx
goyaladitya05: changed both.

tests/python_tests/samples/test_image2video.py
goyaladitya05: added a seperated model id.

samples/python/video_generation/README.md
goyaladitya05: Since its not a part of any release yet, I've cited installing from the main branch.

tests/python_tests/conftest.py
goyaladitya05: Yes, while that is true, my intention is to support legacy exports as well, for Text2Video. Otherwise this instantly becomes a breaking change.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
goyaladitya05: Removed it, it's not needed now. The shape is cached during compile(), so the getter is a plain const return with no fallthrough. The "timestep input not found" check moved into compile(), so a bad export now fails at compile time instead of at first generate.

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: changed, now it wraps the same memory in a rank-4 view instead, same as autoencoder_kl.cpp:35.

tests/python_tests/test_video_generation.py
goyaladitya05: done, switched to a structured image.

tests/python_tests/test_video_generation.py
goyaladitya05: Fixed, it takes the video_generation_model fixture now.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -328,8 +351,76 @@ class Text2VideoPipeline::LTXPipeline {
...
+        OPENVINO_ASSERT(m_pipeline_type == VideoPipelineType::IMAGE_2_VIDEO,
+                        "Image2VideoPipeline requires a VAE encoder. "
+                        "Ensure 'vae_encoder' exists in the model directory.");
+
+        // ov::Tensor copies share the underlying memory, so set_shape() here would promote the
+        // caller's rank-3 tensor in place. Wrap the same memory in a rank-4 view instead.
+        ov::Tensor img = image;
+        if (image.get_shape().size() == 3) {
+            const auto s = image.get_shape();
+            img = ov::Tensor(image.get_element_type(), ov::Shape{1, s[0], s[1], s[2]}, image.data());
```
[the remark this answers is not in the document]
goyaladitya05: image is a const ov::Tensor&, so image.data() resolves to the const void* data() const overload, which binds to the Tensor(const element::Type&, const Shape&, const void* host_ptr, ...) constructor. both exist in `openvino/runtime/tensor.hpp`. Using img.data() instead would take the non-const overload and expose a mutable pointer into the caller's buffer, which is the issue this change was meant to fix.

goyaladitya05:
@sgonorov could you please take a look again? do you propose any further changes?

likholat:
@goyaladitya05 is there a particular reason why this PR is marked as a draft?

goyaladitya05:
> @goyaladitya05 is there a particular reason why this PR is marked as a draft?

I had marked it as draft because I was doing WWB verification after the changes in 0b41a24. I have completed the validation, it works same as before.

tests/python_tests/test_video_generation.py
```diff
@@ -433,3 +535,54 @@ def test_transformer_has_set_adapters_method(self, video_generation_model):
...
+        result1 = pipe.generate(image, "test prompt", **self.GENERATE_KWARGS, generator=ov_genai.CppStdGenerator(42))
+        result2 = pipe.generate(image, "test prompt", **self.GENERATE_KWARGS, generator=ov_genai.CppStdGenerator(42))
+        np.testing.assert_array_equal(result1.video.data, result2.video.data)
+
+    def test_lora_passthrough(self, video_generation_model):
+        adapter_config = ov_genai.AdapterConfig()
+        pipe = ov_genai.Image2VideoPipeline(video_generation_model, "CPU")
+        image = self._make_image()
+        result = pipe.generate(image, "test prompt", **self.GENERATE_KWARGS, adapters=adapter_config)
+        assert result.video is not None
```
likholat, line 609: Please add a test for `reshape()` method of Image2Video Pipeline

goyaladitya05: added

sgonorov (approved):

samples/python/video_generation/README.md
```diff
@@ -21,14 +22,23 @@ Install [../../export-requirements.txt](../../export-requirements.txt) if model
...
 
 > **Note:** For basic video generation without LoRA, `--weight-format int8` produces a smaller model.
 
+### For Image-to-Video
+
+The image-to-video pipeline requires a model exported with optimum-intel from the `main` branch:
+
+```sh
+pip install "optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@main"
+optimum-cli export openvino --model Lightricks/LTX-Video --weight-format fp32 ltx_video_ov_i2v/FP32
```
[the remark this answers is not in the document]
goyaladitya05: the vae encoder is exported regardless of the task.

samples/cpp/video_generation/README.md
```diff
@@ -21,14 +22,23 @@ Install [../../export-requirements.txt](../../export-requirements.txt) if model
...
 
 > **Note:** `--weight-format fp32` is required for LoRA adapter support. For basic video generation without LoRA, `--weight-format int8` produces a smaller model.
 
+### For Image-to-Video
+
+The image-to-video pipeline requires a model exported with optimum-intel from the `main` branch:
+
+```sh
+pip install "optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@main"
+optimum-cli export openvino --model Lightricks/LTX-Video --weight-format fp32 ltx_video_ov_i2v/FP32
```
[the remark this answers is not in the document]
goyaladitya05: the vae encoder is exported regardless of the task.

tests/python_tests/samples/test_image2video.py
```diff
@@ -0,0 +1,31 @@
...
+        ],
+        indirect=["convert_model"],
+    )
+    @pytest.mark.parametrize("download_test_content", ["overture-creations.png"], indirect=True)
+    def test_sample_image2video(self, convert_model, sample_args, download_test_content):
+        py_script = SAMPLES_PY_DIR / "video_generation/image2video.py"
+        py_command = [sys.executable, py_script, convert_model, download_test_content, sample_args]
+        run_sample(py_command)
```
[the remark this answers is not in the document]
goyaladitya05: reduced the num_frames similar to t2v.

goyaladitya05:
@likholat I pushed some changes in [5b5e698](https://github.com/openvinotoolkit/openvino.genai/pull/3907/commits/5b5e6981e37154e19e74ea35f6f90905de17f8c9) to fix CI, could you please run the tests again?

likholat:
build_jenkins

PR #3932, opened 2026-06-03
add dtor for ThreadedCallbackWrapper

pavel-esir:
## Description
Fix process crash when `generate()` throws an exception while a step callback is active.

CVS-187609

## Root Cause

`ThreadedCallbackWrapper` spawns a background worker thread when a `callback` is passed to
`generate()`. If an exception is thrown mid-generation (e.g. wrong tensor element type,
unsupported model configuration), stack unwinding destroys the `shared_ptr<ThreadedCallbackWrapper>`
without ever calling `end()`. The background thread remains blocked on `m_squeue.pull()`, and
destroying a joinable `std::thread` calls `std::terminate()` — causing an immediate process crash
(exit code `0xC0000409` / `-1073740791` on Windows).

This affects all four image generation pipelines that use `ThreadedCallbackWrapper`:
`StableDiffusionPipeline`, `StableDiffusion3Pipeline`, `FluxPipeline`, and `FluxFillPipeline`.

## Checklist:
- [ ] I have made corresponding changes to the documentation. - Not needed


as-suvorov (commented):
Please implement dtor for CB streamer wrapper as well: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching/threaded_streamer.hpp

pavel-esir:
> Please implement dtor for CB streamer wrapper as well: https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/continuous_batching/threaded_streamer.hpp

done

as-suvorov (approved):

PR #3943, opened 2026-06-05
ignore cache_interval_multiplier for non LM models

dtrawins:
## Description
Ignoring the value of cache_interval_multiplier for non LM models. It allows the application to control the default value regardless of the model. Genai library is setting now 8 which is consuming too much memory with prefix caching.
While the value is ignored, the application could set different settings regardless of the model architecture. 
Currently application would need to know if the model is LA but such info is not exposed without full pipeline loading.

CVS-188061

## Checklist:


popovaan (approved):

Wovchena (approved):

SearchSavior:
@dtrawins  Hello, thanks for the work debugging this.

What is expected memory savings on gpu for linear attention models?

PR #3974, opened 2026-06-11
WWB: add llama.cpp chat and n_ctx CLI support with validation and regression tests

rasapala:
## Description
This PR improves WWB llama.cpp text CLI behavior and adds regression coverage.

Changes:

Added llama.cpp text options:
[--llamacpp-chat](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for chat-completions formatting.
[--llamacpp-n-ctx](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for context window size.
Added validation/default handling:
[--llamacpp-chat](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) requires [--llamacpp](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
[--llamacpp-n-ctx](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) requires [--llamacpp](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
With [--llamacpp](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html), default n_ctx=8192 is applied when [--llamacpp-n-ctx](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is not provided.
Non-llama.cpp runs are unaffected by llama.cpp defaults.
Forwarded [llamacpp_n_ctx](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) into llama.cpp pipeline loading.

Added lightweight CLI tests for the new validation and default paths.

Updated README to document [--llamacpp-chat](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [--llamacpp-n-ctx](vscode-file://vscode-app/c:/Users/rasapala/AppData/Local/Programs/Microsoft%20VS%20Code/41dd792b5e/resources/app/out/vs/code/electron-browser/workbench/workbench.html).

Fixed text evaluator comparison alignment for differing GT/prediction lengths.

Usage example:
wwb --target-model ~/models/Qwen3.6-35B-A3B-UD-Q4_K_M.gguf     --gt-data ~/models/reference_long_prompt_limit_5_max_new_tokens_4096.csv --model-type text     --llamacpp     --llamacpp-n-ctx 8192     --num-samples 5 --llamacpp-chat --output llama_chat --max_new_tokens 4096

Similarity on SPR CPU: 0.964
Similarity on older CPU: 0.9567192

## Checklist:


sstrehlk (approved):

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -612,20 +623,27 @@ def genai_gen_chat_text(
...
+    skip_question,
+    use_chat_template=False,
+    empty_adapters=False,
+    num_assistant_tokens=0,
+    assistant_confidence_threshold=0.0,
+):
+    _ = tokenizer
+    _ = empty_adapters
+    _ = num_assistant_tokens
+    _ = assistant_confidence_threshold
```
sbalandi:
```suggestion
```

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -1231,15 +1255,28 @@ def main():
...
+        raise ValueError("--llamacpp-chat requires --llamacpp")
+
+    if args.llamacpp:
+        kwargs["llamacpp_n_ctx"] = 8192 if args.llamacpp_n_ctx is None else args.llamacpp_n_ctx
+    elif args.llamacpp_n_ctx is not None:
+        raise ValueError("--llamacpp-n-ctx requires --llamacpp")
+
     if args.base_model is not None:
+        base_model_kwargs = dict(kwargs)
+        if not (args.model_type in ["text", "text-chat"] and args.llamacpp and not args.hf and not args.genai):
```
sbalandi: Why this condition is needed and why for base model only ?  
llamacpp_n_ctx is allowed to set/is set by tool only with args.llamacpp and model loader functions should validate kwargs inside, so why llamacpp_n_ctx should be removed

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,220 @@
...
+            "--model-type",
+            "text",
+        ],
+        "Ground-truth data file",
+    )
+
+    assert "--llamacpp-n-ctx requires --llamacpp" not in output
+
+
+def test_text_base_model_load_passes_llamacpp_flag(monkeypatch):
```
sbalandi: I would suggest to add wwb test with some small real model via llama_cpp and check output instead of checking args. output sould contains string: `Using llama.cpp API` (logging here https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/model_loaders.py#L237). Also you can add llamacpp_n_ctx value to logging and check it here.

rasapala: Added _get_tiny_llamacpp_model and tests.

tools/who_what_benchmark/whowhatbench/wwb.py
```diff
@@ -1239,6 +1271,7 @@ def main():
             args.ov_config,
             args.hf,
             args.genai,
+            args.llamacpp,
```
sbalandi, line 1399: Don't this args should be added to target model loading as well ?

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,449 @@
...
+        relevance_weight=None,
+        max_new_tokens=128,
+    )
+
+    monkeypatch.setattr(wwb, "parse_args", lambda: args)
+    monkeypatch.setattr(wwb, "load_model", lambda *a, **k: object())
+    monkeypatch.setattr(
+        wwb, "create_evaluator", lambda *_: type("DummyEvaluator", (), {"dump_gt": lambda self, _: None})()
+    )
+    with pytest.raises(ValueError, match="--llamacpp is supported only with --model-type text"):
```
sbalandi: why do you need all arguments and monkeypatch ? why don's just call run_wwb with --llamacpp and wrong --model-type ?

sbalandi: the question for all tests with argument testing

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,449 @@
...
+    captured = {}
+
+    def fake_load_text_llamacpp_pipeline(model_dir, **kwargs):
+        captured["model_dir"] = model_dir
+        captured["kwargs"] = kwargs
+        return object()
+
+    monkeypatch.setattr(model_loaders, "load_text_llamacpp_pipeline", fake_load_text_llamacpp_pipeline)
+
+    model_loaders.load_model(
```
sbalandi, line 122: Why load_model need to be tested ? Why it can't be tested via real model run scenario ?

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,449 @@
...
+        False,
+        False,
+        True,
+        llamacpp_n_ctx=None,
+    )
+
+    assert captured["kwargs"]["llamacpp_n_ctx"] == 8192
+
+
+def test_loader_sets_llamacpp_n_ctx_default_when_key_omitted(monkeypatch):
```
sbalandi: This and previous test are differ only in llamacpp_n_ctx parameter. It seems to me that testing input parameters  for model_loaders is too granular, but if we decided to keep tests, I think it better to parametrize tests.

tools/who_what_benchmark/whowhatbench/wwb.py
rasapala: Yes it was present before for target model line 1317.

rasapala (commented):
.

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
rasapala: Good point.

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
rasapala: because they validate argument-routing logic independently from network/model/runtime noise.

akashchi (approved):
GHA changes LGTM.

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,384 @@
...
+        max_new_tokens=128,
+    )
+
+    monkeypatch.setattr(wwb, "parse_args", lambda: args)
+
+    with pytest.raises(ValueError, match="Options --hf, --genai and --llamacpp are mutually exclusive"):
+        wwb.main()
+
+
+def test_main_csv_only_rejects_llamacpp_n_ctx_without_llamacpp(monkeypatch, tmp_path):
```
sbalandi: Is there any need to call the main() separately? it can be set several required options and call run_wwb, it will also call main

rasapala: The test needs to inject specific (e.g., llamacpp=False, llamacpp_n_ctx=4096, with pre-existing CSV files) via monkeypatch. There's no CLI flag combination that would produce that exact state cleanly — specifically, there's no way to pass llamacpp_n_ctx=4096 while also passing --llamacpp omitted through the CLI without triggering other errors first. So calling main() directly with a mocked parse_args is the right approach here.

sbalandi: And tests should fail on load_model, right ? the test `test_text_llamacpp_n_ctx_requires_llamacpp` is checking similar behavior. Why not just parametrize it and add llamacpp_chat=False ?

rasapala: On parametrization: you can parametrize within the same layer, but not across these two tests unless you also parametrize the call target (wwb.main vs model_loaders.load_model). Adding llamacpp_chat=False does not help this scenario, because that flag is unrelated to the llamacpp_n_ctx validation branch.

sbalandi: ok, then it is just the same test as `test_text_llamacpp_n_ctx_requires_llamacpp`

`--llamacpp-n-ctx requires --llamacpp` is raised in _sanitize_load_kwargs from load_model. load_model know just use_llamacpp and llamacpp_n_ctx, and know nothing about llamacpp_chat, so we don't need to check it. This test check combination llamacpp=False, llamacpp_chat=False, llamacpp_n_ctx=4096, without llamacpp_chat it's the same as use_llamacpp and llamacpp_n_ctx from test_text_llamacpp_n_ctx_requires_llamacpp.
and also if we can't pass something via CLI why we need to check it with CLI and main() ?

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,384 @@
...
+
+    wwb.main()
+
+    assert len(calls) == 1
+    load_args, load_kwargs = calls[0]
+    assert load_args[6] is True
+    assert load_kwargs["llamacpp_n_ctx"] is None
+
+
+def test_main_rejects_mutually_exclusive_backend_flags(monkeypatch):
```
sbalandi: Can the test be replaced to run ?
```
   run_wwb(
        [
            "--base-model",
            str(dummy_base_model),
            "--llamacpp",
            "--hf"
        ]
    )
```

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
```diff
@@ -0,0 +1,384 @@
...
+    monkeypatch.setattr(wwb, "parse_args", lambda: args)
+    monkeypatch.setattr(wwb, "load_model", fake_load_model)
+    monkeypatch.setattr(wwb, "create_evaluator", lambda *_: DummyEvaluator())
+
+    wwb.main()
+
+    assert len(calls) == 1
+    load_args, load_kwargs = calls[0]
+    assert load_args[6] is True
+    assert load_kwargs["llamacpp_n_ctx"] is None
```
sbalandi: Since we have test_text_llamacpp_real_gguf_logs_backend_and_n_ctx I would suggest to delete this , because in real scenarios llamacpp_n_ctx can't be None in after load_model, and usign llamacpp is also checking.

sbalandi: I meant to delete the test completely.

tools/who_what_benchmark/tests/test_cli_text_llamacpp_args.py
rasapala: I can remove the test if you want to have lower code coverage and less tests.

sbalandi: @rasapala could you please show example of how a user might get into the situation covered by this test ?

sbalandi: I wouldn't want to keep in the repository tests that don't have a meaningful purpose

sbalandi (approved):

PR #4002, opened 2026-06-22
Fix heap buffer overflow in GGUF Q4_K/Q6_K dequantizer

TianmengChen:
Related JIRA: [CVS-188918](https://jira.devtools.intel.com/browse/CVS-188918)

A crafted .gguf file can trigger an attacker-controlled heap out-of-bounds write in the K-quant dequantizers. gguf_load_quantized() sizes the output weights/scales/biases buffers from shape.back(), guarded only by OPENVINO_ASSERT(shape.back() % weights_per_block == 0) where weights_per_block is the sub-block size (32 for Q4_K, 16 for Q6_K). extract_q4_k_data()/extract_q6_k_data(), however, iterate tensor.bsize / bytes_per_block full 256-weight super-blocks and write a whole super-block each (128 bytes via unpack_256_4() for Q4_K, 256 weights for Q6_K).

For a K-quant tensor whose last dim is a multiple of the sub-block size but not of the 256-weight super-block (e.g. 32), gguflib pads the data up to a full super-block, so the loop writes past the end of the smaller output buffer: a 112-byte (Q4_K) / 240-byte (Q6_K) heap overflow whose contents are read straight from the file and are thus fully attacker-controlled (CWE-787). Reachable pre-auth at model-load time in the default build via LLMPipeline(path) / Tokenizer(path) / read_model(path).

Require whole 256-weight super-blocks for K-quants in gguf_load_quantized() before dequantizing, so the write loop can never exceed the allocated output buffers.


as-suvorov (approved):

as-suvorov:
@TianmengChen Can we add a test?

as-suvorov:
@TianmengChen Could you please fix linting in PR? This is the only blocking point

PR #4016, opened 2026-06-23
[wwb] Force dtype float32 for models loaded with diffusers

sbalandi:
## Description
transformers load model in "auto" model type depending on the data type in HuggingFace, diffusers is still force torch.float32 for models. As a result, we have text_encoder(s) in torch.float16, and a transformer/vae in torch.float32. When several models in pipeline has different type it leads to error like this:
RuntimeError: mat1 and mat2 must have the same dtype, but got BFloat16 and Float

Let's cast all models to one type.

[CVS-189083](https://jira.devtools.intel.com/browse/CVS-189083)
[CVS-189081](https://jira.devtools.intel.com/browse/CVS-189081)

## Checklist:

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/model_loaders.py
+++ b/tools/who_what_benchmark/whowhatbench/model_loaders.py
@@ -301,9 +301,9 @@ def load_text2image_model(
 
         logger.info("Using HF Transformers API")
         try:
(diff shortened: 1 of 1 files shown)


sbalandi:
@likholat could you, please, take a look ? is it ok use model type float32 or it's better to cast to float16 ?

likholat (approved):

PR #4020, opened 2026-06-25
llm_bench: Add initial support for Kokoro TTS model

RyanMetcalfeInt8:
## Description

CVS-188526

## Checklist:
- [X] Tests have been updated or added to cover the new code. 
- [X] This PR fully addresses the ticket. 
- [N/A] I have made corresponding changes to the documentation.

### DIFF
--- a/tests/python_tests/samples/test_tools_llm_benchmark.py
+++ b/tests/python_tests/samples/test_tools_llm_benchmark.py
@@ -9,6 +9,9 @@ from pathlib import Path
 from test_utils import run_sample
 from data.models import GGUF_MODEL_LIST
 from utils.hugging_face import download_gguf_model
+from utils.constants import get_ov_cache_converted_models_dir
+from utils.kokoro_test_assets import prepare_tiny_kokoro_model_path
+from utils.kokoro_test_assets import prepare_tiny_kokoro_ov_path
 from conftest import SAMPLES_PY_DIR, convert_model, download_test_content
 
 convert_draft_model = convert_model
@@ -57,6 +60,21 @@ video_generation_json = [
         "frame_rate": 25,
     }
 ]
+
+
+@pytest.fixture(scope="module")
+def tiny_kokoro_ov_path() -> Path:
+    converted_models_dir = get_ov_cache_converted_models_dir()
+    tiny_kokoro_model_path = prepare_tiny_kokoro_model_path(converted_models_dir)
+    return prepare_tiny_kokoro_ov_path(converted_models_dir, tiny_kokoro_model_path)
+
+
+@pytest.fixture(scope="module")
+def tiny_kokoro_speaker_embedding_file_path(tiny_kokoro_ov_path: Path) -> str:
+    voice_bin_path = tiny_kokoro_ov_path / "voices" / "tiny_voice.bin"
+    if not voice_bin_path.exists():
+        raise FileNotFoundError(f"Missing tiny Kokoro speaker embedding file at {voice_bin_path}")
+    return str(voice_bin_path)
 
 
 class TestBenchmarkLLM:
@@ -261,6 +279,69 @@ class TestBenchmarkLLM:
             "-m", convert_model,
             "--speaker_embeddings", download_test_content
         ] + sample_args
+        run_sample(benchmark_py_command)
+
+    @pytest.mark.samples
+    @pytest.mark.speech_generation
+    @pytest.mark.parametrize(
+        "sample_args, use_explicit_speaker_embedding",
+        [
+            (
+                [
+                    "-d",
+                    "cpu",
+                    "-n",
+                    "1",
+                    "--task",
+                    "text_to_speech",
+                    "--genai",
+                    "--speech_language",
+                    "en-us",
+                    "--prompt",
+                    "Why is the Sun yellow?",
+                ],
+                True,
+            ),
+            (
+                [
+                    "-d",
+                    "cpu",
+                    "-n",
+                    "1",
+                    "--task",
+                    "text_to_speech",
+                    "--genai",
+                    "--speech_voice",
+                    "tiny_voice",
+                    "--speech_language",
+                    "en-us",
+                    "--prompt",
+                    "Why is the Sun yellow?",
+                ],
+                False,
+            ),
+        ],
+    )
+    def test_python_tool_llm_benchmark_tts_kokoro_genai(
+        self,
+        tiny_kokoro_ov_path: Path,
+        tiny_kokoro_speaker_embedding_file_path: str,
+        sample_args,
+        use_explicit_speaker_embedding,
+    ):
+        benchmark_script = SAMPLES_PY_DIR / "llm_bench/benchmark.py"
+        benchmark_py_command = [
+            sys.executable,
+            benchmark_script,
+            "-m",
+            str(tiny_kokoro_ov_path),
+        ]
+        if use_explicit_speaker_embedding:
+            benchmark_py_command += [
+                "--speaker_embeddings",
+                tiny_kokoro_speaker_embedding_file_path,
+            ]
+        benchmark_py_command += sample_args
         run_sample(benchmark_py_command)
 
 
--- a/tests/python_tests/samples/test_tools_llm_benchmark.py
+++ b/tests/python_tests/samples/test_tools_llm_benchmark.py
@@ -320,9 +320,27 @@ class TestBenchmarkLLM:
                 ],
                 False,
             ),
+            (
+                [
+                    "-d",
+                    "cpu",
+                    "-n",
+                    "1",
+                    "--task",
+                    "text_to_speech",
+                    "--optimum",
+                    "--speech_voice",
+                    "tiny_voice",
+                    "--speech_language",
+                    "en-us",
+                    "--prompt",
+                    "Why is the Sun yellow?",
+                ],
+                False,
+            ),
         ],
     )
-    def test_python_tool_llm_benchmark_tts_kokoro_genai(
+    def test_python_tool_llm_benchmark_tts_kokoro(
         self,
         tiny_kokoro_ov_path: Path,
         tiny_kokoro_speaker_embedding_file_path: str,
--- a/tools/llm_bench/doc/PROMPT.md
+++ b/tools/llm_bench/doc/PROMPT.md
@@ -71,8 +71,6 @@ Prompt file example：
 ## 8. Text To Speech Models
 Supported parameters that can be set are:
 * `prompt` - input prompt text for the speech generation
-* `speech_voice` - optional voice for Kokoro text to speech (default `af_heart`)
-* `speech_language` - optional language for Kokoro text to speech (`en-us`, `en-gb`, `es`, `fr-fr`, `hi`, `it`, `pt-br`, `ja`, `zh`)
 Prompt file example：
 {"prompt": "what is openvino?"}
 {"prompt": "A chat between a curious user and an artificial intelligence assistant."}
--- a/tools/llm_bench/doc/PROMPT.md
+++ b/tools/llm_bench/doc/PROMPT.md
@@ -71,6 +71,8 @@ Prompt file example：
 ## 8. Text To Speech Models
 Supported parameters that can be set are:
 * `prompt` - input prompt text for the speech generation
+* `speech_voice` - optional voice for Kokoro text to speech (default `af_heart`)
+* `speech_language` - optional language for Kokoro text to speech (`en-us`, `en-gb`, `es`, `fr-fr`, `hi`, `it`, `pt-br`, `ja`, `zh`)
 Prompt file example：
 {"prompt": "what is openvino?"}
 {"prompt": "A chat between a curious user and an artificial intelligence assistant."}
--- a/tools/llm_bench/llm_bench_utils/config_class.py
+++ b/tools/llm_bench/llm_bench_utils/config_class.py
@@ -248,7 +248,7 @@ USE_CASES = {
     "ldm_super_resolution": [UseCaseLDMSuperResolution(["ldm-super-resolution"])],
     "text_embed": [UseCaseTextEmbeddings(["qwen3", "bge", "bert", "albert", "roberta", "xlm-roberta"])],
     "text_rerank": [UseCaseTextReranker(["qwen3", "bge", "bert", "albert", "roberta", "xlm-roberta"])],
-    "text_to_speech": [UseCaseTextToSpeech(["speecht5"])],
+    "text_to_speech": [UseCaseTextToSpeech(["speecht5", "kokoro"])],
 }
 
 PA_ATTENTION_BACKEND = "PA"
--- a/tools/llm_bench/llm_bench_utils/get_use_case.py
+++ b/tools/llm_bench/llm_bench_utils/get_use_case.py
@@ -51,10 +51,7 @@ def safe_json_load(file_path: Path) -> Optional[dict]:
 
 
 def resolve_complex_model_types(config):
-    model_type_raw = config.get("model_type")
-    if not isinstance(model_type_raw, str) or model_type_raw == "":
-        return None, None
-    model_type = model_type_raw.lower().replace("_", "-")
+    model_type = config.get("model_type").lower().replace('_', '-')
     if model_type == "gemma3":
         return USE_CASES["visual_text_gen"][0], model_type
     if model_type == "gemma3-text":
--- a/tools/llm_bench/llm_bench_utils/get_use_case.py
+++ b/tools/llm_bench/llm_bench_utils/get_use_case.py
@@ -51,7 +51,10 @@ def safe_json_load(file_path: Path) -> Optional[dict]:
 
 
 def resolve_complex_model_types(config):
-    model_type = config.get("model_type").lower().replace('_', '-')
+    model_type_raw = config.get("model_type")
+    if not isinstance(model_type_raw, str) or model_type_raw == "":
+        return None, None
+    model_type = model_type_raw.lower().replace('_', '-')
     if model_type == "gemma3":
         return USE_CASES["visual_text_gen"][0], model_type
     if model_type == "gemma3-text":
--- a/tools/llm_bench/llm_bench_utils/get_use_case.py
+++ b/tools/llm_bench/llm_bench_utils/get_use_case.py
@@ -54,7 +54,7 @@ def resolve_complex_model_types(config):
     model_type_raw = config.get("model_type")
     if not isinstance(model_type_raw, str) or model_type_raw == "":
         return None, None
-    model_type = model_type_raw.lower().replace('_', '-')
+    model_type = model_type_raw.lower().replace("_", "-")
     if model_type == "gemma3":
         return USE_CASES["visual_text_gen"][0], model_type
     if model_type == "gemma3-text":
--- a/tools/llm_bench/llm_bench_utils/ov_utils.py
+++ b/tools/llm_bench/llm_bench_utils/ov_utils.py
@@ -12,12 +12,13 @@ import copy
 import types
 from llm_bench_utils.hook_common import get_bench_hook
 from llm_bench_utils.hook_forward import MeanStdPair, RawImGenPerfMetrics
-from llm_bench_utils.model_utils import get_version_in_format_to_pars, resolve_kokoro_speaker_embedding
+from llm_bench_utils.model_utils import get_version_in_format_to_pars
 from llm_bench_utils.config_class import UseCaseSpeech2Text, UseCaseTextGen, UseCaseTextReranker, PA_ATTENTION_BACKEND
 from llm_bench_utils.tts_utils import (
     is_kokoro_model_id,
     normalize_kokoro_lang_code,
     DEFAULT_KOKORO_VOICE,
+    resolve_kokoro_speaker_embedding,
 )
 from transformers import pipeline
 import queue
--- a/tools/llm_bench/llm_bench_utils/ov_utils.py
+++ b/tools/llm_bench/llm_bench_utils/ov_utils.py
@@ -12,9 +12,14 @@ import copy
 import types
 from llm_bench_utils.hook_common import get_bench_hook
 from llm_bench_utils.hook_forward import MeanStdPair, RawImGenPerfMetrics
-from llm_bench_utils.model_utils import get_version_in_format_to_pars
+from llm_bench_utils.model_utils import get_version_in_format_to_pars, get_speaker_embeddings
 from llm_bench_utils.config_class import UseCaseSpeech2Text, UseCaseTextGen, UseCaseTextReranker, PA_ATTENTION_BACKEND
-from llm_bench_utils.tts_utils import is_kokoro_model_id, normalize_kokoro_lang_code, DEFAULT_KOKORO_VOICE
+from llm_bench_utils.tts_utils import (
+    is_kokoro_model_id,
+    normalize_kokoro_lang_code,
+    DEFAULT_KOKORO_VOICE,
+    KOKORO_SPEAKER_EMB_SHAPE,
+)
 from transformers import pipeline
 import queue
 from transformers.generation.streamers import BaseStreamer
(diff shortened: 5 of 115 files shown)


tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -19,6 +21,50 @@
...
+    speaker_embeddings = args.get('speaker_embeddings')
+    speech_language = args.get('speech_language', '')
+    speech_voice = args.get('speech_voice', '')
+
+    if use_genai:
+        # Note: For GenAI Kokoro model, the speaker embedding is passed here, even if user specified --speech_voice.
+        generation_result = model.generate(
+            input_text,
+            speaker_embedding=ov.Tensor(speaker_embeddings.numpy()),
+        )
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -158,6 +239,19 @@ def run_text_to_speech_generation_genai(
...
     mem_consumption.update_marker("model")
     model, processor, vocoder, pretrain_time, use_genai = FW_UTILS[framework].create_text_2_speech_model(model_path, device, mem_consumption, **args)
+    # For Kokoro with GenAI, the pipeline requires speaker_embedding as an ov.Tensor.
+    # Auto-resolve from the model's voices/ directory when not provided via CLI.
+    if args.get('is_kokoro_model') and use_genai and args.get('speaker_embeddings') is None:
+        from llm_bench_utils.model_utils import get_speaker_embeddings
+        voice = args.get('speech_voice', '').strip() or DEFAULT_KOKORO_VOICE
+        voice_file = model_path / 'voices' / f'{voice}.bin'
+        if not voice_file.exists():
+            raise RuntimeError(
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: This isn't an issue. For GenAI, it's expected that the user is passing path to optimum-exported path, which should have voices in .bin format, not .pt.

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -941,20 +958,62 @@ def create_text_2_speech_model(model_path, device, memory_data_collector, **kwar
...
+                self._model = model
+
+            def generate(self, prompt, speaker_embeddings=None, language="", voice=""):
+                preprocess_kwargs = {
+                    "text": prompt,
+                    "lang_code": normalize_kokoro_lang_code(language),
+                }
+                if speaker_embeddings is not None:
+                    preprocess_kwargs["speaker_embedding"] = speaker_embeddings.numpy()
+                else:
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

tools/llm_bench/llm_bench_utils/config_class.py
```diff
@@ -248,7 +248,7 @@ class UseCaseTextToSpeech(UseCase):
     "ldm_super_resolution": [UseCaseLDMSuperResolution(["ldm-super-resolution"])],
     "text_embed": [UseCaseTextEmbeddings(["qwen3", "bge", "bert", "albert", "roberta", "xlm-roberta"])],
     "text_rerank": [UseCaseTextReranker(["qwen3", "bge", "bert", "albert", "roberta", "xlm-roberta"])],
-    "text_to_speech": [UseCaseTextToSpeech(["speecht5"])],
+    "text_to_speech": [UseCaseTextToSpeech(["speecht5", "kokoro"])],
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: tests have been added.

tools/llm_bench/requirements.txt
```diff
@@ -11,7 +11,7 @@ transformers[sentencepiece]>=4.40.0,<5.4.0
 diffusers>=0.22.0,<=0.38.0
 # optimum is in dependency list of optimum-intel
 # last supported commit f50a0b326df82d4b05c461a9184a97f616bfa63a
-optimum-intel[nncf]>=1.25.0
+optimum-intel[nncf]>=2.0.0
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -941,20 +958,62 @@ def create_text_2_speech_model(model_path, device, memory_data_collector, **kwar
...
+                    ov_config=ov_config,
+                    trust_remote_code=True,
+                )
+            except Exception as exp:
+                raise RuntimeError(
+                    "Failed to initialize Optimum Kokoro pipeline. "
+                    "Kokoro with Optimum requires a recent optimum-intel build and compatible dependencies. "
+                    "Recommended setup:\n"
+                    "1) pip install --no-deps git+https://github.com/huggingface/optimum-intel.git\n"
+                    "2) pip install kokoro misaki[en] -c constraints-kokoro-optimum.txt\n"
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: fixed.

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -110,19 +183,35 @@ def run_text_to_speech_generation_genai(
...
+            speeches.append(_kokoro_generate_once(model, input_text, args, use_genai=True))
+        out_size = sum(speech.size for speech in speeches)
+        tokenization_time = [0]
+    else:
+        additional_args = (
+            {"speaker_embeddings": ov.Tensor(args["speaker_embeddings"].numpy())}
+            if args.get("speaker_embeddings") is not None
+            else {}
+        )
+        generation_result = model.generate(input_text_list, **additional_args)
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: hmm, I think this may be true but actually this isn't something that I changed. Even on current master, it sets as "speaker_embeddings" argument even though GenAI interface is:
```
def generate(self, text: str, speaker_embedding: typing.Any = None, **kwargs) -> Text2SpeechDecodedResults:
```

Let me look into it a little bit more. Probably yes it should be fixed.

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -158,6 +248,20 @@ def run_text_to_speech_generation_genai(
...
     mem_consumption.update_marker("model")
     model, processor, vocoder, pretrain_time, use_genai = FW_UTILS[framework].create_text_2_speech_model(model_path, device, mem_consumption, **args)
+    # For Kokoro with GenAI, the pipeline requires speaker_embedding as an ov.Tensor.
+    # Auto-resolve from the model's voices/ directory when not provided via CLI.
+    if args.get("is_kokoro_model") and use_genai and args.get("speaker_embeddings") is None:
+        from llm_bench_utils.model_utils import get_speaker_embeddings
+
+        voice = args.get("speech_voice", "").strip() or DEFAULT_KOKORO_VOICE
+        voice_file = model_path / "voices" / f"{voice}.bin"
+        if not voice_file.exists():
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: again, for GenAI path it's expected that `.bin` exists. Not a concern.

tools/llm_bench/task/text_to_speech_generation.py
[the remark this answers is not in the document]
RyanMetcalfeInt8: okay, yes, I ran some tests with SpeechT5 and confirmed that it's ignoring passed in speaker embedding, and always using built-in default. So, I've applied this fix.

tools/who_what_benchmark/setup.py
```diff
@@ -66,7 +66,7 @@ def set_version(base_version: str):
             "spacy==3.8.14",
             "kokoro==0.9.4",
             "transformers==4.57.6",
-            "optimum-intel[nncf] @ https://github.com/huggingface/optimum-intel/archive/a8c4734741e766ef95d7f1a7d1e29a1d4ba2ab8f.tar.gz",
+            "optimum-intel==2.0.0",
         ],
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: reverted.

tests/python_tests/samples/test_tools_llm_benchmark.py
```diff
@@ -263,6 +281,69 @@ def test_python_tool_llm_benchmark_tts(self, convert_model, download_test_conten
...
+                    "tiny_voice",
+                    "--speech_language",
+                    "en-us",
+                    "--prompt",
+                    "Why is the Sun yellow?",
+                ],
+                False,
+            ),
+        ],
+    )
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: optimum test has been added.

tools/llm_bench/llm_bench_utils/get_use_case.py
```diff
@@ -51,7 +51,10 @@ def safe_json_load(file_path: Path) -> Optional[dict]:
 
 
 def resolve_complex_model_types(config):
-    model_type = config.get("model_type").lower().replace('_', '-')
+    model_type_raw = config.get("model_type")
```
sbalandi: please, remove, this function will be removed in this pr https://github.com/openvinotoolkit/openvino.genai/pull/4063 , llm_bench moves away from the concept of automatic model type detection, the recommendation now is to always specify tasks

tools/llm_bench/doc/PROMPT.md
```diff
@@ -71,6 +71,8 @@ Prompt file example：
 ## 8. Text To Speech Models
 Supported parameters that can be set are:
 * `prompt` - input prompt text for the speech generation
+* `speech_voice` - optional voice for Kokoro text to speech (default `af_heart`)
+* `speech_language` - optional language for Kokoro text to speech (`en-us`, `en-gb`, `es`, `fr-fr`, `hi`, `it`, `pt-br`, `ja`, `zh`)
```
sbalandi: Can these parameters be set via json input file?

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -44,18 +104,30 @@ def run_text_to_speech_generation_optimum(
 
     mem_consumption.start(num)
     start = time.perf_counter()
-    if vocoder:
-        result = model.generate(input_tokens, speaker_embeddings=args.get('speaker_embeddings'), vocoder=vocoder)
+    speeches = []
+    if is_kokoro_model:
+        for _ in range(args["batch_size"]):
+            speeches.append(_kokoro_generate_once(model, input_text, args, use_genai=False))
+        out_size = sum(speech.size for speech in speeches)
```
sbalandi, line 72: if kokoro model can't work with batch, please, add warning and fallback to batch_size 1 as in vlm https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/task/visual_language_generation.py#L32
```
    if args['batch_size'] != 1:
        log.warning("Only batch size 1 available for benchmarking with kokoro model")
        args["batch_size"] = 1
```

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -28,14 +81,21 @@ def run_text_to_speech_generation_optimum(
...
-            llm_bench_utils.output_file.output_input_text(in_text, args, model_precision, prompt_index, bs_index, proc_id)
+            llm_bench_utils.output_file.output_input_text(
+                in_text, args, model_precision, prompt_index, bs_index, proc_id
+            )
+    is_kokoro_model = args.get("is_kokoro_model", False)
+    sample_rate = _get_tts_sample_rate(args)
     tok_encode_start = time.perf_counter()
-    input_data = processor(text=input_text_list, return_tensors='pt', padding=True, truncation=True)
+    if is_kokoro_model:
+        input_token_size = len(input_text.split())
```
sbalandi, line 48: please, set tok_encode_time to N/A for kokoro

tools/llm_bench/llm_bench_utils/pt_utils.py
```diff
@@ -182,22 +184,58 @@ def create_text_2_speech_model(model_path, device, memory_data_collector, **kwar
...
+
+                    def generate(self, prompt, speaker_embeddings=None, language="", voice=""):
+                        requested_lang_code = normalize_kokoro_lang_code(language)
+                        if requested_lang_code != self._lang_code:
+                            self._lang_code = requested_lang_code
+                            self._pipeline = KPipeline(lang_code=self._lang_code, model=self._kmodel)
+
+                        selected_voice = voice.strip() if isinstance(voice, str) else ""
+                        if not selected_voice:
+                            selected_voice = DEFAULT_KOKORO_VOICE
```
sbalandi, line 216: could you please move preprocessing from `generate` to separate function ? and call it separately before generate. For PyTorch model and for optimum-intel scenario. llm_bench would like to collect time for generate only on generation step

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -178,7 +292,7 @@ def run_text_2_speech_benchmark(model_path, framework, device, args, num_iters,
              f'prompt nums: {len(text_list)}, prompt idx: {prompt_idx_list}')
 
     tts_hook = None
-    if framework == "ov" and not use_genai:
+    if framework == "ov" and not use_genai and not args.get("is_kokoro_model", False):
```
sbalandi, line 263: Is there a way to check the inference time for a kokoro model?

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: So now that preprocessing (tokenization) is a separate call for optimum / pt case, it's properly timed / set for those. For GenAI case, it's not set (eventually in future PR I can populate metrics as SpeechT5 does).

tools/llm_bench/llm_bench_utils/pt_utils.py
RyanMetcalfeInt8: Yes, this is done. While it works, it wasn't very trivial, as PT routine essentially needs to mimic what optimum-intel does for `preprocess_inputs` -- this is why it expanded into more code than I anticipated.

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: hmm, I'd rather not implement a dedicated `TTSHooks` specifically for Kokoro. In the case of Kokoro, timing the call to `generate` should be timing back-to-back inference call(s), especially since preprocess (tokenization) has been moved out of generate timing. So I don't think it' really necessary, unless I'm missing something here.

tools/llm_bench/llm_bench_utils/get_use_case.py
RyanMetcalfeInt8: done.

tools/llm_bench/doc/PROMPT.md
RyanMetcalfeInt8: ah, good catch. I didn't realize that PROMPT.md is a description of parameters supported via jsonl input. I've removed these as they can only be set via cmd-line input.

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -110,32 +228,58 @@ def run_text_to_speech_generation_genai(
...
 
     start = time.perf_counter()
-    additional_args = {"speaker_embeddings": ov.Tensor(args['speaker_embeddings'].numpy())} if args.get('speaker_embeddings') is not None else {}
-    generation_result = model.generate(input_text_list, **additional_args)
+    speeches = []
+    perf_metrics = None
+    tokenization_time = None
+    if is_kokoro_model:
+        for _ in range(args["batch_size"]):
+            speeches.append(_kokoro_generate_once(model, input_text, args, use_genai=True))
```
sbalandi:
```suggestion
        speeches.append(_kokoro_generate_once(model, input_text, args, use_genai=True))
```

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -147,7 +291,8 @@ def run_text_to_speech_generation_genai(
         batch_size=args['batch_size'],
         prompt_idx=prompt_index
     )
-    log.debug(f'[{num}]Throughput: {perf_metrics.throughput.mean:.4f}')
+    if perf_metrics is not None:
+        log.debug(f"[{num}]Throughput: {perf_metrics.throughput.mean:.4f}")
```
sbalandi: it looks like generate_durations are collected for kokoro https://github.com/openvinotoolkit/openvino.genai/blob/8981d6f848f17985979be0a9224251d181f68c56/src/cpp/src/speech_generation/kokoro_tts_model.cpp#L1091 
f so, throughput can also be calculated https://github.com/openvinotoolkit/openvino.genai/blob/8981d6f848f17985979be0a9224251d181f68c56/src/cpp/src/speech_generation/speech_generation_perf_metrics.cpp#L24

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -158,6 +303,23 @@ def run_text_to_speech_generation_genai(
...
+
+        voice = args.get("speech_voice", "").strip() or DEFAULT_KOKORO_VOICE
+        voice_file = model_path / "voices" / f"{voice}.bin"
+        if not voice_file.exists():
+            raise RuntimeError(
+                f"Kokoro voice file not found: {voice_file}. "
+                f"Pass --speaker_embeddings directly or ensure '{voice}' exists under {model_path / 'voices'}/."
+            )
+        log.info(f"Loading Kokoro speaker embedding from {voice_file}")
+        args["speaker_embeddings"] = get_speaker_embeddings(str(voice_file), expected_shape=KOKORO_SPEAKER_EMB_SHAPE)
```
sbalandi: This code is very similar to optimum-intel, and close to pt, except for creating the .bin file. Is it possible to separate skeaper embedding choosing/loading into a separate common function?

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -19,6 +21,86 @@
...
+        speaker_embeddings=speaker_embeddings,
+        language=speech_language,
+        voice=speech_voice,
+    )
+    return _extract_audio_array(generation_result)
+
+
+def _kokoro_preprocess_once(model, input_text, args, use_genai):
+    if use_genai:
+        return input_text
```
sbalandi: these functions is not used for GenAI (_kokoro_generate_from_preprocessed, _kokoro_generate_once), GenAI branches could be removed

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -19,6 +21,86 @@
...
+    except ImportError:
+        pass
+
+    if hasattr(output, "numpy"):
+        return output.numpy().reshape(-1)
+
+    return np.asarray(output, dtype=np.float32).reshape(-1)
+
+
+def _kokoro_generate_once(model, input_text, args, use_genai):
```
sbalandi: Could you please move all kokoro related function to tts_utils.py ?

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: Okay, yes. In latest commit, `perf_metrics` are returned for Kokoro case as well, and so throughput here will be printed.

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: done.

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: Okay, I combined the majority of the handling between optimum-intel and GenAI case into a common `resolve_kokoro_speaker_embedding`. PT case is a little bit different, so I mainly left that alone.

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: Okay right -- these branches are removed from `kokoro_preprocess_once`, `kokoro_generate_from_preprocessed` -- and `_kokoro_generate_once` has been removed entirely (as it was specific to GenAI, and so this path was commonized with non-Kokoro model path).

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: Okay done.

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -158,6 +303,23 @@ def run_text_to_speech_generation_genai(
...
+        log.warning("Only batch size 1 available for benchmarking with kokoro model")
+        args["batch_size"] = 1
+    # For Kokoro with GenAI, the pipeline requires speaker_embedding as an ov.Tensor.
+    # Auto-resolve from the model's voices/ directory when not provided via CLI.
+    if args.get("is_kokoro_model") and use_genai and args.get("speaker_embeddings") is None:
+        from llm_bench_utils.model_utils import get_speaker_embeddings
+
+        voice = args.get("speech_voice", "").strip() or DEFAULT_KOKORO_VOICE
+        voice_file = model_path / "voices" / f"{voice}.bin"
+        if not voice_file.exists():
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: hmm, I don't think we ever expect path to be to the .xml and not the parent folder... the valid use case is always path to exported model directory, not individual .xml (OV IR)

tools/llm_bench/llm_bench_utils/pt_utils.py
```diff
@@ -182,22 +184,115 @@ def create_text_2_speech_model(model_path, device, memory_data_collector, **kwar
...
+
+                    def _update_language(self, language):
+                        requested_lang_code = normalize_kokoro_lang_code(language)
+                        if requested_lang_code != self._lang_code:
+                            self._lang_code = requested_lang_code
+                            self._pipeline = KPipeline(lang_code=self._lang_code, model=self._kmodel)
+
+                    def preprocess_input(self, prompt, speaker_embeddings=None, language="", voice=""):
+                        self._update_language(language)
+                        quiet_pipeline = KPipeline(lang_code=self._lang_code, model=False)
```
sbalandi, line 212: Why separate pipeline is created here and self._pipeline is not used ?

RyanMetcalfeInt8: Kokoro is a little bit weird, but in order to have a 'clean' preprocess (phonemization + tokenization) path for Kokoro, we essentially need to create a special kind of pipeline object that explicitly disables the KModel (the piece that performs inference). This 'quiet' pipeline (quiet as as in can't generate audio) is used to perform G2P preprocessing, and the output if that is passed to the 'real' pipeline.

Note that optimum-intel uses a similar mechanism for preprocess_inputs: https://github.com/huggingface/optimum-intel/blob/main/optimum/intel/openvino/modeling_text2speech.py#L822

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -158,6 +228,18 @@ def run_text_to_speech_generation_genai(
 def run_text_2_speech_benchmark(model_path, framework, device, args, num_iters, mem_consumption):
     mem_consumption.update_marker("model")
     model, processor, vocoder, pretrain_time, use_genai = FW_UTILS[framework].create_text_2_speech_model(model_path, device, mem_consumption, **args)
+    if args.get("is_kokoro_model", False) and args.get("batch_size", 1) != 1:
+        log.warning("Only batch size 1 available for benchmarking with kokoro model")
+        args["batch_size"] = 1
+    # For Kokoro with GenAI, the pipeline requires speaker_embedding as an ov.Tensor.
+    # Auto-resolve from the model's voices/ directory when not provided via CLI.
+    if args.get("is_kokoro_model") and use_genai and args.get("speaker_embeddings") is None:
```
sbalandi: There are 2 steps: a file loading part, and a preprocessing part. When we measure the preprocessing part, we don't care about the file loading part. But there is a part where we select a voice for pt/optimum, so it turns out that loading happens internally in that cases. Then could you please move the loading of embeddings for GenAI to the run_text_to_speech_generation_genai function to be able to calculate the loading and processing time together in the future, for optimum/pt it's already like that

tools/llm_bench/llm_bench_utils/pt_utils.py
sbalandi: got it, thank you

tools/llm_bench/task/text_to_speech_generation.py
RyanMetcalfeInt8: Sure, see latest commit.

sbalandi (approved):

tools/llm_bench/task/text_to_speech_generation.py
sbalandi: thanks !

PR #4032, opened 2026-06-26
Fix CVS-189485

sunxiaoxia2022:
## Description

This PR adjusts metric reset behavior in speculative decoding fast draft strategy to avoid clearing SD logical statistics at generate reset time.
m_sd_metrics is a debugging and diagnostic metric collector specifically for speculative decoding. Statistics are printed and cleaned up once per step only when the environment variable is set to >warning. Now we get statistics of main and draft models from m_perf_metrics. So the reset for m_sd_metrics here is not needed and can be removed.

CVS-189485

Fixes #(issue)

## Checklist:

### DIFF
--- a/src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
+++ b/src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
@@ -180,11 +180,9 @@ protected:
     std::map<uint64_t, GenerationHandle> m_draft_generations;
 
     void reset_generate_metrics() {
-        m_sd_metrics = SpeculativeDecodingMetrics();
         m_perf_metrics = ov::genai::SDPerModelsPerfMetrics();
-        RawPerfMetrics draft_metrics;
-        draft_metrics.m_inference_durations = {{ MicroSeconds(0.0f) }};
-        m_draft_pipeline->raw_perf_metrics = std::move(draft_metrics);
+        m_draft_pipeline->raw_perf_metrics = RawPerfMetrics{};
+        m_draft_pipeline->raw_perf_metrics.m_inference_durations = {{ MicroSeconds(0.0f) }};
     }
 
     void drop_requests();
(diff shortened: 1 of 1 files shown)


xufang-lisa (approved):

src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
```diff
@@ -180,7 +180,6 @@ class ContinuousBatchingPipeline::SpeculativeDecodingImpl : public ContinuousBat
     std::map<uint64_t, GenerationHandle> m_draft_generations;
 
     void reset_generate_metrics() {
-        m_sd_metrics = SpeculativeDecodingMetrics();
```
songbell, line 183: if removed, is the sd_metrics right for result2 in this case?
auto result1 = pipeline.generate(prompt1, config);  // generate #1
auto result2 = pipeline.generate(prompt2, config);  // generate #2

sunxiaoxia2022: `m_sd_metrics` is a debugging and diagnostic metric collector specifically for speculative decoding. Statistics are printed and cleaned up once per step only when the environment variable is set to `>warning`. Now we get statistics of main and draft models from `m_perf_metrics`. So the reset for `m_sd_metrics` here is not needed and can be removed.

songbell: get it, then I have no problem with the logic here, but I'm curious why the scan reports data race here on this line..

songbell (approved):

src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.hpp
sunxiaoxia2022: Because a lock was added in [step()](https://github.com/openvinotoolkit/openvino.genai/blob/master/src/cpp/src/speculative_decoding/continuous_batching/fast_draft_strategy.cpp#L178), and there is a write operation on `m_sd_metrics`. All other accesses to `m_sd_metrics` are protected by the lock `m_draft_generations_mutex`. Only here there is no lock. So "missing lock" is reported.

as-suvorov (approved):

PR #4048, opened 2026-06-26
[llm_bench] Add text generation chat pipeline

sbalandi:
## Description
- added text generation pipeline for chat mode, pipeline can be triggered by --task text_gen_chat, user may specify prompts for chat via propmts list(-pf) or specify --chat_iter and one prompt, chat will run chat_inter times with one same prompt 
- refactor text generation pipeline

examples how to use:
```
python benchmark.py -m ./models/Llama-2-7b-chat-ov/ --task text_gen_chat --chat_iter 3 -n 2 
python benchmark.py -m ./models/Llama-2-7b-chat-ov/ --task text_gen_chat --chat_iter 3 -n 2 -p "What is OpenVINO?"
python benchmark.py -m ./models/Llama-2-7b-chat-ov/ --task text_gen_chat -n 2 -pf ./prompts/llm_chat.json
```

## Checklist:

### DIFF
--- a/tools/llm_bench/README.md
+++ b/tools/llm_bench/README.md
@@ -194,7 +194,7 @@ python benchmark.py -m models/llama-2-7b-chat/ --draft_model models/TinyLlama-1.
 ```sh
 # chat iteration
 python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat --chat_iter 3
-python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat -pf ./prompts/llm_chat.jsonl
+python benchmark.py -m ./models/llama-2-7b-chat/ -n 2 --task text_gen_chat -pf ./prompts/llm_chat.jsonl
 ```
 
 **Some additional parameters:**
--- a/tools/llm_bench/README.md
+++ b/tools/llm_bench/README.md
@@ -194,7 +194,7 @@ python benchmark.py -m models/llama-2-7b-chat/ --draft_model models/TinyLlama-1.
 ```sh
 # chat iteration
 python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat --chat_iter 3
-python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat --pf ./prompts/llm_chat.jsonl
+python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat -pf ./prompts/llm_chat.jsonl
 ```
 
 **Some additional parameters:**
--- a/tools/llm_bench/README.md
+++ b/tools/llm_bench/README.md
@@ -191,6 +191,12 @@ optimum-cli export openvino --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 models/Ti
 python benchmark.py -m models/llama-2-7b-chat/ --draft_model models/TinyLlama-1.1B-Chat-v1.0 -p "What is openvino?" -n 2 --task text_gen --num_assistant_tokens 5
 ```
 
+```sh
+# chat iteration
+python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat --chat_iter 3
+python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat --pf ./prompts/llm_chat.jsonl
+```
+
 **Some additional parameters:**
 - `--draft_device`: Inference device for Speculative decoding of draft model.
 - `--draft_cb_config`: Path to file with Continuous Batching Scheduler settings or dict for Speculative decoding of draft model.
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -12,6 +12,7 @@ import llm_bench_utils.output_csv
 import llm_bench_utils.output_json
 import task.visual_language_generation as bench_vlm
 import task.text_generation as bench_text
+import task.text_generation_chat as bench_text_chat
 import task.image_generation as bench_image
 import task.video_generation as bench_video
 import task.super_resolution_generation as bench_ldm_sr
@@ -26,14 +27,14 @@ from llm_bench_utils.memory_monitor import MemoryMonitorHandler
 DEFAULT_TORCH_THREAD_NUMS = 16
 
 
-def num_iters_type(x):
+def num_positive_integers(x):
     x = int(x)
     if x < 0:
         raise argparse.ArgumentTypeError("Minimum input value is 0")
     return x
 
 
-def num_infer_count_type(x):
+def num_greater_than_one(x):
     x = int(x)
     if x < 1:
         raise argparse.ArgumentTypeError("Minimum input value is 1")
@@ -89,7 +90,7 @@ def get_argparser():
         "-pi",
         "--prompt_index",
         nargs="+",
-        type=num_iters_type,
+        type=num_positive_integers,
         default=None,
         help="Run the specified prompt index. You can specify multiple prompt indexes, separated by spaces.",
     )
@@ -98,14 +99,14 @@ def get_argparser():
         "-ic",
         "--infer_count",
         default=None,
-        type=num_infer_count_type,
+        type=num_greater_than_one,
         help="set the output token size, the value must be greater than 0.",
     )
     parser.add_argument(
         "-n",
         "--num_iters",
         default=0,
-        type=num_iters_type,
+        type=num_positive_integers,
         help="number of benchmarking iterations, "
         "if the value is greater than 0, the average numbers exclude the first(0th) iteration,\n"
         "if the value equals 0 (default), execute the warm-up iteration(0th iteration).",
@@ -316,7 +317,7 @@ def get_argparser():
         help="Stop the generation even if output token size does not achieve infer_count or max token size ({DEFAULT_OUTPUT_TOKEN_SIZE}}).",
     )
     parser.add_argument(
-        "--set_torch_thread", default=0, type=num_infer_count_type, help="Set the number of Torch thread. "
+        "--set_torch_thread", default=0, type=num_greater_than_one, help="Set the number of Torch thread. "
     )
     parser.add_argument(
         "-tl",
@@ -371,6 +372,7 @@ def get_argparser():
         default=None,
         choices=[
             "text_gen",
+            "text_gen_chat",
             "image_gen",
             "visual_text_gen",
             "speech_to_text",
@@ -479,11 +481,20 @@ def get_argparser():
         default=None,
         help="controller of video frames to process (required frame number if positive or decimation factor if negative)",
     )
+    parser.add_argument(
+        "--chat_iter",
+        type=num_greater_than_one,
+        default=None,
+        help="Use with --task text_gen_chat. The chat will run chat_iter iterations with the one prompt."
+        " Alternative option is setup prompts list in JSONL via -pf option."
+        " The parameter specifies the amount of the chat iterations.",
+    )
     return parser.parse_args()
 
 
 CASE_TO_BENCH = {
     "text_gen": bench_text.run_text_generation_benchmark,
+    "text_gen_chat": bench_text_chat.run_text_generation_benchmark,
     "image_gen": bench_image.run_image_generation_benchmark,
     "video_gen": bench_video.run_video_generation_benchmark,
     "code_gen": bench_text.run_text_generation_benchmark,
@@ -569,7 +580,7 @@ def main():
     log.info(out_str)
 
     try:
-        if model_args["use_case"].task in ["text_gen", "code_gen"]:
+        if model_args["use_case"].task in ["text_gen", "text_gen_chat", "code_gen"]:
             iter_data_list, pretrain_time, iter_timestamp = CASE_TO_BENCH[model_args["use_case"].task](
                 model_path,
                 framework,
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -27,14 +27,14 @@ from llm_bench_utils.memory_monitor import MemoryMonitorHandler
 DEFAULT_TORCH_THREAD_NUMS = 16
 
 
-def num_positive_integers(x):
+def positive_integer(x):
     x = int(x)
     if x < 0:
         raise argparse.ArgumentTypeError("Minimum input value is 0")
     return x
 
 
-def num_at_least_one(x):
+def greater_than_zero(x):
     x = int(x)
     if x < 1:
         raise argparse.ArgumentTypeError("Minimum input value is 1")
@@ -90,7 +90,7 @@ def get_argparser():
         "-pi",
         "--prompt_index",
         nargs="+",
-        type=num_positive_integers,
+        type=positive_integer,
         default=None,
         help="Run the specified prompt index. You can specify multiple prompt indexes, separated by spaces.",
     )
@@ -99,14 +99,14 @@ def get_argparser():
         "-ic",
         "--infer_count",
         default=None,
-        type=num_at_least_one,
+        type=greater_than_zero,
         help="set the output token size, the value must be greater than 0.",
     )
     parser.add_argument(
         "-n",
         "--num_iters",
         default=0,
-        type=num_positive_integers,
+        type=positive_integer,
         help="number of benchmarking iterations, "
         "if the value is greater than 0, the average numbers exclude the first(0th) iteration,\n"
         "if the value equals 0 (default), execute the warm-up iteration(0th iteration).",
@@ -316,7 +316,9 @@ def get_argparser():
         action="store_true",
         help="Stop the generation even if output token size does not achieve infer_count or max token size ({DEFAULT_OUTPUT_TOKEN_SIZE}}).",
     )
-    parser.add_argument("--set_torch_thread", default=0, type=num_at_least_one, help="Set the number of Torch thread. ")
+    parser.add_argument(
+        "--set_torch_thread", default=0, type=greater_than_zero, help="Set the number of Torch thread. "
+    )
     parser.add_argument(
         "-tl",
         "--tokens_len",
@@ -481,7 +483,7 @@ def get_argparser():
     )
     parser.add_argument(
         "--chat_iter",
-        type=num_at_least_one,
+        type=greater_than_zero,
         default=None,
         help="Use with --task text_gen_chat. The chat will run chat-iter iterations with the one prompt."
         " Alternative option is setup prompts list in JSONL via -pf option."
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -493,7 +493,6 @@ def get_argparser():
         "--full_chat",
         action="store_true",
         help="Use with --task text_gen_chat and optimum-intel/PyTorch backends. "
-        "Use with --task text_gen_chat and optimum-intel/PyTorch backends. "
         "Benchmark will send the full chat history as input for generation on each turn. By default, only the new prompt is used.",
     )
 
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -34,7 +34,7 @@ def num_positive_integers(x):
     return x
 
 
-def num_greater_than_one(x):
+def num_at_least_one(x):
     x = int(x)
     if x < 1:
         raise argparse.ArgumentTypeError("Minimum input value is 1")
@@ -99,7 +99,7 @@ def get_argparser():
         "-ic",
         "--infer_count",
         default=None,
-        type=num_greater_than_one,
+        type=num_at_least_one,
         help="set the output token size, the value must be greater than 0.",
     )
     parser.add_argument(
@@ -316,9 +316,7 @@ def get_argparser():
         action="store_true",
         help="Stop the generation even if output token size does not achieve infer_count or max token size ({DEFAULT_OUTPUT_TOKEN_SIZE}}).",
     )
-    parser.add_argument(
-        "--set_torch_thread", default=0, type=num_greater_than_one, help="Set the number of Torch thread. "
-    )
+    parser.add_argument("--set_torch_thread", default=0, type=num_at_least_one, help="Set the number of Torch thread. ")
     parser.add_argument(
         "-tl",
         "--tokens_len",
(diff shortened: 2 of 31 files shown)


tools/llm_bench/task/text_generation_chat.py
```diff
@@ -0,0 +1,654 @@
...
+# ===== Common Utils =====
+def get_chat_input_data(input_text: str | list, args: dict):
+    # if prompts are set as list, let's use it
+    # if prompt is set as single string, let's create chat where prompt will repeat chat_iter times
+    input_data = input_text
+    if not isinstance(input_text, list):
+        if args.get("chat_iter"):
+            input_data = [input_text] * args["chat_iter"]
+        else:
+            log.error("Chat mode can't be started due to incompatible input prompts")
```
avasenin-14: Medium priority: Do I understand correctly, that with 1 prompt string or no explicit prompt and no --chat_iter given, it will log the error but still will iterate over the string in the line 209 later? Over the characters, in that case?

tools/llm_bench/task/text_generation_chat.py
```diff
@@ -0,0 +1,654 @@
...
+    # for optimum-intel statefull model ((),)
+    past_key_values = None
+    tokenized_history: list = []
+
+    past_key_values = None
+    prefix_len = 0
+    full_chat = args.get("full_chat_hist")
+
+    mem_consumption.start(iter_num)
+    for prompt_index, prompt in enumerate(input_data):
```
avasenin-14: Low priority: I'm not sure if users can / will specify empty list in a json.
If it can happen, than get_chat_input_data probably will return empy list, and the iter_data won't be assigned in the loop, and then metrics_print.print_memory_info(iter_num, iter_data, chat_index) might fail without the variable defined.
Do you think it makes sense add a check to get_chat_input_data or initialize the variable before the loop? Or is it too far fetched scenario?

tools/llm_bench/task/text_generation_chat.py
sbalandi: llm_bench will crash

tools/llm_bench/task/text_generation_chat.py
sbalandi: it will not fail, but all resutls will empty, I will add check and crash for that case

krzyczar (commented):
Reviewer:    GNAI (automated review, 2026-06-30)
=====

OVERVIEW
--------
This PR adds a new `text_gen_chat` benchmarking pipeline to `tools/llm_bench`,
allowing users to benchmark LLMs in a multi-turn chat mode. The pipeline can be
driven either by a JSONL prompt file (`-pf`) with lists of turns, or by a single
prompt repeated `--chat_iter` times. The PR also refactors shared helpers in
`task/text_generation.py`, adds `@dataclass` decorators to hook output classes
across 15 versioned files, and propagates a new `chat_idx` field through all
reporting paths (CSV, JSON, log output).

Scope: 32 files changed, ~490 additions, ~310 deletions.

CRITICAL ISSUES (must fix before merge)
=====

[C1] MISSING FILE: task/text_generation_chat.py
------------------------------------------------------
`benchmark.py` imports:
    import task.text_generation_chat as bench_text_chat

and routes `text_gen_chat` to:
    "text_gen_chat": bench_text_chat.run_text_generation_benchmark,

However, `tools/llm_bench/task/text_generation_chat.py` does NOT exist in the
repository (it is not part of this PR's file diff). Selecting `--task text_gen_chat`
will immediately raise a `ModuleNotFoundError`, making the entire new feature
completely unusable. This is a showstopper — the file must be included in the PR.

Action: Add the missing `task/text_generation_chat.py` implementation.

----

[C2] TESTS VALIDATE THE WRONG TASK
------------------------------------------------------
The new test `test_python_tool_llm_benchmark_llm_chat` passes `--task text_gen_chat`
via `sample_args`, but looking at the test body:

    benchmark_py_command = [sys.executable, benchmark_script, "-m", convert_model,
                            *prompt_args] + sample_args

The `sample_args` parametrize lists do include `"--task", "text_gen_chat"`. However,
given issue C1 (missing module), every test invocation will fail with
`ModuleNotFoundError`. Once C1 is fixed, the tests should actually exercise the new
pipeline correctly.

Also, the no-prompt variants (without `-p` or `-pf`) rely on the fallback default
prompt "What is OpenVINO?" being a string, but the chat pipeline presumably expects
a list of strings. This needs verification.

Action: Ensure tests pass end-to-end after C1 is resolved; validate default prompt
handling in chat mode.

MAJOR ISSUES (should fix before merge)
=====

[M1] STRAY DEBUG PRINT STATEMENT IN text_generation.py
------------------------------------------------------
A debug print like:
    print("IIII...")
was left in `run_text_generation_benchmark()`. This will pollute stdout and break
any CI output parsing or log diffing that depends on clean benchmark output.

Action: Remove the debug print.

----

[M2] MISLEADING FUNCTION NAME: num_greater_than_one()
------------------------------------------------------
The renamed function:
    def num_greater_than_one(x):
        x = int(x)
        if x < 1:
            raise argparse.ArgumentTypeError("Minimum input value is 1")
        return x

Actually *allows* the value 1 (rejects only values < 1), so semantically it means
"at least 1" / "positive integer", not "greater than one". The name is actively
misleading. Copilot's review also flagged this.

Action: Rename to `num_at_least_one` or `num_positive_integer` (non-zero positive).

----

[M3] INCONSISTENT / POTENTIALLY WRONG first_token_latency IN CHAT MODE
------------------------------------------------------
In `metrics_print.py`:

    iter_data['first_token_latency'] = tms[0] * 1000 if len(tms) > 0 else -1
    if is_chat_mode and len(tms) > 0:
        iter_data["first_token_latency"] /= iter_data["input_size"]

This normalizes first-token latency by `input_size` in chat mode. While the comment
explains the intent ("prompts can be very long, so we need to calculate latency per
token"), the display suffix for `first_token_infer_latency` (the hook-based first
inference latency) still reads `ms/infer` instead of `ms/token`, making it
inconsistent and misleading.

Furthermore, dividing first-token latency by the *total* accumulated input size
across all chat turns may not be meaningful — the denominator grows with each turn,
producing a metric that decreases over the chat session regardless of actual model
performance.

Action: Clarify what normalization is intended, fix the `ms/infer` suffix for
`first_token_infer_latency` in chat mode, and document the metric definition.

----

[M4] HARDCODED NPU ASSUMPTION FOR full_chat_hist
------------------------------------------------------
In `model_utils.py`:
    model_args["full_chat_hist"] = model_args["devices"] == "NPU"

Full chat history is silently force-enabled for NPU and disabled for everything
else. This is a hidden behavioral difference based on device type with no user
override. If a user wants full history on CPU for a specific benchmark, they cannot.
If NPU no longer requires this special treatment in a future version, this code
will be a silent correctness bug.

Action: Expose `--full_chat_hist` as an explicit CLI flag with NPU as default True,
or at minimum document this behavior prominently.

=====
MINOR ISSUES / IMPROVEMENTS
=====

[N1] TYPO IN WARNING MESSAGE (model_utils.py)
The warning reads:
    "take prompts form prompt_file"
Should be:
    "take prompts from prompt_file"

----

[N2] DEAD CODE: max_length VARIABLE (text_generation.py)
After the refactoring, `max_length` is assigned but never used in
`run_text_generation()`. Remove it to avoid confusion and linting warnings.

----

[N3] DUPLICATE ASSIGNMENT OF prompt_idx (gen_output_data.py)
`prompt_idx` is assigned to `iter_data["prompt_idx"]` twice in `gen_iterate_data()`.
The second assignment is redundant — remove it.

----

[N4] MISSING NEWLINE AT END OF FILE (prompts/llm_chat.jsonl)
The new `prompts/llm_chat.jsonl` file is missing a trailing newline. Many tools
and POSIX standards expect text files to end with a newline.

----

[N5] UseCaseTextGenChat IS A NEAR-DUPLICATE OF UseCaseTextGen (config_class.py)
The new dataclass:
    @dataclass
    class UseCaseTextGenChat(UseCase):
        task = "text_gen_chat"
        ov_cls: type | None = OVModelForCausalLM
        pt_cls: type | None = AutoModelForCausalLM

is identical in structure to `UseCaseTextGen`, differing only in the `task` string.
Consider having `UseCaseTextGenChat` inherit from `UseCaseTextGen` or use a factory
pattern to reduce duplication and keep the two in sync automatically.

----

[N6] output_csv.py COMPARISON WITH UseCase OBJECT
`output_comments()` compares `use_case` against string values, but
`model_args['use_case']` may be a `UseCase` object rather than a plain string.
Verify that `use_case` is properly `.task`-accessed before string comparison in
this function (or that the caller passes the correct type).

----

[N7] INCONSISTENT ARGUMENT NAMING: --chat_iter vs --chat-iter
The PR description uses `--chat-iter` (hyphen) in the usage examples:
    python benchmark.py ... --task text_gen_chat --chat-iter 3
But the actual argument name defined in `get_argparser()` is `--chat_iter` (underscore).
While argparse normalizes both at parse time, the inconsistency in documentation
is confusing. Standardize on one form (prefer hyphen in CLI docs, since argparse
converts hyphens to underscores internally).

----

[N8] LARGE HOOK FILE CHURN (15 files, 45 lines total)
Adding `@dataclass` to `GenerateDecoderOnlyOutput` / `GenerateEncoderDecoderOutput`
/ `GenerateBeamDecoderOnlyOutput` / `GenerateBeamEncoderDecoderOutput` across 15
versioned hook files is a good cleanup, but it is entirely unrelated to the chat
pipeline feature. Consider splitting this into a separate PR to keep the diff
focused and make bisect/revert easier if needed.

----

[N9] ARGUMENT VALIDATION: --chat_iter ALLOWS VALUE 1
`num_greater_than_one` allows value=1. A single chat iteration makes no practical
sense for a "chat" pipeline (you need at least 2 turns for a real conversation).
Consider enforcing `chat_iter >= 2` or at minimum document why 1 is valid.

POSITIVE ASPECTS
=====

(+) Feature is well-motivated: benchmarking multi-turn chat is an important use
    case distinct from single-turn text generation.

(+) Both Optimum and GenAI backends are handled, and the tests cover both paths.

(+) Extraction of `save_input_data_to_file()`, `print_input_info()`, and
    `print_generated_output()` from text_generation.py into shared helpers is a
    clean refactor that reduces duplication between text and chat pipelines.

(+) `chat_idx` is propagated consistently through gen_output_data, metrics_print,
    output_json, and output_csv, keeping reporting coherent.

(+) README.md and PROMPT.md are updated with usage examples, which is good practice.

(+) The conflict handling between `--chat_iter` and `--pf` is explicit (warning +
    graceful fallback to file), which is user-friendly.

(+) Test parametrization covers 4 scenarios (optimum with/without `-p`,
    genai with/without `-p`) giving reasonable coverage.

SUMMARY & RECOMMENDATION
=====

VERDICT: REQUEST CHANGES

The PR cannot be merged in its current state due to the missing
`task/text_generation_chat.py` file (C1), which makes the entire feature
non-functional. The stray debug print (M1) and misleading function name (M2) also
need to be addressed.

Priority of fixes:
  1. [C1] Add the missing task/text_generation_chat.py — BLOCKER
  2. [C2] Verify tests pass end-to-end — BLOCKER
  3. [M1] Remove debug print — HIGH
  4. [M2] Fix misleading function name — MEDIUM
  5. [M3] Fix chat-mode latency metric inconsistency — MEDIUM
  6. [M4] Make full_chat_hist configurable — MEDIUM
  7. [N1-N9] Minor/cosmetic fixes — LOW

Once the missing implementation file is added and the above issues are resolved,
this will be a solid contribution to llm_bench.

sbalandi:
> ### [C1] MISSING FILE: task/text_generation_chat.py
the file exists, https://github.com/openvinotoolkit/openvino.genai/pull/4048/changes#diff-168dffd70e4ac00b9532eab6d0931fa1445fd76478ac9d0f492efe513f934815R4-R5

> ### [C2] TESTS VALIDATE THE WRONG TASK
ci is green, so tests are passed

> ### [M1] STRAY DEBUG PRINT STATEMENT IN text_generation.py
it was removed, don't see it now 

> ### [M2] MISLEADING FUNCTION NAME: num_greater_than_one()
renamed

> ### [M3] INCONSISTENT / POTENTIALLY WRONG first_token_latency IN CHAT MODE
Comment updated, normalization should be done by input size, input size is a prompt for each chat iteration, not the entire chat, suffix should be `ms/token`.

> ### [M4] HARDCODED NPU ASSUMPTION FOR full_chat_hist
good point, fixed 

> #### [N1] TYPO IN WARNING MESSAGE (model_utils.py) The warning reads: "take prompts form prompt_file" Should be: "take prompts from prompt_file"
> #### [N2] DEAD CODE: max_length VARIABLE (text_generation.py) After the refactoring, `max_length` is assigned but never used in `run_text_generation()`. Remove it to avoid confusion and linting warnings.
> #### [N3] DUPLICATE ASSIGNMENT OF prompt_idx (gen_output_data.py) `prompt_idx` is assigned to `iter_data["prompt_idx"]` twice in `gen_iterate_data()`. The second assignment is redundant — remove it.
don't see it now, look like was fixed before

> #### [N4] MISSING NEWLINE AT END OF FILE (prompts/llm_chat.jsonl) The new `prompts/llm_chat.jsonl` file is missing a trailing newline. Many tools and POSIX standards expect text files to end with a newline.
don't thinks it's necessary for jsonl

> #### [N5] UseCaseTextGenChat IS A NEAR-DUPLICATE OF UseCaseTextGen (config_class.py)
fixed

> #### [N6] output_csv.py COMPARISON WITH UseCase OBJECT `output_comments()` compares `use_case` against string values, but `model_args['use_case']` may be a `UseCase` object rather than a plain string. Verify that `use_case` is properly `.task`-accessed before string comparison in this function (or that the caller passes the correct type).
fixed 

> #### [N7] INCONSISTENT ARGUMENT NAMING: --chat_iter vs --chat-iter The PR description uses `--chat-iter` (hyphen) in the usage examples: python benchmark.py ... --task text_gen_chat --chat-iter 3 But the actual argument name defined in `get_argparser()` is `--chat_iter` (underscore). While argparse normalizes both at parse time, the inconsistency in documentation is confusing. Standardize on one form (prefer hyphen in CLI docs, since argparse converts hyphens to underscores internally).
fixed 

> #### [N8] LARGE HOOK FILE CHURN (15 files, 45 lines total) Adding `@dataclass` to `GenerateDecoderOnlyOutput` / `GenerateEncoderDecoderOutput` / `GenerateBeamDecoderOnlyOutput` / `GenerateBeamEncoderDecoderOutput` across 15 versioned hook files is a good cleanup, but it is entirely unrelated to the chat pipeline feature. Consider splitting this into a separate PR to keep the diff focused and make bisect/revert easier if needed.
agree, but it's required for that PR, so I would like to keep it here

> #### [N9] ARGUMENT VALIDATION: --chat_iter ALLOWS VALUE 1 `num_greater_than_one` allows value=1. A single chat iteration makes no practical sense for a "chat" pipeline (you need at least 2 turns for a real conversation). Consider enforcing `chat_iter >= 2` or at minimum document why 1 is valid.
Why not, for GenAI it different input parameter, for example

tools/llm_bench/task/text_generation_chat.py
sbalandi: moved to raise

krzyczar (approved):

tools/llm_bench/llm_bench_utils/llm_hook_beam_search/hook_beam_search_v4_40.py
```diff
@@ -21,11 +21,13 @@
 )
 from transformers.utils import ModelOutput
 import llm_bench_utils.hook_beam_search as hook_beam
+from dataclasses import dataclass
 
 
 logger = log.getLogger(__name__)
 
 
+@dataclass
```
as-suvorov, line 30: How this `@dataclass` used?

tools/llm_bench/llm_bench_utils/metrics_print.py
```diff
@@ -55,9 +58,15 @@ def print_metrics(
         output_str = ' '.join([prefix, output_str])
         log.info(output_str)
     if tms is not None:
-        iter_data['first_token_latency'] = tms[0] * 1000 if len(tms) > 0 else -1
-        iter_data['other_tokens_avg_latency'] = sum(tms[1:]) / (len(tms) - 1) * 1000 if len(tms) > 1 else -1
-        first_token_latency = 'NA' if iter_data['first_token_latency'] == -1 else f"{iter_data['first_token_latency']:.2f} ms"
+        iter_data["first_token_latency"] = tms[0] * 1000 if len(tms) > 0 else -1
+        if is_chat_mode and len(tms) > 0 and iter_data["input_size"] != "" and int(iter_data["input_size"]) > 0:
+            iter_data["first_token_latency"] /= iter_data["input_size"]
```
as-suvorov, line 63: Please explain why division required here?

tools/llm_bench/llm_bench_utils/metrics_print.py
```diff
@@ -227,45 +275,92 @@ def output_avg_statis_tokens(prompt_dict, prompt_idx_list, iter_data_list, batch
...
+            if avg_1st_token_latency >= 0:
+                avg_1st_token_latency = (
+                    f"{avg_1st_token_latency:.2f} ms/{display_unit}" if chat_mode else f"{avg_1st_token_latency:.2f} ms"
+                )
+            else:
+                avg_1st_token_latency = "NA"
+            avg_2nd_tokens_latency = (
+                "NA" if avg_2nd_tokens_latency < 0 else f"{avg_2nd_tokens_latency:.2f} ms/{display_unit}"
+            )
+            avg_2nd_token_tput = "NA" if avg_2nd_tokens_latency == "NA" else f"{avg_2nd_token_tput:.2f} {tput_unit}s/s"
```
as-suvorov, line 317:
```suggestion
            avg_2nd_token_output = "NA" if avg_2nd_tokens_latency == "NA" else f"{avg_2nd_token_tput:.2f} {tput_unit}s/s"
```

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -241,6 +241,18 @@ def analyze_args(args):
...
+    if model_args["use_case"].task == "text_gen_chat":
+        if args.chat_iter is not None and args.prompt_file is not None:
+            log.warning(
+                "`--chat_iter` can't be combined with `--prompt_file`, llm_bench will ignore `--chat_iter` and take prompts from prompt_file for generation"
+            )
+            model_args["chat_iter"] = None
+        elif args.chat_iter is None and args.prompt_file is None:
+            model_args["chat_iter"] = 1
+
+    model_args["full_chat"] = args.full_chat or model_args["devices"] == "NPU"
```
as-suvorov: Why NPU forcing full_chat needed?

tools/llm_bench/benchmark.py
```diff
@@ -26,14 +27,14 @@
 DEFAULT_TORCH_THREAD_NUMS = 16
 
 
-def num_iters_type(x):
+def num_positive_integers(x):
```
as-suvorov: nitpick
```suggestion
def positive_integer(x):
```

tools/llm_bench/benchmark.py
```diff
@@ -26,14 +27,14 @@
...
     x = int(x)
     if x < 0:
         raise argparse.ArgumentTypeError("Minimum input value is 0")
     return x
 
 
-def num_infer_count_type(x):
+def num_at_least_one(x):
```
as-suvorov: nitpick
```suggestion
def greater_than_zero(x):
```

tools/llm_bench/README.md
```diff
@@ -185,6 +185,12 @@ optimum-cli export openvino --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 models/Ti
 python benchmark.py -m models/llama-2-7b-chat/ --draft_model models/TinyLlama-1.1B-Chat-v1.0 -p "What is openvino?" -n 2 --task text_gen --num_assistant_tokens 5
 ```
 
+```sh
+# chat iteration
+python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat --chat_iter 3
+python benchmark.py -m ./models/llama-2-7b-chat/ -p "What is openvino?" -n 2 --task text_gen_chat -pf ./prompts/llm_chat.jsonl
```
as-suvorov: Can we use `-pf ./prompts/llm_chat.jsonl` by default if `--task text_gen_chat` specified?
How `-p "What is openvino?"` and `-pf ./prompts/llm_chat.jsonl` works together?

tools/llm_bench/task/text_generation_chat.py
```diff
@@ -0,0 +1,656 @@
...
+    # === Save perf data ===
+    iter_data_list.extend(chat_iter_data_list)
+
+    if args["output_dir"] is not None:
+        llm_bench_utils.output_file.output_gen_text(
+            chat_history, args, model_precision, chat_index, iter_num, batchsize_idx=0, proc_id=proc_id, is_chat=True
+        )
+
+
+def run_text_generation_genai_chat_mode(
```
as-suvorov: `run_text_generation_chat_optimum` and `run_text_generation_genai_chat_mode` do the same thing. It make sense to consider logic sharing. How complicated will be to implement adapter classes for optimum and genai to have common interface so both can be run with `run_text_generation_chat_mode`?

tools/llm_bench/llm_bench_utils/llm_hook_beam_search/hook_beam_search_v4_40.py
sbalandi: it fails if not set 
`TypeError: llm_bench_utils.llm_hook_sample.hook_sample_v5_3.GenerateDecoderOnlyOutput is not a dataclass. This is a subclass of ModelOutput and so must use the @dataclass decorator.`

tools/llm_bench/llm_bench_utils/metrics_print.py
sbalandi: we benchmark chat, chat includes several prompts, most likely of different sizes, for example:
prompt1: What is OpenVINO?
prompt2: 1+1=

so, we can't generalize different first token latency as it will measure different amount of tokens, but we can divide it and generalize/estimate time per token

tools/llm_bench/llm_bench_utils/model_utils.py
sbalandi: NPU don't have manipulation with kv_cache, they pass whole history each time (correct me if it was changed)

tools/llm_bench/llm_bench_utils/metrics_print.py
sbalandi: it's throughput, please, see on [str 299](https://github.com/sbalandi/openvino.genai/blob/93bc4b2ebca412e449a02e8e4469ee13eb2fa5b4/tools/llm_bench/llm_bench_utils/metrics_print.py#L299C17-L299C36)

tools/llm_bench/benchmark.py
sbalandi: fixed

tools/llm_bench/benchmark.py
sbalandi: fixed

tools/llm_bench/README.md
sbalandi: it's mistake, removed
if --task text_gen_chat is specified llm_bench will use internal prompt by default and --chat_iter 1. If the user has specified a prompt in any way (-p/-pf), llm_bench will use it

tools/llm_bench/task/text_generation_chat.py
sbalandi: they are the same in steps, but the implementation of the steps are different. I thought about this, but decided that commenting and visually highlighting the steps would be about the same as separating them into a separate class. I'll try to do adapters and write

sbalandi: added commit with classes

avasenin-14 (approved):

tools/llm_bench/llm_bench_utils/metrics_print.py
as-suvorov: What is `iter_data["first_token_latency"]` ? Is it array or scalar?

as-suvorov (approved):

Issue #3968, opened 2026-06-10
High memory usage on Qwen3.5


WizardlyBump17:
I have noticed that Qwen3.5 uses tons of memory. 

For example, Echo9Zulu/Qwen3.5-0.8B-int8_asym-ov wanted to use 18.4GB for 110k tokens even with `{"KV_CACHE_PRECISION": "i4"}`; `ovms --rest_port 8080 --model_repository_path /models/ --source_model Echo9Zulu/Qwen3.5-0.8B-int8_asym-ov --target_device GPU --task text_generation --plugin_config '{"KV_CACHE_PRECISION": "i4"}'`
<img width="1282" height="40" alt="Image" src="https://github.com/user-attachments/assets/72a1f117-a8b7-4db9-a375-0f90c32195ab" />

OpenVINO/Qwen3-4B-int4-ov, a much bigger model, wanted to use only 9.3GB for the same task. `ovms --rest_port 8080 --model_repository_path /models/ --source_model OpenVINO/Qwen3-4B-int4-ov --target_device GPU --task text_generation --plugin_config '{"KV_CACHE_PRECISION": "i4"}'`. config.json edited to allow bigger prompts.
<img width="1282" height="49" alt="Image" src="https://github.com/user-attachments/assets/dd95bb3a-ff05-4ba0-8b30-dfa5d90c0049" />

It would be very nice if Qwen3.5 used less memory

droans:
I've run quite a few tests related to the memory usage on Qwen3.5/3.6. I have noticed that, on average, the VRAM usage is about 1GB per 1K tokens. Importantly, it seems that setting `KV_CACHE_PRECISION` doesn't have much of an effect on this while adjusting `INFERENCE_PRECISION_HINT` will. 

However, there are many levers that can effect this. 

OV is built using OCL for the GPU backend. I accidentally built it with Level Zero and noticed something interesting. When set to use the PA attention backend, the difference was massive - around 5.5K tokens per gigabyte of VRAM at peak usage, 12.6K at plateau. Going even further, by setting `use_sparse_attention=True`, `max_num_batched_tokens=2048`, and `max_num_seq=2048`, we were looking at around 13.9K peak, 30.9K plateau). 

This was only true for PA, though. L0 SDPA still showed the same memory issues. 

[...]

PR #4051, opened 2026-06-27
[CB] Reserve dynamic cache capacity for later/larger prompts in one reallocation (#3968)

exzile:
## Summary

Reduces continuous-batching cache **memory churn / instability** in dynamic allocation mode for long prompts. Complements #4050 (relates to #3968).

## Problem

In dynamic cache-allocation mode (`cache_size == 0` and `num_kv_blocks == 0`), the cache is pre-sized to fit incoming prompts **only on the first scheduling round** (`_initialize_cache`, gated by `has_token_capacity()`). A prompt that arrives **later and is larger** than what the cache was sized for instead grows `m_cache_growth_num_tokens` (256) at a time across its whole prefill. Each growth **reallocates and copies the entire KV (and linear-attention) cache**, so a long prompt causes O(n²) copy work and the unstable, stepwise memory growth reported for hybrid models (e.g. the "memory jumps 23→27 GB when the larger prompt runs second" in #3968).

## Fix

When a prompt-phase sequence is entering and the cache is already growing dynamically, reserve capacity in a **single reallocation** via `ensure_sequence_token_capacity` (which only grows when the required level exceeds current capacity, so it is idempotent).

The reservation targets **all active sequences (running and pending)**, not just the arriving prompt: `ensure_sequence_token_capacity` grows to the *absolute sum* of the supplied targets, so it must include the footprint already held by running sequences — otherwise a prompt arriving alongside running ones would be under-reserved and fall back to chunked growth. The work is skipped on pure generation steps and is a no-op once capacity is sufficient.

- Factored per-sequence target-building into a shared helper (`_build_sequence_token_targets`) used by both `_initialize_cache` and the new `_reserve_dynamic_capacity_for_active_sequences`.
- **Fixed-budget mode is unaffected** (it never enters dynamic allocation).
- Note: the grow-and-copy itself (allocate new tensor → copy old in → replace) is the existing pattern shared by the KV and linear-attention managers; this change targets the realloc **frequency**, which drives the instability — not the single-realloc copy.

## Validation

- **Unit tests** (new): a later, larger prompt reserves its full capacity in one `schedule()` round; and a new prompt reserves capacity **on top of a still-running sequence** (the concurrent case).
- **CB unit suite** passes (one unrelated, pre-existing flaky timestamp-LRU `BlockManager` test aside — confirmed flaky on master, passes in isolation).
- **End-to-end** on `OpenVINO/Qwen3.5-9B-int8-ov` (Intel Arc GPU, dynamic allocation): a large prompt issued **after** a small one now prefills in the same time as when issued **first** (ratio ~0.94; previously slower due to incremental growth), with byte-identical output.

## Notes

Draft for maintainer feedback. Focused scheduler change in the dynamic-allocation path; happy to adjust the pre-size heuristic or gating if a different approach is preferred.


exzile:
**Relationship to #4050:** both PRs address facets of #3968 but are **independent and can merge in any order** — they touch disjoint files (#4050: `cache_orchestrator.hpp`, `scheduler_config.hpp`, `pipeline_impl.cpp`; this PR: `scheduler.hpp`, `tests/cpp/scheduler.cpp`), so there is no merge conflict and no ordering dependency. #4050 reduces the linear-attention checkpoint size (adaptive interval) and surfaces OOM as a clear error; this PR removes the incremental-growth realloc churn in dynamic mode. They are complementary.

rkazants (commented):
Is it AI code generated? Which model and prompts are used?

exzile:
@rkazants I use Claude Opus/Sonnet 5 for code reviews, bug/memory leaks searches, refactor suggestions.

popovaan (approved):

exzile:
Re: the Copilot review comment on `_reserve_dynamic_capacity_for_active_sequences()` (scheduler.hpp:637) — checked into this.

The concern: `_build_sequence_token_targets()` sizes each active sequence by `prompt_len` + `max_new_tokens` (an estimate of *final* length), not by the sequence's *actual current* cache usage. In theory, if that estimate under-shoots a running sequence's real footprint, the computed `required_blocks` could come out `<=` the current total, so `ensure_sequence_token_capacity()` wouldn't grow — even though a newly-arriving prompt still needs room.

This is worth flagging, but it's not a correctness hazard, for two reasons:

1. **`ensure_sequence_token_capacity()` is monotonic-grow-only** (`block_manager.hpp`): it only calls `increase_block_count()` when `required_blocks > get_total_block_count()`, never shrinks. So an under-estimated target simply means "doesn't grow as much as it ideally would," not "grows incorrectly" or "frees blocks it shouldn't."
2. **The pre-existing `_try_increase_cache()` fallback is untouched.** If this one-shot reservation under-shoots, the old chunked-growth safety net (the exact mechanism this PR is trying to avoid triggering *as often*) still catches the shortfall on the next scheduling step, same as it did before this PR existed.

So the worst case for this edge case is graceful degradation to pre-PR behavior (a few extra small reallocations for that specific sequence), not a new bug. And the `seq_length = prompt_len * multiplier` / `min(..., prompt_len + max_new)` heuristic itself is pre-existing code (unchanged from `_initialize_cache`) — this PR just reuses it via the new shared helper, so it's not introducing a new estimation strategy, just applying the existing one in a second call site.

Happy to tighten the heuristic to use actual current sequence length (`get_num_processed_tokens()` or similar) instead of the prompt/max_new estimate as a follow-up, but don't think it should block this PR — it fixes a real, reproducible O(n²) reallocation/memory-spike issue (openvinotoolkit/openvino.genai#3968) without weakening any existing correctness guarantee.

apaniukov:
build_jenkins

PR #4055, opened 2026-06-29
llm_bench: add Windows-native memory sampler

krzyczar:
Introduce MemorySamplerW, a new Windows-specific memory sampler that queries per-process memory counters directly through the Win32 GetProcessMemoryInfo API (psapi.dll) instead of relying on the psutil abstraction layer.

Key changes
-----------
* Module-level Win32 ctypes setup (guarded by sys.platform == 'win32'):
  - Defines _PROCESS_MEMORY_COUNTERS_EX ctypes.Structure mirroring the PROCESS_MEMORY_COUNTERS_EX struct from psapi.h.
  - Loads kernel32.dll and psapi.dll exactly once at import time, keeping per-sample overhead to a minimum.
  - Declares explicit argtypes / restype for OpenProcess, GetProcessMemoryInfo, and CloseHandle to prevent silent 64-bit pointer truncation.

* MemorySamplerW class (inherits MemorySampler):
  - Tracks five metrics per sample for the target process and all its recursive children:
      wset     – Working Set (WorkingSetSize): physical RAM pages currently
                 resident; equivalent to RSS on Linux and to the
                 "Memory (Working Set)" column in Task Manager Details.
      priv     – Private Bytes (PrivateUsage): committed virtual memory
                 private to the process; maps to psutil's .private field
                 and the Task Manager "Commit Size" column.
      pagefile – Page-File Usage (PagefileUsage): system page-file bytes
                 consumed by the process.
      sys      – System-wide used physical memory (total - available),
                 same definition as in MemorySampler5.
      nsys     – sys expressed as percentage of total installed RAM.
  - _query_win_mem(pid) static helper opens a process handle, fills the
    PROCESS_MEMORY_COUNTERS_EX struct, closes the handle, and returns
    (WorkingSetSize, PrivateUsage, PagefileUsage) in bytes; gracefully
    returns (0, 0, 0) when the process has already exited or access is
    denied.
  - collect(marker) aggregates parent + all recursive children, reads
    psutil.virtual_memory() once for the system-wide counters, then
    delegates to MemorySampler.aggregate_and_format() for statistics
    bookkeeping (min / max / mean / CV).

Motivation
----------
On Windows, psutil's memory_full_info() requires SeDebugPrivilege for child processes and may fail silently in restricted environments.  The Win32 path uses PROCESS_QUERY_INFORMATION | PROCESS_VM_READ access rights, which are available to standard users for processes they own, and exposes Working Set and Private Bytes metrics that closely match what Windows Task Manager reports — making benchmark results directly comparable to Task Manager observations.

## Description

CVS-175953

## Checklist:

### DIFF
--- a/tools/llm_bench/llm_bench_utils/memory_monitor.py
+++ b/tools/llm_bench/llm_bench_utils/memory_monitor.py
@@ -883,6 +883,250 @@ class MemorySampler(dict, SamplerTiming):
         return tuple(map(self.format_to_export, mapargs))
 
 
+# ---------------------------------------------------------------------------
+# Windows-specific ctypes setup for MemorySamplerW
+#
+# All objects are defined at module level so they are created only once and
+# reused across every collect() call, keeping sampling overhead minimal.
+#
+# The key Win32 call chain is:
+#   OpenProcess(PROCESS_QUERY_INFORMATION | PROCESS_VM_READ, FALSE, pid)
+#       → HANDLE hProcess
+#   GetProcessMemoryInfo(hProcess, &PROCESS_MEMORY_COUNTERS_EX, sizeof(...))
+#       → fills the struct with per-process memory counters
+#   CloseHandle(hProcess)
+#
+# PROCESS_MEMORY_COUNTERS_EX lives in psapi.dll (also re-exported by
+# kernel32.dll on modern Windows) and mirrors the C struct from psapi.h.
+# ---------------------------------------------------------------------------
+if sys.platform == "win32":
+    import ctypes
+    import ctypes.wintypes as _wintypes
+
+    _PROCESS_QUERY_INFORMATION = 0x0400
+    _PROCESS_VM_READ = 0x0010
+
+    class _PROCESS_MEMORY_COUNTERS_EX(ctypes.Structure):
+        """
+        Mirrors the Win32 ``PROCESS_MEMORY_COUNTERS_EX`` struct (psapi.h).
+
+        Fields of interest
+        ------------------
+        WorkingSetSize  : Physical RAM pages currently resident for the
+                          process (= RSS / what Task Manager "Memory" column
+                          shows after adding the Working Set column in Details).
+        PrivateUsage    : Committed virtual memory private to the process
+                          (= Private Bytes / Commit Size in Task Manager).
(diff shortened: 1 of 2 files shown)


avasenin-14:
Hi @krzyczar , thank you for the PR!
Can you please check the failed tests and the Copilot comments?
Once these are resolved we can review the PR.

sbalandi:
windows-specific metrics look good, but it's probably worth checking with the plugins to see if they're useful for them. Have you compared them with the current one ? Is system memory, for example, similar ?

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -857,7 +1101,8 @@ def collect(self, marker):
 class MemoryMarkerMonitor(list):
     def __init__(self, marker_queue, process_id, sampling_interval, path_prefix):
         log.info("Memory worker: MemorySampler5 init...")
-        self.sampler = MemorySampler5(process_id)
```
avasenin-14, line 926: Hi @krzyczar , can you please help me understand why do you think it is better to add a new class instead of adding a fallback for some metrics in the MemorySampler5?
Also was it the only place where MemorySampler5 was used?

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -817,6 +817,250 @@ def aggregate_and_format(self, marker, values):
...
+            # The target process may have exited between the parent query
+            # and the child iteration — treat as a partial sample.
+            pass
+
+        # ── System-wide memory ───────────────────────────────────────────
+        vm       = psutil.virtual_memory()
+        sys_mem  = vm.total - vm.available
+        nsys_mem = 100.0 * sys_mem / vm.total
+
+        vals = wset_total, priv_total, pagefile_total, sys_mem, nsys_mem
```
avasenin-14: Do you think collecting different types of metrics might break some validation comparison with previous releases?

avasenin-14:
As far as I understand the feature was merged in https://github.com/openvinotoolkit/openvino.genai/pull/4243. Please reopen this PR if I'm wrong.

krzyczar:
thanks

PR #4059, opened 2026-06-30
[GGUF] Support Mistral (mistral3 / Devstral) with correct q/k RoPE layout + YaRN

### LINKED ISSUE #4058: [GGUF] Mistral GGUF models (e.g. Devstral / arch "mistral3") unsupported and produce garbage
### Summary

GGUF models with Mistral architecture (`general.architecture = "mistral"` / `"mistral3"`, e.g. **Devstral-Small-2512**) cannot be served correctly by OpenVINO GenAI's native GGUF reader:

1. **Hard failure on load:** the architecture dispatch in `gguf_modeling.cpp` whitelists only `llama`, `qwen2`, `qwen3`, so loading throws `OPENVINO_THROW("Unsupported model architecture 'mistral3'")`.
2. **Even after whitelisting, generation is garbage** — the root cause is q/k weight layout (details below).

### Environment
- openvino.genai: master (also reproduced on the 2026.3.0 dev build)
- Model: `Devstral-Small-2-24B-Instruct-2512-Q4_K_M.gguf` (`general.architecture = mistral3`)
- OS: Windows 11, CPU/GPU

### Root cause

Mistral is a **NORM-type RoPE** architecture (llama.cpp `rope type = 0`), like Llama. For NORM-type models, llama.cpp stores the q/k projection weights **interleaved (permuted)**. GenAI's `rotate_half` uses the **NEOX "halves"** convention, so it relies on `reorder_interleaved_format()` to de-interleave q/k weights on load.

That de-interleaving is gated in `building_blocks.cpp` on the arch name containing `"llama"`:

```cpp
bool reorder = false;
if (configs.at("architecture").find("llama") != std::string::npos) {
    reorder = true;
}
```

`"mistral3"` does not match, so `reorder` stays `false` → Mistral's interleaved q/k weights are fed to `rotate_half` in the wrong layout → broken attention → incoherent output. (Qwen2/Qwen3 are NEOX-type / `rope type = 2` and correctly need no reorder, which is why they work.)

Additionally, Devstral uses **YaRN** RoPE scaling (`rope.scaling.type = yarn`, factor 48, original context 8192) which the reader currently ignores, so long-context support is missing.

### Reproduce
Serve the Devstral GGUF (or any Mistral GGUF) and request a completion. Without the fix: `Unsupported model architecture 'mistral3'` on load, or — with only the whitelist added — output like `"The capital of France is" -> "a French colony of F of the United Kingdom of New Hampshire..."`.

### Proposed fix
- Add `mistral`/`mistral3` to the GGUF arch whitelist.
- Enable q/k `reorder` (de-interleave) for `mistral` as well (it is NORM-type like `llama`).
- Read `rope.scaling.*` and apply YaRN frequency interpolation for long context (mscale = 1.0 for this model, matching llama.cpp's `yarn_attn_factor`).

After the fix, output matches the llama.cpp reference (`"The capital of France is Paris."`). PR to follow.

exzile:
### Tickets
Fixes #4058

### Details
Enables OpenVINO GenAI's native GGUF reader to serve **Mistral** GGUF models (`general.architecture = "mistral"` / `"mistral3"`, e.g. **Devstral-Small-2512**), which previously failed with `Unsupported model architecture 'mistral3'` and — once whitelisted — produced garbage output.

Three changes:

1. **Arch whitelist** (`gguf_modeling.cpp`): route `mistral`/`mistral3` through `create_language_model`.

2. **q/k weight de-interleave for Mistral** (`building_blocks.cpp`) — *the correctness fix*. The `reorder` flag that converts llama.cpp's NORM-type interleaved q/k layout into the NEOX "halves" layout `rotate_half` expects was only enabled for arch names containing `"llama"`. Mistral is also NORM-type (`rope type 0`), so without `reorder` its q/k weights were applied with the wrong layout → broken attention → garbage. Now enabled for `mistral`/`mistral3` too. (Qwen2/Qwen3 are NEOX-type / `rope type 2` and correctly need no reorder.)

3. **YaRN RoPE frequency interpolation** (`gguf.cpp`, `building_blocks.cpp/.hpp`, `gguf_modeling.cpp`): read `rope.scaling.{type,factor,original_context_length,yarn_beta_fast,yarn_beta_slow,yarn_log_multiplier}` from GGUF metadata and apply NTK-by-parts interpolation in `init_rope` for long-context support. The attention-magnitude `mscale` is left at 1.0 to match llama.cpp's effective `yarn_attn_factor` for this model (verified from llama.cpp: `yarn_attn_factor = 1.0000 (mscale == 1.0)`). `attention.temperature_scale` is read into config and left as a documented TODO (no-op below `original_context_length`).

`llama`, `qwen2`, `qwen3` paths are unaffected (defaults are no-ops; reorder behavior unchanged for them).

### Validation
Validated against a standard llama.cpp CPU reference build on the same Devstral-Small-2512 Q4_K_M GGUF:
- Chat: "What is the capital of France?" → **"The capital of France is Paris."** (matches reference)
- Code: "Write a Python function for the nth Fibonacci number" → clean, correct, idiomatic Python (was incoherent before)

### Notes
`attention.temperature_scale` full wiring is left as a follow-up (only affects context beyond `original_context_length`).


apaniukov:
This one probably requires tokenizer conversion adjustments as well. I don't see `tokenizer.ggml.pre==tekken` support in GenAI GGUF reader.

exzile:
> This one probably requires tokenizer conversion adjustments as well. I don't see `tokenizer.ggml.pre==tekken` support in GenAI GGUF reader.
@apaniukov  Good catch, pushed a commit with the implementation.

TianmengChen:
@exzile can you please add a pytest?

rkazants (changes requested):
Was it AI code generated? Which model did you use for generation? Please share prompts.

I disagree with Rope representation

exzile:
@rkazants. 
On provenance: yes, I developed this with AI assistance (Claude Opus 4.8 / Sonnet 5) and then validated it myself token-for-token against a llama.cpp CPU reference build on Devstral-Small-2512 Q4_K_M (chat + code prompts match). Happy to share the reasoning and validation steps, and to walk through any part in detail.

Before creating the PR, I also code reviewed, refactored, scan for bugs & memory leaks, just in case.

On the RoPE representation — I want to make sure I address the right thing before reworking it, since a few things fall under that heading:

q/k weight de-interleave (reorder) — this isn't new in this PR. The existing code already permutes q/k from llama.cpp's interleaved (NORM-type) layout into the "halves" layout rotate_half expects, gated on arch.find("llama"). Mistral is also NORM-type (rope type 0), so I extended the same gate to mistral/mistral3. Without it, attention gets the wrong q/k layout → garbage (that's the root cause of #4058).
YaRN — NTK-by-parts frequency interpolation in init_rope, with mscale left at 1.0 to match llama.cpp's effective yarn_attn_factor for this model.
Could you point to which part you disagree with — the weight-permutation approach vs. handling the layout in the graph, the gating signal, or the YaRN math?

If the concern is that the layout is chosen by an architecture-name substring, I agree that's brittle and I'm happy to refactor it to select the de-interleave based on the model's RoPE type (NORM vs NEOX) via an explicit mapping rather than arch.find(...) — that keeps behavior identical while making the representation explicit. If you have a preferred canonical representation in mind, I'll align to it.

exzile:
@TianmengChen 
Added a pytest: `tests/python_tests/test_gguf_mistral.py` (`test_mistral_arch_matches_llama_reference`).

Since there's no CI-sized `mistral3` model and `transformers`' GGUF loader can't act as a reference for the `mistral3` arch, I test the fix by **equivalence to the llama path** instead — no external model needed:

- Mistral is llama-family: identical tensor layout and NORM-type (interleaved) q/k RoPE weights. This PR routes `mistral`/`mistral3` through the same `create_language_model` path and applies the same q/k de-interleave (`reorder`) that `llama` already uses.
- The test takes the small llama-arch GGUF already in the suite (`SmolLM2-135M`), rewrites `general.architecture` (and the `llama.*` hyperparameter keys) to `mistral` / `mistral3`, and asserts OV GenAI generates **identical** output to the original llama load, across `GGUF_PIPELINE_TYPES`.
- This exercises exactly the two code paths the PR changes: the arch whitelist (`gguf_modeling.cpp`) and the `reorder` branch (`building_blocks.cpp`). **Before** this PR, `mistral3` failed to load and `mistral` skipped the reorder → output diverges; **after**, both match the llama reference. So the assertion genuinely distinguishes fixed vs. broken.

The GGUF relabel helper does a faithful field/type + tensor copy (following llama.cpp's `gguf_new_metadata`) and I validated it against `gguf==0.19.0` (already in `tests/python_tests/requirements.txt`).

Note: this covers the correctness fix (whitelist + NORM reorder), the root cause of the garbage output in #4058. The tekken pre-tokenizer is covered by the C++ unit test in `tests/cpp/utils.cpp`. YaRN scaling is a no-op below `original_context_length` and was validated end-to-end against llama.cpp on Devstral (per the PR description); happy to add a dedicated YaRN metadata-injection check too if you'd like.

exzile (commented):
@rkazants 
On the RoPE representation — I'd like to make sure I address the right concern before changing anything, since "representation" could point at a few different things here. Could you clarify which you disagree with?

The q/k reorder de-interleave (building_blocks.cpp) — extending the existing llama reorder=true path to mistral/mistral3. Both are NORM-type rope (rope type 0), so this de-interleaves llama.cpp's interleaved q/k layout into the NEOX "halves" layout rotate_half expects. This is the actual correctness fix for the garbage output; it mirrors what the llama path already does.

The YaRN frequency interpolation (init_rope) — computing NTK-by-parts inv-freqs on the host and baking them into constants, rather than representing YaRN via a parametric RoPE op / graph-level scaling. If this is the objection, I'm happy to rework it toward whatever representation you'd prefer.
Something else in how cos/sin are materialized in multi_head_attention.

If it's (2)/(3), a pointer to the representation you'd want (existing op, or a reference model that does it the "right" way in the GGUF reader) would save a round-trip. I can also split the correctness fix (1) out into its own PR so Devstral serving lands independently of the YaRN discussion, if that's easier to review.

exzile:
@TianmengChen we going to let this go unmerged?

dorloff:
@exzile , thank you for your contribution and sorry for the lack of communication. Please note that there is ongoing work on a new approach for GGUF support - [the GGUF frontend](https://github.com/openvinotoolkit/openvino/tree/master/src/frontends/gguf). So this architecture should be supported soon, together with several others.

PR #4060, opened 2026-06-30
Add supports_paged_attention() model capability query

exzile:
## Summary
Adds an authoritative, structural way to ask whether a model can be served by the continuous batching (paged attention) pipeline.

Continuous batching applies `ov::pass::SDPAToPagedAttention`, which rewrites `ScaledDotProductAttention` ops into `PagedAttention`. Some architectures are exported without any `ScaledDotProductAttention` op (e.g. certain Gemma exports) and therefore cannot be converted — building the CB pipeline throws. Today the only way for a consumer to discover this is to attempt construction and **match the exception text**, which is brittle (an upstream message change silently breaks it).

This adds a capability query consumers can call *before* constructing the pipeline:

- `utils::supports_paged_attention(const std::shared_ptr<ov::Model>&)` — scans for `ScaledDotProductAttention` ops, placed next to `has_linear_attention_states()`.
- A public `OPENVINO_GENAI_EXPORTS ov::genai::supports_paged_attention(...)` wrapper in `continuous_batching_pipeline.hpp` so OpenVINO Model Server (and other consumers) can query it through the public API.

## Motivation
Requested by @mzegla in the OVMS Gemma/PagedAttention auto-fallback discussion (openvinotoolkit/model_server#4336) as the clean, cross-repo home for this check, instead of an error-message match on the OVMS side. Once this lands, OVMS will consume it directly.

## Notes
- The check mirrors exactly what `SDPAToPagedAttention` requires (presence of `ScaledDotProductAttention`). If the transformation's requirements expand in the future, this helper is the single place to keep in sync.
- cc @mzegla


src/cpp/src/utils.cpp
```diff
@@ -461,6 +462,19 @@ bool has_linear_attention_states(const std::shared_ptr<ov::Model>& model) {
     return get_cache_types(*model).has_linear();
 }
 
+bool supports_paged_attention(const std::shared_ptr<ov::Model>& model) {
+    // Continuous batching applies ov::pass::SDPAToPagedAttention, which rewrites
+    // ScaledDotProductAttention ops into PagedAttention. A model exported without any
+    // ScaledDotProductAttention op (e.g. some Gemma exports) cannot be converted and
+    // must use a non-paged pipeline instead.
+    for (const auto& op : model->get_ops()) {
+        if (ov::is_type<ov::op::v13::ScaledDotProductAttention>(op)) {
```
apaniukov: This is a necessary condition, but it is not sufficient. A model has to successfully pass SDPAtoPA transformation that adds a bunch of PA related inputs, removes attention_mask/beam_idx inputs (if those inputs still present after the transformation, then the model doesn't support PA), and flattens input_ids input.

exzile: Good catch — you're right that the SDPA op is necessary but not sufficient. Reworked the check to be outcome-based: it now clones the model, runs `ov::pass::SDPAToPagedAttention` on the clone, and reports support only if the transformation actually produced a `PagedAttentionExtension` op and left no `attention_mask`/`beam_idx` inputs behind. The op-presence scan is gone. Cloning keeps the caller's model intact for a non-paged fallback. Updated the unit tests accordingly — a bare ScaledDotProductAttention (without the stateful KV-cache/attention_mask/beam_idx structure) now correctly reports `false`.

src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
```diff
@@ -384,4 +384,18 @@ OPENVINO_GENAI_EXPORTS std::pair<std::string, ov::Any> videos_metadata_batches(
...
+ * Continuous batching requires the model to pass the SDPAToPagedAttention transformation.
+ * This runs that transformation on a clone of the model and reports whether it succeeded
+ * (a PagedAttention op was produced and the attention_mask/beam_idx inputs were removed).
+ * Some architectures (e.g. certain Gemma exports) cannot be converted; callers can use this
+ * to choose a non-paged pipeline instead of attempting construction and handling a failure.
+ *
+ * @param model The model to inspect.
+ * @return true if the model supports paged attention (continuous batching), false otherwise.
+ */
+OPENVINO_GENAI_EXPORTS bool supports_paged_attention(const std::shared_ptr<ov::Model>& model);
```
rkazants, line 405: What will it report if only some part of the pipeline works in PA mode but not the whole? For example, omni pipeline includes language decoder in PA and talker in SDPA?

src/cpp/src/utils.cpp
```diff
@@ -461,6 +463,39 @@ bool has_linear_attention_states(const std::shared_ptr<ov::Model>& model) {
...
+    // (on a clone, leaving the caller's model untouched for a non-paged fallback) and verify it
+    // produced a PagedAttention op with no leftover attention_mask/beam_idx inputs.
+    std::shared_ptr<ov::Model> clone;
+    try {
+        clone = model->clone();
+        ov::pass::SDPAToPagedAttention().run_on_model(clone);
+    } catch (const std::exception&) {
+        // Transformation failed -> the model cannot be served with paged attention.
+        return false;
+    }
```
rkazants, line 494: try-catch is expensive block in terms of performance. Can we rely that this transformation does not throw exception. Need to discuss with transformation team to have resilient transformation.

src/cpp/src/utils.cpp
```diff
@@ -461,6 +463,39 @@ bool has_linear_attention_states(const std::shared_ptr<ov::Model>& model) {
...
+                return false;
+            }
+        }
+    }
+    // A successful conversion inserts at least one PagedAttentionExtension op.
+    for (const auto& op : clone->get_ops()) {
+        if (std::string(op->get_type_name()) == "PagedAttentionExtension") {
+            return true;
+        }
+    }
```
rkazants, line 509: Expensive code block, we need to walk through the whole graph. I believe we should have method of transformation that reports if the transformation was applied or not.

@CuriousPanCake, @mryzhov, please advise.

CuriousPanCake: Right now we don't have meaningful codes on a transformation callback return, but we can add checks on the transformation level for presence of `PagedAttentionExtension` and absence of `attention_mask` & `beam_idx`. If both conditions are true, we would return `true`, otherwise `false`, which would mean something is wrong.

src/cpp/src/utils.cpp
apaniukov: No exceptions from transformation is not reliable indicator.

src/cpp/src/utils.cpp
apaniukov: > If both conditions are true, we would return true, otherwise false, which would mean something is wrong.

Is there a way to avoid model cloning?

CuriousPanCake: > Is there a way to avoid model cloning?

We would check this on the original model within the transformation, no cloning will take place.

src/cpp/include/openvino/genai/continuous_batching_pipeline.hpp
exzile: Good question - supports_paged_attention() takes a single ov::Model, so it reports per-model, not per-pipeline. For a composite pipeline like omni (PA-capable language decoder + SDPA-only talker), the caller would call it once on each sub-model's ov::Model and combine the results according to that pipeline's own logic (e.g. the language decoder must support PA for CB to apply; the talker's SDPA-only status doesn't block that). This function doesn't attempt to represent "does the whole pipeline support PA" as one bool - it's a per-model capability query, and orchestration across a pipeline's sub-models is left to the caller.

If you'd like, I can add a doc comment on the declaration making that per-model scope explicit, and/or add a test case showing two ov::Models in the same pipeline reporting differently (PA-capable decoder=true, SDPA-only talker=false) to make the intended usage pattern concrete.

exzile: Pushed the doc comment clarifying the per-model (not per-pipeline) scope.

On the test case: I looked at adding a composite-pipeline test (PA-capable decoder=true, SDPA-only talker=false) and want to flag something before doing it - the existing test file only has "false" cases (bare SDPA, no SDPA at all). There's no fixture anywhere in the suite for a model that actually passes SDPAToPagedAttention (that requires a properly-shaped stateful graph with attention_mask/beam_idx inputs and KV-cache state via ReadValue/Assign, matching what a real LLM export produces). Hand-building one just for this test risks being subtly wrong in a way that's hard to verify without a real reference model.

Given that, I'd rather not fabricate a "true" fixture here. If a true-case test is wanted, I think it's better scoped as follow-up work using a real small exported model (or coordinating with whoever owns SDPAToPagedAttention test fixtures) rather than a synthetic graph I assemble by hand. Let me know if you'd like me to pursue that, or if the doc comment plus this thread is enough clarification for now.

exzile:
@apaniukov I don't want this to be buried after the work I put into it. Can you tell me what needs to be done to get this moving forward? Thanks

exzile:
Bumping before the stale bot closes this — there's a concrete decision to make and I'm happy to do the work either way.

Summarizing where we landed:
- The op-presence scan was replaced with an outcome-based check (run `SDPAToPagedAttention`, confirm a `PagedAttentionExtension` is produced and no `attention_mask`/`beam_idx` inputs remain). Per-model scope is now documented.
- The remaining concern (@rkazants, @apaniukov) is cost/robustness: try-catch isn't a reliable signal, and clone + full-graph walk is expensive.
- @CuriousPanCake proposed the cleaner design: perform that same check **inside the transformation** on the original model (no clone), reporting whether it successfully applied. @apaniukov agreed.

I think that transformation-level API is the right home for this, but it lives in openvino core rather than this PR. **How would you like to proceed?**

1. I open an openvino-core PR adding the "did SDPAToPagedAttention apply?" report to the transformation (per @CuriousPanCake's sketch), and this GenAI wrapper becomes a thin call into it. Happy to take that on if @CuriousPanCake / @mryzhov are OK with me implementing it.
2. Land this GenAI-side implementation now as an interim capability query and refactor onto the transformation-level API once it exists.

For context: on the OVMS side (openvinotoolkit/model_server#4336), @mzegla confirmed the detection should live in GenAI with OVMS as a pure consumer — so this PR is the agreed home. Just need the preferred sequencing so it doesn't stall.

PR #4070, opened 2026-07-02
[llm_bench] Add BenchPrompt/BenchPrompter and unify scheduling loop

krzyczar:
## Description

[llm_bench] Add BenchPrompt/BenchPrompter and unify text-gen scheduling loop
    
    ## Motivation
    
    The original prompt-handling code in llm_bench was spread across several
    task-specific helper functions (get_text_prompt, get_vlm_prompt,
    get_image_prompt, get_video_gen_prompt) with no shared abstraction.
    Additionally, run_text_generation_benchmark contained two almost-identical
    scheduling branches (subsequent=False / subsequent=True) that diverged over
    time, including a positional-argument bug in the subsequent=True path where
    tokens_len and streaming arguments were missing from the text_gen_fn call.
    
    ## Changes
    
    ### tools/llm_bench/llm_bench_utils/prompt_utils.py
    
    Introduce two new classes that provide a reusable, task-agnostic prompt
    abstraction layer:
    
    **BenchPrompt(dict)**
    - Handler for a single multimedia prompt entry.
    - Handler for a single multimedia prompt entry.
    - Accepts either a plain string (text-only) or a dict with any combination
      of well-known media keys: prompt, media, mask_image, video, audio,
      negative_prompt.
    - Structural validation on construction: requires a non-empty 'prompt' key.
    - Lazy media probing (probe() / __repr__): on first access, inspects all
      attached media files and populates internal metadata:
        - image  -> (width, height) via PIL
        - video  -> (frames, height, width) via make_video_tensor (same
                    decimation path as during actual benchmarking)
        - audio  -> (duration_sec, sample_rate) via librosa
        - mask   -> non-zero pixel fraction (%)
    - __repr__ produces a human-readable, size-annotated description, e.g.:
        BenchPrompt(text:7t + image:512x512)
        BenchPrompt(text:12t + video:640x480@30f)
        BenchPrompt(text:5t + audio:30.0s@44100Hz)
        BenchPrompt(text:9t + image:1024x768/35.2%)
    - MEDIA_KEYS class constant documents all recognised keys.
    
    **BenchPrompter(list)**
    - Container for multiple BenchPrompt objects.
    - Parses command-line args and/or .jsonl prompt files using the existing
      get_param_from_file / parse_*_json_data infrastructure, dispatching on
      task type (text_gen, visual_text_gen, image_gen, video_gen, ...).
    - active_pairs property: applies args['prompt_index'] filtering and returns
      (p_idx, BenchPrompt) pairs in original-list order.
    - iter_schedule(num_iters) generator: yields (num, p_idx, prompt) triples
      in the correct scheduling order:
        - subsequent=False (default): iterate all prompts inside each iteration
        - subsequent=True:            complete all iterations for one prompt
                                      before advancing to the next
      num=0 is always the warm-up iteration; num=1..num_iters are the timed
      benchmark iterations.
    
    ### tools/llm_bench/task/text_generation.py
    
    Refactor run_text_generation_benchmark to use BenchPrompter:
    
    - Replace get_text_prompt import with BenchPrompter import.
    - Replace manual prompt-index filtering logic with BenchPrompter + its
      active_pairs property.
    - Replace the duplicated if args['subsequent'] is False / else scheduling
      block (~74 lines) with a single unified for loop over
      prompter.iter_schedule(num_iters) (~35 lines).
    - Fix the pre-existing positional-argument bug in the subsequent=True path:
      tokens_len and streaming were not forwarded to text_gen_fn; the unified
      path always passes all required arguments correctly.

  (plus refactor of another pipelines)
    
    ## Backward compatibility
    
    All existing functionality is preserved:
    - prompt loading from command line and .jsonl files works as before.
    - prompt_index filtering behaviour is unchanged.
    - Both subsequent scheduling modes produce the same iteration/prompt order.
    - No changes to public function signatures visible to other tasks.

## Checklist:


tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+
+    Scheduling modes
+    ----------------
+    ``subsequent=False`` *(default)*
+        Outer loop = iteration numbers, inner loop = prompts.
+        All prompts are run before advancing to the next iteration.
+
+    ``subsequent=True``
+        Outer loop = prompts, inner loop = iteration numbers.
+        All iterations for one prompt complete before moving to the next.
```
sbalandi:
```suggestion
    ``subsequent=False`` *(default)*
        Outer loop = iteration numbers, inner loop = prompts.
        Prompts are processed in interleaved manner. Prompts are processed in subsequent manner. All prompts are run before advancing to the next iteration.

    ``subsequent=True``
        Outer loop = prompts, inner loop = iteration numbers.
        Prompts are processed in subsequent manner. All iterations for one prompt complete before moving to the next.
```

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+
+    In both modes ``num=0`` is the warm-up iteration.
+
+    Parameters
+    ----------
+    args : dict
+        Full benchmark args dict (as produced by
+        ``model_utils.analyze_args``). Relevant keys:
+
+            'use_case'      - object with a ``.task`` attribute
```
sbalandi:
```suggestion
            'use_case'      - object describing a pipeline type includes classes for handling the pipeline in optimum/PyTorch scenarios and the name of the pipeline type (``.task`` attribute)
```

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+        """
+        Populate the list with :class:`BenchPrompt` objects.
+
+        Selects the appropriate ``input_key`` for
+        :func:`get_param_from_file` and the matching parse function based
+        on the task type derived from ``args['use_case'].task``.
+        """
+        args = self._args
+        use_case = args.get("use_case")
+        task = getattr(use_case, "task", "text_gen") if use_case else "text_gen"
```
sbalandi: I would raise an error if no use case was found. On the one hand, this should be impossible, on the other, the benchmark doesn't know how to interpret the model in this case, it's some kind of undefined behavior. Therefore, it's better to throw an error than to run something unpredictable.

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+# DONE: refactor code in tools/llm_bench/task/text_generation.py to use new
+#       classes.
+# DONE: move prompt creation functions into BenchPrompter (see static methods
+#       below).
+#
+# All task files have been migrated to use BenchPrompt / BenchPrompter.
+# NOTE: speech_to_text loads the raw waveform lazily inside the iteration
+#       loop (not at prompt-construction time) since audio decoding is
+#       device-sampling-rate-dependent and cannot be pre-computed.
+#
```
sbalandi: please, remove

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+        Token count is estimated as the whitespace-split word count of the
+        text prompt (exact token count would require the tokenizer).
+        """
+        self.probe()
+        parts = []
+
+        # ---- text (optional) ----
+        if self.get("prompt"):
+            token_est = len(self["prompt"].split())
+            parts.append(f"text:{token_est}t")
```
sbalandi: These aren't tokens, they're symbols. if tokens are needed, you can pass a tokenizer here and encode a prompt, or update the prompt later in the main functions after tokenization when the tokens is known.

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+
+        # ---- image (optionally decorated with mask coverage fraction) ----
+        if self.get("media"):
+            if self._image_size:
+                w, h = self._image_size
+                if self.get("mask_image"):
+                    frac = self._get_mask_fraction(self["mask_image"])
```
sbalandi: why this function is called here, not in probe() ?

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -161,19 +161,15 @@ def run_text_2_speech_benchmark(model_path, framework, device, args, num_iters,
...
-            if 0 <= i < len(input_text_list):
-                text_list.append(input_text_list[i])
-                prompt_idx_list.append(i)
-    if len(input_text_list) == 0:
-        raise RuntimeError('==Failure prompts is empty ==')
+
+    # Build the prompt schedule via BenchPrompter, which handles both
+    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
+    # in a single unified iter_schedule() loop, eliminating the previous
+    # duplicated if/else branches.
```
sbalandi: I would remove this here and move it above the loop `for num, p_idx, prompt in prompter.iter_schedule(num_iters):` and leave smth like that, I believe  no one will remember how it was before
```suggestion
    # iterate over all prompt with BenchPrompter, loop run in two mode:
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
```

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+            input_key = "prompt"
+
+        output_data_list, is_json_data = get_param_from_file(args, input_key)
+
+        if is_json_data:
+            # Parse raw jsonl dicts according to task type.
+            # Note: parse_text_json_data returns plain strings; all others
+            # return dicts – both are accepted by BenchPrompt.__init__.
+            if task == "visual_text_gen":
+                raw_list = parse_vlm_json_data(output_data_list)
```
sbalandi: Shouldn't resolve_media_file_path be added here like in get_vlm_prompt ? (the same for parse_image_json_data, parse_video_json_data)

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+            text_param_list = parse_text_json_data(output_data_list)
+            if len(text_param_list) > 0:
+                for text in text_param_list:
+                    text_list.append(text)
+        else:
+            text_list.append(output_data_list[0])
+        return text_list
+
+    @staticmethod
+    def get_vlm_prompt(args):
```
sbalandi: Why do we need to keep these functions? It's not used inside the tool anymore, and there is not practice to use it outside

sbalandi:
These are quite big changes, it would be good to run validation with at least one model for each active pipeline to be sure that nothing was broken

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+        self.probe()
+        parts = []
+
+        # ---- text (optional) ----
+        if self.get("prompt"):
+            token_est = len(self["prompt"].split())
+            parts.append(f"text:{token_est}t")
+
+        # ---- image (optionally decorated with mask coverage fraction) ----
+        if self.get("media"):
```
sbalandi: it looks like representation should be updated after real media files are loaded 
`[ WARNING ] BenchPrompt: cannot probe image 'https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11': [Errno 2] No such file or directory: 'https://github.com/openvinotoolkit/openvino_notebooks/assets/29454499/d5fbbd1a-d484-415c-88cb-9986625b7b11'`
https://github.com/openvinotoolkit/openvino.genai/actions/runs/28935331643/job/85846800065?pr=4070#step:12:711

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: done

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: ok

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: ok

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: nice!

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: t -> w

tools/llm_bench/task/image_generation.py
```diff
@@ -228,26 +234,18 @@ def run_image_generation_benchmark(model_path, framework, device, args, num_iter
...
+        mem_consumption.update_marker(f"step-{num}-{p_idx}")
+        prefix = prompter.get_prefix(num, p_idx)
+        prompt.introduce_in_stdout(num, prefix)
+        iter_timestamp[num][p_idx]["start"] = datetime.datetime.now().isoformat()
+        image_gen_fn(prompt, num, p_idx, pipe, args, iter_data_list, proc_id, mem_consumption, callback)
+        if iter_data_list:
+            iter_data_list[-1]["prompt_repr"] = repr(prompt)
+        iter_timestamp[num][p_idx]["end"] = datetime.datetime.now().isoformat()
+        log.info(
+            f"{prefix}[P{p_idx}] start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
avasenin-14:
```suggestion
            f"{prefix} start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
Prefix already returns f"[{num}][P{p_idx}]"

tools/llm_bench/task/speech_to_text_generation.py
```diff
@@ -182,36 +186,26 @@ def run_speech_2_txt_benchmark(model_path, framework, device, args, num_iters, m
...
+        input_param["speech_param"] = prompt  # BenchPrompt dict carries language/timestamp
+        input_param["iter_idx"] = num
+        input_param["raw_speech"] = raw_speech
+        iter_timestamp[num][p_idx]["start"] = datetime.datetime.now().isoformat()
+        run_speech_2_txt_generation(input_param, args, md5_list, iter_data_list)
+        if iter_data_list:
+            iter_data_list[-1]["prompt_repr"] = repr(prompt)
+        iter_timestamp[num][p_idx]["end"] = datetime.datetime.now().isoformat()
+        log.info(
+            f"{prefix}[P{p_idx}] start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
avasenin-14:
```suggestion
            f"{prefix} start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
Prefix already returns f"[{num}][P{p_idx}]"

tools/llm_bench/task/super_resolution_generation.py
```diff
@@ -70,54 +71,45 @@ def run_ldm_super_resolution_benchmark(model_path, framework, device, args, num_
...
+        iter_timestamp[num][p_idx]["start"] = datetime.datetime.now().isoformat()
+        run_ldm_super_resolution(
+            prompt, num, pipe, args, framework, iter_data_list, p_idx, tm_list, proc_id, mem_consumption
+        )
+        if iter_data_list:
+            iter_data_list[-1]["prompt_repr"] = repr(prompt)
+        iter_timestamp[num][p_idx]["end"] = datetime.datetime.now().isoformat()
+        tm_list.clear()
+        log.info(
+            f"{prefix}[P{p_idx}] start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
avasenin-14:
```suggestion
            f"{prefix} start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
Prefix already returns f"[{num}][P{p_idx}]"

tools/llm_bench/task/video_generation.py
```diff
@@ -394,9 +408,20 @@ def run_video_generation_benchmark(model_path, framework, device, args, num_iter
...
+        mem_consumption.update_marker(f"step-{num}-{p_idx}")
+        prefix = prompter.get_prefix(num, p_idx)
+        prompt.introduce_in_stdout(num, prefix)
+        iter_timestamp[num][p_idx]["start"] = datetime.datetime.now().isoformat()
+        iter_data, _ = video_gen_pipeline.run(prompt, num, p_idx, proc_id, None)
+        iter_data_list.append(iter_data)
+        iter_data["prompt_repr"] = repr(prompt)
+        iter_timestamp[num][p_idx]["end"] = datetime.datetime.now().isoformat()
+        log.info(
+            f"{prefix}[P{p_idx}] start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
avasenin-14:
```suggestion
            f"{prefix} start: {iter_timestamp[num][p_idx]['start']}, end: {iter_timestamp[num][p_idx]['end']}"
```
Prefix already returns f"[{num}][P{p_idx}]"

avasenin-14 (changes requested):

tools/llm_bench/task/image_generation.py
```diff
@@ -196,7 +195,14 @@ def run_image_generation_genai(image_param, num, image_id, pipe, args, iter_data
 
 
 def run_image_generation_benchmark(model_path, framework, device, args, num_iters, mem_consumption):
-    image_list, prompt_idx_list = collect_prompts_step(args, get_image_prompt)
+    # Build the prompt schedule via BenchPrompter, which handles both
+    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
+    # in a single unified iter_schedule() loop, eliminating the previous
+    # duplicated if/else branches.  Also fixes the bug in the subsequent=True
+    # path where `callback` was not forwarded to image_gen_fn.
```
sbalandi:
```suggestion
    # Build the prompt schedule via BenchPrompter, which handles both
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
    # in a single unified iter_schedule() loop.  
```
Also fixes the bug in the subsequent=True path where `callback` was not forwarded to image_gen_fn. - please, move to PR description, after code will change the bug will gone, for what keep memories of it

tools/llm_bench/task/text_embeddings.py
```diff
@@ -155,19 +155,13 @@ def run_text_embddings_benchmark(model_path, framework, device, args, num_iters,
...
-        for i in args['prompt_index']:
-            if 0 <= i < len(input_text_list):
-                text_list.append(input_text_list[i])
-                prompt_idx_list.append(i)
-    if len(input_text_list) == 0:
-        raise RuntimeError('==Failure prompts is empty ==')
+    # Build the prompt schedule via BenchPrompter, which handles both
+    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
+    # in a single unified iter_schedule() loop, eliminating the previous
+    # duplicated if/else branches.
```
sbalandi:
```suggestion
    # Build the prompt schedule via BenchPrompter, which handles both
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
    # in a single unified iter_schedule() loop.
```

tools/llm_bench/task/text_generation.py
```diff
@@ -641,19 +641,16 @@ def run_text_generation_benchmark(model_path, framework, device, tokens_len, str
...
-                text_list.append(input_text_list[i])
-                prompt_idx_list.append(i)
-    if len(input_text_list) == 0:
-        raise RuntimeError('==Failure prompts is empty ==')
+
+    # Build the prompt schedule via BenchPrompter, which handles both
+    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
+    # in a single unified iter_schedule() loop, eliminating the previous
+    # duplicated if/else branches and fixing the positional-argument bug
+    # that existed in the subsequent=True path.
```
sbalandi:
```suggestion
    # Build the prompt schedule via BenchPrompter, which handles both
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
    # in a single unified iter_schedule() loop.
```
fixing the positional-argument bug that existed in the subsequent=True path - please, move to description, which arguments were passed wrong ?

tools/llm_bench/task/text_reranker.py
```diff
@@ -366,7 +366,14 @@ def run_text_reranker_benchmark(
     mem_consumption.update_marker("model")
     model, tokenizer, pretrain_time, bench_hook, use_genai = FW_UTILS[framework].create_text_reranker_model(model_path, device, mem_consumption, **args)
     iter_data_list = []
-    text_list, prompt_idx_list = collect_prompts_step(args, get_text_prompt)
+
+    # Build the prompt schedule via BenchPrompter, which handles both
+    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
+    # in a single unified iter_schedule() loop, eliminating the previous
+    # duplicated if/else branches.
```
sbalandi:
```suggestion
    # Build the prompt schedule via BenchPrompter, which handles both
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
    # in a single unified iter_schedule() loop.
```

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -243,19 +243,15 @@ def run_text_2_speech_benchmark(model_path, framework, device, args, num_iters,
...
```
sbalandi:
```suggestion
    # Build the prompt schedule via BenchPrompter, which handles both
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
    # in a single unified iter_schedule() loop.
```

tools/llm_bench/task/visual_language_generation.py
```diff
@@ -292,19 +291,14 @@ def run_visual_language_generation_benchmark(model_path, framework, device, args
...
-        for i in args['prompt_index']:
-            if 0 <= i < len(input_image_text_list):
-                image_text_list.append(input_image_text_list[i])
-                prompt_idx_list.append(i)
-    if len(input_image_text_list) == 0:
-        raise RuntimeError('==Failure prompts is empty ==')
+    # Build the prompt schedule via BenchPrompter, which handles both
+    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
+    # in a single unified iter_schedule() loop, eliminating the previous
+    # duplicated if/else branches.
```
sbalandi:
```suggestion
    # Build the prompt schedule via BenchPrompter, which handles both
    # subsequent=False (iter-major) and subsequent=True (prompt-major) modes
    # in a single unified iter_schedule() loop.
```

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +91,561 @@ def load_image_genai(image_path):
...
+            audio:30.0s@44100Hz      <- no text prompt
+
+        Word count (whitespace-split) of the text prompt is shown with a
+        ``w`` suffix.  Exact token count would require the model tokenizer.
+        """
+        self.probe()
+        parts = []
+
+        # ---- text (optional) ----
+        if self.get("prompt"):
```
sbalandi, line 306: In general, words aren't very informative for evaluating a prompt. We don't know how many chunks the tokenizer will break a phrase into. prompt might have the same number of words but a different number of tokens. Maybe the metrics could be updated in the function itself after tokenization, like you did for VLM in extract_prompt_data ?

tools/llm_bench/task/image_generation.py
krzyczar: fixed

tools/llm_bench/task/speech_to_text_generation.py
krzyczar: fixed

tools/llm_bench/task/super_resolution_generation.py
krzyczar: fixed

tools/llm_bench/task/video_generation.py
krzyczar: done

tools/llm_bench/llm_bench_utils/output_csv.py
```diff
@@ -132,6 +132,8 @@ def gen_data_to_csv(
     result["infer_count"] = iter_data["infer_count"]
     result["generation_time(s)"] = round(generation_time, 5) if generation_time != "" else generation_time
     result["output_size"] = iter_data["output_size"]
+    result["output_length"] = iter_data["output_size"]  # symmetric name for prompt_length
```
sbalandi: why output_length is needed ? what the difference with output_size ?

tools/llm_bench/llm_bench_utils/output_csv.py
```diff
@@ -149,15 +151,23 @@ def gen_data_to_csv(
...
+    result[f"max_sys_mem({mem_unit.value})"] = round(sys_mem, 5) if sys_mem != "" else sys_mem
+    result[f"max_increase_rss_mem({mem_unit.value})"] = (
+        round(rss_mem_increase, 5) if rss_mem_increase != "" else rss_mem_increase
+    )
+    result[f"max_increase_sys_mem({mem_unit.value})"] = (
+        round(sys_mem_increase, 5) if sys_mem_increase != "" else sys_mem_increase
+    )
+    result["prompt_idx"] = iter_data["prompt_idx"]
+    result["prompt_repr"] = iter_data.get("prompt_repr", "")
+    result["prompt_length"] = iter_data.get("prompt_length", "")
```
sbalandi: Why it is needed ? we have **input_size**, leave one of them, please

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +93,646 @@ def load_image_genai(image_path):
...
+            return
+        self._probed = True
+
+        # Image probing is skipped when the prompt carries audio. speech_to_text
+        # prompts store their audio path under the 'audio' key (populated by
+        # BenchPrompter._load_prompts); a 'media' key on such a prompt would be
+        # an audio path, not an image, and PIL-probing it would emit spurious
+        # warnings and waste work. Guarding on 'audio' keeps probe() correct
+        # regardless of which key the audio path arrives under.
+        if self.get("media") and not self.get("audio"):
```
sbalandi, line 191: the media is not checking later, so if media can be audio, then it's necessary to check it in case of audio

tools/llm_bench/llm_bench_utils/output_csv.py
avasenin-14: if someone has relied on indexes to read the csv, they'd be surprised. This does look logical on the other hand to group similar fields together, so I'm not sure what is the right thing to do here.

avasenin-14 (commented):
I'm good with this PR as soon as you resolve the sbalandi comment about tasks tests.

tools/llm_bench/llm_bench_utils/output_csv.py
krzyczar: ok removed

krzyczar: ok i added READEME lets discuss about this

tools/llm_bench/llm_bench_utils/output_json.py
```diff
@@ -72,6 +75,8 @@ def write_result(report_file, model, framework, device, model_args, iter_data_li
             "detokenization_time": detokenization_time,
             "prompt_idx": iter_data["prompt_idx"],
             "chat_idx": iter_data["chat_idx"],
+            "prompt_repr": iter_data.get("prompt_repr", ""),
+            "input_tokens": iter_data.get("input_tokens", ""),
```
sbalandi, line 79: what the different between input_tokens and input_size ?

sbalandi: and why it is divided to batch size ? 
If a batch is specified, the model will take a prompt repeated batch_size times, it will work on a specific amount of data. I would remove this field, but if you think we really need to have data without a batch, it can be reflected in name somehow, like _before_batch, _no_batched

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -9,26 +9,22 @@
...
+import llm_bench_utils.metrics_print as metrics_print
 from pathlib import Path
+from dataclasses import dataclass, field
+from typing import Callable, Optional, Union
 import openvino as ov
 import math
 import cv2
 
 
-def get_text_prompt(args):
```
avasenin-14, line 19: I think https://github.com/krzyczar/openvino.genai/blob/6f5ccc18163f2f98d4b5f5cd04c1f907ce069126/tools/llm_bench/task/text_generation_chat.py#L586 still uses that function, we probably need to remove it from there too.

avasenin-14 (commented):
In general I'm good with it.
Please check out Sofia's comments, and also pls try to run some small job to make sure it is working around as it was before?

tools/llm_bench/llm_bench_utils/prompt_utils.py
```diff
@@ -97,86 +85,709 @@ def load_image_genai(image_path):
...
+                            )
+            else:
+                raw_list = parse_text_json_data(output_data_list)
+        else:
+            # For speech_to_text the raw value is an audio-file path which
+            # must be stored under 'media' (not 'prompt').
+            if task == "speech_to_text":
+                raw_list = [{"media": item} for item in output_data_list]
+            else:
+                raw_list = output_data_list
```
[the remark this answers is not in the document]
krzyczar: Copilot proposed mirroring (entry["audio"] = entry["media"], keeping both). The code renames instead. I think renaming is the better of the two:

  extract_prompt_data calls load_image() on whatever sits under media. If speech prompts kept a media key holding a .wav path, that becomes a hazard for anything that routes a speech prompt through image-aware
  code — the Qwen3-Omni path does send speech prompts into the VLM benchmark, and only avoids the problem because it explicitly rebuilds entries as {"prompt": ..., "audio": ...}. That's a latent trap rather
  than a live bug, but renaming removes the ambiguity at the source instead of depending on every downstream reader to remember the not self.get("audio") guard.

  The guard is still worth keeping, and is kept — it's load-bearing for a genuinely multimodal entry that carries both an image and audio, where the rename doesn't apply.

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: Good catch — you're right that the guard skips media without ever checking it, which makes its own justification incoherent.
  
I went through the producers: media is never an audio path. _PROMPT_SPECS["speech_to_text"] routes audio to the audio key on both branches (rename={"media": "audio"} for JSONL, nonjson_wrap for CLI), and parse_speech_json_data never emits an audio key of its own — so a speech prompt has audio and no media. Every other producer of media stores an image (visual_text_gen, visual_text_gen_chat, text_embed, image_gen, and ldm_super_resolution via rename={"prompt": "media"}).

So the guard never fired for speech, and instead broke the case it wasn't considering: a VLM prompt can carry an image and audio at once — PROMPT.md §5 lists them as independent keys, parse_vlm_json_data copies them independently, and extract_prompt_data loads both. The guard silently dropped the image:
 
  ▎ before:  text:2w + audio:30.0s@44100Hz
  ▎ after:   text:2w + image:512x384 + audio:30.0s@44100Hz
  
 Fixed — the two modalities are now probed and rendered independently, and the misleading comment is gone. Speech, super-resolution and image-only VLM prompts are unchanged.

tools/llm_bench/llm_bench_utils/output_json.py
krzyczar: Agreed — removing it. prompt_repr stays, since word counts and media dimensions aren't derivable from any other column. The README section shrinks accordingly

tools/llm_bench/llm_bench_utils/prompt_utils.py
krzyczar: Right — text_generation_chat.py and visual_language_generation_chat.py were the two holdouts, which is exactly the problem: dropping get_text_prompt while they still used it is what broke --task text_gen_chat. Both are now migrated to BenchPrompter,

PR #4105, opened 2026-07-08
Covierity fixes (July)

sgonorov:
Coverity bug and performance fixess

[CVS-190263](https://jira.devtools.intel.com/browse/CVS-190263)

### DIFF
--- a/samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
+++ b/samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
@@ -1,6 +1,9 @@
 // Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
+#include <iomanip>
+#include <iostream>
+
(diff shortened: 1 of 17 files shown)


samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
```diff
@@ -51,6 +54,8 @@ int main(int argc, char* argv[]) try {
 
     std::cout << result << "\n";
 
+    const std::ios_base::fmtflags cout_flags = std::cout.flags();
+    const std::streamsize cout_precision = std::cout.precision();
```
as-suvorov, line 58: Let's revert this one and mark coverity issue as false positive or something.
I think that such restoring doesn't make sense in a sample. I prefer to keep it as simple and readable as possible.

as-suvorov (changes requested):

Wovchena:
Moved to https://github.com/openvinotoolkit/openvino.genai/pull/4119

PR #4129, opened 2026-07-10
Restore Python < 3.11 compatibility for av

skuros:
## Description

CVS-190670

The dependency set was resolving av 18.0.0, and av 18 no longer supports Python 3.10.
Failing convert jobs in this path still run on Python 3.10, so installation failed at av first, which left the environment partially installed and caused downstream runtime errors later in the pipeline.
Pinning av 17.1.0 for Python 3.10 restores successful dependency installation and makes those jobs deterministic again.
For newer runtimes, Python 3.11+ can still use av 18.0.0, so this is backward-compatible and does not block forward upgrades.

## Checklist:


sbalandi:
Are some tests crashing? Could you please update the description with some explanation ?

sbalandi (approved):

PR #4160, opened 2026-07-15
Fix undefined symbol for format<int> in libopenvino_genai.so

jwludzik:
## Description
Move the `format()` template definition from gguf.cpp into gguf.hpp as `inline` to eliminate a cross-TU vague linkage dependency.

GCC's IPA-SRA pass under -O3 can localize the instantiation inside gguf.cpp.o, leaving no externally visible symbol for building_blocks.cpp.o to link against at runtime.

This fixes building wheel package with libopenvino_genai.so in Ubuntu26.04 for NPU project

## Checklist:

### DIFF
--- a/src/cpp/src/gguf_utils/gguf.cpp
+++ b/src/cpp/src/gguf_utils/gguf.cpp
@@ -13,18 +13,6 @@
 // https://github.com/antirez/gguf-tools/blob/af7d88d808a7608a33723fba067036202910acb3/gguflib.h#L102-L108
 constexpr int gguf_array_header_size = 12;
 
-template <typename... Args>
-std::string format(std::string fmt, Args... args) {
-    size_t bufferSize = 1000;
-    char* buffer = new char[bufferSize];
-    int n = sprintf(buffer, fmt.c_str(), args...);
-    assert(n >= 0 && n < (int)bufferSize - 1);
-
-    std::string fmtStr(buffer);
-    delete[] buffer;
-    return fmtStr;
-}
-
 std::optional<uint32_t> dtype_to_gguf_tensor_type(const ov::element::Type& dtype) {
     switch (dtype) {
     case ov::element::f64:
--- a/src/cpp/src/gguf_utils/gguf.hpp
+++ b/src/cpp/src/gguf_utils/gguf.hpp
@@ -28,7 +28,16 @@ using GGUFLoad = std::tuple<std::unordered_map<std::string, GGUFMetaData>,
                             std::unordered_map<std::string, gguf_tensor_type>>;
 
 template <typename... Args>
-std::string format(std::string fmt, Args... args);
+std::string format(std::string fmt, Args... args) {
+    size_t bufferSize = 1000;
+    char* buffer = new char[bufferSize];
+    int n = sprintf(buffer, fmt.c_str(), args...);
+    OPENVINO_ASSERT(n >= 0 && n < (int)bufferSize - 1);
+
+    std::string fmtStr(buffer);
+    delete[] buffer;
+    return fmtStr;
+}
 
 ov::Shape get_shape(const gguf_tensor& tensor);
 


src/cpp/src/gguf_utils/gguf.hpp
```diff
@@ -28,7 +28,16 @@ using GGUFLoad = std::tuple<std::unordered_map<std::string, GGUFMetaData>,
                             std::unordered_map<std::string, gguf_tensor_type>>;
 
 template <typename... Args>
-std::string format(std::string fmt, Args... args);
+inline std::string format(std::string fmt, Args... args) {
```
olpipi: inline key word is not needed for template functions
```suggestion
std::string format(std::string fmt, Args... args) {
```

src/cpp/src/gguf_utils/gguf.cpp
```diff
@@ -13,18 +13,6 @@
 // https://github.com/antirez/gguf-tools/blob/af7d88d808a7608a33723fba067036202910acb3/gguflib.h#L102-L108
 constexpr int gguf_array_header_size = 12;
 
-template <typename... Args>
-std::string format(std::string fmt, Args... args) {
```
olpipi, line 17: As I can see format() function is used in src/cpp/src/gguf_utils/gguf.cpp only.
So you probably have no reason to move it to header. Or you plan to reuse it in other translation units?

olpipi (commented):
Please check how much does it affect binary size.

src/cpp/src/gguf_utils/gguf.cpp
jwludzik: It is used in [src/cpp/src/gguf_utils/building_blocks.cpp#L877](https://github.com/openvinotoolkit/openvino.genai/blob/c637ed85efebf1a44d5f0433845849a2d80b353c/src/cpp/src/gguf_utils/building_blocks.cpp#L877)

src/cpp/src/gguf_utils/gguf.hpp
jwludzik: I agree, I will delete `inline`

jwludzik:
> Please check how much does it affect binary size.

In Ubuntu24.04 the difference between libopenvino_genai.so with and without patch is 4 bytes. The size is decreased by 4 bytes with a patch.

src/cpp/src/gguf_utils/gguf.hpp
jwludzik: Done

src/cpp/src/gguf_utils/gguf.hpp
```diff
@@ -28,7 +28,16 @@ using GGUFLoad = std::tuple<std::unordered_map<std::string, GGUFMetaData>,
                             std::unordered_map<std::string, gguf_tensor_type>>;
 
 template <typename... Args>
-std::string format(std::string fmt, Args... args);
+std::string format(std::string fmt, Args... args) {
+    size_t bufferSize = 1000;
+    char* buffer = new char[bufferSize];
+    int n = sprintf(buffer, fmt.c_str(), args...);
+    assert(n >= 0 && n < (int)bufferSize - 1);
```
as-suvorov:
```suggestion
    OPENVINO_ASSERT(n >= 0 && n < (int)bufferSize - 1);
```

as-suvorov (changes requested):

src/cpp/src/gguf_utils/gguf.hpp
jwludzik: Done

olpipi (approved):
LGTM

as-suvorov (approved):

jwludzik:
@as-suvorov , @olpipi , can you help me to merge this change? I am not authorized to merge it. Thanks in advance

as-suvorov:
build_jenkins

PR #4170, opened 2026-07-17
Fix stack buffer overflow in GGUF loader (ndim bounds check)

TianmengChen:
A crafted .gguf file can trigger a stack-based buffer overflow in the vendored gguflib gguf_get_tensor(): it reads the per-tensor dimension count (ndim) as a 32-bit value straight from the file and the only bound is assert(ndim <= GGUF_TENSOR_MAX_DIM), which is compiled out under NDEBUG (release builds, shipped PyPI wheels). The subsequent loop then writes ndim 8-byte values past the fixed uint64_t dim[8] member of the caller's stack-allocated gguf_tensor, overwriting saved registers, the return address and the stack canary with attacker-controlled values (CWE-787). Reachable pre-auth at model-load time in the default build via LLMPipeline(path) / Tokenizer(path) / read_model(path).

Fixes:
- Patch the fetched gguflib so gguf_get_tensor() performs a hard runtime check on ndim and fails the read instead of overflowing. The overwrite completes inside gguf_get_tensor() before the caller regains control, so this must be fixed in the library itself. The patch is applied idempotently at FetchContent time.
- Add a defense-in-depth bounds check in load_arrays() so an out-of-bounds dimension count can never reach get_shape().

CVS-188917


src/cpp/cmake/patches/gguflib-ndim-bounds-check.patch
```diff
@@ -0,0 +1,18 @@
...
+     tensor->ndim = *num_dim;
++    /* Security hardening (OpenVINO GenAI): reject a tensor whose declared
++     * dimension count exceeds the fixed-size dim[] array. The assert() below
++     * is compiled out under NDEBUG (release builds, shipped wheels), so without
++     * this hard runtime check a crafted .gguf with ndim > GGUF_TENSOR_MAX_DIM
++     * overflows the caller's stack-allocated gguf_tensor. Fail the read instead. */
++    if (tensor->ndim > GGUF_TENSOR_MAX_DIM) {
++        tensor->name = NULL;
++        return 0;
++    }
```
Wovchena, line 15: The original GGUF repo should be patched instead. GenAI can be updated to fetch the patched version after that

TianmengChen: If so, we can wait until original repo accept relative PR described in JIRA [CVS-188917](https://jira.devtools.intel.com/browse/CVS-188917), otherwise we should fork the original repo.

Wovchena: Yeah, I've read the ticket AFTER I started this conversation :( So we wait for Jaroslaw's reply there

as-suvorov:
@TianmengChen Thank you for the PR! We will wait for gguf-utils fix. If patching will be required on GenAI side we will re-open this PR. Closing PR for now.

PR #4175, opened 2026-07-17
[WWB] Split 43 minute tests suite

avasenin-14:
## Description

[CVS-190881](https://jira.devtools.intel.com/browse/CVS-190881)

The main purpose of this PR is to split the long running (40-45 minutes) test in CI suite. I've split the test suite in two, moving the images, video and vlm tests to a separate step that would run in parallel. Locally the split went ~20 minutes for both.

Optional change: registered marks in WWB tests to avoid warnings about unregistered marks:
```python
  C:\actions-runner\_work\openvino.genai\openvino.genai\src\tools\who_what_benchmark\tests\test_cli_vlm.py:227: PytestUnknownMarkWarning: Unknown pytest.mark.transformers_dependent - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    @pytest.mark.transformers_dependent(
```

## Checklist:


sbalandi (approved):

tools/who_what_benchmark/tests/pytest.ini
```diff
@@ -0,0 +1,7 @@
+[pytest]
+
+markers =
+    nanollava: VLM tests that require the nanoLLaVA stack (older transformers/diffusers).
+    transformers_lower_v5: tests that require transformers < 5.0.
+    transformers_dependent: tests sensitive to the installed transformers version.
+    wwb_rerank: reranking benchmark tests.
```
as-suvorov: Marker not used, let's remove it from test as well
```suggestion
```

as-suvorov (approved):

tools/who_what_benchmark/tests/pytest.ini
avasenin-14: Done, also removed the mark from tools/who_what_benchmark/test_cli_reranking.py test, as it wasn't used in CI.

avasenin-14:
Copilot is telling that pytest.ini won't be picked up from repo dir, but it is not true. The search goes from test upward in directories until the config is found (https://docs.pytest.org/en/stable/reference/customize.html#rootdir), in which case it should be determined correctly.
Also tested locally from repo root - it does supress warnings as expected.

PR #4185, opened 2026-07-20
Add HF CPU fallback for visual-text loaders

sswierze:
Normalizes the HF visual-text loader behavior to follow the existing text fallback path on non-CUDA devices, avoiding invalid GPU device-map handling for Intel GPU runs in nat_vs_genai.

This keeps the change minimal and localized to the who_what_benchmark visual-text HF loader path.

### DIFF
--- a/tools/who_what_benchmark/whowhatbench/model_loaders.py
+++ b/tools/who_what_benchmark/whowhatbench/model_loaders.py
@@ -72,6 +72,12 @@ def _sanitize_load_kwargs(model_type, use_hf, use_genai, use_llamacpp, kwargs):
     return sanitized_kwargs
 
 
+def _get_hf_device_map(device: str) -> str:
+    if not torch.cuda.is_available() or device.strip().lower() == "cpu":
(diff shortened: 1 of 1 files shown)


tools/who_what_benchmark/whowhatbench/model_loaders.py
```diff
@@ -72,6 +72,12 @@ def _sanitize_load_kwargs(model_type, use_hf, use_genai, use_llamacpp, kwargs):
     return sanitized_kwargs
 
 
+def _get_hf_device_map(device: str) -> str:
+    if not torch.cuda.is_available() or device.strip().lower() == "cpu":
```
sbalandi, line 76: if device will not be set in cmd it will be set to CPU. If user will set cuda to device, after your pr, pipeline will be reset to CPU anyway, which is not expected behavior. 
So, changes should be done in cmd , not in wwb , please, update script for running wwb

sbalandi (changes requested):

PR #4212, opened 2026-07-23
[llm_bench] Fix KeyError: 'chat_idx' for text_embed/text_rerank CSV report

dylanneve1:
### Problem

Running `llm_bench` with `-r` (CSV report) on `--task text_embed` or `--task text_rerank` crashes **after** inference completes:

```
File "tools/llm_bench/llm_bench_utils/output_csv.py", line 158, in gen_data_to_csv
    result["chat_idx"] = iter_data["chat_idx"]
KeyError: 'chat_idx'
```

`gen_data_to_csv` reads `iter_data["chat_idx"]` unconditionally. That key is only populated by the text-generation chat pipeline (`gen_iterate_data(chat_idx="")`); the embedding task (`embed_iterate_data`) and the reranking pipeline build their `iter_data` without it. So embedding/reranking benchmarks run to completion, produce valid results, then fail in the report writer.

The writer is only reached when `-r` is passed, which is why the existing `text_embed`/`text_rerank` sample tests (run without `-r`) don't catch it.

### Fix

Read `chat_idx` defensively via `.get("chat_idx", "")`, matching the `""` default the text-gen path already uses.

### Testing

Reproduced the `KeyError` by driving `embed_iterate_data` → `gen_data_to_csv` directly (no model/device needed); confirmed the row writes cleanly with the fix.


sbalandi (approved):

sbalandi:
@dylanneve1 thank you for fixing !

PR #4213, opened 2026-07-23
Use single tokenizer for VLM CB backend. Gemma4-12b-it template fix

as-suvorov:
## Description

Fix Gemma 4 chat templates containing Python-style implicit multiline string concatenation, which minja does not support.

The Gemma 4 inputs embedder now normalizes adjacent string literals before parsing. Continuous Batching also passes its existing tokenizer to the VLM inputs embedder, ensuring that the patched template is shared and avoiding duplicate tokenizer loading.

### Performance

Local Debug build, Gemma 4 12B, CPU, five runs after warmup:

| Revision | Mean load time | Median |
|---|---:|---:|
| `master` | 1575.2 ms | 1574 ms |
| This PR | 1504.6 ms | 1499 ms |

Loading time improved by **70.6 ms (4.48%)** on average.

Validated by building and running `benchmark_vlm` with the Gemma 4 Continuous Batching pipeline. No documentation changes are required because the public API is unchanged.

CVS-191107

## Checklist:
- [NA] Tests have been updated or added to cover the new code. - tested with current scope

### DIFF
--- a/src/cpp/src/visual_language/inputs_embedder.hpp
+++ b/src/cpp/src/visual_language/inputs_embedder.hpp
@@ -35,6 +35,11 @@ struct NormalizedPrompt {
 class InputsEmbedder {
 public:
     InputsEmbedder(const std::filesystem::path& model_dir,
+                   const Tokenizer& tokenizer,
+                   const std::string& device,
+                   const ov::AnyMap device_config);
+
+    InputsEmbedder(const std::filesystem::path& model_dir,
                    const std::string& device,
                    const ov::AnyMap device_config);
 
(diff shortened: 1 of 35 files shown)


src/cpp/src/visual_language/inputs_embedder.hpp
```diff
@@ -34,6 +34,11 @@ struct NormalizedPrompt {
...
 class InputsEmbedder {
 public:
+    InputsEmbedder(const std::filesystem::path& model_dir,
+                   const Tokenizer& tokenizer,
+                   const std::string& device,
+                   const ov::AnyMap device_config);
+
     InputsEmbedder(const std::filesystem::path& model_dir,
                    const std::string& device,
                    const ov::AnyMap device_config);
```
apaniukov, line 44: I think we can replace the old constructor instead of adding another one.

yatarkan (approved):

src/cpp/src/visual_language/inputs_embedder.hpp
as-suvorov: To be addressed in further PRs

PR #4218, opened 2026-07-23
[OMNI] Finalize Talker public API and fix ChatHistory, volume, and chunk-size issues

sgonorov:
## Description

This PR finalizes the public `Talker` API for Qwen3-Omni speech generation and fixes several bugs found while testing the speech path end to end.

Talker public API (CVS-191234):
- Add `Talker::get_speech_config()` / `set_speech_config()` 
- Add a property-bag `generate(vlm_result, ov::AnyMap properties = {})` overload on `TalkerBase`
- Add a `ModelsMap` constructor `Talker(models_map, config, config_dir_path, device_mapping, properties)` for blob deployment and per-submodel device placement

Bugs fixed on this branch:
- ChatHistory + speech produced different Thinker text than the text-only path. Enabling `return_audio` applied the chat template before multimodal normalization, then routed the string through the prompt overload and misplaced image/audio tokens. Speech ChatHistory now stays on the native history path, and the continuous-batching ChatHistory path captures the prompt-token slice and forwards hidden states when `return_omni_outputs` is set.
- Volume came out wrong because the next talker-step input added the first-code embedding twice. `cp_embed_sum` already includes it, so removing the redundant add gives correct audio amplitude.
- Last, the `audio_chunk_frames` docstring claimed a fixed 1920 samples per frame. It isn't fixed. The code2wav vocoder trims its convolutional warmup from the first frame of each decode call, so a chunk of N frames yields `1920*N - 555` samples. Corrected the C++ header, the pybind docstring, and the `.pyi` stub.

CVS-191234

## Checklist:
- [x] Tests have been updated or added to cover the new code. Added to `tests/python_tests/test_omni_pipeline.py`: `test_speech_config_accessors_are_talker_only` (accessors present on `Talker`, absent on `TalkerBase`), `test_talker_blob_ctor_signature` (the `ModelsMap` overload resolves), and `TestOmniChatHistorySpeechParity` (a `real_models`-marked test asserting identical Thinker text with `return_audio` off vs on).
- [x] This PR fully addresses the ticket. The Talker public API from CVS-191234 is complete. The AnyMap `generate` overload is implemented in C++ but not yet bound to Python, since the talker's `generate` is not exposed to Python (OmniPipeline drives generation).
- [x] I have made corresponding changes to the documentation. Updated the `audio_chunk_frames` docstring across the C++ header, the pybind binding, and the `.pyi` stub.

### DIFF
--- a/src/cpp/include/openvino/genai/omni/talker.hpp
+++ b/src/cpp/include/openvino/genai/omni/talker.hpp
@@ -72,9 +72,10 @@ public:
                                   const OmniSpeechStreamerVariant& speech_streamer = std::monostate{}) = 0;
 
     /// @brief Property-bag speech generation. `properties` recognizes `speech_streamer` plus any
-    /// `OmniTalkerSpeechConfig` field; unspecified fields fall back to the talker's stored config
-    /// (see the concrete backend's get/set_speech_config). Convenience wrapper over the typed
-    /// overload for callers that build config from an ov::AnyMap.
+    /// `OmniTalkerSpeechConfig` field. How unspecified fields are resolved is backend-defined; the
+    /// default `Talker` falls back to its stored config (see `Talker::get_speech_config` /
+    /// `set_speech_config`). Convenience wrapper over the typed overload for callers that build
+    /// config from an ov::AnyMap.
     virtual TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {}) = 0;
 
     /// @brief List names of speakers exposed by this talker.
@@ -122,7 +123,8 @@ public:
     /// @param config Stored default speech config (see get/set_speech_config).
     /// @param config_dir_path Directory with `config.json` (codec/token IDs, speakers) and optional
     ///        `generation_config.json` (sampling defaults).
-    /// @param device_mapping Submodel name -> device. Submodels absent from the map load on "CPU".
+    /// @param device_mapping Submodel name -> device. Entries absent from this map fall back to
+    ///        "CPU"; submodels absent from `models_map` remain unavailable and fail the availability check.
     /// @param properties Passed to ov::Core::compile_model() for every submodel.
     Talker(const ModelsMap& models_map,
                     const OmniTalkerSpeechConfig& config,
--- a/src/cpp/include/openvino/genai/omni/talker.hpp
+++ b/src/cpp/include/openvino/genai/omni/talker.hpp
@@ -4,11 +4,13 @@
 #pragma once
 
 #include <filesystem>
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
 
 #include "openvino/core/any.hpp"
+#include "openvino/genai/common_types.hpp"
 #include "openvino/genai/omni/speech_streamer_base.hpp"
 #include "openvino/genai/omni/talker_perf_metrics.hpp"
 #include "openvino/genai/omni/talker_speech_config.hpp"
@@ -69,6 +71,12 @@ public:
                                   const OmniTalkerSpeechConfig& talker_speech_config,
                                   const OmniSpeechStreamerVariant& speech_streamer = std::monostate{}) = 0;
 
+    /// @brief Property-bag speech generation. `properties` recognizes `speech_streamer` plus any
+    /// `OmniTalkerSpeechConfig` field; unspecified fields fall back to the talker's stored config
+    /// (see the concrete backend's get/set_speech_config). Convenience wrapper over the typed
+    /// overload for callers that build config from an ov::AnyMap.
+    virtual TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {}) = 0;
+
     /// @brief List names of speakers exposed by this talker.
     /// Returns an empty vector when the backend does not enumerate named speakers.
     virtual std::vector<std::string> list_speakers() const = 0;
@@ -108,14 +116,38 @@ public:
                     const std::string& device,
                     const ov::AnyMap& properties = {});
 
+    /// @brief Construct from in-memory model IRs (blob deployment / per-submodel device placement).
+    /// @param models_map Model name -> (IR string, weights tensor). Keys: `text_embeddings`,
+    ///        `talker`, `talker_text_embeddings`, `talker_projections`, `code_predictor`, `code2wav`.
+    /// @param config Stored default speech config (see get/set_speech_config).
+    /// @param config_dir_path Directory with `config.json` (codec/token IDs, speakers) and optional
+    ///        `generation_config.json` (sampling defaults).
+    /// @param device_mapping Submodel name -> device. Submodels absent from the map load on "CPU".
+    /// @param properties Passed to ov::Core::compile_model() for every submodel.
+    Talker(const ModelsMap& models_map,
+                    const OmniTalkerSpeechConfig& config,
+                    const std::filesystem::path& config_dir_path,
+                    const std::map<std::string, std::string>& device_mapping,
+                    const ov::AnyMap& properties = {});
+
     ~Talker() override;
 
     TalkerResults generate(const VLMDecodedResults& vlm_result,
                           const OmniTalkerSpeechConfig& talker_speech_config,
                           const OmniSpeechStreamerVariant& speech_streamer = std::monostate{}) override;
 
+    TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {}) override;
+
     std::vector<std::string> list_speakers() const override;
     ov::Tensor get_speaker_embedding(const std::string& name) const override;
+
+    /// @brief Return the talker's stored default speech config. Seeds the AnyMap generate()
+    /// overload: fields absent from its `properties` fall back to this config.
+    OmniTalkerSpeechConfig get_speech_config() const;
+
+    /// @brief Set the talker's stored default speech config (validated).
+    /// @throws when `config` fails OmniTalkerSpeechConfig validation.
+    void set_speech_config(const OmniTalkerSpeechConfig& config);
 
 private:
     class Impl;
--- a/src/cpp/include/openvino/genai/omni/talker.hpp
+++ b/src/cpp/include/openvino/genai/omni/talker.hpp
@@ -174,7 +174,6 @@ public:
 private:
     class Impl;
     std::unique_ptr<Impl> m_impl;
-    OmniTalkerSpeechConfig m_speech_config;
 };
 
 }  // namespace ov::genai
--- a/src/cpp/include/openvino/genai/omni/talker.hpp
+++ b/src/cpp/include/openvino/genai/omni/talker.hpp
@@ -48,6 +48,9 @@ struct OPENVINO_GENAI_EXPORTS TalkerResults {
  * The default implementation `Talker` (declared in this header) wraps the
  * existing Qwen3-Omni Talker + CodePredictor + Code2Wav stack.
  *
+ * Pure interface: no data members and no default implementations, so a backend is free to
+ * store (or not store) a default speech config however it likes.
+ *
  * @note This is a preview API and is subject to change.
  */
 class OPENVINO_GENAI_EXPORTS TalkerBase {
@@ -75,32 +78,24 @@ public:
                                    const OmniTalkerSpeechConfig& talker_speech_config,
                                    const OmniSpeechStreamerVariant& speech_streamer = std::monostate{}) = 0;
 
-    /// @brief Property-bag convenience wrapper over the typed generate(). `properties` recognizes
-    /// `speech_streamer`, a whole `talker_speech_config`, and the individual `OmniTalkerSpeechConfig`
-    /// fields (`return_audio`, `speaker`, `audio_chunk_frames`, ...). Unrecognized keys throw.
+    /// @brief Property-bag form of generate(). `properties` recognizes `speech_streamer`, a whole
+    /// `talker_speech_config`, and the individual `OmniTalkerSpeechConfig` fields (`return_audio`,
+    /// `speaker`, `audio_chunk_frames`, ...). Unrecognized keys throw.
     ///
-    /// Fields absent from `properties` fall back to the stored default config (`get_speech_config()`);
-    /// recognized entries overlay it. Not virtual — parsing lives here so custom backends get the
-    /// property-bag entry point for free without reimplementing it.
+    /// Fields absent from `properties` fall back to the backend's stored default config
+    /// (`get_speech_config()`); recognized entries overlay it. A backend that recognizes a
+    /// different property set is free to parse `properties` its own way.
     ///
     /// @note `generate(vlm, {})` is ambiguous against the typed overload; pass `ov::AnyMap{}` explicitly.
-    TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {});
+    virtual TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties) = 0;
 
-    /// @brief Generate against a VLM result and an arbitrary number of ov::Property instances.
-    /// Sugar over the property-bag overload: `generate(vlm, ov::genai::speech_streamer(s))`.
-    template <typename... Properties>
-    ov::util::EnableIfAllStringAny<TalkerResults, Properties...> generate(const VLMDecodedResults& vlm_result,
-                                                                          Properties&&... properties) {
-        return generate(vlm_result, ov::AnyMap{std::forward<Properties>(properties)...});
-    }
-
-    /// @brief Return the talker's stored default speech config. Seeds the property-bag generate()
+    /// @brief Return the backend's stored default speech config. Seeds the property-bag generate()
     /// overload: fields absent from its `properties` fall back to this config.
-    OmniTalkerSpeechConfig get_speech_config() const;
+    virtual OmniTalkerSpeechConfig get_speech_config() const = 0;
 
-    /// @brief Set the talker's stored default speech config (validated).
+    /// @brief Set the backend's stored default speech config.
     /// @throws when `config` fails OmniTalkerSpeechConfig validation.
-    void set_speech_config(const OmniTalkerSpeechConfig& config);
+    virtual void set_speech_config(const OmniTalkerSpeechConfig& config) = 0;
 
     /// @brief List names of speakers exposed by this talker.
     /// Returns an empty vector when the backend does not enumerate named speakers.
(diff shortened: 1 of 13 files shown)


src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -108,15 +125,40 @@ class OPENVINO_GENAI_EXPORTS Talker : public TalkerBase {
...
+    /// @param models_map Model name -> (IR string, weights tensor). Keys: `text_embeddings`,
+    ///        `talker`, `talker_text_embeddings`, `talker_projections`, `code_predictor`, `code2wav`.
+    /// @param config Stored default speech config (see get/set_speech_config).
+    /// @param config_dir_path Directory with `config.json` (codec/token IDs, speakers) and optional
+    ///        `generation_config.json` (sampling defaults).
+    /// @param device_mapping Submodel name -> device. Entries absent from this map fall back to
+    ///        "CPU"; submodels absent from `models_map` remain unavailable and fail the availability check.
+    /// @param properties Passed to ov::Core::compile_model() for every submodel.
+    Talker(const ModelsMap& models_map,
+                    const OmniTalkerSpeechConfig& config,
```
Wovchena: Fix the alignment

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -108,15 +125,40 @@ class OPENVINO_GENAI_EXPORTS Talker : public TalkerBase {
...
+                    const std::map<std::string, std::string>& device_mapping,
+                    const ov::AnyMap& properties = {});
+
     ~Talker() override;
 
     TalkerResults generate(const VLMDecodedResults& vlm_result,
                           const OmniTalkerSpeechConfig& talker_speech_config,
                           const OmniSpeechStreamerVariant& speech_streamer = std::monostate{}) override;
 
+    TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {}) override;
```
Wovchena, line 158: Add templated overload

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -108,15 +125,40 @@ class OPENVINO_GENAI_EXPORTS Talker : public TalkerBase {
...
     std::vector<std::string> list_speakers() const override;
     ov::Tensor get_speaker_embedding(const std::string& name) const override;
 
+    /// @brief Return the talker's stored default speech config. Seeds the AnyMap generate()
+    /// overload: fields absent from its `properties` fall back to this config.
+    OmniTalkerSpeechConfig get_speech_config() const;
+
+    /// @brief Set the talker's stored default speech config (validated).
+    /// @throws when `config` fails OmniTalkerSpeechConfig validation.
+    void set_speech_config(const OmniTalkerSpeechConfig& config);
```
Wovchena: Add virtual `get_speech_config()` and `set_speech_config()` to `TalkerBase` and mark here as override

src/cpp/src/omni/talker.cpp
```diff
@@ -34,8 +36,51 @@ std::vector<ov::Tensor> split_hidden_states(const std::vector<ov::Tensor>& per_s
     return per_token;
 }
 
+/// @brief Resolve a talker property-bag into a typed config plus optional streamer.
+/// Starts from `base` (the caller's default), overlays recognized `properties`, and rejects
+/// unknown keys so typos surface immediately instead of being silently dropped.
+struct ResolvedTalkerProperties {
+    OmniTalkerSpeechConfig config;
+    OmniSpeechStreamerVariant speech_streamer{std::monostate{}};
```
Wovchena:
```suggestion
    OmniSpeechStreamerVariant speech_streamer;
```
Not used value

src/cpp/src/omni/talker.cpp
```diff
@@ -34,8 +36,51 @@ std::vector<ov::Tensor> split_hidden_states(const std::vector<ov::Tensor>& per_s
...
+ResolvedTalkerProperties resolve_talker_properties(const OmniTalkerSpeechConfig& base, const ov::AnyMap& properties) {
+    ResolvedTalkerProperties out{base, std::monostate{}};
+    ov::AnyMap leftover;
+    for (const auto& [key, value] : properties) {
+        if (key == ov::genai::speech_streamer.name()) {
+            out.speech_streamer = value.as<OmniSpeechStreamerVariant>();
+        } else if (key == ov::genai::talker_speech_config.name()) {
+            out.config = value.as<OmniTalkerSpeechConfig>();
+        } else {
+            OPENVINO_ASSERT(is_omni_talker_speech_config_key(key),
```
Wovchena, line 65: `is_omni_talker_speech_config_key()` lists many more keys but they are ignored in `resolve_talker_properties()`

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -65,9 +68,39 @@ class OPENVINO_GENAI_EXPORTS TalkerBase {
...
+    /// property-bag entry point for free without reimplementing it.
+    ///
+    /// @note `generate(vlm, {})` is ambiguous against the typed overload; pass `ov::AnyMap{}` explicitly.
+    TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {});
+
+    /// @brief Generate against a VLM result and an arbitrary number of ov::Property instances.
+    /// Sugar over the property-bag overload: `generate(vlm, ov::genai::speech_streamer(s))`.
+    template <typename... Properties>
+    ov::util::EnableIfAllStringAny<TalkerResults, Properties...> generate(const VLMDecodedResults& vlm_result,
+                                                                          Properties&&... properties) {
```
Wovchena: Templates don't make sense for interfaces because you can't have an instance of `TalkerBase`

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -79,6 +112,11 @@ class OPENVINO_GENAI_EXPORTS TalkerBase {
     /// `OmniTalkerSpeechConfig::speaker` (the Tensor alternative of the variant).
     /// @throws when the backend has no named speakers, or `name` doesn't match.
     virtual ov::Tensor get_speaker_embedding(const std::string& name) const = 0;
+
+protected:
+    /// Stored default speech config shared by every backend. The property-bag generate() overlays
+    /// `properties` on top of it; get/set_speech_config() read and write it.
+    OmniTalkerSpeechConfig m_speech_config;
```
Wovchena: Don't enforce members in `TalkerBase`. Let the users decide if they want it. Same for the getters implementations

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -65,9 +68,39 @@ class OPENVINO_GENAI_EXPORTS TalkerBase {
...
+        return generate(vlm_result, ov::AnyMap{std::forward<Properties>(properties)...});
+    }
+
+    /// @brief Return the talker's stored default speech config. Seeds the property-bag generate()
+    /// overload: fields absent from its `properties` fall back to this config.
+    OmniTalkerSpeechConfig get_speech_config() const;
+
+    /// @brief Set the talker's stored default speech config (validated).
+    /// @throws when `config` fails OmniTalkerSpeechConfig validation.
+    void set_speech_config(const OmniTalkerSpeechConfig& config);
```
Wovchena: `virtual` is missing for 
1. TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties = {});
2. OmniTalkerSpeechConfig get_speech_config() const;
3. void set_speech_config(const OmniTalkerSpeechConfig& config);

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -55,8 +55,19 @@ struct OPENVINO_GENAI_EXPORTS OmniTalkerSpeechConfig {
...
-    /// Must be >= 1. Each frame is 80ms of audio at 24 kHz (1920 samples).
-    std::size_t audio_chunk_frames = 1;
+    /// Must be >= 1. At steady state each frame decodes to 1920 samples (80ms at 24 kHz),
+    /// but the code2wav vocoder trims its convolutional warmup from the first frame of every
+    /// decode call, so a chunk of N frames yields 1920*N - 555 samples, not 1920*N. This is a
+    /// property of the vocoder graph, not a miscount. Larger chunks amortize the fixed warmup
+    /// cost; very small chunks (e.g. 1) also risk audible seams between independently decoded
+    /// chunks in streaming mode.
+    ///
+    /// Default is 4: the per-frame talker+code_predictor decode cost is fixed (~77ms on GPU),
```
Wovchena, line 65: What's the `transformers`'s default?

src/cpp/src/omni/talker.cpp
```diff
@@ -34,8 +36,67 @@ std::vector<ov::Tensor> split_hidden_states(const std::vector<ov::Tensor>& per_s
...
```
Wovchena, line 65: https://github.com/openvinotoolkit/openvino.genai/pull/4218#discussion_r3664876390 is still applicable

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/talker.hpp
sgonorov: Ok, template moved to Talker.

src/cpp/include/openvino/genai/omni/talker.hpp
sgonorov: Done

src/cpp/include/openvino/genai/omni/talker.hpp
sgonorov: Fixed

src/cpp/src/omni/talker.cpp
sgonorov: Removed some dead code. Other leftovers are collected and go to talker speech config update.

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
sgonorov: It doesn't stream as far as i know, so there's no similar defaults.

src/cpp/include/openvino/genai/omni/talker.hpp
```diff
@@ -104,22 +132,49 @@ class OPENVINO_GENAI_EXPORTS Talker : public TalkerBase {
...
+    OmniTalkerSpeechConfig get_speech_config() const override;
+    void set_speech_config(const OmniTalkerSpeechConfig& config) override;
 
     std::vector<std::string> list_speakers() const override;
     ov::Tensor get_speaker_embedding(const std::string& name) const override;
 
 private:
     class Impl;
     std::unique_ptr<Impl> m_impl;
+    OmniTalkerSpeechConfig m_speech_config;
```
Wovchena: Move `m_speech_config` to `Impl`

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
```diff
@@ -90,10 +101,6 @@ struct OPENVINO_GENAI_EXPORTS OmniTalkerSpeechConfig {
     /// @brief CodePredictor top-k override (must be >= 1 when set).
     /// Checkpoint default from `generation_config.json -> cp_top_k` if present.
     std::optional<std::size_t> cp_top_k;
-
-    /// @brief CodePredictor repetition penalty override (must be > 0 when set; 1.0 = no penalty).
-    /// Checkpoint default from `generation_config.json -> cp_repetition_penalty` if present.
-    std::optional<float> cp_repetition_penalty;
```
Wovchena, line 96: Mark it as deprecated and remove in the next major release.
Why is it removed instead of being implemented?

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
```diff
@@ -126,6 +126,53 @@ Qwen3OmniSpeechPipeline::Qwen3OmniSpeechPipeline(const std::filesystem::path& mo
...
+        auto dev_it = device_mapping.find(name);
+        const std::string& device = dev_it != device_mapping.end() ? dev_it->second : default_device;
+
+        // Force FP32 inference precision on GPU (see disk ctor): GPU FP16 shifts talker logits,
+        // changing sampled tokens and corrupting speech length.
+        ov::AnyMap compilation_props = properties;
+        if (device == "GPU" || device.find("GPU") == 0) {
+            compilation_props[ov::hint::inference_precision.name()] = ov::element::f32;
+            GENAI_DEBUG("Speech: forcing FP32 precision for %s on GPU", name.c_str());
+        }
```
Wovchena: Reminder: https://github.com/openvinotoolkit/openvino.genai/pull/4218/changes#r3656714196

Wovchena (changes requested):

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
sgonorov: It's a preview API, so i think we can skip and remove it. It's not really implemented,

src/cpp/include/openvino/genai/omni/talker.hpp
sgonorov: Fixed

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
sgonorov: Fixed

src/cpp/include/openvino/genai/omni/talker_speech_config.hpp
Wovchena: Link a ticket in this conversation to remember to describe the removal in release notes

Wovchena: Unresolved because no ticket is linked.
There's still unanswered question:
> Why is it removed instead of being implemented?

sgonorov: https://jira.devtools.intel.com/browse/CVS-191234 has a comment.
It's not implemented because it's inside the graph now, so we have to also change optimum-intel in order to implement it.

Wovchena: This ticket doesn't state directly that the changes should be described in release notes. Ask Maksim if this ticket is enough

sgonorov: https://jira.devtools.intel.com/browse/CVS-192824

src/cpp/src/visual_language/qwen3_omni/speech_pipeline.cpp
```diff
@@ -1021,6 +1053,25 @@ TalkerResults Qwen3OmniSpeechPipeline::generate_speech(const std::vector<int64_t
         return build_result(ov::Tensor{});
     }
 
+    if (streaming) {
```
Wovchena, line 1056: The streamed_chunks accumulation changes waveforms[0], when a speech streamer is passed, from a single full-sequence codes_to_wav pass to the concatenation. I think that's the right destination, but it needs one prerequisite that isn't in this PR yet, so follow-up is needed.

`code2wav` is a convolutional upsampler. When we call it on `codes[a:b]` in isolation, the model has no history to convolve codes before `a`. It responds by trimming its warmup region, which is exactly the fixed 555-sample shortfall this PR documents so well.

That has two consequences that only show up once you chunk. The shortfall is paid per decode call rather than once per utterance, so concatenating chunks yields audio shorter than the real waveform, with the missing samples sitting at the beginning of each chunk's span — i.e. a discontinuity at every seam rather than a trim at the end. And the samples that are produced near each chunk's start were computed against zero-padding instead of the frames that actually precede them, so they differ from what the full pass would have produced there even after the lengths are reconciled.

What left context does. Instead of decoding codes[s:e], decode codes[s-L:e] for some small number of preceding frames L that we've already emitted audio for, then throw away the leading samples those context frames produce. Concretely, with the warmup trimmed from the front of each call, the decoded output for codes[s-L:e] starts at (s-L)*1920 + 555, so the samples belonging to this chunk begin at offset L*1920 - 555 in it. Emitting from there yields exactly 1920 * chunk_frames samples per chunk.

L is then a quality/cost knob: it wants to be large enough to cover the vocoder's receptive field, and each context frame is re-decoded and discarded, so larger L means more vocoder work per emitted chunk. Worth choosing deliberately rather than defaulting to the smallest value that makes the lengths add up.

This is standard practice for streaming vocoders rather than anything exotic — vLLM-Omni does the same thing under the name codec_left_context_frames, and their implementation is a useful reference for the shape of it: chunked_decode. One caveat if you read it: their comments attribute the trim to the end of the sequence rather than the front, so the slice offset is worth confirming against our exported graph rather than copying.

UPD: Ticket CVS-192826

Wovchena, line 1056: Follow-up on the above: expose the left-context size as a config knob, don't hardcode it

If the left-context prefix lands, I'd suggest the number of context frames becomes a field on OmniTalkerSpeechConfig next to audio_chunk_frames — say audio_chunk_left_context_frames — loaded and validated the same way as the other speech knobs, rather than a constant in speech_pipeline.cpp.

Three reasons it belongs in config:
- It's a property of the vocoder graph, not of our pipeline. Upstream vLLM-Omni treats it exactly this way, as a per-model deploy setting (codec_left_context_frames), and the values differ sharply by model — Qwen3-Omni and Qwen3-TTS use 25, while MiniCPM-o uses 3 with the same chunk size. If we ever load a different speech stack through TalkerBase, this is a number that has to move with it.
- It's the quality/cost knob for chunked streaming. Context frames are re-decoded and then discarded, so raising it buys smoother seams at the price of vocoder work per emitted chunk. That trade depends on the deployment and on audio_chunk_frames, so users should be able to make it without a rebuild.
- It shouldn't be derived from audio_chunk_frames. Upstream deliberately keeps the same context size across very different chunk sizes — the context requirement comes from the receptive field, the chunk size from latency and throughput targets. Two independent knobs.

 For the default I'd suggest 25 as in vLLM, and I think we can justify it without measuring anything: our code2wav is the OpenVINO export of the same Qwen3-Omni weights upstream validated 25 against, and receptive field is a property of the graph, so their validated value carries over. Their contributor guide calls 25 the "validated default from Qwen3-TTS and Qwen3-Omni" (reference).
One implementation note: the value wants clamping to the history that actually exists, since the first chunk has none — upstream does min(configured, available) and passes 0 for the first chunk, which also keeps the leading-sample arithmetic correct for that case.

Wovchena (approved):

PR #4226, opened 2026-07-27
Add eagle3 sample test on windows

sunxiaoxia2022:
## Description

Add eagle3 sample test on windows

CVS-###

Fixes #(issue)

## Checklist:


.github/workflows/mac.yml
```diff
@@ -595,6 +595,10 @@ jobs:
             marker: 'speech_generation'
             cmd: 'tests/python_tests/samples'
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).Speech_generation_samples.test }}
+          - name: 'Eagle3 decoding'
+            marker: 'eagle3_decoding'
+            cmd: 'tests/python_tests/samples'
+            run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).speculative_decoding.test }}
```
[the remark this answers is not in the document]
Wovchena: I forgot that Mac already have a comment near xfail:
> Update mac.yml to run the test separately similar to linux.yml after resolution

Sorry.

Please keep Windows only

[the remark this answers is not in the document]
sunxiaoxia2022: removed.

samples/cpp/text_generation/speculative_decoding_lm.cpp
```diff
@@ -12,7 +12,7 @@ int main(int argc, char* argv[]) try {
     }
 
     ov::genai::GenerationConfig config;
-    config.max_new_tokens = 100;
+    config.max_new_tokens = 40;
```
Wovchena: Is this because greedy and speculative decoding results don't match?

linux.yml had this test enabled for a while, so the failure may indicate there's a new bug

Wovchena (approved):

sunxiaoxia2022:
Hi @Wovchena, I reproduced this issue on local PTL windows platform(CPU).
The result is as follows:
image
They differ starting from the 47th token.
I also tested on local XOEN Linux machine. The results from `greedy_causal_lm` and `speculative_decoding_lm` are the same. The result is as follows:

> Okay, the user wants me to add 2 and 3 using the provided code. Let me check the code first. The function add takes two parameters, a and b, and returns their sum. So, if I call add(2, 3), it should return 5.
> Wait, but the question is phrased as "Can you please add 2 and 3?" So maybe they just want the result of the function call. The code is already written,

The results from Linux XEON and windows PTL are different.

I added log includes output token, logit and 2nd token, logit. The data are as follows:
Windows:
image
Linux:
image
From above table, 1st output token and logit from `greedy_causal_lm` and `speculative_decoding_lm` are the same. Because they use the same kernel. Although from 2nd to the 46th tokens are the same on windows, all output tokens are the same on Linux. But the logit from 2nd token is different. Because they use different kernel. The 47th token on windows is different due to top1 and top2 logit are very close. The logit of `code` is higher than `function` from `greedy_causal_lm`, but `function` is higher than `code` from `speculative_decoding_lm`.
The logit depends not only on the kernel but also on the platform. For example, the same algorithm might use different instruction sets on different platforms. The logit may be different.
So I think It is unreasonable to compare two results from `greedy_causal_lm` and `speculative_decoding_lm` that are exactly the same. The test should pass as long as the result is reasonable. What do you think?

Wovchena:
The value change for
`code`: |36.123901 - 35.790699| / 36.123901 = 0.9%
`function`: |35.5851364 - 36.127686| / 35.5851364 = 1.5%

That looks too large just for a kernel switch. Are these percentages within the acceptable range?

sunxiaoxia2022:
> The value change for `code`: |36.123901 - 35.790699| / 36.123901 = 0.9% `function`: |35.5851364 - 36.127686| / 35.5851364 = 1.5%
> 
> That looks too large just for a kernel switch. Are these percentages within the acceptable range?

Hi @Wovchena I cannot say whether these percentages are acceptable directly.
So, I compared the logits of the `greedy_causal_lm` across some platforms.
image
They differed starting from the 46th token.
The logit difference of `if` in Linux_XEON and Linux_RPL: ABS(34.841248 - 34.567177)/34.841248 = 0.8%
The logit difference of `if` in Linux_XEON and Windows_PTL: ABS(34.841248 - 34.227341)/34.841248 = 1.7%
And I calculated the logit differences of top1_token, top2_token between `greedy_causal_lm` and `speculative_decoding_lm` from 1st token to the 46th token. Because they are the same.
image
The average difference of top1 is 0.35, top2=0.41, max difference of top1 is 0.96, top2 is 1.69. And the distribution of the differences is also reasonable.
So I think the The value change for `code` and `function` are acceptable.

Wovchena:
> The test should pass as long as the result is reasonable. What do you think?

There's no simple way to enforce that. It requires accuracy comparison.

Can you come up with a different test prompt to pass the test on Linux and Windows?
Alternatively you can replace test model with tiny random just for Windows with a comment summarizing your research and a link to this PR.

Wovchena:
Can you try forcing precision when test models are converted with `optimum-cli` in the test? Maybe that will align the results

Wovchena (approved):
Workaround for ScatterNDUpdate CI fails: https://github.com/openvinotoolkit/openvino.genai/pull/4271

tests/python_tests/samples/test_speculative_decoding_lm.py
```diff
@@ -66,13 +66,16 @@ def test_sample_speculative_decoding_lm(self, convert_model, convert_draft_model
         env["OPENVINO_LOG_LEVEL"] = "0"
         _run_spec_case(convert_model, convert_draft_model, sample_args, env)
 
-test_prompt = """Code:
+test_prompt = (
+    "1+1="
+    if sys.platform == "win32"
+    else """Code:
```
Wovchena, line 76: Add a comment summarizing your research and a link to this PR. This knowledge will be forgotten otherwise and the branching will look strange

sunxiaoxia2022: Ok, updated. Please take a look, thank you!

Wovchena (approved):

PR #4227, opened 2026-07-27
[Speculative decoding] Fix Eagle3 KV cache reorder model precision mismatch

liubo-intel:
## Description
Fixes a `ParameterMismatch` crash in Eagle3 speculative decoding where the
KV cache reorder helper model expected FP16 tensors while the main model's
compiled KV cache was actually BF16. The reorder model's precision is now
derived from the main model's compiled `key_cache.*` port instead of the
`kv_cache_precision` hint, and propagated consistently everywhere it's used
so plugins don't fall back to a stale/mismatched value. Verified locally
with Qwen3-8B target + Eagle3 draft on CPU, which now generates successfully.

CVS-191392

### DIFF
--- a/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
+++ b/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
@@ -6,7 +6,6 @@
 #include "eagle3_strategy.hpp"
 #include "openvino/pass/pa_kv_reorder_fusion.hpp"
 #include "speculative_decoding/eagle3_model_transforms.hpp"
-#include "logger.hpp"
 
 namespace ov::genai {
 KVUpdateWrapper::KVUpdateWrapper(const ov::genai::ModelDesc& kv_model_desc) {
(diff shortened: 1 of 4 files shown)


Wovchena (approved):

as-suvorov (approved):

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
```diff
@@ -92,23 +91,26 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
...
     ov::pass::PaKVReorderFusion(kv_cache_precision).run_on_model(kv_model);
-    // add rt_info for real kv precision into kv_model
+    // Pass the resolved (post u4->u8) precision explicitly instead of forwarding the raw
+    // ov::hint::kv_cache_precision value from main_model_desc.properties: for u4 caches that raw
+    // value would still say "u4" while PaKVReorderFusion above already rewrote the key_cache./
+    // value_cache. parameters to u8, and some plugins (e.g. GPU) give an explicitly user-set
+    // kv_cache_precision property priority over the auxiliary_kv_cache_precision rt_info, which
+    // would reintroduce a precision mismatch between the property and the actual parameter type.
+    kv_model_desc.properties[ov::hint::kv_cache_precision.name()] = kv_cache_precision;
     m_kv_update_wrapper = std::make_shared<KVUpdateWrapper>(kv_model_desc);
```
songbell, line 114: I suppose this will not work for GPU 4bit kv cache because the input tensors will always be resolved to 8bit, the kv update model does not know it needs to handle 4bit data. @liubo-intel

songbell:
please do not merge until the last comment is resolved.

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
liubo-intel: Hi, @songbell : how about using the following method(commit: 5191e2d )to cover both CPU and GPU? : The reorder model's Parameter type now always matches the main model's compiled key_cache/value_cache ports (fixing the bf16 CPU crash), while a separate auxiliary_kv_cache_precision rt_info carries the plugin's logical precision only when it's a sub-byte type the ports can't express (fixing GPU 4-bit cache corruption)

songbell (approved):

liubo-intel:
Hi, @as-suvorov : Regarding the CI test failures in these PRs (the Python (VLM (qwen3-vl and qwen3-next)) Tests (wheel) job), restart the CI and rebase the branch seem not work, and it seems that the failures are not related to the changes introduced by this PR. I also noticed that some other PRs are failing on the same CI tests.
Do you have any suggestions about this?thanks

songbell:
regression of this PR : https://github.com/openvinotoolkit/openvino.genai/pull/4411 @liubo-intel  please help review

PR #4237, opened 2026-07-28
StatefulLLMPipeline: negotiate NPU chat-history reuse via continuous prefill

dylanneve1:
### Details
- On NPU, `m_use_full_chat_history` stops being a device rule: when the compiled model advertises `NPUW_LLM_CONTINUOUS_PREFILL_SUPPORTED`, chat turns reuse the existing CPU/GPU cache machinery plus a negotiation step, so every turn re-prefills only the new tokens instead of the entire conversation.
- Per turn: `align_cache_and_history()` computes the common prefix, the pipeline validates the complete request (history within `MAX_PROMPT_LEN`, history + response budget within KV capacity), proposes the prefix through `npuw_stored_tokens_state`, reads the granted keep back, resizes `CacheState` to the grant and slices the input there. The physical `trim_kv_cache` path stays CPU/GPU only.
- Capability is read directly from the property in both NPU construction paths. A `false` value keeps full-history behaviour; property lookup failures propagate. The stored-tokens state is not used for capability discovery.
- Failed turns are not rolled back. A failed pending prefill requires a request reset and full-history prefill for recovery; resetting the chat clears its state. Cancellation stays a committed operation and rolls forward through the next alignment.

Companion plugin PRs (all merged): openvinotoolkit/openvino#37146, openvinotoolkit/openvino#37147, openvinotoolkit/openvino#37148, openvinotoolkit/openvino#37149 (protocol, transaction coordinator, KV strategy continuation, block-based KV continuation).

### Validation
- 3-turn greedy chat on qwen3 via `NPUW_DEVICES=CPU` (chunk 256, prompt window 2048): transcripts token-identical with the feature on and off; NPUW logs confirm delta-only prefills at the granted keep on turns 2 and 3.
- C++ syntax check of the complete modified `pipeline_stateful.cpp` against local OpenVINO headers, plus `git diff --check`. Runtime chat tests were not rerun for the property-read simplification.

### Tickets
- EISW-227938

### DIFF
--- a/src/cpp/src/llm/pipeline_stateful.cpp
+++ b/src/cpp/src/llm/pipeline_stateful.cpp
@@ -23,7 +23,9 @@ StatefulLLMPipeline::StatefulLLMPipeline(
     if (execution_devices[0].find("NPU") != std::string::npos) {
         OPENVINO_ASSERT(execution_devices.size() == 1u);
         m_is_npu = true;
+        m_use_full_chat_history = true;
         m_max_prompt_len = compiled_model.get_property("NPUW_LLM_MAX_PROMPT_LEN").as<uint32_t>();
+        init_npu_continuous_prefill(compiled_model);
     }
 }
 
@@ -74,6 +76,9 @@ StatefulLLMPipeline::StatefulLLMPipeline(
         utils::KVDesc kv_desc;
         std::tie(compiled_model, kv_desc) = utils::compile_decoder_for_npu(model, *filtered_properties, kv_pos);
         m_max_prompt_len = kv_desc.max_prompt_len;
+        // The capability is only known after compilation, so full chat history stops
+        // being a device rule here and becomes the fallback.
+        init_npu_continuous_prefill(compiled_model);
     } else {
        compiled_model = utils::singleton_core().compile_model(model, device, *filtered_properties);
     }
@@ -156,6 +161,7 @@ DecodedResults StatefulLLMPipeline::generate(
     auto start_time = std::chrono::steady_clock::now();
 
     GenerationConfig config = resolve_generation_config(generation_config);
+    NPUTurnGuard npu_turn_guard(*this);
 
     TokenizedInputs encoded_input;
     auto tokenization_start_time = start_time;
@@ -176,6 +182,8 @@ DecodedResults StatefulLLMPipeline::generate(
                 encoded_input = new_chat_tokens;
             } else {
                 ov::genai::align_cache_and_history(new_chat_tokens.input_ids, m_cache_state);
+                if (m_npu_continuous_prefill)
+                    negotiate_npu_history_reuse(new_chat_tokens.input_ids.get_shape().at(1), config);
                 encoded_input = get_chat_encoded_input(new_chat_tokens.input_ids, m_cache_state);
             }
         } else if (config.apply_chat_template && !m_tokenizer.get_chat_template().empty()) {
@@ -212,6 +220,8 @@ DecodedResults StatefulLLMPipeline::generate(
                 encoded_input = new_chat_tokens;
             } else {
                 ov::genai::align_cache_and_history(new_chat_tokens.input_ids, m_cache_state);
+                if (m_npu_continuous_prefill)
+                    negotiate_npu_history_reuse(new_chat_tokens.input_ids.get_shape().at(1), config);
                 encoded_input = get_chat_encoded_input(new_chat_tokens.input_ids, m_cache_state);
             }
             // TODO: Forbid LoRA config change if we are in the chat mode, because it requires regenerating the history with LoRA applied
@@ -240,6 +250,7 @@ DecodedResults StatefulLLMPipeline::generate(
         tokenization_start_time,
         chat_template_duration_us
     );
+    npu_turn_guard.disarm();
 
     if (is_chat_conversation) {
         if (m_chat_generation_finish_status == ov::genai::GenerationStatus::CANCEL) {
@@ -273,6 +284,8 @@ DecodedResults StatefulLLMPipeline::generate(
     auto start_time = std::chrono::steady_clock::now();
 
     GenerationConfig config = resolve_generation_config(generation_config);
+    // Snapshots the pre-turn history before it is replaced below.
+    NPUTurnGuard npu_turn_guard(*this);
 
     OPENVINO_ASSERT(config.apply_chat_template, "Chat template must be applied when using ChatHistory in generate method.");
     OPENVINO_ASSERT(!m_tokenizer.get_chat_template().empty(), "Chat template must not be empty when using ChatHistory in generate method.");
@@ -307,9 +320,11 @@ DecodedResults StatefulLLMPipeline::generate(
         encoded_input = new_chat_tokens;
     } else {
         ov::genai::align_cache_and_history(new_chat_tokens.input_ids, m_cache_state);
+        if (m_npu_continuous_prefill)
+            negotiate_npu_history_reuse(new_chat_tokens.input_ids.get_shape().at(1), config);
         encoded_input = get_chat_encoded_input(new_chat_tokens.input_ids, m_cache_state);
     }
-    return get_decoded_results(
+    auto decoded_results = get_decoded_results(
         encoded_input,
         config,
         streamer,
@@ -317,6 +332,8 @@ DecodedResults StatefulLLMPipeline::generate(
         tokenization_start_time,
         PerfMetrics::get_microsec(tokenization_start_time - template_start_time)
     );
+    npu_turn_guard.disarm();
+    return decoded_results;
 }
 
 EncodedResults StatefulLLMPipeline::generate(
@@ -370,6 +387,11 @@ EncodedResults StatefulLLMPipeline::generate(
         data->attention_mask.copy_to(attention_mask);
     }
 
+    // Resolved before the alignment block so the negotiation can validate the
+    // response budget, and the turn guard snapshots history before it grows.
+    GenerationConfig config = resolve_generation_config(generation_config);
+    NPUTurnGuard npu_turn_guard(*this);
+
     if (is_chat_conversation && m_chat_input_type == ov::genai::utils::GenerationChatInputsType::ENCODED_INPUTS)
         std::copy(input_ids.data<int64_t>(), input_ids.data<int64_t>() + input_ids.get_size(), std::back_inserter(m_tokenized_chat_history));
 
@@ -382,13 +404,13 @@ EncodedResults StatefulLLMPipeline::generate(
     if (is_chat_conversation && m_chat_input_type == ov::genai::utils::GenerationChatInputsType::ENCODED_INPUTS) {
         ov::Tensor new_chat_tokens = ov::Tensor{ov::element::i64, {1, m_tokenized_chat_history.size()}, m_tokenized_chat_history.data()};
         ov::genai::align_cache_and_history(new_chat_tokens, m_cache_state);
+        if (m_npu_continuous_prefill)
+            negotiate_npu_history_reuse(new_chat_tokens.get_shape().at(1), config);
 
         auto encoded_input = get_chat_encoded_input(new_chat_tokens, m_cache_state);
         input_ids = encoded_input.input_ids;
         attention_mask = encoded_input.attention_mask;
     }
-
-    GenerationConfig config = resolve_generation_config(generation_config);
 
     auto batch_size = input_ids.get_shape().at(0);
 
@@ -416,10 +438,14 @@ EncodedResults StatefulLLMPipeline::generate(
                     "but you have '" + std::to_string(num_inputs) + "' inputs");
 
     if (is_chat_conversation) {
-        if (m_use_full_chat_history)
+        if (m_use_full_chat_history) {
             reset_state();
-        else
+        } else if (!m_npu_continuous_prefill) {
             ov::genai::utils::trim_kv_cache(m_model_runner, m_cache_state, m_adapter_controller);
+        }
+        // With continuous prefill the negotiation already issued the one command for
+        // this turn. The physical trim path stays CPU and GPU only, and no second
+        // state write may happen here.
     }
 
     size_t cache_len = 0;
@@ -494,6 +520,11 @@ EncodedResults StatefulLLMPipeline::generate(
     ov::genai::EncodedResults& result = finish_info.results;
     m_chat_generation_finish_status = finish_info.streaming_finish_status;
 
+    // Inference completed, including a cancelled one, which is a successful physically
+    // committed operation and must not go through the failure rollback.
+    npu_turn_guard.disarm();
+    m_forced_full_prefill = false;
+
     if (is_chat_conversation) {
         m_cache_state.num_tokens_to_trim = 0;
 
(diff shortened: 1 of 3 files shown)


src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -31,7 +31,69 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+                m_armed = false;
+            }
+        }
+        ~NPUTurnGuard() {
+            if (m_armed) {
+                m_pipeline.m_npu_turn_guard_active = false;
+                m_pipeline.on_npu_turn_failure(std::move(m_history_snapshot),
+                                               std::move(m_tokenized_history_snapshot));
+            }
+        }
```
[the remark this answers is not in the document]
dylanneve1: Fixed in 8042cc01 (before this thread was revisited): the rollback is wrapped in try/catch with a no-throw comment, and on_npu_turn_failure sets the forced-full-prefill flag first so recovery intent survives even if touching the runner throws again.

src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -31,7 +31,69 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+    bool m_forced_full_prefill = false;
+    // Guards against nested arming when generate(EncodedInputs) runs inside the
+    // string or ChatHistory entry points.
+    bool m_npu_turn_guard_active = false;
+
     void reset_state();
+
+    // Reads the continuous prefill capability from the compiled model and flips
+    // m_use_full_chat_history accordingly. Called from every construction path.
+    void init_npu_continuous_prefill(ov::CompiledModel& compiled_model);
```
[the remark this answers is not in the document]
dylanneve1: Addressed in b4293eb4: init_npu_continuous_prefill now takes const ov::CompiledModel&.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -530,6 +561,115 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
...
+    // looking for npuw_stored_tokens_state in query_state(), because every existing
+    // plugin build publishes that state and would give a false positive. A plugin
+    // that predates the feature has no such property, which keeps the existing
+    // full-history behaviour automatically.
+    bool supported = false;
+    try {
+        supported = compiled_model.get_property("NPUW_LLM_CONTINUOUS_PREFILL_SUPPORTED").as<bool>();
+    } catch (...) {
+        supported = false;
+    }
```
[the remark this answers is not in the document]
dylanneve1: Narrowed the catch to ov::Exception, which is what a plugin without the feature throws for an unsupported configuration key, and everything else propagates now.

The supported_properties route does not work here though: NPUW_LLM_CONTINUOUS_PREFILL_SUPPORTED is computed on demand in LLMCompiledModel::get_property and is never added to the property map, so it is not listed in supported_properties and checking there would always report it as absent.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -240,6 +250,7 @@ DecodedResults StatefulLLMPipeline::generate(
         tokenization_start_time,
         chat_template_duration_us
     );
+    npu_turn_guard.disarm();
 
     if (is_chat_conversation) {
         if (m_chat_generation_finish_status == ov::genai::GenerationStatus::CANCEL) {
```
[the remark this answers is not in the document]
dylanneve1: Right, moved it. The guard is now disarmed after the history is updated in the string path and after the tokenized history and the trim bookkeeping in the encoded path, so a throw while committing the turn still rolls the conversation back and forces a full prefill on the next one.

src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -31,7 +31,73 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+            if (pipeline.m_npu_continuous_prefill && pipeline.is_chat_conversation &&
+                !pipeline.m_npu_turn_guard_active) {
+                m_armed = true;
+                pipeline.m_npu_turn_guard_active = true;
+                // ChatHistory copies share their JsonContainer storage, so the
+                // snapshot needs explicit deep copies to survive in-place mutation.
+                m_history_snapshot = ChatHistory(pipeline.m_history.get_messages().copy());
+                m_history_snapshot.set_tools(pipeline.m_history.get_tools().copy());
+                m_history_snapshot.set_extra_context(pipeline.m_history.get_extra_context().copy());
+                m_tokenized_history_snapshot = pipeline.m_tokenized_chat_history;
```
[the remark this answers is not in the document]
dylanneve1: Fixed. Both snapshots are taken first and the guard only sets m_armed and m_npu_turn_guard_active once they succeeded, so a throwing copy leaves the flag untouched.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -530,6 +561,115 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
...
 
+void StatefulLLMPipeline::init_npu_continuous_prefill(const ov::CompiledModel& compiled_model) {
+    // The capability must come from the read-only property. It must not be probed by
+    // looking for npuw_stored_tokens_state in query_state(), because every existing
+    // plugin build publishes that state and would give a false positive. A plugin
+    // that predates the feature has no such property, which keeps the existing
+    // full-history behaviour automatically.
+    bool supported = false;
+    try {
+        supported = compiled_model.get_property("NPUW_LLM_CONTINUOUS_PREFILL_SUPPORTED").as<bool>();
```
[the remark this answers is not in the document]
dylanneve1: Added the coverage in tests/python_tests/test_llm_pipeline_static.py. Those tests drive the NPUW pipeline over NPUW_DEVICES=CPU like the rest of that file, so they run on an ordinary CI machine without an NPU.

They cover the fallback when the plugin refuses the capability, chat parity against a full-history pipeline for string, encoded and ChatHistory inputs, the rejection of a bounded request that does not fit the KV cache together with an unbounded one that must still be accepted, the recovery after a rejected turn, a failed turn not leaking into the next chat, and a cancelled turn the chat goes on from.

Checked with NPUW debug logging that the parity tests really do continue: the keep is granted in whole chunks, so the questions are long enough for the history to cross one and the runs show grants of 64 and 128 tokens rather than a full prefill every turn.

src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -31,7 +31,73 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+    bool m_npu_continuous_prefill = false;
+    // Past KV capacity of the largest generate variant, used to validate an explicitly
+    // bounded response budget before proposing.
+    size_t m_kv_cache_capacity = std::numeric_limits<size_t>::max();
+    // Set after a failed turn. The next attempt skips the proposal and sends the full
+    // history, which satisfies the plugin's pending reset.
+    bool m_forced_full_prefill = false;
+    // Guards against nested arming when generate(EncodedInputs) runs inside the
+    // string or ChatHistory entry points.
+    bool m_npu_turn_guard_active = false;
```
AlexanderKalistratov: Don't need that.
If we failed we failed.

src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -31,7 +31,73 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+    // Proposes the post-alignment common prefix to the plugin, reads the grant back
+    // and resizes the cache state to it, so slicing happens at the granted value.
+    void negotiate_npu_history_reuse(size_t full_history_len, const GenerationConfig& config);
+    // Single recovery path for any failure between proposal and inference. Resets the
+    // plugin channel, restores the pre-turn history and forces a full prefill next.
+    void on_npu_turn_failure(ChatHistory history_snapshot, std::vector<int64_t> tokenized_history_snapshot);
+
+    // Restores conversation state when a turn throws anywhere between the proposal
+    // and the end of inference. Armed only at the outermost generate entry point.
+    class NPUTurnGuard {
```
AlexanderKalistratov: Don't need that.

src/cpp/src/llm/pipeline_stateful.hpp
```diff
@@ -31,7 +31,73 @@ class StatefulLLMPipeline final : public LLMPipelineImplBase {
...
+
+    // Reads the continuous prefill capability from the compiled model and flips
+    // m_use_full_chat_history accordingly. Called from every construction path.
+    void init_npu_continuous_prefill(const ov::CompiledModel& compiled_model);
+    // Proposes the post-alignment common prefix to the plugin, reads the grant back
+    // and resizes the cache state to it, so slicing happens at the granted value.
+    void negotiate_npu_history_reuse(size_t full_history_len, const GenerationConfig& config);
+    // Single recovery path for any failure between proposal and inference. Resets the
+    // plugin channel, restores the pre-turn history and forces a full prefill next.
+    void on_npu_turn_failure(ChatHistory history_snapshot, std::vector<int64_t> tokenized_history_snapshot);
```
AlexanderKalistratov: No.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -156,6 +161,7 @@ DecodedResults StatefulLLMPipeline::generate(
     auto start_time = std::chrono::steady_clock::now();
 
     GenerationConfig config = resolve_generation_config(generation_config);
+    NPUTurnGuard npu_turn_guard(*this);
```
AlexanderKalistratov: No.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -240,6 +250,7 @@ DecodedResults StatefulLLMPipeline::generate(
         tokenization_start_time,
         chat_template_duration_us
     );
+    npu_turn_guard.disarm();
```
AlexanderKalistratov: Remove.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -273,6 +284,8 @@ DecodedResults StatefulLLMPipeline::generate(
     auto start_time = std::chrono::steady_clock::now();
 
     GenerationConfig config = resolve_generation_config(generation_config);
+    // Snapshots the pre-turn history before it is replaced below.
+    NPUTurnGuard npu_turn_guard(*this);
```
AlexanderKalistratov: Remove.

AlexanderKalistratov (changes requested):
Please remove `NPUTurnGuard`.

src/cpp/src/llm/pipeline_stateful.hpp
dylanneve1: Removed

src/cpp/src/llm/pipeline_stateful.hpp
dylanneve1: Removed

src/cpp/src/llm/pipeline_stateful.hpp
dylanneve1: Removed

src/cpp/src/llm/pipeline_stateful.cpp
dylanneve1: Removed

src/cpp/src/llm/pipeline_stateful.cpp
dylanneve1: Removed

src/cpp/src/llm/pipeline_stateful.cpp
dylanneve1: Removed

dylanneve1:
Done @AlexanderKalistratov

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -530,6 +547,92 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
...
+        if (st.get_name() == "npuw_stored_tokens_state") {
+            channel = st;
+            break;
+        }
+    }
+    OPENVINO_ASSERT(channel.has_value(), "npuw_stored_tokens_state is missing while continuous prefill is enabled.");
+
+    ov::Tensor proposal(ov::element::i64, {1});
+    proposal.data<int64_t>()[0] = static_cast<int64_t>(k_common);
+    channel->set_state(proposal);
```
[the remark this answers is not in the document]
AlexanderKalistratov: If something throws we are dead

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -541,6 +558,92 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
...
+    // The capability must come from the read-only property. It must not be probed by
+    // looking for npuw_stored_tokens_state in query_state(), because every existing
+    // plugin build publishes that state and would give a false positive. The property
+    // is computed on demand and is not listed in supported_properties either, so
+    // asking for it is the only way to probe it. A plugin that predates the feature
+    // answers with an ov::Exception for an unsupported configuration key, which keeps
+    // the existing full-history behaviour automatically and is the only failure
+    // expected here; anything else propagates.
+    bool supported = false;
+    try {
```
AlexanderKalistratov, line 571: We are expecting this property to exist. Not sure if we actually need `try-catch` here

tests/python_tests/test_llm_pipeline_static.py
```diff
@@ -475,6 +475,218 @@ def generate_with_chat_history(pipe: LLMPipeline, questions: list[str]) -> ChatH
...
+# Continuous prefill reuses the part of the chat history the plugin still holds in its
+# KV cache and sends only the new tokens. It requires chunked prefill, which the plugin
+# enables when the chunk is smaller than the prompt window. The chunk is also the
+# granularity of the granted keep, so a small one keeps it non-zero for short prompts.
+CONTINUOUS_PREFILL_CONFIG: dict = {
+    **DEFAULT_CONFIG,
+    "MAX_PROMPT_LEN": 1024,
+    "MIN_RESPONSE_LEN": 128,
+    "NPUW_LLM_PREFILL_CHUNK_SIZE": 64,
+    "NPUW_LLM_ENABLE_CONTINUOUS_PREFILL": "YES",
```
dmatveev, line 491: Why we need a knob here?

dmatveev: We need to take care of not introducing extra knobs

tests/python_tests/test_llm_pipeline_static.py
dmatveev, line 1: Sounds like this is a test for a test - it only tests chat history case just works with continuous prefill.

I'd rather reuse a test for chat history and extend it to test one more configuration with continuous prefill enabled.

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -23,7 +23,9 @@ StatefulLLMPipeline::StatefulLLMPipeline(
     if (execution_devices[0].find("NPU") != std::string::npos) {
         OPENVINO_ASSERT(execution_devices.size() == 1u);
         m_is_npu = true;
+        m_use_full_chat_history = true;
         m_max_prompt_len = compiled_model.get_property("NPUW_LLM_MAX_PROMPT_LEN").as<uint32_t>();
+        init_npu_continuous_prefill(compiled_model);
```
dmatveev, line 28: why we'd need to do it here?

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -541,6 +558,81 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
     m_history.push_back({{"role", "system"}, {"content", system_message}});
 }
 
+void StatefulLLMPipeline::init_npu_continuous_prefill(const ov::CompiledModel& compiled_model) {
+    // Read the capability directly: the stored-tokens state exists regardless of
+    // support, and this property is not listed in supported_properties.
+    const bool supported = compiled_model.get_property("NPUW_LLM_CONTINUOUS_PREFILL_SUPPORTED").as<bool>();
```
dmatveev, line 564: why it is _SUPPORTED? shouldn't it be the property that enabled it?

if it couldn't be applied, it shouldn't be reported?

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -541,6 +558,81 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
...
+        // already pending would be a protocol error.
+        return;
+    }
+
+    // Propose the common prefix and read the grant back. The write is not idempotent,
+    // the plugin clamps by capacity, chunk alignment and its prefill watermark, so
+    // slicing must happen at the granted value, never at the proposed one.
+    std::optional<ov::VariableState> channel;
+    for (auto& st : m_model_runner.query_state()) {
+        if (st.get_name() == "npuw_stored_tokens_state") {
```
dmatveev, line 615: Is it state named properly anywhere? Should we use some mnemonic instead?

dmatveev: cc: @AsyaPronina

src/cpp/src/llm/pipeline_stateful.cpp
```diff
@@ -541,6 +558,81 @@ void StatefulLLMPipeline::start_chat(const std::string& system_message) {
...
 
+void StatefulLLMPipeline::init_npu_continuous_prefill(const ov::CompiledModel& compiled_model) {
+    // Read the capability directly: the stored-tokens state exists regardless of
+    // support, and this property is not listed in supported_properties.
+    const bool supported = compiled_model.get_property("NPUW_LLM_CONTINUOUS_PREFILL_SUPPORTED").as<bool>();
+    m_npu_continuous_prefill = supported;
+    m_use_full_chat_history = !supported;
+    if (supported) {
+        m_kv_cache_capacity = utils::get_npu_kv_cache_capacity(compiled_model);
+    }
```
dmatveev, line 569: What's the portable way to get the kv-cache size on CPU/GPU with stateful (SDPA) backend?

PR #4240, opened 2026-07-29
[llm_bench] Add support of Qwen3-VL-Embedding-8B

avasenin-14:
## Description

[CVS-188814](https://jira.devtools.intel.com/browse/CVS-188814)

Add Qwen3-VL-Embedding-8B support for llm_bench.

## Checklist:
- [x]  I have made corresponding changes to the documentation.


tools/llm_bench/task/text_embeddings.py
```diff
@@ -89,35 +148,56 @@ def run_text_embeddings_optimum(input_text, num, model, tokenizer, args, iter_da
...
         log.info(out_str)
 
     mem_consumption.start(num)
     start = time.perf_counter()
-    model.embed_documents(input_text_list)
+    if is_qwen3_vl:
+        media_kwargs = {}
+        if embedding_prompt:
+            media_kwargs["embedding_prompt"] = embedding_prompt
+        if images:
```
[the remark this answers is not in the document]
avasenin-14: I think not: https://github.com/avasenin-14/openvino.genai/blob/dee34c6bf030cf968a259651999938a1dd8f7357/src/cpp/include/openvino/genai/rag/embedding_pipeline.hpp#L103-L104? If I'm not mixing up classes.

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -114,6 +122,20 @@ def read_wav(filepath, sampling_rate):
     return raw_speech[0]
 
 
+def resolve_model_dir(model_path):
```
as-suvorov: What is expected input here?

tools/llm_bench/README.md
```diff
@@ -318,13 +318,35 @@ python benchmark.py -m models/bge-small-en-v1.5/ -n 2 --task text_embed
...
+- `--embedding_prompt`: Instruction/system prompt used to guide embedding generation. Distinct from `-p`, which supplies the content being embedded. For Qwen3-VL-Embedding this defaults to `"Represent the user's input."`.
 
-> **Supported Text Embeddings model types:**: bge, bert, albert, roberta, xlm-roberta, qwen3
+> **Supported Text Embeddings model types:**: bge, bert, albert, roberta, xlm-roberta, qwen3, qwen3-vl
+
+### Compare Multimodal (Qwen3-VL) Embedding models
+Qwen3-VL-Embedding shares its architecture with Qwen3-VL text generation; the embedding behavior requires `--task text_embed` explicitly.
+```sh
+# convert model to OpenVINO IR format
+optimum-cli export openvino --model Qwen/Qwen3-VL-Embedding-8B --task feature-extraction --trust-remote-code models/Qwen3-VL-Embedding-8B
```
as-suvorov: Please double check if `--trust-remote-code` required

tools/llm_bench/README.md
```diff
@@ -318,13 +318,35 @@ python benchmark.py -m models/bge-small-en-v1.5/ -n 2 --task text_embed
...
+# text-only embedding
+python benchmark.py -m models/Qwen3-VL-Embedding-8B -n 2 --task text_embed -p "Describe OpenVINO"
+# image embedding
+python benchmark.py -m models/Qwen3-VL-Embedding-8B -n 2 --task text_embed -p "Represent this image" --media cat.png
+# video embedding
+python benchmark.py -m models/Qwen3-VL-Embedding-8B -n 2 --task text_embed -p "Represent this video" --video video.mp4 -vf 4
+# media-only embedding: -p is optional when media is provided
+python benchmark.py -m models/Qwen3-VL-Embedding-8B -n 2 --task text_embed --media cat.png
```
as-suvorov: Should the task be generic as well? `--task embed`? It should be also aligned with `wwb`

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -728,7 +731,15 @@ def create_genai_text_embed_model(model_path, device, memory_data_collector, **k
...
-    pipe = openvino_genai.TextEmbeddingPipeline(model_path, device.upper(), config, **ov_config)
+    if is_qwen3_vl:
+        if not hasattr(openvino_genai, "EmbeddingPipeline"):
+            raise RuntimeError(
+                "Qwen3-VL-Embedding requires openvino_genai.EmbeddingPipeline, which is unavailable "
+                "in the installed openvino_genai. Please upgrade openvino_genai."
+            )
+        pipe = openvino_genai.EmbeddingPipeline(model_path, device.upper(), text_embedding_config=config, **ov_config)
+    else:
+        pipe = openvino_genai.TextEmbeddingPipeline(model_path, device.upper(), config, **ov_config)
```
as-suvorov, line 765:
```suggestion
```
Let's always use `EmbeddingPipeline`. It became recommended default

tools/llm_bench/task/text_embeddings.py
```diff
@@ -12,36 +12,100 @@
...
+def run_text_embeddings_optimum(
+    entry, num, model, tokenizer, args, iter_data_list, prompt_index, bench_hook, proc_id, mem_consumption
+):
+    is_qwen3_vl = _is_qwen3_vl(args)
+    embedding_prompt = _resolve_embedding_prompt(args, is_qwen3_vl)
+    batch_size = args["batch_size"]
 
     tok_encode_start = time.perf_counter()
-    input_data = tokenizer(input_text_list, return_tensors='pt', **tokenizer_kwargs)
+    if is_qwen3_vl:
```
pavel-esir: Please make it more general. E.g. `is_multimodal`

This is applicable not only for qwen for other future multimodal models.
```suggestion
    if is_multimodal:
```

tools/llm_bench/README.md
pavel-esir: Agree with @as-suvorov let's have more general name `--task embed`.

We can allow to set `text_embed` for backward compatibility so that we don't break components that already run llm_bench with `--task text_embed`, but let's have recommended new value as `--task embed`.

tools/llm_bench/task/embedding.py
```diff
@@ -12,36 +12,100 @@
 import llm_bench_utils.output_csv
```
pavel-esir, line 12: I think it's better to rename file from `text_embedding.py` -> `embedding.py` / `multimodal_embedding.py`. Since it became related to multimodal embedding not just for text.

pavel-esir (commented):
Thanks for the PR. In general looks good. The main comment is regarding naming. Let's make naming more consisted with the the fact that actually we are doing not just text embedding but embedding for several modalities.

After that correction i think we can merge

tools/llm_bench/llm_bench_utils/model_utils.py
avasenin-14: Either a model dir or an .xml file from IR. I've got it from https://github.com/openvinotoolkit/openvino.genai/pull/4240/changes#diff-02c17a223e95b1b38883d669197902ef4898b329a03b9187fcf0427ac56f5346L749, but I'm not sure of the historical reasons of when we need to pass the .xml file. Do you think ti is needed still? If we can get away with model dir only, that would simplify the task :)

tools/llm_bench/README.md
avasenin-14: Not required, removed

tools/llm_bench/README.md
avasenin-14: No problems, I can add the general task here. In WWB we have distinct tasks for text/image/video, but we can theoretically consolidate them, I think they already are using one underlying function. Just not in this PR , if it's ok? To avoid growing the scope (and lenghtening review time).

tools/llm_bench/llm_bench_utils/ov_utils.py
avasenin-14: Added for all models for now, but with a fallback still, so the older pipelines would still work.

tools/llm_bench/task/embedding.py
avasenin-14: Sure, done

tools/llm_bench/README.md
avasenin-14: Added generic `embed` task.

tools/llm_bench/task/text_embeddings.py
avasenin-14: Done

pavel-esir (approved):

tools/llm_bench/llm_bench_utils/model_utils.py
as-suvorov: If we can align inputs of tool in general to always accept model dir, it make sense to simplify.
If we cannot align in general it's better to have such helper.
If we should have it:
1. please add function typing
2. please add short comment explaining possible inputs

tools/llm_bench/llm_bench_utils/model_utils.py
```diff
@@ -113,16 +118,22 @@ def get_param_from_file(args, input_key):
...
+    p = Path(model_path)
+    if p.is_dir():
+        return p
+    if not p.name.endswith("xml"):
+        return p
+    if (p.parent / "config.json").is_file():
+        return p.parent
+    try:
+        return p.parents[2]
```
as-suvorov, line 140: What `model_path` covered here?

tools/llm_bench/llm_bench_utils/model_utils.py
avasenin-14: Added typing and a comment.
I'd be happy to accept only dir as we do in WWB, but I'm afraid it might break some legacy conventions, so let's wait for Sofia maybe? I'll ask her if she remembers why we needed .xml paths.

tools/llm_bench/llm_bench_utils/model_utils.py
avasenin-14: I moved the existing code into helper:
```
    model_path = Path(model_path)
    if model_path.name.endswith('xml'):
        model_path = model_path.parents[2]
```
, soI'm not sure what was the directory layout in the case of an xml. I've removed the other checks now, left only these.

as-suvorov (approved):

PR #4243, opened 2026-07-30
llm_bench: add Windows GPU and full memory samplers

krzyczar:
llm_bench: add win-gpu and full memory samplers, switch MemorySamplerBase to memory_info
Rework process-based memory monitoring so the extra Windows-GPU and
Linux-full metrics extend the existing psutil-based sampler instead of
reimplementing per-process counters via the native Win32 API. The
--memory_sampler option selects the implementation: base (default),
win-gpu, or full.

MemorySamplerBase (--memory_sampler base, cross-platform, default)
  - Renamed from MemorySampler5.
  - Replace the `uss` metric with `vms`.
  - Collect via psutil.memory_info() instead of memory_full_info(). USS was
    the only field requiring memory_full_info(); its computation walks the
    working set / smaps and can raise or stall on Windows. rss/vms are
    available from memory_info() on every platform and `priv` (private) stays
    hasattr-guarded, so Linux (no `private`) keeps working.
  - Observed metrics: rss, vms, priv, sys, nsys.
  - Factor the parent+children summation into _collect_ram_values() so
    subclasses reuse the exact same RAM collection.

MemorySamplerWinGPU (--memory_sampler win-gpu, Windows)
  - New subclass of MemorySamplerBase: inherits all RAM/system metrics
    (psutil) and appends per-adapter GPU memory gpu_<index>_ded (dedicated
    VRAM) and gpu_<index>_shr (shared system RAM) via the WMI
    Win32_PerfFormattedData_GPUPerformanceCounters_GPUAdapterMemory perf
    counters (Windows 10 1709+). Shared-pool coverage makes integrated GPUs
    report real usage instead of a constant 0. Vendor-agnostic (NVIDIA/AMD/
    Intel). The `wmi` package is optional; GPU metrics are disabled with a
    warning when it is absent.
  - The WMI connection is cached; the GPU reading is additionally cached and
    refreshed at most once per second, so an expensive (100-500 ms) query
    never runs on every process sample. A metric/value count mismatch is
    logged loudly but kept non-fatal.
  - No native/ctypes Win32 memory API: GetProcessMemoryInfo,
    PROCESS_MEMORY_COUNTERS_EX and OpenProcess are not used; all RAM counters
    come from psutil.

MemorySamplerFull (--memory_sampler full, Linux)
  - New subclass of MemorySamplerBase that adds uss, pss and swap from
    psutil.memory_full_info() (reads /proc/<pid>/smaps). More accurate "real"
    footprint at the cost of speed.

Sampler selection
  - Add the --memory_sampler {base,win-gpu,full} option (default base),
    normalised with str.lower. win-gpu is used only on Windows and full only
    on Linux; both fall back to MemorySamplerBase with a warning on other
    platforms. Thread sampler_type through MemoryMarkerHandler ->
    background_worker -> MemoryMarkerMonitor.
  - background_worker: return after a failed MemoryMarkerMonitor init to avoid
    an UnboundLocalError when using `mmm` further down.

### DIFF
--- a/tools/llm_bench/benchmark.py
+++ b/tools/llm_bench/benchmark.py
@@ -175,6 +175,31 @@ def get_argparser():
         type=str,
         help="Path to store memory consumption logs and chart.",
     )
+    parser.add_argument(
+        "--memory_sampler",
+        default="base",
+        choices=["base", "win-gpu", "full"],
+        type=str.lower,  # normalise e.g. 'WIN-GPU'/'Full' -> 'win-gpu'/'full' before choices validation
+        required=False,
+        help="Memory sampler implementation to use when process-based monitoring is active\n"
+        "(--memory_consumption 3 or 4).\n"
+        "Possible values:\n"
+        "  base (default) — MemorySamplerBase: cross-platform sampler built on\n"
+        "                   psutil.memory_info(). Collects RSS, VMS, Private and\n"
+        "                   system-wide RAM. Works on Linux, macOS and Windows.\n"
+        "  win-gpu        — MemorySamplerWinGPU: same RAM metrics as base plus, when\n"
+        "                   the optional *wmi* package is installed (pip install wmi),\n"
+        "                   two per-GPU-adapter metrics: gpu_<index>_ded (dedicated\n"
+        "                   VRAM) and gpu_<index>_shr (shared system RAM). Sourced\n"
+        "                   from GPUAdapterMemory perf counters (Windows 10 1709+),\n"
+        "                   so integrated GPUs report real usage via the shared pool\n"
+        "                   instead of a constant 0. Windows only; falls back to\n"
+        "                   MemorySamplerBase on other platforms.\n"
+        "  full           — MemorySamplerFull: same RAM metrics as base plus uss, pss\n"
+        "                   and swap from psutil.memory_full_info() (reads /proc smaps).\n"
+        "                   More accurate 'real' footprint but slower. Linux only;\n"
+        "                   falls back to MemorySamplerBase on other platforms.",
+    )
     parser.add_argument("-bs", "--batch_size", type=int, default=1, required=False, help="Batch size value")
     parser.add_argument(
         "--num_beams",
--- a/tools/llm_bench/llm_bench_utils/memory_monitor.py
+++ b/tools/llm_bench/llm_bench_utils/memory_monitor.py
@@ -1060,10 +1060,14 @@ class MemorySamplerWinGPU(MemorySamplerBase):
     _GPU_ADAPTER_CLS = "Win32_PerfFormattedData_GPUPerformanceCounters_GPUAdapterMemory"
 
     @staticmethod
-    def _import_wmi():
+    def import_wmi():
         """
-        Import and return the optional ``wmi`` module, or ``None`` when it is
-        unavailable (non-Windows platform or package not installed).
+        Import and return the ``wmi`` module, or raise ``RuntimeError`` when
+        win-gpu cannot be honored -- a non-Windows platform, or the optional
+        ``wmi`` package not installed. Raising here (rather than returning
+        ``None``) means callers don't each have to re-check the result: an
+        explicit ``--memory_sampler win-gpu`` request fails loudly instead of
+        silently degrading to a RAM-only sampler.
 
         GPU memory is read from the WMI performance-counter class
         ``Win32_PerfFormattedData_GPUPerformanceCounters_GPUAdapterMemory``
@@ -1075,19 +1079,19 @@ class MemorySamplerWinGPU(MemorySamplerBase):
         (psutil).
         """
         if sys.platform != "win32":
-            return None
+            raise RuntimeError(
+                f"--memory_sampler win-gpu is only available on Windows (current platform: {sys.platform!r})."
+            )
         try:
             import wmi  # type: ignore[import]
-
-            log.debug("MemorySamplerWinGPU: wmi module loaded — GPU memory metrics (gpu_<index>) enabled.")
-            return wmi
         except ImportError:
-            log.debug(
-                "MemorySamplerWinGPU: wmi module not found "
-                "— GPU memory metrics (gpu_<index>) will be disabled. "
+            raise RuntimeError(
+                "--memory_sampler win-gpu requires the 'wmi' package for per-GPU memory "
+                "metrics (gpu_<index>_ded/shr), but it is not installed. "
                 "Install it with:  pip install wmi"
             )
-            return None
+        log.debug("MemorySamplerWinGPU: wmi module loaded — GPU memory metrics (gpu_<index>) enabled.")
+        return wmi
 
     def __init__(self, process_id):
         self.process_id = process_id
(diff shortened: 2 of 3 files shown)


avasenin-14 (approved):
I'm good with it. The usage of memory_info and class inheritance looks better for me, and I've seen you run the jobs to test that the new functionality is working.

tools/llm_bench/benchmark.py
```diff
@@ -175,6 +175,31 @@ def get_argparser():
         type=str,
         help="Path to store memory consumption logs and chart.",
     )
+    parser.add_argument(
+        "--memory_sampler",
+        default="base",
+        choices=["base", "win-gpu", "full"],
+        type=str.lower,  # normalise e.g. 'WIN-GPU'/'Full' -> 'win-gpu'/'full' before choices validation
```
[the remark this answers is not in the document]
krzyczar: fixed

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,399 @@ def __init__(self, process_id):
...
+            try:
+                self._wmi_conn = _wmi_module.WMI()
+                self._discover_gpu_adapters()
+            except Exception as exc:
+                log.warning(f"MemorySamplerWinGPU: WMI GPU detection failed ({exc}) — gpu_<index> metrics disabled.")
+        else:
+            # MemorySamplerWinGPU is only instantiated when the user explicitly
+            # asks for --memory_sampler W, so if the optional 'wmi' package is
+            # missing they should be told why GPU metrics are absent — the
+            # import-time notice is only DEBUG level and easily missed.
```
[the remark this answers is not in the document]
krzyczar: fixed

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,399 @@ def __init__(self, process_id):
...
+        import wmi as _wmi_module  # type: ignore[import]
+
+        _wmi_available = True
+        log.debug("MemorySamplerWinGPU: wmi module loaded — GPU memory metrics (gpu_<index>) enabled.")
+    except ImportError:
+        log.debug(
+            "MemorySamplerWinGPU: wmi module not found "
+            "— GPU memory metrics (gpu_<index>) will be disabled. "
+            "Install it with:  pip install wmi"
+        )
```
as-suvorov: Only used by `MemorySamplerWinGPU` so let's move it to a class

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,399 @@ def __init__(self, process_id):
...
+        else:
+            # MemorySamplerWinGPU is only instantiated when the user explicitly
+            # asks for --memory_sampler win-gpu, so if the optional 'wmi' package
+            # is missing they should be told why GPU metrics are absent — the
+            # import-time notice is only DEBUG level and easily missed.
+            log.warning(
+                "MemorySamplerWinGPU: optional 'wmi' package not installed "
+                "— per-GPU memory metrics (gpu_<index>_ded/shr) are disabled. "
+                "Install it with:  pip install wmi"
+            )
```
as-suvorov: If `MemorySamplerWinGPU` requested explicitly what is the reason not to fail if it's missing?

as-suvorov (changes requested):

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: true, will fix it

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: ok

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,395 @@ def __init__(self, process_id):
...
+        ``Win32_PerfFormattedData_GPUPerformanceCounters_GPUAdapterMemory``
+        (Windows 10 1709+), which exposes per-adapter dedicated AND shared GPU
+        memory currently in use.  Shared coverage is what makes integrated GPUs
+        report non-zero usage.  psutil has no GPU-memory API, so WMI is used
+        here; there is deliberately NO ctypes / native-Win32 memory code —
+        every RAM / system counter comes from :class:`MemorySamplerBase`
+        (psutil).
+        """
+        if sys.platform != "win32":
+            return None
```
as-suvorov: Should we better fail if win gpu sampler was requested on non win platform?

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,395 @@ def __init__(self, process_id):
...
+        try:
+            import wmi  # type: ignore[import]
+
+            log.debug("MemorySamplerWinGPU: wmi module loaded — GPU memory metrics (gpu_<index>) enabled.")
+            return wmi
+        except ImportError:
```
as-suvorov, line 1087: `_import_wmi` caller always check if `None` returned and raise. Let's better raise inside import so caller don't have to duplicate check?

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,395 @@ def __init__(self, process_id):
...
+        # asks for it (e.g. --memory_sampler win-gpu), so a missing 'wmi'
+        # package means the requested GPU metrics cannot be produced. Fail
+        # loudly rather than silently degrading to a RAM-only sampler.
+        wmi_module = self._import_wmi()
+        if wmi_module is None:
+            raise RuntimeError(
+                "MemorySamplerWinGPU requires the 'wmi' package for per-GPU memory metrics "
+                "(gpu_<index>_ded/shr), but it is not available "
+                f"(platform={sys.platform!r}). Install it with:  pip install wmi"
+            )
```
as-suvorov: Let's remove check here if raised inside import
```suggestion
```

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -1062,13 +1426,26 @@ def __init__(self, args):
...
+        # on Windows but the required 'wmi' package is missing. The sampler
+        # itself runs in a background daemon whose exceptions are caught and
+        # logged there (background_worker), so without this check the benchmark
+        # would silently continue with no GPU metrics. Terminate here instead
+        # so the missing dependency is impossible to miss.
+        if sampler_type == "win-gpu" and sys.platform == "win32" and MemorySamplerWinGPU._import_wmi() is None:
+            raise RuntimeError(
+                "--memory_sampler win-gpu requires the 'wmi' package for per-GPU memory metrics "
+                "(gpu_<index>_ded/shr), but it is not installed. Install it with:  pip install wmi"
+            )
```
as-suvorov:
```suggestion
        if sampler_type == "win-gpu":
            # import_wmi raise if wmi is not available
            MemorySamplerWinGPU.import_wmi()
```

tools/llm_bench/llm_bench_utils/memory_monitor.py
```diff
@@ -899,31 +899,395 @@ def __init__(self, process_id):
...
+      process-sampling interval.
+    """
+
+    # WMI performance-counter class exposing per-adapter GPU memory.  Unlike
+    # Win32_VideoController.DedicatedUsage (dedicated VRAM only), it also
+    # reports the shared pool, so integrated GPUs report real usage.
+    _GPU_ADAPTER_CLS = "Win32_PerfFormattedData_GPUPerformanceCounters_GPUAdapterMemory"
+
+    @staticmethod
+    def _import_wmi():
```
as-suvorov: Let's make it public as called externally
```suggestion
    def import_wmi():
```

as-suvorov (changes requested):

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: ok

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: done

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: ok

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: done

tools/llm_bench/llm_bench_utils/memory_monitor.py
krzyczar: extended

as-suvorov (approved):

as-suvorov:
build_jenkins

PR #4015, opened 2026-06-23
[wwb] Add suport for Qwen-Omni-MoE

avasenin-14:
## Description

CVS-188611

Add support for HF and Optimum Qwen-Omni-MoE models.
**How to reproduce:**
1. The Optimum support change is in https://github.com/huggingface/optimum-intel/pull/1700:
```
optimum-cli export openvino -m Qwen/Qwen3-Omni-30B-A3B-Instruct Qwen3-Omni-30B-A3B-Instruct --trust-remote-code
```
2. Collect the ground truth for HuggingFace public model:
```
python3 tools/who_what_benchmark/whowhatbench/wwb.py \
  --base-model Qwen/Qwen3-Omni-30B-A3B-Instruct \
  --gt-data gt-data.csv \
  --model-type visual-text \
  --hf
```
3. Run benchmark for Optimum comparison with the ground truth:
```
python3 tools/who_what_benchmark/whowhatbench/wwb.py \
  --target-model Qwen3-Omni-30B-A3B-Instruct  \
  --gt-data gt-data.csv \
  --model-type visual-text
```
4. Pull request with change for GenAI: https://github.com/openvinotoolkit/openvino.genai/pull/3868.Build and run benchmark for GenAI comparison with the ground truth:
```
python3 -m pip wheel . --no-deps -w dist/ \
  --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/pre-release \
  --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly

python3 -m pip install \
  dist/openvino_genai-2026.3.0.0-cp312-cp312-linux_x86_64.whl \
  --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/pre-release \
  --extra-index-url https://storage.openvinotoolkit.org/simple/wheels/nightly

python3 tools/who_what_benchmark/whowhatbench/wwb.py \
  --target-model ~/models/Qwen3-Omni-30B-A3B-Instruct \
  --gt-data gt_data.csv \
  --model-type visual-text \
  --genai
  ```

**Noticable changes:** 
- for HF models loading `Qwen3OmniMoeForConditionalGeneration` and `Qwen3OmniForConditionalGeneration` classes are used, as Omni models are moved out of `AutoModelForCausalLM` class;
- for Optimum models loading `OVModelForMultimodalLM` class is used instead of `OVModelForCausalLM`;
- chat templates for Omni models are moved from a tokenizer to a processor.

**What this PR does not cover:**
[...]

PR #4077, opened 2026-07-03
[WWB] Add Qwen3-Omni speech generation support (MoE and dense models)

avasenin-14:
## Description

[CVS-188611](https://jira.devtools.intel.com/browse/CVS-188611)

Adds Qwen3-Omni speech generation support to `who_what_benchmark` (WWB).
Qwen3-Omni is not a pure TTS model, the prompt goes through an LLM (thinker) module first before reaching audio (talker), so some additions were needed to account for that.

- Route `qwen3_omni` / `qwen3_omni_moe` models through the speech-generation
  loaders for the HF (transformers), Optimum (`OVModelForMultimodalLM`), and
  GenAI (`openvino_genai.OmniPipeline`) backends.
- Add `Qwen3OmniSpeechWrapper` and `GenAIOmniSpeechWrapper` that adapt the
  Omni `generate(..., speaker=..., return_audio=True)` output to the existing
  speech-generation evaluator result shape (`.speeches` / `.output_sample_rate`).
- Select the voice via a named speaker (e.g. `Ethan`, `Chelsie`) instead of an
  xvector speaker embedding; reject `speaker_embedding` and `--speech-language`
  since they are not supported by Qwen3-Omni.

## Checklist:
- [x] This PR follows [GenAI Contributing guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing).
- [x] Tests have been updated or added to cover the new code. 
- [ ] This PR fully addresses the ticket.
- [ ] I have made corresponding changes to the documentation.

PR #4245, opened 2026-07-30
[wwb] Seed RNG for speech-generation

sgonorov:
## Description

WWB generates the speech-generation reference and target audio in separate passes. On the HF path the models sample from the global torch RNG, and nothing pinned it between passes. The two passes drew different noise as a result, so comparing an identical model against itself scored well below 1.000: Speaker 0.857, Acoustic 0.839, Overall 0.834 on Qwen3-Omni. Calling `torch.manual_seed` by hand before generation brought it back to 1.000.

WWB already has a `--seed` argument (default 42), but it was only wired to text-to-image and text-to-video. This PR passes it into the speech-generation evaluator and pins the RNG once per prompt, right before each generation, so both passes draw the same noise. GenAI and Optimum don't use the global torch RNG, so for those backends this only adds determinism and leaves the output unchanged. The `--seed` help text now mentions speech generation.

CVS-190714

## Checklist:
- [x] Tests have been updated or added to cover the new code. Added `tools/who_what_benchmark/tests/test_unit_speechgen.py` with three unit tests: `manual_seed` is called once per prompt before each generation, a seeded evaluator yields identical audio across two runs, and `seed=None` leaves the global RNG untouched.
- [x] This PR fully addresses the ticket. The Qwen3-Omni HF path from the ticket lives in PR #4015; because that path also generates through the shared `SpeechGenerationEvaluator`, it inherits the seeding once merged.

### DIFF
--- a/tools/who_what_benchmark/tests/test_unit_speechgen.py
+++ b/tools/who_what_benchmark/tests/test_unit_speechgen.py
@@ -1,0 +1,99 @@
+# Copyright (C) 2023-2026 Intel Corporation
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+from types import SimpleNamespace
+
+import numpy as np
+import pytest
+
+
+class _Speech:
+    def __init__(self, data):
+        self.data = data
+
+
+class _SpeechResult:
+    def __init__(self, data, sample_rate=16000):
+        self.speeches = [_Speech(data)]
+        self.output_sample_rate = sample_rate
+
+
+class _RngModel:
+    """Fake TTS model whose output depends on the global torch RNG."""
+
+    def get_speaker_embedding_shape(self):
+        return (1, 512)
+
+    def generate(self, prompt, speaker_embedding=None, **kwargs):
+        import torch
+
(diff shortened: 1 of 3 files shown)


avasenin-14 (approved):

avasenin-14:
Linter is picking on unused libraries, otherwise looks good for me.

sgonorov:
Closing in favor of #4077, which adds Qwen3-Omni speech-generation support and already makes both backends deterministic by decoding greedily (talker `do_sample=False` / `talker_top_k=1`, thinker `do_sample=False`, CodePredictor `cp_top_k=1`) plus seeding `torch.manual_seed`. That removes the sampler as a source of HF-vs-GenAI divergence entirely, rather than only pinning the RNG per prompt as this PR did.

A small follow-up on top of #4077 forces the HF talker's CodePredictor greedy as well — its `do_sample=True` is hardcoded in `prepare_inputs_for_generation` and `talker_do_sample=False` does not reach it. CVS-190714 is tracked there.

PR #4247, opened 2026-07-31
[Omni] Enable Python subclassing of VLMPipelineBase/TalkerBase and add DI tests

sgonorov:
## Description

`OmniPipeline` takes a `VLMPipelineBase` and a `TalkerBase` through its dependency-injection constructor. The catch is that both base classes were registered in pybind11 without a constructor or a trampoline, so neither could be subclassed from Python. This PR makes both bases subclassable and adds functional tests that inject Python-defined children into `OmniPipeline`.

- Add a `PyTalkerBase` trampoline and expose `TalkerBase.__init__` / `generate` / `list_speakers` / `get_speaker_embedding`. A talker written in Python then dispatches from C++.
- Add a `PyVLMPipelineBase` trampoline and expose `VLMPipelineBase.__init__` plus its `generate`, tokenizer, generation-config, chat-template, and Qwen3-Omni capability-query methods. The many C++ `generate` overloads collapse to a single Python `generate`. The narrower overloads forward to the `videos_metadata`-aware one; the property-bag overloads unpack the bag and forward to it as well, rejecting only `audio_streamer` / `speech_streamer`, which no Python-defined subclass can consume.
- Defer the `TalkerBase.generate` binding until `OmniTalkerSpeechConfig` and `TalkerResults` are registered, so the generated `.pyi` renders those Python types in the signature instead of raw C++ names.
- Extend `tests/python_tests/test_omni_pipeline.py`, following the existing pipeline test patterns. It covers subclass instantiation and dispatch for both bases, plus DI tests that build `OmniPipeline(vlm, talker)` from Python mocks. Those DI tests assert three things: the speech path drives both stages, the text-only path skips the talker, and the constructor rejects a VLM that reports no audio output or no hidden-states collection.

CVS-191233

## Checklist:
- [x] Tests have been updated or added to cover the new code. Added `TestCustomTalkerSubclass`, `TestCustomVLMSubclass`, and `TestOmniPipelineDependencyInjection` in `tests/python_tests/test_omni_pipeline.py`.

### DIFF
--- a/.github/workflows/test_matrices/linux/wheel_tests.yml
+++ b/.github/workflows/test_matrices/linux/wheel_tests.yml
@@ -92,7 +92,6 @@
 - name: Omni
   cmd: |-
     python -m pip install transformers==5.1
-    python -m pip install --no-deps git+https://github.com/huggingface/optimum-intel.git@6898d97242498b7dc4367a12555e4f7e52b01d16
     python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
   timeout: 90
   components:
--- a/.github/workflows/test_matrices/linux/wheel_tests.yml
+++ b/.github/workflows/test_matrices/linux/wheel_tests.yml
@@ -89,6 +89,16 @@
   components:
     - visual_language
 
+- name: Omni
+  cmd: |-
+    python -m pip install transformers==5.1
+    python -m pip install --no-deps git+https://github.com/huggingface/optimum-intel.git@6898d97242498b7dc4367a12555e4f7e52b01d16
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
+  timeout: 90
+  components:
+    - visual_language
+    - speech_generation
+
 - name: Video Generation
   cmd: |-
     python -m pytest -v ./tests/python_tests/test_video_generation.py --override-ini cache_dir=/mount/caches/pytest/
--- a/.github/workflows/test_matrices/mac/wheel_tests.yml
+++ b/.github/workflows/test_matrices/mac/wheel_tests.yml
@@ -36,6 +36,14 @@
 #   components:
 #     - continuous_batching
 
+- name: Omni (API)
+  cmd: |-
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py -k "not TestOmniPipelineRealModel"
+  timeout: 40
+  components:
+    - visual_language
+    - speech_generation
+
 - name: GGUF Reader tests
   cmd: |-
     python -m pytest -v ./tests/python_tests/test_gguf_reader.py
--- a/.github/workflows/test_matrices/manylinux/wheel_tests.yml
+++ b/.github/workflows/test_matrices/manylinux/wheel_tests.yml
@@ -92,6 +92,8 @@
 
 - name: Omni
   cmd: |-
+    python -m pip install transformers==5.1
+    python -m pip install --no-deps git+https://github.com/huggingface/optimum-intel.git@6898d97242498b7dc4367a12555e4f7e52b01d16
     python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
   timeout: 90
   components:
--- a/.github/workflows/test_matrices/manylinux/wheel_tests.yml
+++ b/.github/workflows/test_matrices/manylinux/wheel_tests.yml
@@ -93,7 +93,6 @@
 - name: Omni
   cmd: |-
     python -m pip install transformers==5.1
-    python -m pip install --no-deps git+https://github.com/huggingface/optimum-intel.git@6898d97242498b7dc4367a12555e4f7e52b01d16
     python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
   timeout: 90
   components:
--- a/.github/workflows/test_matrices/manylinux/wheel_tests.yml
+++ b/.github/workflows/test_matrices/manylinux/wheel_tests.yml
@@ -93,7 +93,7 @@
 - name: Omni
   cmd: |-
     python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
-  timeout: 60
+  timeout: 90
   components:
     - visual_language
     - speech_generation
--- a/.github/workflows/test_matrices/manylinux/wheel_tests.yml
+++ b/.github/workflows/test_matrices/manylinux/wheel_tests.yml
@@ -90,6 +90,14 @@
   components:
     - visual_language
 
+- name: Omni
+  cmd: |-
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
+  timeout: 60
+  components:
+    - visual_language
+    - speech_generation
+
 - name: Video Generation
   cmd: |-
     python -m pytest -v ./tests/python_tests/test_video_generation.py --override-ini cache_dir=/mount/caches/pytest/
--- a/src/cpp/src/omni/talker_speech_config.cpp
+++ b/src/cpp/src/omni/talker_speech_config.cpp
@@ -10,6 +10,7 @@
 #include <nlohmann/json.hpp>
 
 #include "omni/talker_speech_config_utils.hpp"
+#include "openvino/genai/omni/pipeline.hpp"
 #include "utils.hpp"
 
 namespace ov {
@@ -139,6 +140,53 @@ void validate_omni_talker_speech_config(const OmniTalkerSpeechConfig& config) {
                     config.audio_chunk_frames,
                     ". Max allowed: ",
                     kAudioChunkFramesUpperBound);
+}
+
+namespace {
+
+std::string join_recognized_keys() {
+    std::string joined = "speech_streamer, talker_speech_config";
+    for (const auto& key : omni_talker_speech_config_keys()) {
+        joined += ", ";
+        joined += key;
+    }
+    return joined;
+}
+
+}  // namespace
+
+ResolvedTalkerProperties resolve_talker_properties(const OmniTalkerSpeechConfig& base, const ov::AnyMap& properties) {
+    ResolvedTalkerProperties out{base, std::monostate{}};
+    ov::AnyMap leftover;
+    for (const auto& [key, value] : properties) {
+        if (key == ov::genai::speech_streamer.name()) {
+            // Python kwargs arrive already unwrapped to a concrete alternative, C++ callers pass the variant.
+            if (value.is<std::shared_ptr<OmniSpeechStreamerBase>>()) {
+                out.speech_streamer = value.as<std::shared_ptr<OmniSpeechStreamerBase>>();
+            } else if (value.is<std::function<StreamingStatus(const ov::Tensor&)>>()) {
+                out.speech_streamer = value.as<std::function<StreamingStatus(const ov::Tensor&)>>();
+            } else {
+                out.speech_streamer = value.as<OmniSpeechStreamerVariant>();
+            }
+        } else if (key == ov::genai::talker_speech_config.name()) {
+            out.config = value.as<OmniTalkerSpeechConfig>();
+        } else {
+            OPENVINO_ASSERT(is_omni_talker_speech_config_key(key),
+                            "TalkerBase::generate: unrecognized property '",
+                            key,
+                            "'. Recognized keys: ",
+                            join_recognized_keys(),
+                            ".");
+            leftover.emplace(key, value);
+        }
+    }
+    if (!leftover.empty()) {
+        update_omni_talker_speech_config(out.config, leftover);
+    }
+    // Values supplied via properties bypass set_speech_config(), so this is the only guard
+    // on the AnyMap path.
+    validate_omni_talker_speech_config(out.config);
+    return out;
 }
 
 }  // namespace genai
--- a/src/python/openvino_genai/py_openvino_genai.pyi
+++ b/src/python/openvino_genai/py_openvino_genai.pyi
@@ -4499,6 +4499,12 @@ class TalkerBase:
             is Talker. Subclasses must override generate(), list_speakers(), and
             get_speaker_embedding().
     """
+    def __init__(self) -> None:
+        ...
+    def generate(self, vlm_result: VLMDecodedResults, talker_speech_config: OmniTalkerSpeechConfig, speech_streamer: collections.abc.Callable[[openvino._pyopenvino.Tensor], StreamingStatus] | openvino_genai.py_openvino_genai.OmniSpeechStreamerBase | None = None) -> TalkerResults:
+        """
+        Run speech generation against a VLM result. Override in a subclass. Returns TalkerResults.
+        """
     def get_speaker_embedding(self, name: str) -> openvino._pyopenvino.Tensor:
         ...
     def list_speakers(self) -> list[str]:
@@ -5762,7 +5768,30 @@ class VLMPipeline(VLMPipelineBase):
 class VLMPipelineBase:
     """
     Abstract base of VLM-style pipelines.
+    
+            Subclass to plug a custom thinker into OmniPipeline via its dependency-injection
+            constructor. A subclass must override generate(), get_tokenizer(), get_generation_config(),
+            set_generation_config(), set_chat_template(), and the Qwen3-Omni capability queries
+            supports_hidden_states_collection() and is_audio_output_enabled().
     """
+    def __init__(self) -> None:
+        ...
+    def generate(self, prompt: str | openvino_genai.py_openvino_genai.ChatHistory, images: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], videos: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], audios: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], videos_metadata: collections.abc.Sequence[VideoMetadata] = [], generation_config: GenerationConfig = ..., streamer: collections.abc.Callable[[str], int | None] | openvino_genai.py_openvino_genai.StreamerBase | None = None) -> VLMDecodedResults:
+        """
+        Generate a VLM response. prompt may be a str or a ChatHistory. Override in a subclass.
+        """
+    def get_generation_config(self) -> GenerationConfig:
+        ...
+    def get_tokenizer(self) -> Tokenizer:
+        ...
+    def is_audio_output_enabled(self) -> bool:
+        ...
+    def set_chat_template(self, chat_template: str) -> None:
+        ...
+    def set_generation_config(self, config: GenerationConfig) -> None:
+        ...
+    def supports_hidden_states_collection(self) -> bool:
+        ...
 class VLMRawPerfMetrics:
     """
     
--- a/src/python/openvino_genai/py_openvino_genai.pyi
+++ b/src/python/openvino_genai/py_openvino_genai.pyi
@@ -4679,9 +4679,23 @@ class TalkerBase:
     """
     def __init__(self) -> None:
         ...
+    @typing.overload
     def generate(self, vlm_result: VLMDecodedResults, talker_speech_config: OmniTalkerSpeechConfig, speech_streamer: collections.abc.Callable[[openvino._pyopenvino.Tensor], StreamingStatus] | openvino_genai.py_openvino_genai.OmniSpeechStreamerBase | None = None) -> TalkerResults:
         """
         Run speech generation against a VLM result. Override in a subclass. Returns TalkerResults.
+        """
+    @typing.overload
+    def generate(self, vlm_result: VLMDecodedResults, **kwargs) -> TalkerResults:
+        """
+                                Run speech generation against a VLM result, configured by keyword arguments.
+        
+                                Property-bag form of generate(). Recognized kwargs are the OmniTalkerSpeechConfig
+                                fields (return_audio, speaker, speaker_embedding, audio_chunk_frames,
+                                max_new_tokens, rng_seed, talker_temperature, talker_top_k,
+                                talker_repetition_penalty, cp_temperature, cp_top_k), plus speech_streamer.
+                                Unrecognized keys raise. Fields not given fall back to get_speech_config().
+        
+                                :return: TalkerResults
         """
     def get_speaker_embedding(self, name: str) -> openvino._pyopenvino.Tensor:
         ...
@@ -5987,9 +6001,15 @@ class VLMPipelineBase:
     """
     def __init__(self) -> None:
         ...
+    @typing.overload
     def generate(self, prompt: str | openvino_genai.py_openvino_genai.ChatHistory, images: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], videos: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], audios: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], videos_metadata: collections.abc.Sequence[VideoMetadata] = [], generation_config: GenerationConfig = ..., streamer: collections.abc.Callable[[str], int | None] | openvino_genai.py_openvino_genai.StreamerBase | None = None) -> VLMDecodedResults:
         """
         Generate a VLM response. prompt may be a str or a ChatHistory. Override in a subclass.
+        """
+    @typing.overload
+    def generate(self, prompt: str | openvino_genai.py_openvino_genai.ChatHistory, **kwargs) -> VLMDecodedResults:
+        """
+        Generate a VLM response from a property bag: images, videos, audios, videos_metadata, generation_config, streamer, or any GenerationConfig field as a keyword argument. Reduces to the typed generate() the subclass overrides.
         """
     def get_generation_config(self) -> GenerationConfig:
         ...
--- a/src/python/openvino_genai/py_openvino_genai.pyi
+++ b/src/python/openvino_genai/py_openvino_genai.pyi
@@ -4677,6 +4677,12 @@ class TalkerBase:
             generate(), get_speech_config(), set_speech_config(), list_speakers(), and
             get_speaker_embedding().
     """
+    def __init__(self) -> None:
+        ...
+    def generate(self, vlm_result: VLMDecodedResults, talker_speech_config: OmniTalkerSpeechConfig, speech_streamer: collections.abc.Callable[[openvino._pyopenvino.Tensor], StreamingStatus] | openvino_genai.py_openvino_genai.OmniSpeechStreamerBase | None = None) -> TalkerResults:
+        """
+        Run speech generation against a VLM result. Override in a subclass. Returns TalkerResults.
+        """
     def get_speaker_embedding(self, name: str) -> openvino._pyopenvino.Tensor:
         ...
     def get_speech_config(self) -> OmniTalkerSpeechConfig:
@@ -5973,7 +5979,30 @@ class VLMPipeline(VLMPipelineBase):
 class VLMPipelineBase:
     """
     Abstract base of VLM-style pipelines.
+    
+            Subclass to plug a custom thinker into OmniPipeline via its dependency-injection
+            constructor. A subclass must override generate(), get_tokenizer(), get_generation_config(),
+            set_generation_config(), set_chat_template(), and the Qwen3-Omni capability queries
+            supports_hidden_states_collection() and is_audio_output_enabled().
     """
+    def __init__(self) -> None:
+        ...
+    def generate(self, prompt: str | openvino_genai.py_openvino_genai.ChatHistory, images: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], videos: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], audios: collections.abc.Sequence[openvino._pyopenvino.Tensor] = [], videos_metadata: collections.abc.Sequence[VideoMetadata] = [], generation_config: GenerationConfig = ..., streamer: collections.abc.Callable[[str], int | None] | openvino_genai.py_openvino_genai.StreamerBase | None = None) -> VLMDecodedResults:
+        """
+        Generate a VLM response. prompt may be a str or a ChatHistory. Override in a subclass.
+        """
+    def get_generation_config(self) -> GenerationConfig:
+        ...
+    def get_tokenizer(self) -> Tokenizer:
+        ...
+    def is_audio_output_enabled(self) -> bool:
+        ...
+    def set_chat_template(self, chat_template: str) -> None:
+        ...
+    def set_generation_config(self, config: GenerationConfig) -> None:
+        ...
+    def supports_hidden_states_collection(self) -> bool:
+        ...
 class VLMRawPerfMetrics:
     """
     
--- a/src/python/py_omni_pipeline.cpp
+++ b/src/python/py_omni_pipeline.cpp
@@ -173,12 +173,11 @@ public:
                                speech_streamer);
     }
 
-    // OmniPipeline only ever calls the typed overload above, and the property-bag form would need
-    // the internal config-parsing helpers that these bindings deliberately don't include.
-    ov::genai::TalkerResults generate(const VLMDecodedResults&, const ov::AnyMap&) override {
-        OPENVINO_THROW(
-            "Python TalkerBase subclasses implement generate(vlm_result, talker_speech_config, speech_streamer). "
-            "The property-bag generate(vlm_result, properties) overload is not available from Python.");
+    // Reduce the property bag the same way Talker does, then dispatch through the typed
+    // override, so a Python backend answers both C++ overloads with one Python method.
+    ov::genai::TalkerResults generate(const VLMDecodedResults& vlm_result, const ov::AnyMap& properties) override {
+        auto resolved = ov::genai::resolve_talker_properties(get_speech_config(), properties);
+        return generate(vlm_result, resolved.config, resolved.speech_streamer);
     }
 
     OmniTalkerSpeechConfig get_speech_config() const override {
@@ -433,6 +432,29 @@ void init_omni_pipeline(py::module_& m) {
                     py::arg("talker_speech_config"),
                     py::arg("speech_streamer") = std::monostate{},
                     "Run speech generation against a VLM result. Override in a subclass. Returns TalkerResults.");
+
+    talker_base.def("generate",
+                    [](TalkerBase& self, const VLMDecodedResults& vlm_result, const py::kwargs& kwargs) {
+                        const ov::AnyMap properties = pyutils::kwargs_to_any_map(kwargs);
+                        ov::genai::TalkerResults res;
+                        {
+                            py::gil_scoped_release rel;
+                            res = self.generate(vlm_result, properties);
+                        }
+                        return res;
+                    },
+                    py::arg("vlm_result"),
+                    R"(
+                        Run speech generation against a VLM result, configured by keyword arguments.
+
+                        Property-bag form of generate(). Recognized kwargs are the OmniTalkerSpeechConfig
+                        fields (return_audio, speaker, speaker_embedding, audio_chunk_frames,
+                        max_new_tokens, rng_seed, talker_temperature, talker_top_k,
+                        talker_repetition_penalty, cp_temperature, cp_top_k), plus speech_streamer.
+                        Unrecognized keys raise. Fields not given fall back to get_speech_config().
+
+                        :return: TalkerResults
+                    )");
 
     py::class_<OmniDecodedResults, VLMDecodedResults>(m, "OmniDecodedResults",
         R"(Omni-specific decoded results including speech outputs.
--- a/src/python/py_omni_pipeline.cpp
+++ b/src/python/py_omni_pipeline.cpp
@@ -152,6 +152,34 @@ auto omni_generate_history_docstring = R"(
     :type videos_metadata: list[VideoMetadata]
 )";
 
+// Trampoline enabling Python subclasses of TalkerBase (injected via the OmniPipeline DI ctor).
+class PyTalkerBase : public TalkerBase {
+public:
+    using TalkerBase::TalkerBase;
+
+    ov::genai::TalkerResults generate(const VLMDecodedResults& vlm_result,
+                                      const OmniTalkerSpeechConfig& talker_speech_config,
+                                      const ov::genai::OmniSpeechStreamerVariant& speech_streamer) override {
+        py::gil_scoped_acquire acquire;
+        PYBIND11_OVERRIDE_PURE(ov::genai::TalkerResults,
+                               TalkerBase,
+                               generate,
+                               vlm_result,
+                               talker_speech_config,
+                               speech_streamer);
+    }
+
+    std::vector<std::string> list_speakers() const override {
+        py::gil_scoped_acquire acquire;
+        PYBIND11_OVERRIDE_PURE(std::vector<std::string>, TalkerBase, list_speakers);
+    }
+
+    ov::Tensor get_speaker_embedding(const std::string& name) const override {
+        py::gil_scoped_acquire acquire;
+        PYBIND11_OVERRIDE_PURE(ov::Tensor, TalkerBase, get_speaker_embedding, name);
+    }
+};
+
 py::object call_omni_generate(OmniPipeline& pipe,
                               const std::string& prompt,
                               const std::vector<ov::Tensor>& images,
@@ -211,12 +239,16 @@ py::object call_omni_generate_history(OmniPipeline& pipe,
 }  // namespace
 
 void init_omni_pipeline(py::module_& m) {
-    py::class_<TalkerBase, std::shared_ptr<TalkerBase>>(m, "TalkerBase",
+    // Register the base and the Talker subclass first (Talker needs its base registered), but defer
+    // the generate() binding until OmniTalkerSpeechConfig and TalkerResults are registered below —
+    // pybind resolves argument/return types at .def() time and would otherwise emit raw C++ names.
+    auto talker_base = py::class_<TalkerBase, PyTalkerBase, std::shared_ptr<TalkerBase>>(m, "TalkerBase",
         R"(Abstract speech-output backend for OmniPipeline.
 
         Subclass to plug a custom talker into OmniPipeline. The default implementation
         is Talker. Subclasses must override generate(), list_speakers(), and
         get_speaker_embedding().)")
+        .def(py::init<>())
         .def("list_speakers", &TalkerBase::list_speakers)
         .def("get_speaker_embedding", &TalkerBase::get_speaker_embedding, py::arg("name"));
 
@@ -320,6 +352,14 @@ void init_omni_pipeline(py::module_& m) {
         .def(py::init<>())
         .def_readonly("waveforms", &ov::genai::TalkerResults::waveforms)
         .def_readonly("perf_metrics", &ov::genai::TalkerResults::perf_metrics);
+
+    // Deferred from the TalkerBase registration so OmniTalkerSpeechConfig and TalkerResults render
+    // as Python types in the signature rather than raw C++ names.
+    talker_base.def("generate", &TalkerBase::generate,
+                    py::arg("vlm_result"),
+                    py::arg("talker_speech_config"),
+                    py::arg("speech_streamer") = std::monostate{},
+                    "Run speech generation against a VLM result. Override in a subclass. Returns TalkerResults.");
 
     py::class_<OmniDecodedResults, VLMDecodedResults>(m, "OmniDecodedResults",
         R"(Omni-specific decoded results including speech outputs.
--- a/src/python/py_omni_pipeline.cpp
+++ b/src/python/py_omni_pipeline.cpp
@@ -417,7 +417,18 @@ void init_omni_pipeline(py::module_& m) {
 
     // Deferred from the TalkerBase registration so OmniTalkerSpeechConfig and TalkerResults render
     // as Python types in the signature rather than raw C++ names.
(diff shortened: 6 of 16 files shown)


src/python/py_omni_pipeline.cpp
```diff
@@ -152,6 +152,34 @@ auto omni_generate_history_docstring = R"(
...
+
+    std::vector<std::string> list_speakers() const override {
+        py::gil_scoped_acquire acquire;
+        PYBIND11_OVERRIDE_PURE(std::vector<std::string>, TalkerBase, list_speakers);
+    }
+
+    ov::Tensor get_speaker_embedding(const std::string& name) const override {
+        py::gil_scoped_acquire acquire;
+        PYBIND11_OVERRIDE_PURE(ov::Tensor, TalkerBase, get_speaker_embedding, name);
+    }
```
Wovchena, line 201: Reminder to add get/set_speech_config() from https://github.com/openvinotoolkit/openvino.genai/pull/4218/

tests/python_tests/test_omni_pipeline.py
Wovchena, line 1: Great tests! They cover
> Additionally test that users can define their working VLMPiepelineBase and TalkerBase children and pass them to OmniPipeline

part of the ticket but there's also
> Follow other pipeline patterns.

which should be addressed

Wovchena (changes requested):

src/python/py_omni_pipeline.cpp
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

src/python/py_omni_pipeline.cpp
```diff
@@ -156,6 +156,52 @@ auto omni_generate_history_docstring = R"(
...
+                               talker_speech_config,
+                               speech_streamer);
+    }
+
+    // OmniPipeline only ever calls the typed overload above, and the property-bag form would need
+    // the internal config-parsing helpers that these bindings deliberately don't include.
+    ov::genai::TalkerResults generate(const VLMDecodedResults&, const ov::AnyMap&) override {
+        OPENVINO_THROW(
+            "Python TalkerBase subclasses implement generate(vlm_result, talker_speech_config, speech_streamer). "
+            "The property-bag generate(vlm_result, properties) overload is not available from Python.");
```
Wovchena: If `OmniPipeline` pipeline doesn't use it it doesn't mean other users won't. Keep C++ and Python API aligned

The equivalent of AnyMap is kwargs in Python. See for example https://github.com/openvinotoolkit/openvino.genai/blob/2e24cc9374aca1ded220a8dbbe2ac3cc3174773e/src/python/py_vlm_pipeline.cpp#L542

What was the problem with "the internal config-parsing helpers"?

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,56 @@
...
+   exercising virtual dispatch, the text vs speech branching, and the constructor
+   capability guards end to end.
+
+2. Real-model tests against `optimum-intel-internal-testing/tiny-random-qwen3-omni`,
+   covering the path-based constructor, text-only and speech generation, the
+   ChatHistory overload, and the speaker APIs.
+
+The real-model tier currently skips: the tiny checkpoint is a full omni model
+(talker_config, code2wav_config, enable_audio_output), but exporting it needs
+support that the pinned dependencies do not have yet. With transformers==5.0.0 the
```
Wovchena: Install a required `transformers` version in .yml workflows to fix the export

.github/workflows/test_matrices/manylinux/wheel_tests.yml
Wovchena, line 1: linux and mac workflows are missing

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    """Path to an exported tiny Qwen3-Omni model, or skip when the pinned deps cannot produce one.
+
+    Three distinct gaps are reported separately so a future failure points at the right dependency:
+    a transformers version that cannot export Qwen3-Omni at all (checked before any download), the
+    export itself raising, and the export succeeding without any talker submodels. Only the known
+    transformers 5.0.x config bug is turned into a skip — any other export failure propagates so a
+    real regression cannot hide behind a skipped test.
+    """
+    from optimum.utils.import_utils import is_transformers_version
+    from utils.atomic_download import AtomicDownloadManager
```
Wovchena: Why are imports scattered through function calls? Move them to the top

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    align_with_optimum_cli = {"padding_side": "left", "truncation_side": "left"}
+    processor = retry_request(
+        lambda: transformers.AutoProcessor.from_pretrained(
+            model_cached,
+            trust_remote_code=True,
+            **align_with_optimum_cli,
+        )
+    )
+    # No trust_remote_code here: test_vlm_pipeline.py grants it per model, and Qwen3-Omni is not on
+    # that list — the architecture is native to transformers >= 4.57, so the export needs no custom code.
```
Wovchena: Why is `test_vlm_pipeline.py` relevant? Why did you leave the comment at all?

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    model = retry_request(
+        lambda: OVModelForVisualCausalLM.from_pretrained(
+            model_cached, compile=False, device="CPU", export=True, load_in_8bit=False
+        )
+    )
+
+    tokenizer = processor.tokenizer
+    if tokenizer.chat_template is None:
+        tokenizer.chat_template = processor.chat_template
```
Wovchena: There's just a single model. You know upfront if `chat_template is None`. Remove the condition

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    """
+    from optimum.utils.import_utils import is_transformers_version
+    from utils.atomic_download import AtomicDownloadManager
+
+    # Same gate as test_vlm_pipeline.py's _maybe_skip_unsupported_model_export, applied before the
+    # snapshot download so an unsupported environment costs nothing.
+    if is_transformers_version("<", "4.57.0"):
+        pytest.skip(
+            "ValueError: The current version of Transformers does not allow for the export of Qwen3-Omni. "
+            "Minimum required is 4.57.0."
```
Wovchena: Why do we still care about 4.* versions? `tests/python_tests/requirements.txt` already points to 5.0.0: https://github.com/openvinotoolkit/openvino.genai/blob/2e24cc9374aca1ded220a8dbbe2ac3cc3174773e/tests/python_tests/requirements.txt#L5

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+
+@pytest.fixture(scope="module")
+def omni_pipe(omni_model_path: Path) -> ov_genai.OmniPipeline:
+    """Pipeline built once per module — loading the model per test dominates the suite runtime.
+
+    Reused across the real-model tests the way test_vlm_pipeline.py reuses ov_pipe_model. Safe to
+    share: none of these tests enters chat mode or mutates the stored configs, so no state carries
+    between them.
+    """
+    return ov_genai.OmniPipeline(str(omni_model_path), "CPU")
```
Wovchena: C++ has `std::filesystem::path` based constructor. Make Python also have it so the cast to `str` isn't needed

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    def test_generate_text_only(self, omni_pipe: ov_genai.OmniPipeline) -> None:
+        """With return_audio=False the pipeline decodes text and emits no waveform."""
+        result = omni_pipe.generate(
+            "Describe this.",
+            text_config=_text_config(),
+            talker_speech_config=_talker_speech_config(return_audio=False),
+        )
+
+        assert isinstance(result, ov_genai.OmniDecodedResults)
+        assert len(result.texts) == 1, "greedy decode must produce exactly one sequence"
```
Wovchena: Compare the results with optimum inference

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    Covers what the mock-based DI tests cannot: the path-based constructor, the real
+    thinker/talker composition, and the speaker APIs backed by actual model data.
+    """
+
+    def test_constructs_from_path(self, omni_pipe: ov_genai.OmniPipeline) -> None:
+        """The path-based ctor builds both stages from a single model directory."""
+        assert omni_pipe.get_vlm() is not None, "path ctor must build a VLM stage"
+        assert omni_pipe.get_talker() is not None, "path ctor must build a talker stage"
+
+    def test_generate_text_only(self, omni_pipe: ov_genai.OmniPipeline) -> None:
```
Wovchena, line 955: Add a test verifying that altering generation config in different ways alters the result

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+
+        assert len(result.speech_result.waveforms) == 1, "return_audio=True must produce one waveform"
+        waveform = np.array(result.speech_result.waveforms[0].data, dtype=np.float32).reshape(-1)
+        assert waveform.size > 0, "waveform must not be empty"
+        assert np.isfinite(waveform).all(), "waveform must contain no NaN or Inf"
+
+    def test_generate_from_chat_history(self, omni_pipe: ov_genai.OmniPipeline) -> None:
+        """The ChatHistory overload accepts structured turns and leaves the caller's history intact."""
+        history = ov_genai.ChatHistory()
+        history.append({"role": "user", "content": "Describe this."})
```
Wovchena, line 1097: include an image, video and audio to test all modalities. Compare the results with optimum inference

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+        assert len(result.texts) == 1
+        assert history.get_messages() == messages_before, "ChatHistory messages should not be mutated after generate."
+
+    def test_speaker_apis(self, omni_pipe: ov_genai.OmniPipeline) -> None:
+        """list_speakers() reports the checkpoint's voices and each resolves to an embedding."""
+        speakers = omni_pipe.get_talker().list_speakers()
+
+        assert speakers, "the checkpoint declares speakers, so list_speakers() must not be empty"
+        embedding = omni_pipe.get_talker().get_speaker_embedding(speakers[0])
+        assert embedding.get_size() > 0, f"speaker {speakers[0]!r} must resolve to a non-empty embedding"
```
Wovchena, line 1172: Test that altering embeddings produces a different speech

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +197,383 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+
+
+class TestOmniPipelineRealModel:
+    """OmniPipeline driven by an exported tiny Qwen3-Omni checkpoint.
+
+    Covers what the mock-based DI tests cannot: the path-based constructor, the real
+    thinker/talker composition, and the speaker APIs backed by actual model data.
+    """
+
+    def test_constructs_from_path(self, omni_pipe: ov_genai.OmniPipeline) -> None:
```
Wovchena, line 883: Test that VLM and Talker constructed from `ModelsMap` generate the same results as from path

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

.github/workflows/test_matrices/manylinux/wheel_tests.yml
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

src/python/py_omni_pipeline.cpp
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed.

text+image does not match optimum though. Both stacks build the same 57-token input, but the generated ids differ; text-only matches exactly, id for id. Left as a strict xfail — can't tell yet whether it's a real preprocessing difference or just this checkpoint's random weights flipping the greedy argmax, and it can't be arbitrated locally since optimum only recognizes model_type qwen3_omni_moe while every full Qwen3-Omni export here is dense qwen3_omni.

Compared on token ids rather than decoded text: the checkpoint pairs a ~151k-vocab model head with a 769-entry tokenizer, so everything decodes to '' and a text comparison could not fail.

video and audio are strict xfails for checkpoint defects too — video_processor.patch_size=14 vs 16 elsewhere, and audio_token_id=9 vs <|AUDIO|>=267 with no <|audio_pad|>.

.github/workflows/test_matrices/linux/wheel_tests.yml
```diff
@@ -79,6 +79,16 @@
   components:
     - visual_language
 
+- name: Omni
+  cmd: |-
+    python -m pip install transformers==5.1
+    python -m pip install --no-deps git+https://github.com/huggingface/optimum-intel.git@6898d97242498b7dc4367a12555e4f7e52b01d16
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
```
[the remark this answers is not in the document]
Wovchena: Pin a specific version without `*`

.github/workflows/test_matrices/linux/wheel_tests.yml
```diff
@@ -79,6 +79,15 @@
   components:
     - visual_language
 
+- name: Omni
+  cmd: |-
+    python -m pip install transformers==5.1
```
Wovchena, line 84: Reminder https://github.com/openvinotoolkit/openvino.genai/pull/4247/changes#r3866591514

.github/workflows/test_matrices/mac/wheel_tests.yml
```diff
@@ -41,6 +41,14 @@
 #   components:
 #     - continuous_batching
 
+- name: Omni (API)
+  cmd: |-
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py -k "not TestOmniPipelineRealModel"
```
Wovchena, line 46: Why does mac skip `TestOmniPipelineRealModel`?

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -4631,6 +4631,26 @@ class TalkerBase:
             generate(), get_speech_config(), set_speech_config(), list_speakers(), and
             get_speaker_embedding().
     """
+    def __init__(self) -> None:
+        ...
+    @typing.overload
+    def generate(self, vlm_result: VLMDecodedResults, talker_speech_config: OmniTalkerSpeechConfig, speech_streamer: collections.abc.Callable[[openvino._pyopenvino.Tensor], StreamingStatus] | openvino_genai.py_openvino_genai.OmniSpeechStreamerBase | None = None) -> TalkerResults:
```
Wovchena: Reminder https://github.com/openvinotoolkit/openvino.genai/pull/4247/changes#r3866591473 Although instead of changing .pyi directly, fix bindings

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    def test_rejects_backend_without_hidden_states(self) -> None:
+        """The constructor must reject a VLM that cannot collect hidden states the talker needs."""
+        vlm = RecordingVLM(hidden_states=False)
+        with pytest.raises(RuntimeError, match="supports_hidden_states_collection"):
+            ov_genai.OmniPipeline(vlm, RecordingTalker())
+
+
+def _export_tiny_omni_model(target_dir: Path) -> None:
+    """Export the tiny Qwen3-Omni checkpoint to OpenVINO IR under ``target_dir``."""
+    model_cached = snapshot_download(OMNI_MODEL_ID)  # required to avoid HF rate limits
```
Wovchena, line 590:
```suggestion
    model_cached = snapshot_download(OMNI_MODEL_ID)
```
Looks like a regular download to me without any rate limit handling

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    processor = retry_request(
+        lambda: transformers.AutoProcessor.from_pretrained(
+            model_cached,
+            trust_remote_code=True,
+            **align_with_optimum_cli,
+        )
+    )
+    model = retry_request(
+        lambda: OVModelForVisualCausalLM.from_pretrained(
+            model_cached, compile=False, device="CPU", export=True, load_in_8bit=False
```
Wovchena, line 601: `model_cached` returned by `snapshot_download()` is a local path. No need to retry. Retry `snapshot_download()` instead

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+            message = str(error)
+            if not (
+                is_transformers_version(">=", "5.0")
+                and is_transformers_version("<", "5.1")
+                and "Qwen3OmniMoeTalkerCodePredictorConfig" in message
+                and "use_sliding_window" in message
+            ):
+                raise
+            logger.info("Tiny Qwen3-Omni export hit the known transformers 5.0.x config bug: %s", message)
+            pytest.skip(f"Cannot export {OMNI_MODEL_ID} with the pinned dependencies: AttributeError: {message}")
```
Wovchena, line 641: Why are the messages reported to different channels? Merge everything into a skip reason

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+
+    missing = [name for name in TALKER_ARTIFACTS if not (model_dir / name).exists()]
+    if missing:
+        pytest.skip(
+            f"{OMNI_MODEL_ID} exported without the talker stage (missing {', '.join(missing)}); "
+            "the installed optimum-intel exported no talker stage; huggingface/optimum-intel#1700 is what "
+            "added it, so a revision predating that one cannot produce these files."
```
Wovchena, line 648: Update optimum-intel and remove the skip. We should use one of the latest commits. Note, there's a similar work in https://github.com/openvinotoolkit/openvino.genai/pull/4381

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    config = ov_genai.GenerationConfig()
+    config.max_new_tokens = max_new_tokens
+    config.do_sample = False
+    return config
+
+
+def _sampling_text_config(rng_seed: int, max_new_tokens: int = 20) -> ov_genai.GenerationConfig:
+    """Multinomial counterpart of _text_config, seeded so one run reproduces itself.
+
+    The high temperature and ignore_eos come from test_llm_pipeline.py's rng_seed pair: they keep
```
Wovchena, line 675: 1. There's just 1 `rng_seed`. What does `pair` refer to?
2. The whole docstring looks obvious except for `rng_seed pair`. Consider removing the docstring.

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,117 @@
...
+OMNI_REAL_MODEL_ENV = "OMNI_REAL_MODEL_PATH"
+
+# Written by the Talker export; without them OmniPipeline's path ctor cannot build the speech stage.
+TALKER_ARTIFACTS = (
+    "openvino_talker_model.xml",
+    "openvino_code_predictor_model.xml",
+    "openvino_code2wav_model.xml",
+)
+
+MEDIA_EDGE = 64
```
Wovchena, line 81:
```suggestion
FRAME_RESOLUTION = 64
```
or FRAME_WIDTH_AND_HEIGHT.

The word edge is too broad as well as media. Media could refer to audio

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+        assert from_map.texts == from_path.texts, "the thinker stages must agree before waveforms can be compared"
+
+        path_waveform = _single_waveform(from_path)
+        map_waveform = _single_waveform(from_map)
+
+        assert map_waveform.shape == path_waveform.shape, (
+            f"waveform lengths diverged ({map_waveform.size} vs {path_waveform.size} samples), so the two talkers "
+            "sampled different codec tokens"
+        )
+        assert np.allclose(map_waveform, path_waveform, rtol=0.0, atol=1e-4), (
```
Wovchena, line 950: `map_waveform` and `path_waveform` must match exactly

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    assert len(waveforms) == 1, f"return_audio=True must produce exactly one waveform, got {len(waveforms)}"
+    waveform = np.array(waveforms[0].data, dtype=np.float32).reshape(-1)
+    assert waveform.size > 0, "waveform must not be empty"
+    assert np.isfinite(waveform).all(), "waveform must contain no NaN or Inf"
+    return waveform
+
+
+def _waveforms_differ(left: np.ndarray, right: np.ndarray) -> bool:
+    """Whether two waveforms are different audio, by length or by samples."""
+    return left.shape != right.shape or not np.allclose(left, right)
```
Wovchena, line 725: `pytest` offers `approx()`. Use it instead: https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.approx

You already use it in other places

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+
+
+@pytest.fixture(scope="module")
+def omni_audio() -> ov.Tensor:
+    """Deterministic mono PCM as a 1-D float32 tensor at 16 kHz — what the audio encoder validates."""
+    seconds = np.arange(AUDIO_SAMPLES, dtype=np.float32) / AUDIO_SAMPLE_RATE
+    return ov.Tensor(np.sin(2.0 * np.pi * 440.0 * seconds).astype(np.float32))
+
+
+def _single_waveform(result: ov_genai.OmniDecodedResults) -> np.ndarray:
```
Wovchena, line 713:
```suggestion
def _extract_assert_single_waveform(result: ov_genai.OmniDecodedResults) -> np.ndarray:
```

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,117 @@
...
+from utils.network import retry_request
+
+logger = logging.getLogger(__name__)
+
+pytestmark = pytest.mark.omni
+
+OMNI_MODEL_ID = "optimum-intel-internal-testing/tiny-random-qwen3-omni"
+
+# Points at a full Qwen3-Omni OpenVINO export for the speech tests the tiny checkpoint cannot host.
+OMNI_REAL_MODEL_ENV = "OMNI_REAL_MODEL_PATH"
```
Wovchena, line 72: This env var isn't set by tests. Remove it

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+        ),
+    )
+    return _single_waveform(result)
+
+
+@pytest.fixture(scope="module")
+def real_omni_pipe() -> ov_genai.OmniPipeline:
+    """Pipeline over a full Qwen3-Omni export, for tests the tiny checkpoint cannot host.
+
+    Pointed at by OMNI_REAL_MODEL_PATH rather than a models/ list file: those hold HF ids consumed by
```
Wovchena, line 747: This looks like a dead code. The tests don't set `OMNI_REAL_MODEL_PATH` env var. We don't need tests which can't be run on GitHub Actions

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    if not (model_dir / "openvino_talker_model.xml").exists():
+        pytest.skip(f"{OMNI_REAL_MODEL_ENV}={model_dir} has no talker export")
+
+    return ov_genai.OmniPipeline(model_dir, "CPU")
+
+
+class _CapturingStreamer(ov_genai.StreamerBase):
+    """Records the raw token ids GenAI generates.
+
+    Decoded text is useless as a comparison target on this checkpoint: its tokenizer covers ids
```
Wovchena, line 766: > on this checkpoint

On what checkpoint?

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +258,969 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+        text_only = _talker_speech_config(return_audio=False)
+
+        short_result = omni_pipe.generate("Describe this.", text_config=short_config, talker_speech_config=text_only)
+        long_result = omni_pipe.generate("Describe this.", text_config=long_config, talker_speech_config=text_only)
+
+        short_tokens = short_result.perf_metrics.get_num_generated_tokens()
+        long_tokens = long_result.perf_metrics.get_num_generated_tokens()
+
+        assert 0 < short_tokens <= 4, f"max_new_tokens=4 must cap the decode, got {short_tokens} tokens"
+        assert 0 < long_tokens <= 24, f"max_new_tokens=24 must cap the decode, got {long_tokens} tokens"
```
Wovchena, line 987: Why is there a range of allowed values instead of the value set as `max_new_tokens`?

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,117 @@
...
+    "limitation test_tools_llm_benchmark.py xfails its omni text_to_speech cases on, for both "
+    "--optimum and --genai."
+)
+
+VIDEO_INPUT_XFAIL_REASON = (
+    "tiny-random-qwen3-omni's processor_config.json declares video_processor.patch_size=14 (a stale "
+    "Qwen2VL default) while image_processor and thinker_config.vision_config both say 16, so the video "
+    "patch dim is 2*3*14*14=1176 against an encoder built for 1536 and the infer request fails "
+    "shape.compatible(). Setting that one value to 16 makes video work on this checkpoint, and real "
+    "Qwen3-Omni models already ship 16 — the encoder and the video path itself are fine."
```
Wovchena, line 109: Report that so they update tiny-random-qwen3-omni. While you are waiting for the update, patch `video_processor.patch_size` during export. Same for `AUDIO_INPUT_XFAIL_REASON`

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,117 @@
...
+)
+
+OPTIMUM_IMAGE_XFAIL_REASON = (
+    "GenAI and optimum agree exactly on generated token ids for a text-only prompt but diverge once an "
+    "image is attached, while still agreeing on the 57-token input length — so they build the same "
+    "sequence and merge different image embeddings into it. Undetermined whether that is a real "
+    "preprocessing difference or this checkpoint's random weights making the greedy argmax flip on "
+    "numerical noise: it cannot be arbitrated locally, because optimum only recognizes model_type "
+    "qwen3_omni_moe and every full Qwen3-Omni export at hand is dense qwen3_omni."
```
Wovchena, line 118: Link a ticket

src/python/py_utils.cpp
```diff
@@ -499,7 +499,7 @@ ov::Any py_object_to_any(const py::object& py_obj, std::string property_name) {
         auto streamer = py::cast<ov::genai::pybind::utils::PyBindStreamerVariant>(py_obj);
         return ov::genai::streamer(pystreamer_to_streamer(streamer)).second;
     } else if ((py::isinstance<py::function>(py_obj) || py::isinstance<ov::genai::OmniSpeechStreamerBase>(py_obj)) &&
-               property_name == "audio_streamer") {
+               (property_name == "audio_streamer" || property_name == "speech_streamer")) {
```
Wovchena, line 502: Why are there 2 different names for same property?

src/python/py_vlm_pipeline.cpp
```diff
@@ -324,10 +479,72 @@ An input image without explicit slicing metadata counts as one slice.)")
...
+                const std::variant<std::string, ov::genai::ChatHistory>& prompt,
+                const py::kwargs& kwargs) -> ov::genai::VLMDecodedResults {
+                 const ov::AnyMap config_map = pyutils::kwargs_to_any_map(kwargs);
+                 ov::genai::VLMDecodedResults res;
+                 {
+                     py::gil_scoped_release rel;
+                     if (const auto* prompt_str = std::get_if<std::string>(&prompt)) {
+                         res = self.generate(*prompt_str, config_map);
+                     } else {
+                         res = self.generate(std::get<ov::genai::ChatHistory>(prompt), config_map);
```
Wovchena, line 533: `ChatHistory` arg should be named `history`. See `src/python/py_vlm_pipeline.cpp` as an example

src/cpp/src/omni/talker_speech_config.cpp
```diff
@@ -141,5 +142,52 @@ void validate_omni_talker_speech_config(const OmniTalkerSpeechConfig& config) {
...
+}
+
+}  // namespace
+
+ResolvedTalkerProperties resolve_talker_properties(const OmniTalkerSpeechConfig& base, const ov::AnyMap& properties) {
+    ResolvedTalkerProperties out{base, std::monostate{}};
+    ov::AnyMap leftover;
+    for (const auto& [key, value] : properties) {
+        if (key == ov::genai::speech_streamer.name()) {
+            // Python kwargs arrive already unwrapped to a concrete alternative, C++ callers pass the variant.
```
Wovchena, line 163: Why can't Python bindings wrap the arg with variant so there's no extra handling?

Wovchena (changes requested):

.github/workflows/test_matrices/linux/wheel_tests.yml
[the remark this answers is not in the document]
sgonorov: Fixed

.github/workflows/test_matrices/linux/wheel_tests.yml
sgonorov: Fixed

.github/workflows/test_matrices/mac/wheel_tests.yml
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

src/python/py_vlm_pipeline.cpp
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: I'd rather do this in a follow-up PR.

tests/python_tests/test_omni_pipeline.py
sgonorov: Let's track this separately, outside this PR.

src/python/py_utils.cpp
sgonorov: It predates this PR — separate change, I think.

src/cpp/src/omni/talker_speech_config.cpp
sgonorov: I think we should move it to a separate PR.

tests/python_tests/test_omni_pipeline.py
sgonorov: Tracked in https://jira.devtools.intel.com/browse/CVS-194799 — covers the `video_processor.patch_size` mismatch, the `audio_token_id` inconsistency and the out-of-range talker role token ids, plus reporting them so the checkpoint gets regenerated.

tests/python_tests/test_omni_pipeline.py
sgonorov: https://jira.devtools.intel.com/browse/CVS-194800

src/python/py_utils.cpp
sgonorov: Correction to my earlier reply: this does not predate the PR. Master accepts only `audio_streamer`; the `speech_streamer` name is added here. Tracked together with the related unwrap handling in https://jira.devtools.intel.com/browse/CVS-194802

src/cpp/src/omni/talker_speech_config.cpp
sgonorov: Tracked in https://jira.devtools.intel.com/browse/CVS-194802, together with the duplicate `speech_streamer` / `audio_streamer` property name — same root cause: the binding stores the unwrapped concrete alternative rather than the variant.

.github/workflows/test_matrices/mac/wheel_tests.yml
```diff
@@ -36,6 +36,17 @@
 #   components:
 #     - continuous_batching
 
+# OmniPipeline needs continuous batching, which is only supported on X64 or ARM with SVE
+# support, so the real-model tests are left to the other platforms.
+- name: Omni (API)
+  cmd: |-
+    python -m pip install transformers==5.1.0
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py -k "not TestOmniPipelineRealModel"
```
Wovchena: Follow up on https://github.com/openvinotoolkit/openvino.genai/pull/4247/changes#r3893654711: Mac still skips `TestOmniPipelineRealModel`

.github/workflows/test_matrices/linux/wheel_tests.yml
```diff
@@ -89,6 +89,15 @@
...
     - visual_language
 
+- name: Omni
+  cmd: |-
+    python -m pip install transformers==5.1.0
+    python -m pytest -v ./tests/python_tests/test_omni_pipeline.py --override-ini cache_dir=/mount/caches/pytest/
+  timeout: 90
+  components:
+    - visual_language
+    - speech_generation
```
Wovchena: Omni files changes aren't labeled at all. Add a new `omni` component.

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +233,871 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+    return model_dir
+
+
+@pytest.fixture(scope="module")
+def omni_pipe(omni_model_path: Path) -> ov_genai.OmniPipeline:
+    """Pipeline built once per module — loading the model per test dominates the suite runtime.
+
+    Reused across the real-model tests the way test_vlm_pipeline.py reuses ov_pipe_model. Safe to
+    share: none of these tests enters chat mode or mutates the stored configs, so no state carries
+    between them.
```
Wovchena:
```suggestion
```
The comment looks irrelevant

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,92 @@
...
+OPTIMUM_IMAGE_XFAIL_REASON = (
+    "GenAI and optimum agree exactly on generated token ids for a text-only prompt but diverge once an "
+    "image is attached, while still agreeing on the 57-token input length — so they build the same "
+    "sequence and merge different image embeddings into it. Undetermined whether that is a real "
+    "preprocessing difference or this checkpoint's random weights making the greedy argmax flip on "
+    "numerical noise: it cannot be arbitrated locally, because optimum only recognizes model_type "
+    "qwen3_omni_moe and every full Qwen3-Omni export at hand is dense qwen3_omni."
+)
+
+AUDIO_INPUT_XFAIL_REASON = (
```
Wovchena: Create the tickets for xfails and link here

tests/python_tests/test_omni_pipeline.py
```diff
@@ -160,3 +233,871 @@ def test_talker_blob_ctor_signature(self) -> None:
...
+                "the audio encoder must run when audios are passed"
+            )
+
+    def test_speaker_apis(self, omni_pipe: ov_genai.OmniPipeline) -> None:
+        """list_speakers() reports the checkpoint's voices and each resolves to an embedding."""
+        speakers = omni_pipe.get_talker().list_speakers()
+
+        assert speakers, "the checkpoint declares speakers, so list_speakers() must not be empty"
+        embedding = omni_pipe.get_talker().get_speaker_embedding(speakers[0])
+        assert embedding.get_size() > 0, f"speaker {speakers[0]!r} must resolve to a non-empty embedding"
```
Wovchena: Follow up on https://github.com/openvinotoolkit/openvino.genai/pull/4247/changes#r3829016168
I don't see
> Test that altering embeddings produces a different speech

being fixed

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,92 @@
...
+OMNI_MODEL_ID = "optimum-intel-internal-testing/tiny-random-qwen3-omni"
+
+FRAME_RESOLUTION = 64
+VIDEO_FRAMES = 4
+AUDIO_SAMPLE_RATE = 16000
+AUDIO_SAMPLES = AUDIO_SAMPLE_RATE
+
+OPTIMUM_COMPARE_TOKENS = 6
+
+NO_WAVEFORM_XFAIL_REASON = (
```
Wovchena: What is the supposed fix for this? If there's no fix, it doesn't make sense to keep the codd

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,92 @@
...
+    "tiny-random-qwen3-omni's talker role token ids are outside its tokenizer's range (im_start 151644, "
+    "user 872, assistant 77091, against a max id of 769), so build_talker_input finds no segments and "
+    "speech generation raises before reaching code2wav — see "
+    "test_speech_generation_rejects_unmatched_role_tokens, which locks that error in. Same checkpoint "
+    "limitation test_tools_llm_benchmark.py xfails its omni text_to_speech cases on, for both "
```
Wovchena: `test_tools_llm_benchmark.py` is huge. Link a specific place or remove the reference.
Same for `AUDIO_INPUT_XFAIL_REASON`

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,92 @@
...
+    "Qwen3-Omni models already ship 16 — the encoder and the video path itself are fine."
+)
+
+OPTIMUM_IMAGE_XFAIL_REASON = (
+    "GenAI and optimum agree exactly on generated token ids for a text-only prompt but diverge once an "
+    "image is attached, while still agreeing on the 57-token input length — so they build the same "
+    "sequence and merge different image embeddings into it. Undetermined whether that is a real "
+    "preprocessing difference or this checkpoint's random weights making the greedy argmax flip on "
+    "numerical noise: it cannot be arbitrated locally, because optimum only recognizes model_type "
+    "qwen3_omni_moe and every full Qwen3-Omni export at hand is dense qwen3_omni."
```
Wovchena: Link tickets in all xfail reasons

src/python/py_utils.cpp
```diff
@@ -499,7 +504,7 @@ ov::Any py_object_to_any(const py::object& py_obj, std::string property_name) {
...
```
Wovchena: Follow up on https://github.com/openvinotoolkit/openvino.genai/pull/4247/changes#r3903111399

This seems to be limited to this PR. Fix it right away instead of creating a ticket

src/cpp/src/omni/talker_speech_config.cpp
Wovchena: Follow up on https://github.com/openvinotoolkit/openvino.genai/pull/4247/changes#r3903160567

> the binding stores the unwrapped concrete alternative rather than the variant.


That's exacrty my question: 
> Why can't Python bindings wrap the arg with variant so there's no extra handling?

tests/python_tests/test_omni_pipeline.py
```diff
@@ -14,19 +14,92 @@
...
+from huggingface_hub import snapshot_download
+from optimum.intel.openvino import OVModelForVisualCausalLM
+from optimum.utils.import_utils import is_transformers_version
 
 import openvino_genai as ov_genai
+from utils.atomic_download import AtomicDownloadManager
+from utils.constants import get_ov_cache_converted_models_dir
+from utils.network import retry_request
+
+pytestmark = pytest.mark.omni
```
Wovchena: This mark isn't used. yml workflows specify a file path instead. Remove the mark from here and `tests/python_tests/pytest.ini`

Wovchena (changes requested):

.github/workflows/test_matrices/mac/wheel_tests.yml
sgonorov: Fixed

.github/workflows/test_matrices/linux/wheel_tests.yml
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

src/python/py_utils.cpp
sgonorov: Fixed

src/cpp/src/omni/talker_speech_config.cpp
sgonorov: Fixed

tests/python_tests/test_omni_pipeline.py
sgonorov: Fixed

Issue #3805, opened 2026-05-05
Add Speculative Decoding Miss Rate Metric to LLM Benchmark Reporting


acastrogonzalez94:
## Summary

Add Speculative Decoding Miss Rate Metric to LLM Benchmark Reporting

---

## Description

When running LLM benchmarks with speculative decoding (i.e., using a draft model via `--draft_model` and `--num_assistant_tokens`), the tool currently has no visibility into how well the draft model is performing. Specifically, there is no way to tell what percentage of draft tokens are being rejected by the target model — a critical signal for tuning and debugging speculative decoding setups.

We'd like to add a **miss rate metric** to the benchmark output that is automatically calculated and reported whenever speculative decoding is active.

### What is miss rate?

Miss rate represents the percentage of draft tokens that were generated by the draft model but ultimately rejected by the main model. A high miss rate indicates the draft model is a poor fit for the given task or prompts, while a low miss rate signals an efficient and well-matched draft/main model pair.

---

## Suggested Approach

> The implementation team is free to handle this however makes sense — the logic below is provided as a reference only.

In `tools/llm_bench/task/text_generation.py`, after collecting `inference_durations` and `num_tokens`, the following logic can serve as a starting point:

```python
# Calculate speculative decoding miss rate
total_output_tokens = num_tokens // args['batch_size']
total_inferences = len(inference_durations)
if args.get('draft_model', '') and args.get('num_assistant_tokens', 0) > 0:
    assistant_tokens = int(args['num_assistant_tokens'])
    draft_tokens_generated = (total_inferences - 1) * assistant_tokens  # -1 for first token
    accepted_draft_tokens = total_output_tokens - total_inferences
    rejected_draft_tokens = draft_tokens_generated - accepted_draft_tokens
    miss_rate = (rejected_draft_tokens / draft_tokens_generated * 100) if draft_tokens_generated > 0 else 0

    log.info(f"{prefix} Draft Model Performance:")
    log.info(f"{prefix}   Draft tokens generated: {draft_tokens_generated}")
    log.info(f"{prefix}   Draft tokens rejected (misses): {rejected_draft_tokens}")
    log.info(f"{prefix}   Miss rate: {miss_rate:.1f}%")
```

This was validated using the following benchmark run:

```bash
taskset -c 4-11 python3 .../llm_bench/benchmark.py \
  -m=<main_model_path> \
  -d=GPU \
  -pf=<prompts_file> \
[...]

PR #4065, opened 2026-07-01
Qwen35 MTP Support

apaniukov:
## Description
Support MTP speculative decoding for Qwen3.5

CVS-187942

Based on:
- Optimum-Intel PR: https://github.com/huggingface/optimum-intel/pull/1814
- Prompt Lookup PR:https://github.com/openvinotoolkit/openvino.genai/pull/4056

Code example:
```python
import time
from pathlib import Path

import openvino_genai as ovg

MODEL_DIR = Path("Qwen3.5-2B")
MAX_NEW_TOKENS = 30
NUM_ASSISTANT_TOKENS = 1

PROMPTS = [
    "1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26, 27, 28,",
    "Once upon a time there was a little girl named Goldilocks. She went for a walk in the forest. Pretty soon, she came upon a house. She knocked and",
    "The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the lazy dog. The quick brown fox jumps over the",
]


def make_config():
    config = ovg.GenerationConfig()
    config.max_new_tokens = MAX_NEW_TOKENS
    config.num_assistant_tokens = NUM_ASSISTANT_TOKENS
    return config


def generate(pipe, prompt, config):
    t0 = time.perf_counter()
    result = pipe.generate([prompt], [config])[0]
    elapsed = time.perf_counter() - t0
    # ContinuousBatchingPipeline.generate on string prompts returns decoded text per sequence.
    text = result.m_generation_ids[0]
    return text, elapsed


def report_timing(tag, n_tokens, elapsed):
    ms_per_tok = 1000 * elapsed / n_tokens if n_tokens else float("nan")
    tok_per_s = n_tokens / elapsed if elapsed else float("nan")
    print(f"  {tag:9s} {elapsed:.3f}s for {n_tokens} tokens ({ms_per_tok:.1f} ms/tok, {tok_per_s:.1f} tok/s)")


def main():
    scheduler_config = ovg.SchedulerConfig()

    print("Loading pipelines...")
    baseline_pipe = ovg.ContinuousBatchingPipeline(str(MODEL_DIR), scheduler_config, "CPU", {})
    mtp_pipe = ovg.ContinuousBatchingPipeline(
        str(MODEL_DIR),
        scheduler_config,
        "CPU",
[...]

PR #4248, opened 2026-07-31
[Speculative Decoding] : Add speculative decoding metrics reporting to llm_bench and wwb

PrashantTakale369:
- Uses the already-exposed `extended_perf_metrics` (`SDPerModelsPerfMetrics`) from the generate result : no C++/binding changes needed.
- Draft counts are : proposed = `draft_model_metrics.get_num_generated_tokens()`, accepted = `get_num_accepted_tokens()`; miss rate = ( proposed − accepted ) / proposed. The draft counters accumulate across generate calls of the same pipeline, so the previous total is tracked to report per call values.
- Non-speculative runs are unaffected : no extra log line, no extra CSV columns.
- In wwb, `genai_gen_text` now passes the prompt as a list so `generate` returns `DecodedResults` (which carries the metrics) instead of a plain string; the answer text and similarity scores are unchanged.

### Tested with

SmolLM2-360M (main) + SmolLM2-135M (draft) on CPU:
- llm_bench with `--draft_model` then  metric line appears ; without it no line
- Values cross checked against GenAI's internal debug output (`OPENVINO_LOG_LEVEL=4`)
- wwb with/without draft model :  CSV columns appear only for the speculative run [ as requested by @sbalandi ]

Metric line
image

CSV columns
image

#3805


tools/llm_bench/task/text_generation.py
```diff
@@ -367,7 +367,31 @@ def token_printer():
         generated_tokens = np.array(generation_result.tokens)
 
     perf_metrics = generation_result[0].perf_metrics if cb_pipeline else generation_result.perf_metrics
-    return generated_tokens, perf_metrics, end - start
+    extended_perf_metrics = generation_result[0].extended_perf_metrics if cb_pipeline else generation_result.extended_perf_metrics
+    return generated_tokens, perf_metrics, extended_perf_metrics, end - start
+
+
+# the draft model raw metrics are accumulated over generate() calls of the same pipeline,
+# while num_accepted_tokens is reported per call, so track the previous total to get per-call values
```
as-suvorov: It looks like we need to fix it on GenAI side. @pavel-esir Could you please take a look at it

tools/llm_bench/task/text_generation.py
```diff
@@ -367,7 +367,31 @@ def token_printer():
...
+    if not hasattr(extended_perf_metrics, "get_num_accepted_tokens"):
+        return None
+    total_draft_generated = extended_perf_metrics.draft_model_metrics.get_num_generated_tokens()
+    num_draft_generated = total_draft_generated - sd_prev_draft_generated.get(id(model), 0)
+    sd_prev_draft_generated[id(model)] = total_draft_generated
+    num_accepted = extended_perf_metrics.get_num_accepted_tokens()
+    if num_draft_generated <= 0:
+        return None
+    return {
+        "num_draft_generated": num_draft_generated,
```
sbalandi: let's print more metrics for main and draft model: generate duration, TTFT, TTST, TPOT, first infer/other infer latency, throughput

tools/llm_bench/llm_bench_utils/metrics_print.py
sbalandi, line 1: Could you please update tests for llm_bench with SD pipeline and add assert, that string "Speculative decoding: draft tokens generated:" exists in the output ? Tests is placed here https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_tools_llm_benchmark.py#L143 and https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_tools_llm_benchmark.py#L165 .
Example of assert: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_tools_llm_benchmark.py#L1004

tools/llm_bench/task/text_generation.py
PrashantTakale369: @sbalandi Sure, makes sense. I’ll add these metrics for both the main and draft models,

tools/llm_bench/llm_bench_utils/metrics_print.py
PrashantTakale369: Ok I’ll update the `llm_bench` tests for the SD pipeline and add an assertion to verify that `Speculative decoding: draft tokens generated` is present in the output.
and also I’ll add the assertion to the relevant SD tests following the existing pattern in the test file.

tools/llm_bench/task/text_generation.py
PrashantTakale369: > let's print more metrics for main and draft model: generate duration, TTFT, TTST, TPOT, first infer/other infer latency, throughput

Hi @sbalandi Added for both main and draft model : generate duration, TTFT, TTST, TPOT, first infer latency, other infers latency, throughput, plus generated tokens and inference count.

Values are computed in get_sd_per_model_metrics()
Printed by print_sd_per_model_metrics(), called [once per model](https://github.com/openvinotoolkit/openvino.genai/blob/7801da211c797c2044dd2fc7440fe6c82800a574/tools/llm_bench/llm_bench_utils/metrics_print.py#L124-L125)

image

tools/llm_bench/llm_bench_utils/metrics_print.py
PrashantTakale369: > Could you please update tests for llm_bench with SD pipeline and add assert, that string "Speculative decoding: draft tokens generated:" exists in the output ? Tests is placed here https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_tools_llm_benchmark.py#L143 and https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_tools_llm_benchmark.py#L165 . Example of assert: https://github.com/openvinotoolkit/openvino.genai/blob/master/tests/python_tests/samples/test_tools_llm_benchmark.py#L1004

Added the assert to both tests you linked, following the pattern from L1004:

`test_python_tool_llm_benchmark_speculative (master L143)`
`test_python_tool_llm_benchmark_sd_generation_config (master L165)`
`test_python_tool_llm_benchmark_sd_generation_config_json_string ` : added this one too, since it exercises the same SD path via an inline JSON string and would otherwise be an inconsistent gap

tools/llm_bench/task/text_generation.py
```diff
@@ -391,7 +391,88 @@ def token_printer():
...
+    num_accepted = extended_perf_metrics.get_num_accepted_tokens()
+    # a negative value means the counters got out of sync, so the per call value is unknown
+    if num_draft_generated < 0:
+        return None
+    return {
+        "num_draft_generated": num_draft_generated,
+        "num_accepted": num_accepted,
+        "acceptance_rate": num_accepted / num_draft_generated * 100 if num_draft_generated > 0 else 0.0,
+        "miss_rate": (num_draft_generated - num_accepted) / num_draft_generated * 100 if num_draft_generated > 0 else 0.0,
+        "main_model": main_model,
```
sbalandi: Sorry for the confusion, but we have this PR: https://github.com/openvinotoolkit/openvino.genai/pull/4065 , it will be merged soon , several metrics are introduced there 

1. draft_candidate_tokens `extended_perf_metrics->get_num_draft_tokens()`, could you please add it ?
2. rejected_tokens `extended_perf_metrics->get_num_rejected_tokens()`, could you please add and use in miss_rate calculation as  `(rejected_tokens / draft_candidate_tokens * 100 if draft_candidate_tokens > 0 else 0.0,` ?
3. acceptance_rate  `extended_perf_metrics->get_draft_acceptance_rate()`, it's better to use it as `acceptance_rate`
4. num_draft_generated remains, but a new draft model-specific metric has been added that means the same thing but has a different name: `get_num_draft_processed_tokens()`. I'd suggest renaming metric here too, as this more closely reflects their meaning
5. and also `get_draft_to_main_inference_duration_ratio()` is added, could you please add it ?

tools/llm_bench/llm_bench_utils/metrics_print.py
sbalandi: great, thank you !

tools/llm_bench/task/text_generation.py
PrashantTakale369: > Sorry for the confusion, but we have this PR: #4065 , it will be merged soon , several metrics are introduced there
> 
> 1. draft_candidate_tokens `extended_perf_metrics->get_num_draft_tokens()`, could you please add it ?
> 2. rejected_tokens `extended_perf_metrics->get_num_rejected_tokens()`, could you please add and use in miss_rate calculation as  `(rejected_tokens / draft_candidate_tokens * 100 if draft_candidate_tokens > 0 else 0.0,` ?
> 3. acceptance_rate  `extended_perf_metrics->get_draft_acceptance_rate()`, it's better to use it as `acceptance_rate`
> 4. num_draft_generated remains, but a new draft model-specific metric has been added that means the same thing but has a different name: `get_num_draft_processed_tokens()`. I'd suggest renaming metric here too, as this more closely reflects their meaning
> 5. and also `get_draft_to_main_inference_duration_ratio()` is added, could you please add it ?

@sbalandi added / change

1. Now `draft_candidate_tokens` from `get_num_draft_tokens()`
2. `rejected_tokens` from `get_num_rejected_tokens()`, used in miss_rate
3. `acceptance_rate` now from `get_draft_acceptance_rate()`
4. renamed `num_draft_generated` → `draft_processed_tokens` (llm_bench + wwb + tests)
5. added `draft_to_main_inference_duration_ratio`

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -17,6 +17,52 @@
...
+    extended_perf_metrics = getattr(answer, "extended_perf_metrics", None)
+    if not hasattr(extended_perf_metrics, "get_num_accepted_tokens"):
+        return None
+    draft_model_metrics = extended_perf_metrics.draft_model_metrics
+    # some backends accumulate the metrics over generate() calls of the same pipeline, others reset them,
+    # so the already reported iterations are detected by the draft model latencies to know which case it is
+    durations = draft_model_metrics.raw_metrics.m_durations
+    prev_count, prev_marker = sd_prev_counters.get((id(model), 'durations'), (0, None))
+    is_accumulated = 0 < prev_count <= len(durations) and durations[prev_count - 1] == prev_marker
+    sd_prev_counters[(id(model), 'durations')] = (len(durations), durations[-1] if len(durations) > 0 else None)
```
sbalandi: could you please convert function to the class and check if accumulated once on the first iteration ? and also, please, move it to the another file utils.py

tools/who_what_benchmark/whowhatbench/text_evaluator.py
```diff
@@ -17,6 +17,52 @@
...
+        return None
+    token_numbers = {
+        "draft_processed_tokens": draft_processed_tokens,
+        "num_accepted": num_accepted,
+    }
+    # the candidate token metrics are not reported by every pipeline
+    if hasattr(extended_perf_metrics, "get_num_draft_tokens"):
+        candidates = get_delta('candidates', extended_perf_metrics.get_num_draft_tokens())
+        try:
+            rejected = get_delta('rejected', extended_perf_metrics.get_num_rejected_tokens())
```
sbalandi: could you please also collected get_draft_acceptance_rate in this case ?

tools/llm_bench/task/text_generation.py
```diff
@@ -391,7 +391,141 @@ def token_printer():
         generated_tokens = np.array(generation_result.tokens)
 
     perf_metrics = generation_result[0].perf_metrics if cb_pipeline else generation_result.perf_metrics
-    return generated_tokens, perf_metrics, end - start
+    extended_perf_metrics = generation_result[0].extended_perf_metrics if cb_pipeline else generation_result.extended_perf_metrics
+    return generated_tokens, perf_metrics, extended_perf_metrics, end - start
+
+
+# counter values of each model already reported, see get_sd_counter_delta()
```
sbalandi: could you please create metrcis_collection_utils.py or smth like that and move code here ?

sbalandi:
have created a task to the fix for the history accumulation CVS#194060
I would merge the code as is with workarounds now, what do you think @as-suvorov ?
FYI @as-suvorov

sbalandi:
@PrashantTakale369 the lint problem may be fixed on the commit creation step with pre-commit hook - https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#code-quality

sbalandi:
@PrashantTakale369 it was decided that we need to fix the problem with metric accumulation in the SD GenAI pipelines first. So, please, remove the logic with the workaround for metric accumulation management.
The merge of the current PR is currently being postponed until a pull request with a fix is ​​created and merged.

tools/who_what_benchmark/whowhatbench/text_evaluator.py
PrashantTakale369: @sbalandi 
- Converted the function into a class : `SDMetricsCollector`
- And Moved it to `utils.py` as requested
- On the accumulation : tracing the code, and the counters do accumulate across iterations the same pipeline is reused for every prompt in the loop, so the underlying counters carry over from one `generate()` call to the next instead of resetting per prompt

PrashantTakale369: @sbalandi also this is the same root cause you flagged for `llm_bench` tracked as CVS#194060  
so rather than adding a workaround here too I left it to be fixed at the source

tools/who_what_benchmark/whowhatbench/text_evaluator.py
PrashantTakale369: `SDMetricsCollector` now collects `get_draft_acceptance_rate() `and reports it as the `acceptance_rate` column.

tools/llm_bench/task/text_generation.py
PrashantTakale369: @sbalandi created `tools/llm_bench/llm_bench_utils/metrics_collection_utils.py` and moved the SD metric collection code there. 
- `text_generation.py `now just imports `get_sd_metrics` from it.

PrashantTakale369:
> @PrashantTakale369 it was decided that we need to fix the problem with metric accumulation in the SD GenAI pipelines first. So, please, remove the logic with the workaround for metric accumulation management. The merge of the current PR is currently being postponed until a pull request with a fix is ​​created and merged.

@sbalandi The metric accumulation workaround has been removed from this PR so we’ll keep this PR on hold until the pipeline fix is created and merged.

sbalandi:
> > @PrashantTakale369 it was decided that we need to fix the problem with metric accumulation in the SD GenAI pipelines first. So, please, remove the logic with the workaround for metric accumulation management. The merge of the current PR is currently being postponed until a pull request with a fix is ​​created and merged.
> 
> @sbalandi The metric accumulation workaround has been removed from this PR so we’ll keep this PR on hold until the pipeline fix is created and merged.

ok, thank you !

PR #4255, opened 2026-08-03
[CB] Sum GPU memory across devices for multi-GPU KV cache validation

### LINKED ISSUE #4254: [CB] Multi-GPU (HETERO pipeline-parallel): KV cache memory validation only counts first execution device
## Problem

When running continuous batching on multiple GPUs via HETERO pipeline-parallel (e.g., `HETERO:GPU.0,GPU.1` with `MODEL_DISTRIBUTION_POLICY=PIPELINE_PARALLEL`), the KV cache memory budget used for `cache_size` / `num_kv_blocks` normalization only accounts for the **first** execution device.

In `src/cpp/src/continuous_batching/pipeline_impl.cpp` (`ContinuousBatchingImpl::initialize_pipeline`):

```cpp
const std::string execution_device = execution_devices[0];
...
auto get_available_memory = [&execution_device](const std::string& /*device*/, size_t num_cache_tensors) -> size_t {
    if (execution_device.find("GPU") != std::string::npos) {
        return utils::get_available_gpu_memory(execution_device, num_cache_tensors);
    }
    return get_available_cpu_memory();
};
auto cache_orchestrator = CacheOrchestrator::create(infer_request, normalized_config, get_available_memory);
```

Since the KV cache is distributed across all GPUs, `cache_size` validation undercounts available memory: setting `scheduler_config.cache_size` above a single GPU's free VRAM (but within the combined total) fails validation even though the cache would fit.

The `all_gpu_device` variable introduced in PR #2227 (pipeline-parallel support) is computed a few lines above but is never used for memory aggregation.

This was previously reported/patched in a fork: https://github.com/ambeckley/openvino.genai/commit/dd2c914e6863dac93b984b6cf78c5d689206e6c1 — the code has since been refactored (CacheOrchestrator), but the limitation remains.

## Proposed fix

The refactor makes this a one-lambda change at the call site; `CacheOrchestrator` is untouched since it receives a `std::function<size_t(const std::string&, size_t)>`:

```cpp
auto get_available_memory = [&execution_device, &execution_devices, all_gpu_device](
                                const std::string& /*device*/, size_t num_cache_tensors) -> size_t {
    // Multi-GPU (e.g. HETERO:GPU.0,GPU.1 pipeline-parallel): the KV cache is
    // distributed across all execution devices, so aggregate their free memory
    // instead of validating cache_size against the first device only.
    if (all_gpu_device && execution_devices.size() > 1) {
        size_t total_mem_size = 0;
        for (const auto& dev : execution_devices) {
            total_mem_size += utils::get_available_gpu_memory(dev, num_cache_tensors);
        }
        return total_mem_size;
    }
    if (execution_device.find("GPU") != std::string::npos) {
        return utils::get_available_gpu_memory(execution_device, num_cache_tensors);
    }
    return get_available_cpu_memory();
};
```

Why this is safe:

- `compile_model` runs before the lambda executes, and `utils::get_available_gpu_memory` derives free memory from `ov::intel_gpu::memory_statistics` — actual used VRAM including the already-resident weights — so summing per-device free memory is correct w.r.t. model footprint.
- Single-GPU and CPU paths are byte-identical to today.

One design consideration for review: with pipeline parallelism each GPU holds an equal share of KV layers, so the true capacity bound is closer to `num_devices * min(free_i)` than `sum(free_i)`. On identical, evenly loaded GPUs these are equal; on imbalanced setups the sum is optimistic. Happy to implement either variant.

## Reproduction

1. Two GPUs, e.g. 2x Intel Arc with 16GB each
2. Load any LLM with `HETERO:GPU.0,GPU.1` + `MODEL_DISTRIBUTION_POLICY=PIPELINE_PARALLEL` and continuous batching
3. Set `scheduler_config.cache_size` (GB) above one GPU's free VRAM but below the combined total
4. Before: validation error claiming insufficient memory. After: validates and runs.

duncanmcqueen:
## Description

When running continuous batching on multiple GPUs via HETERO pipeline-parallel (e.g., `HETERO:GPU.0,GPU.1` with `MODEL_DISTRIBUTION_POLICY=PIPELINE_PARALLEL`), the available-memory query used for `cache_size` / `num_kv_blocks` normalization only counts the **first** execution device (`execution_devices[0]`). Since the KV cache is distributed across all GPUs, `cache_size` validation undercounts available memory and rejects values that would fit across the combined VRAM.

This uses the existing `all_gpu_device` flag (introduced in #2227 for pipeline-parallel support, but not previously used for memory aggregation) to sum `utils::get_available_gpu_memory` over all GPU execution devices. The change is confined to the `get_available_memory` lambda at the `CacheOrchestrator::create` call site; single-GPU and CPU paths are byte-identical.

Based on https://github.com/ambeckley/openvino.genai/commit/dd2c914e6863dac93b984b6cf78c5d689206e6c1 by @ambeckley, rebased onto the CacheOrchestrator refactor.

Why summing per-device free memory is safe here: `compile_model` runs before this lambda executes, and `utils::get_available_gpu_memory` derives free memory from `ov::intel_gpu::memory_statistics` — actual used VRAM including the already-resident weights.

Design note for reviewers: with pipeline parallelism each GPU holds an equal share of KV layers, so the strict capacity bound is closer to `num_devices * min(free_i)` than `sum(free_i)`. On identical, evenly loaded GPUs these coincide; the sum matches the original patch's behavior and can be tightened if preferred.

## Testing

Not yet tested on multi-GPU hardware by this author — validation help welcome. Reproduction steps in the linked issue: 2 GPUs, `HETERO:GPU.0,GPU.1` + `PIPELINE_PARALLEL`, set `scheduler_config.cache_size` above one GPU's free VRAM but below the combined total; fails before, should pass after.

Fixes #4254


apaniukov (approved):

apaniukov:
build_jenkins

apaniukov:
build_jenkins

apaniukov:
build_jenkins

apaniukov:
build_jenkins

PR #4258, opened 2026-08-03
[Whisper]: Accept plain language codes in addition to wrapped tokens

byekrang:
## Description

Whisper's `language` generation parameter previously accepted only the wrapped token form and rejected plain language codes:

```python
language="<|en|>"  # accepted
language="en"      # rejected
```

This change makes both forms valid while preserving backward compatibility:

```python
language="<|en|>"  # still accepted
language="en"      # now also accepted
```

Language resolution is handled by the new Whisper-specific helper:

`ov::genai::utils::get_or_throw_token_id_by_language(lang_to_id, language)`

The helper accepts either a plain language code, such as `"en"`, or an already wrapped Whisper language token, such as `"<|en|>"`. Plain input is normalized to the wrapped form and looked up against the wrapped keys stored in `lang_to_id`. It returns the resolved token ID directly and throws if the normalized language is not present.

The `lang_to_id` keys use the wrapped format found in the converted model's `generation_config.json`; plain map keys are not supported. Invalid language values remain rejected during generation-configuration validation before generation begins.

Both Whisper generation paths are updated consistently:

* the regular pipeline in `whisper.cpp`
* the static/NPU pipeline in `pipeline_static.cpp`

`WhisperGenerationConfig::validate()` uses the same helper, ensuring that plain and wrapped values are handled consistently at both the configuration and pipeline layers.

The change is limited to Whisper. Qwen3-ASR language handling, where values such as `"English"` are treated as free text with different semantics, uses a separate code path and is unchanged.

Public API comments, Python docstrings, generated Python stubs, and the JavaScript type description have been updated to describe both accepted formats.

GitHub Pages documentation under `site/docs/` will be added to this draft PR in a separate commit before it is marked ready for review.

CVS-192021

## Checklist:

* [x] This PR follows [[GenAI Contributing guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing)](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing).

  * Added `test_whisper_language_validate_accepts_plain_and_wrapped`, a model-free test that verifies validation accepts `"en"` and `"<|en|>"`, rejects unknown values, and does not mutate the configured language.
  * Updated `test_language_mode` to exercise plain and wrapped language formats end-to-end with both the `WHISPER` and generic `ASR` pipeline types.
  * Updated `test_static_whisper_language_de` to exercise both `"de"` and `"<|de|>"` through the static `pipeline_static.cpp` configured-language path using NPUW:CPU.

  * Plain Whisper language codes are accepted.
  * Wrapped language tokens remain supported.
  * Invalid language values remain rejected.
  * Qwen3-ASR behavior is unchanged.

  * Public API comments, Python docstrings and stubs, and the JavaScript type description are included in the current commit.
  * GitHub Pages documentation and its preview link will be added in a separate commit to this PR.

### DIFF
--- a/site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
+++ b/site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
@@ -162,8 +162,8 @@ ASR models can automatically detect the language of the input audio, or you can 
         result = pipe.generate(raw_speech)
         print(result.languages[0])  # e.g. "en", "fr", "de"
 
-        # Explicitly specify language (English). Use "English" for Qwen3-ASR models.
-        result = pipe.generate(raw_speech, language="<|en|>")
+        # Explicitly specify language. In the form of `en`, `<|en|>` for Whisper and `English` for Qwen3-ASR
+        result = pipe.generate(raw_speech, language="en")
         print(result.languages[0])  # "en"
 
         # French speech sample
@@ -181,7 +181,7 @@ ASR models can automatically detect the language of the input audio, or you can 
             auto result = pipe.generate(raw_speech);
             std::cout << result.languages[0] << "\n";  // e.g. "en", "fr", "de"
 
-            // Explicitly specify language (English). Use "English" for Qwen3-ASR models.
+            // Explicitly specify language. In the form of `en`, `<|en|>` for Whisper and `English` for Qwen3-ASR
             result = pipe.generate(raw_speech, ov::genai::language("<|en|>"));
             std::cout << result.languages[0] << "\n";  // "en"
 
--- a/src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
+++ b/src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
@@ -13,8 +13,8 @@ public:
     explicit ASRGenerationConfig(const std::filesystem::path& json_path);
 
     /**
-     * @brief Language token to use for generation
-     * In the form of <|en|> for Whisper models. Can be set for multilingual models only.
+     * @brief Language to use for generation
+     * In the form of `en`, `<|en|>` for Whisper models. Can be set for multilingual models only.
      * In the form of English for Qwen3-ASR models.
      */
     std::optional<std::string> language = std::nullopt;
--- a/src/cpp/include/openvino/genai/whisper_generation_config.hpp
+++ b/src/cpp/include/openvino/genai/whisper_generation_config.hpp
@@ -45,7 +45,7 @@ public:
 
     bool is_multilingual = true;
 
-    // Language token to use for generation in the form of <|en|>.
+    // Language to use for generation. In the form of `en`, `<|en|>`.
     // You can find all the possible language tokens in the generation_config.json lang_to_id dictionary.
     // Can be set for multilingual models only.
     std::optional<std::string> language = std::nullopt;
--- a/src/cpp/src/whisper/pipeline_static.cpp
+++ b/src/cpp/src/whisper/pipeline_static.cpp
@@ -265,9 +265,7 @@ ov::genai::SotTokensResult prepare_sot_tokens(ov::Tensor& encoder_hidden_state,
     std::string language;
     if (config.language.has_value()) {
         language = *config.language;
-        if (config.lang_to_id.count(language)) {
-            language_token_id = config.lang_to_id.at(language);
-        }
+        language_token_id = ov::genai::utils::find_token_id_by_language(config.lang_to_id, language);
     } else {
         language_token_id = detect_language(encoder_hidden_state, decoder, config, raw_metrics);
         language = ov::genai::utils::find_language_by_token_id(config.lang_to_id, language_token_id);
--- a/src/cpp/src/whisper/pipeline_static.cpp
+++ b/src/cpp/src/whisper/pipeline_static.cpp
@@ -265,7 +265,7 @@ ov::genai::SotTokensResult prepare_sot_tokens(ov::Tensor& encoder_hidden_state,
     std::string language;
     if (config.language.has_value()) {
         language = *config.language;
-        language_token_id = ov::genai::utils::find_token_id_by_language(config.lang_to_id, language);
+        language_token_id = ov::genai::utils::get_or_throw_token_id_by_language(config.lang_to_id, language);
     } else {
         language_token_id = detect_language(encoder_hidden_state, decoder, config, raw_metrics);
         language = ov::genai::utils::find_language_by_token_id(config.lang_to_id, language_token_id);
--- a/src/cpp/src/whisper/whisper.cpp
+++ b/src/cpp/src/whisper/whisper.cpp
@@ -260,9 +260,7 @@ ov::genai::SotTokensResult prepare_sot_tokens(ov::Tensor& encoder_hidden_state,
     std::string language;
     if (config.language.has_value()) {
         language = *config.language;
-        if (config.lang_to_id.count(language)) {
-            language_token_id = config.lang_to_id.at(language);
-        }
+        language_token_id = ov::genai::utils::find_token_id_by_language(config.lang_to_id, language);
     } else {
         auto [language_token, infer_ms] = decoder->detect_language(encoder_hidden_state, config);
         language_token_id = language_token;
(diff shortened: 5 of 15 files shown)


tests/python_tests/test_whisper_pipeline.py
```diff
@@ -458,8 +458,9 @@ def test_max_new_tokens(model_descr, sample_from_dataset, pipeline_type):
 
 @pytest.mark.transformers_lower_v5(reason="CVS-185784")
 @pytest.mark.parametrize("model_descr", get_whisper_models_list(tiny_only=True))
-@pytest.mark.parametrize("language", ["fr", "de"])
-def test_language_mode(model_descr, language, pipeline_type):
+@pytest.mark.parametrize("language_format", ["plain", "wrapped"])
+def test_language_mode(model_descr, language_format, pipeline_type):
+    language = "fr"
```
as-suvorov: Please keep `fr` and `de` languages for this test. Let's parametrize both, language and format

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
```diff
@@ -163,7 +167,7 @@ ASR models can automatically detect the language of the input audio, or you can
         print(result.languages[0])  # e.g. "en", "fr", "de"
 
         # Explicitly specify language (English). Use "English" for Qwen3-ASR models.
```
as-suvorov:
```suggestion
        # Explicitly specify language. In the form of `en`, `<|en|>` for Whisper and `English` for Qwen3-ASR
```

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
```diff
@@ -152,6 +152,10 @@ Beam search support depends on the selected model architecture. Qwen3-ASR models
 
 ASR models can automatically detect the language of the input audio, or you can specify the language to improve accuracy. The detected (or specified) language is available in the result via the `languages` field:
 
+:::info
+For Whisper models, `language` accepts either a plain language code such as `"en"` or the wrapped token form `"<|en|>"`. Qwen3-ASR models instead expect a language name such as `"English"`.
+:::
+
```
as-suvorov:
```suggestion
Let's keep it in a sample comments
```

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
```diff
@@ -13,8 +13,9 @@ class OPENVINO_GENAI_EXPORTS ASRGenerationConfig : public GenerationConfig {
     explicit ASRGenerationConfig(const std::filesystem::path& json_path);
 
     /**
-     * @brief Language token to use for generation
-     * In the form of <|en|> for Whisper models. Can be set for multilingual models only.
+     * @brief Language to use for generation
+     * For Whisper models, accepts either a plain language code such as "en" or the wrapped token
+     * form "<|en|>". Can be set for multilingual models only.
```
as-suvorov:
```suggestion
     * In the form of `en`, `<|en|>` for Whisper models. Can be set for multilingual models only.
```
Please also update python bindings

src/cpp/include/openvino/genai/whisper_generation_config.hpp
```diff
@@ -45,7 +45,8 @@ class OPENVINO_GENAI_EXPORTS WhisperGenerationConfig : public GenerationConfig {
 
     bool is_multilingual = true;
 
-    // Language token to use for generation in the form of <|en|>.
+    // Language to use for generation. Accepts either a plain language code such as "en" or the
+    // wrapped token form "<|en|>".
```
as-suvorov:
```suggestion
    // Language to use for generation. In the form of `en`, `<|en|>`.
```

src/cpp/src/whisper/whisper.cpp
```diff
@@ -256,13 +256,14 @@ ov::genai::SotTokensResult prepare_sot_tokens(ov::Tensor& encoder_hidden_state,
         return {std::vector<int64_t>{config.decoder_start_token_id}, "en"};
     }
 
-    int64_t language_token_id = 0;
+    int64_t language_token_id = -1;  // Assigned in both branches; initialized for static analysis.
```
as-suvorov: Why default changed?

src/cpp/src/whisper/pipeline_static.cpp
```diff
@@ -261,13 +261,14 @@ ov::genai::SotTokensResult prepare_sot_tokens(ov::Tensor& encoder_hidden_state,
...
+    int64_t language_token_id = -1;  // Assigned in both branches; initialized for static analysis.
     std::string language;
     if (config.language.has_value()) {
         language = *config.language;
-        if (config.lang_to_id.count(language)) {
-            language_token_id = config.lang_to_id.at(language);
-        }
+        const auto resolved = ov::genai::utils::find_token_id_by_language(config.lang_to_id, language);
+        OPENVINO_ASSERT(resolved.has_value(),
+                        "'language' " + language + " must be provided in 'lang_to_id' map.");
```
as-suvorov: `ov::genai::utils::find_token_id_by_language` used with `OPENVINO_ASSERT` only. Consider to move assert to this util and make result non optional.

as-suvorov (changes requested):

src/cpp/src/whisper/whisper.cpp
byekrang: Thank you for pointing this out. I initialized it to `-1` because I thought it would make the unassigned state more explicit, since `0` may be a valid token ID.

The current control flow assigns `language_token_id` before it is used, so I agree that the sentinel is not functionally necessary. If you prefer, I can revert it to `0` to preserve the original initialization and keep the change minimal.

src/cpp/src/whisper/pipeline_static.cpp
byekrang: Wow, that makes sense. Thank you!

src/cpp/src/whisper/whisper_utils.cpp
```diff
@@ -110,6 +110,33 @@ std::string find_language_by_token_id(const std::map<std::string, int64_t>& lang
...
 
+int64_t find_token_id_by_language(const std::map<std::string, int64_t>& lang_to_id, const std::string& language) {
+    // Exact-key-first keeps custom lang_to_id maps (whose keys need not be
+    // wrapped) working: try the value exactly as provided.
+    if (auto it = lang_to_id.find(language); it != lang_to_id.end()) {
+        return it->second;
+    }
+
+    // Otherwise translate between the wrapped ("<|en|>") and plain ("en") forms
+    // so either input resolves regardless of the map's key style. Only the exact
```
as-suvorov: I'm not use if we need to support both `<|en|>` and `en` keys format. The configs are usually hardcoded and we can rely on them. I would prefer to have a single format for user provided maps as well.
I believe just:
```
const bool is_wrapped = ...
std::string wrapped_language = language;
if (!is_wrapped) {
    language = "<|" + language + "|>"
}

// find in map
```

Please let me know if you have any concern regarding this and prefer to use `alternative` approach

src/cpp/src/whisper/whisper_utils.hpp
```diff
@@ -33,6 +36,11 @@ ov::genai::WhisperGenerationConfig prepare_per_generate_config(
 
 std::string find_language_by_token_id(const std::map<std::string, int64_t>& lang_to_id, int64_t token_id);
 
+// Resolve a Whisper language token id, accepting either the wrapped form
+// ("<|en|>") or the plain code ("en"). Throws if neither form exists in
+// lang_to_id.
+int64_t find_token_id_by_language(const std::map<std::string, int64_t>& lang_to_id, const std::string& language);
```
as-suvorov:
```suggestion
int64_t get_or_throw_token_id_by_language(const std::map<std::string, int64_t>& lang_to_id, const std::string& language);
```

src/cpp/src/whisper/whisper_utils.cpp
byekrang: That makes sense!

as-suvorov (commented):
I didn't find updates for previous comments, so I unresolved them

as-suvorov (approved):

PR #3915, opened 2026-05-27
fix(python): Remove incompatible keep_alive from def_readwrite in GenerationConfig

frenchwr:
## Description

The `py::keep_alive<1, 2>()` policy is no longer supported with pybind11's `def_readwrite()` method ([link](https://github.com/pybind/pybind11/pull/5533)), causing a static assertion failure during compilation. This policy is unnecessary as pybind11 automatically handles lifetime management for `def_readwrite` attributes.

This is similar to the error fixed by this PR in OpenVINO: https://github.com/openvinotoolkit/openvino/pull/34468

I've tested that this fixes compatibility with pybind11 when building Python bindings for OpenVINO GenAI 2026.1.0.0.

To reproduce the error (building environment: Ubuntu 24.04), build OpenVINO GenAI as an extra module:

```
git clone --branch 2026.1.0 --recursive https://github.com/openvinotoolkit/openvino.git
git clone --branch 2026.1.0.0 --recursive https://github.com/openvinotoolkit/openvino.genai.git
cmake   -DCMAKE_INSTALL_PREFIX=$(pwd)/install -DENABLE_PYTHON=ON   -DPython3_EXECUTABLE=/usr/bin/python3.12   -DOPENVINO_EXTRA_MODULES=$(pwd)/openvino.genai   -DCPACK_ARCHIVE_COMPONENT_INSTALL=OFF   -S ./openvino   -B ./build
cmake --build ./build -j
.
.
In file included from /home/ubuntu/openvino-genai-python/openvino.genai/src/python/py_generation_config.cpp:5:
/home/ubuntu/openvino-genai-python/openvino/src/bindings/python/thirdparty/pybind11/include/pybind11/pybind11.h: In instantiation of ‘pybind11::class_<type_, options>& pybind11::class_<type_, options>::def_property_static(const char*, const pybind11::cpp_function&, const pybind11::cpp_function&, const Extra& ...) [with Extra = {pybind11::is_method, pybind11::return_value_policy, pybind11::keep_alive<1, 2>}; type_ = ov::genai::GenerationConfig; options = {}]’:
[...]

PR #4261, opened 2026-08-04
python: remove invalid keep_alive policy from parsers property

stevegrubb:
pybind11 3.0.4 rejects keep_alive<1, 2> on the parsers def_readwrite property with a compile-time static assertion. This stops the Python-binding build; it is not a warning and is independent of selecting C++17 or C++20. The policy assumes callable argument positions that a generated property setter does not expose.

Remove the call policy. GenerationConfig stores parsers as shared_ptr values, so assigning the vector already keeps each parser alive. Add a regression test that assigns parsers, releases the original Python references, and reads or uses the stored property.


as-suvorov:
Thank you for PR, to be addressed in: https://github.com/openvinotoolkit/openvino.genai/pull/3915

PR #4266, opened 2026-08-04
cmake: make GenAI install destinations configurable

stevegrubb:
OpenVINO GenAI hard-codes an archive-oriented runtime directory hierarchy for libraries, headers, executables, CMake metadata, Python modules, and license files. This forces system packagers to relocate files after installation and can leave exported target paths inconsistent with the final layout.

Expose cache variables for each install destination while preserving the current paths as defaults. Use the same values in install rules, exported targets, and package configuration so an alternate GNU-style layout remains a valid CMake package.

Example use:
cmake -S . -B build \
    -DCMAKE_INSTALL_PREFIX=/usr \
    -DOPENVINO_GENAI_INSTALL_LIBDIR=/usr/lib64 \
    -DOPENVINO_GENAI_INSTALL_BINDIR=/usr/bin \
    -DOPENVINO_GENAI_INSTALL_INCLUDEDIR=/usr/include \
    -DOPENVINO_GENAI_INSTALL_CMAKEDIR=/usr/lib64/cmake/OpenVINOGenAI \
    -DOPENVINO_GENAI_INSTALL_PYTHONDIR=/usr/lib64/python3.14/site-packages \
    -DOPENVINO_GENAI_INSTALL_LICENSEDIR=/usr/share/licenses/openvino-genai

Then check it by:
cmake -N -LA build | grep '^OPENVINO_GENAI_INSTALL_'

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,8 +102,12 @@ if(EXISTS "${OpenVINOGenAI_SOURCE_DIR}/tests/cpp" AND ENABLE_TESTS)
     add_subdirectory(tests/cpp)
 endif()
 
-install(FILES LICENSE DESTINATION docs/licensing COMPONENT licensing_genai RENAME LICENSE-GENAI)
-install(FILES third-party-programs.txt DESTINATION docs/licensing COMPONENT licensing_genai RENAME third-party-programs-genai.txt)
+set(OPENVINO_GENAI_INSTALL_LICENSEDIR "docs/licensing" CACHE PATH
+    "Directory used to install OpenVINO GenAI license files")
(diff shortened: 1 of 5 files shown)


as-suvorov:
@stevegrubb Thank you for PR. I believe we should be aligned with OpenVINO regarding installation options. Could you please check if OpenVINO supports such configurable paths? And if not we probably need to address it for OpenVINO first

stevegrubb:
Actually, hold off on merging until I update it. Yes, openvino itself can be installed anywhere. But I'd like to align this patch with the way openvino does it so that they match. I'll post an update later.

stevegrubb:
I think this will be cleanest if I close this PR and open a new one so that the conversation doesn't start one way and then suddenly the patch is totally different. I'll open a new PR in few minutes.

PR #4274, opened 2026-08-05
[Whisper]: Fix beam search with timestamps

byekrang:
## Description

  Whisper beam search with `return_timestamps=True` produced output that diverged from the Optimum/Hugging Face reference. Greedy decoding was
  unaffected and remains unchanged. The existing beam-search test previously ran without timestamps, so the defect was not exercised.

  **Root cause** — three independent issues in the timestamp/beam-scoring path:

  1. `process_whisper_timestamp_logits` computed `log_softmax` using beam row `0` for every beam, while applying masks to the correct row
  identified by `batch_idx`. Under multi-beam search, every beam after the first made its timestamp decision using the wrong distribution.
  2. The "force a timestamp" rule compared the aggregate timestamp probability against the maximum over the entire vocabulary, including timestamp
  tokens, instead of comparing it against the most likely non-timestamp token.
  3. Beam scores were produced by applying `log_softmax` after the suppression and timestamp `-inf` masks. This renormalized the surviving tokens,
  shifted each row by a different constant, and changed the relative cumulative scores between beams.

  **Changes:**

  * `whisper/logit_processor.cpp`

    * Use `batch_idx` for timestamp probability computation.
    * Restrict the best-token comparison to the non-timestamp range `[0, timestamp_begin)`.

  * `whisper/whisper.cpp`

    * For beam search, normalize each logits row over the full vocabulary before applying Whisper masks, via the new `logits_row_to_logprobs()`
  helper.
    * Move `SequenceGroup` construction from `whisper_generate()` into `decode()`, where it is actually used, and mark the request as carrying
  log-probabilities when beam search is active.

  * `sequence_group.hpp`

    * Add the internal `LogitsType` enum (`RAW` or `LOG_PROBS`) as a per-request member, with a getter and setter, describing what the logits
  tensor handed to the sampler contains.

  * `sampling/sampler.cpp`

    * Add `read_row()`, which reads a logits row without rescaling it.
    * In `GroupBeamSearcher::select_next_tokens`, use the sequence group's `LogitsType` to either normalize raw logits as before or consume
  precomputed log-probabilities without normalizing them again.

  There are no public API or binding changes. `LogitsType` is internal.

  CVS-167239

  ## Checklist

  * [x] This PR follows the [GenAI Contributing
  guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing).
  * [x] Tests have been updated to cover the new behavior. `tests/python_tests/test_whisper_pipeline.py::test_beam_search` is now parametrized
  over `return_timestamps=[False, True]`, so it keeps the previous no-timestamps coverage and adds the timestamps case for both the WHISPER and
  ASR pipelines across two samples. The four timestamps-enabled cases failed before the fix and pass after it. Broader validation showed no change
  to greedy decoding output.
  * [x] Documentation update is not required because this change fixes internal behavior without changing the public API or user-facing
  configuration.


src/cpp/src/sampling/sampler.cpp
```diff
@@ -70,6 +70,24 @@ std::vector<Token> log_softmax(const ov::Tensor& logits, size_t batch_idx) {
...
 
+// Read a row of precomputed log-probabilities without modifying the scores.
+std::vector<Token> read_row_as_logprobs(const ov::Tensor& logits, size_t batch_idx) {
+    ov::Shape shape = logits.get_shape();
+    OPENVINO_ASSERT(shape.size() == 3);
+    size_t batch = shape[0], seq_len = shape[1], vocab_size = shape[2];
+    OPENVINO_ASSERT(batch_idx < batch, "Logits batch size doesn't match the number of beams");
+
+    size_t batch_offset = batch_idx * seq_len * vocab_size, sequence_offset = (seq_len - 1) * vocab_size;
+    const float* beam_logits = logits.data<const float>() + batch_offset + sequence_offset;
```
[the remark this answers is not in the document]
byekrang: `read_row_as_logprobs()` already validates that the tensor is rank 3 and that `batch_idx` is within the batch dimension. In addition, `ov::Tensor::data<const float>()` validates compatibility between the requested fundamental type and the tensor element type and throws on mismatch, so an explicit `f32` assertion would duplicate that check. This is also consistent with the existing sibling `log_softmax()`, which uses `data<const float>()` without an additional element-type assertion.

src/cpp/src/sampling/sampler.cpp
```diff
@@ -338,7 +357,9 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
         std::vector<Beam> candidates;
         candidates.reserve(group_size * 2 * group_size);
         for (const Beam& beam : group.ongoing) {
-            std::vector<Token> tokens = log_softmax(logits, beam.m_global_beam_idx);
+            std::vector<Token> tokens = m_beam_score_input == BeamScoreInput::LOG_PROBABILITIES
```
as-suvorov: I believe `beam_score_input` can be obtained from sequence group so storing in sampler not required.

src/cpp/src/sampling/sampler.cpp
```diff
@@ -70,6 +70,27 @@ std::vector<Token> log_softmax(const ov::Tensor& logits, size_t batch_idx) {
     return tokens;
 }
 
+// Read a row of precomputed log-probabilities without modifying the scores.
+static std::vector<Token> read_row_as_logprobs(const ov::Tensor& logits, size_t batch_idx) {
```
as-suvorov:
```suggestion
static std::vector<Token> read_row_as_logprobs(const ov::Tensor& logits, const size_t batch_idx) {
```

src/cpp/src/sampling/sampler.cpp
```diff
@@ -70,6 +70,27 @@ std::vector<Token> log_softmax(const ov::Tensor& logits, size_t batch_idx) {
...
+    ov::Shape shape = logits.get_shape();
```
as-suvorov:
```suggestion
    const ov::Shape shape = logits.get_shape();
```
we usually use const by default

src/cpp/src/sampling/sampler.cpp
```diff
@@ -70,6 +70,27 @@ std::vector<Token> log_softmax(const ov::Tensor& logits, size_t batch_idx) {
...
+    OPENVINO_ASSERT(shape.size() == 3);
+
+    size_t batch = shape[0], seq_len = shape[1], vocab_size = shape[2];
+    OPENVINO_ASSERT(batch_idx < batch, "Logits batch size doesn't match the number of beams");
+    OPENVINO_ASSERT(seq_len > 0);
+    OPENVINO_ASSERT(vocab_size > 0);
+
+    size_t batch_offset = batch_idx * seq_len * vocab_size,
+           sequence_offset = (seq_len - 1) * vocab_size;
```
as-suvorov: Please consider assign non trivial expressions with a dedicated assignments. We don't have code style forcing, so it's just a reccommendation.
```suggestion
    const size_t batch_offset = batch_idx * seq_len * vocab_size;
    const size_t sequence_offset = (seq_len - 1) * vocab_size;
```

src/cpp/src/whisper/whisper.cpp
```diff
@@ -351,7 +387,12 @@ WhisperGenerateResult whisper_generate(const ov::genai::WhisperGenerationConfig&
             chunk_sot_tokens.push_back(config.no_timestamps_token_id);
         }
 
-        SequenceGroup::Ptr sequence_group = std::make_shared<SequenceGroup>(0, chunk_sot_tokens, config);
+        // Keep the sampler's score representation consistent with process_whisper_logits().
+        const bool is_beam_search = config.is_beam_search();
+        const BeamScoreInput beam_score_input =
+            is_beam_search ? BeamScoreInput::LOG_PROBABILITIES : BeamScoreInput::LOGITS;
+        SequenceGroup::Ptr sequence_group =
+            std::make_shared<SequenceGroup>(0, chunk_sot_tokens, config, beam_score_input);
```
as-suvorov: let's move `sequence_group` creation into `decode`, it seems not used in this function besides passing to decode.

src/cpp/src/whisper/whisper.cpp
```diff
@@ -27,22 +29,56 @@ using ov::genai::MicroSeconds;
...
+    const ov::Shape shape = logits.get_shape();
+    OPENVINO_ASSERT(shape.size() == 3);
+    OPENVINO_ASSERT(batch_idx < shape[0], "Logits batch size doesn't match the number of beams");
+    OPENVINO_ASSERT(shape[1] > 0);
+    OPENVINO_ASSERT(shape[2] > 0);
+
+    const size_t vocab_size = shape[2];
+    const size_t batch_offset = batch_idx * shape[1] * vocab_size;
+    const size_t sequence_offset = (shape[1] - 1) * vocab_size;
+    float* row = logits.data<float>() + batch_offset + sequence_offset;
```
as-suvorov:
```suggestion
    const float* row = logits.data<float>() + batch_offset + sequence_offset;
```

src/cpp/src/whisper/whisper.cpp
```diff
@@ -27,22 +29,56 @@ using ov::genai::MicroSeconds;
...
 void process_whisper_logits(ov::Tensor logits,
                             const ov::genai::WhisperGenerationConfig& config,
                             const bool return_timestamps,
                             const std::map<size_t, std::vector<int64_t>>& batch_to_generated_ids) {
     const bool initial_step = batch_to_generated_ids.empty();
     const size_t batch_size = logits.get_shape().at(0);
 
+    // Normalize before masking so beam scores remain comparable across rows, matching HF beam search.
+    const bool is_beam_search = config.is_beam_search();
+    const std::vector<int64_t> empty_generated_ids;
```
as-suvorov: Do we need this variable?

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -831,14 +831,16 @@ def test_longform_audio_with_word_level_timestamps(model_descr, sample_from_data
     "sample_from_dataset", [*get_fixture_params_for_n_whisper_dataset_samples(n=2, long_form=True)], indirect=True
 )
 @pytest.mark.xfail(condition=(sys.platform == "darwin"), reason="Ticket - 173169")
-def test_beam_search(model_descr, sample_from_dataset, pipeline_type):
-    # use only 30 seconds of audio due to beam search results wrong with enabled timestamps
-    # ticket: 167239
+@pytest.mark.parametrize("return_timestamps", [False, True])
+def test_beam_search(model_descr, sample_from_dataset, pipeline_type, return_timestamps):
+    # Cut to 30 seconds to keep this regression focused and limit CI cost.
+    # With timestamps enabled, beam search previously produced text that diverged from the reference.
```
as-suvorov:
```suggestion
```

src/cpp/src/sequence_group.hpp
```diff
@@ -426,18 +432,21 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
...
     // output seq len at current iteration
     size_t m_output_seq_len = 0;
 
     size_t m_num_streamed_tokens = 0, m_stream_window_size = 0;
     TimePoint m_start_time = std::chrono::steady_clock::now();
     PerfMetrics m_perf_metrics;
 
-    SequenceGroup(uint64_t request_id, const ov::genai::GenerationConfig& sampling_params)
+    SequenceGroup(uint64_t request_id, const ov::genai::GenerationConfig& sampling_params,
+                  BeamScoreInput beam_score_input = BeamScoreInput::LOGITS)
```
as-suvorov: Let's keep current ctor.
```suggestion
    SequenceGroup(uint64_t request_id, const ov::genai::GenerationConfig& sampling_params)
```

src/cpp/src/sequence_group.hpp
```diff
@@ -453,8 +462,9 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
     using CPtr = std::shared_ptr<const SequenceGroup>;
 
     // const_cast is safe as ov::Tensor only views the data and doesn't modify it.
-    SequenceGroup(uint64_t request_id, const TokenIds& input_ids, const ov::genai::GenerationConfig& sampling_params)
-        : SequenceGroup(request_id, ov::Tensor(ov::element::i64, ov::Shape{input_ids.size()}, const_cast<int64_t*>(input_ids.data())), sampling_params, std::nullopt, std::nullopt) {
+    SequenceGroup(uint64_t request_id, const TokenIds& input_ids, const ov::genai::GenerationConfig& sampling_params,
+                  BeamScoreInput beam_score_input = BeamScoreInput::LOGITS)
```
as-suvorov:
```suggestion
    SequenceGroup(uint64_t request_id, const TokenIds& input_ids, const ov::genai::GenerationConfig& sampling_params)
```

src/cpp/src/sequence_group.hpp
```diff
@@ -464,8 +474,9 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
                   const std::optional<std::unordered_map<std::string, ov::Tensor>>& lm_extra_inputs = std::nullopt,
                   const std::optional<ov::Tensor>& position_ids = std::nullopt,
                   const std::optional<int64_t>& rope_delta = std::nullopt,
-                  const std::optional<ov::Tensor>& prompt_ids = std::nullopt)
-        : SequenceGroup(request_id, sampling_params) {
+                  const std::optional<ov::Tensor>& prompt_ids = std::nullopt,
+                  BeamScoreInput beam_score_input = BeamScoreInput::LOGITS)
```
as-suvorov:
```suggestion
                  const std::optional<ov::Tensor>& prompt_ids = std::nullopt)
```

src/cpp/src/sampling/sampler.cpp
```diff
@@ -338,7 +360,9 @@ void Sampler::GroupBeamSearcher::select_next_tokens(const ov::Tensor& logits,
         std::vector<Beam> candidates;
         candidates.reserve(group_size * 2 * group_size);
         for (const Beam& beam : group.ongoing) {
-            std::vector<Token> tokens = log_softmax(logits, beam.m_global_beam_idx);
+            std::vector<Token> tokens = m_beam_score_input == BeamScoreInput::LOG_PROBABILITIES
+                                            ? read_row_as_logprobs(logits, beam.m_global_beam_idx)
```
as-suvorov:
```suggestion
                                            ? read_row(logits, beam.m_global_beam_idx)
```

src/cpp/src/sequence_group.hpp
```diff
@@ -426,18 +432,21 @@ class SequenceGroup  : public std::enable_shared_from_this<SequenceGroup> {
     size_t m_num_validation_tokens = 0;
     // flag to enable/disable token generation, e.g. in speculative decoding scenario
     bool m_is_gen_paused = false;
+    BeamScoreInput m_beam_score_input = BeamScoreInput::LOGITS;
```
as-suvorov: I couldn't find an elegant way to pass logits/logprobs info to sampler. Let's store type as a member and implement type setter.

```suggestion
    LogitsType m_logits_type = LogitsType::RAW;
```

@apaniukov Maybe you can suggest a better solution?

src/cpp/src/sequence_group.hpp
```diff
@@ -30,6 +30,12 @@ enum class SequenceGroupType {
     EMBEDDINGS
 };
 
+// Representation of scores consumed by beam search.
+enum class BeamScoreInput {
+    LOGITS,
+    LOG_PROBABILITIES
+};
```
as-suvorov:
```suggestion
enum class LogitsType {
    RAW,
    LOG_PROBS
};
```
or
```suggestion
enum class ScoreType {
    LOGITS,
    LOG_PROBS
};
```

as-suvorov (changes requested):

src/cpp/src/whisper/whisper.cpp
byekrang: const float* wouldn't work here because this helper normalizes the row in place (row[i] -= log_sum). I use float* const row instead, so the pointer itself is const while the underlying logits remain mutable.

as-suvorov: Sure, my bad

as-suvorov (approved):

as-suvorov:
@apaniukov Could you please review sequence group and sampler changes.

PR #4277, opened 2026-08-06
support setting num_assistant_tokens to 0 for eagle3

xufang-lisa:
## Description
This PR updates GenerationConfig::num_assistant_tokens from size_t to std::optional<size_t> so that an unset value and an explicitly configured 0 can be handled distinctly. For both Stateful and Continuous Batching Eagle3 pipelines, num_assistant_tokens=0 enables main-model-only generation.

The most important changes are:

**Core API and Logic Refactoring:**

* Changed `num_assistant_tokens` in `GenerationConfig` from `size_t` to `std::optional<size_t>`, and updated all related logic to check for presence and value, ensuring explicit handling of unset or zero values. 

**Validation and Error Handling:**

* Strengthened assertions throughout speculative and tree search code to ensure `num_assistant_tokens` is set and valid when required, and clarified mutual exclusivity with `assistant_confidence_threshold`. 

**Speculative Decoding and Pipeline Integration:**

* Updated speculative decoding strategies (`DFlash`, `Eagle3`, and `FastDraft`) to use the new optional type, including logic to set defaults, handle main-only requests, and propagate the parameter correctly. 

**Draft/Main Pipeline Coordination:**

* Improved validation and coordination between main and draft pipelines, especially in `Eagle3` and `FastDraft` strategies, to ensure correct request pairing and handling of cases where speculative decoding is disabled per-request. 

CVS-191892

## Checklist:


src/cpp/src/speculative_decoding/stateful/stateful_pipeline_base.cpp
```diff
@@ -28,7 +28,7 @@ void StatefulSpeculativePipelineBase::ensure_num_assistant_tokens_is_set(Generat
         "Stateful (non-Continuous Batching) Speculative Decoding pipeline only supports num_assistant_tokens, "
         "not assistant_confidence_threshold. Set assistant_confidence_threshold to 0.f or remove its specification.");
 
-    if (config.num_assistant_tokens == 0) {
+    if (!config.num_assistant_tokens.has_value() || config.num_assistant_tokens.value() == 0) {
         config.num_assistant_tokens = DEFAULT_NUM_ASSISTANT_TOKENS;
     }
```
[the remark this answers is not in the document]
xufang-lisa: Thanks for the review. The `num_assistant_tokens = 0` main-only behavior is intended only for the continuous batching Eagle3 path. The stateful Eagle3 changes are just interface adaptations for the GenerationConfig num_assistant_tokens type change, not part of the main-only semantics change.

src/cpp/src/speculative_decoding/continuous_batching/pipeline_impl.cpp
```diff
@@ -655,7 +655,9 @@ void ContinuousBatchingPipeline::ContinuousBatchingForSpeculativeDecodingImpl::m
                 request->pause_generation(true);
             } else if (request->get_num_processed_tokens() == 0 && sampling_params.num_return_sequences > 1) {
                 request->pause_generation(true);
-            } else if (sampling_params.num_assistant_tokens <= generated_tokens_cnt && sampling_params.assistant_confidence_threshold == 0.f) {
+            } else if (sampling_params.assistant_confidence_threshold == 0.f &&
+                       sampling_params.num_assistant_tokens.has_value() &&
+                       sampling_params.num_assistant_tokens.value() <= generated_tokens_cnt) {
                 request->pause_generation(true);
```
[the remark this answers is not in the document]
xufang-lisa: In the current design, `num_assistant_tokens=0` means “main-only”, so such requests will not enter this logic in multistep() at all. I updated the condition to require `num_assistant_tokens > 0` here mainly as a defensive check to prevent this speculative pause logic from being accidentally applied to main-only requests if the flow is reused or misrouted in the future.

src/cpp/src/generation_config.cpp
```diff
@@ -90,7 +90,11 @@ GenerationConfig::GenerationConfig(const std::filesystem::path& json_path) {
 
     // assistant generation
     read_json_param(data, "assistant_confidence_threshold", assistant_confidence_threshold);
-    read_json_param(data, "num_assistant_tokens", num_assistant_tokens);
+    if (data.contains("num_assistant_tokens")) {
+        size_t num_assistant_tokens_value = 0;
+        read_json_param(data, "num_assistant_tokens", num_assistant_tokens_value);
+        num_assistant_tokens = num_assistant_tokens_value;
+    }
```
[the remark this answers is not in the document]
xufang-lisa: Updated.

src/cpp/src/generation_config.cpp
```diff
@@ -155,7 +159,11 @@ void GenerationConfig::update_generation_config(const ov::AnyMap& properties) {
 
     // assistant generation
     read_anymap_param(properties, "assistant_confidence_threshold", assistant_confidence_threshold);
-    read_anymap_param(properties, "num_assistant_tokens", num_assistant_tokens);
+    if (properties.count("num_assistant_tokens") > 0) {
+        size_t num_assistant_tokens_value = 0;
+        read_anymap_param(properties, "num_assistant_tokens", num_assistant_tokens_value);
+        num_assistant_tokens = num_assistant_tokens_value;
+    }
```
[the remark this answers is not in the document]
xufang-lisa: Updated.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1219,9 +1221,10 @@ GenerationConfig StatefulEagle3LLMPipeline::resolve_generation_config(OptionalGe
                     ") exceeds compile-time limit (",
                     m_compile_config.max_branching_factor,
                     "). Set MAX_BRANCHING_FACTOR in draft_model properties to increase.");
-    OPENVINO_ASSERT(config.num_assistant_tokens <= m_compile_config.max_assistant_tokens,
+    OPENVINO_ASSERT(config.num_assistant_tokens.has_value(), "num_assistant_tokens must be set for Eagle3.");
+    OPENVINO_ASSERT(config.num_assistant_tokens.value() <= m_compile_config.max_assistant_tokens,
                     "num_assistant_tokens (",
-                    config.num_assistant_tokens,
+                    config.num_assistant_tokens.value(),
```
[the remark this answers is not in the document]
xufang-lisa: Thanks, I removed the redundant` has_value()` check. `ensure_tree_params_is_set()` already defaults `num_assistant_tokens` in the non-tree-search path, so this keeps the defaulting behavior without making it a hard requirement.

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
```diff
@@ -198,6 +199,40 @@ void ContinuousBatchingPipeline::Eagle3DecodingImpl::align_request_pair_processe
     }
 }
 
+void ContinuousBatchingPipeline::Eagle3DecodingImpl::validate_awaiting_requests(
+    const std::vector<SequenceGroup::Ptr>& main_awaiting_requests,
+    const std::vector<SequenceGroup::Ptr>& draft_awaiting_requests) const {
+    std::unordered_set<uint64_t> main_request_ids;
+    size_t expected_draft_requests = 0;
```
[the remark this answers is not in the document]
xufang-lisa: Thanks for the suggestion. Updated: `validate_awaiting_requests` now does only the size check, and the per-ID `unordered_set` validation was removed. This keeps the same failure behavior (`OPENVINO_ASSERT` on mismatch) with lower hot-path overhead.

tests/python_tests/test_continuous_batching.py
```diff
@@ -805,6 +805,80 @@ def eagle3_model_paths() -> tuple[Path, Path]:
...
+def test_eagle3_num_assistant_tokens_zero_matches_main_only(eagle3_model_paths: tuple[Path, Path]):
+    main_model_path, draft_model_path = eagle3_model_paths
+
+    scheduler_config = dict_to_scheduler_config({"dynamic_split_fuse": False, "max_num_batched_tokens": sys.maxsize})
+    sd_pipe = create_ov_pipeline(
+        main_model_path,
+        pipeline_type=PipelineType.SPECULATIVE_DECODING,
+        draft_model_path=draft_model_path,
+        scheduler_config=scheduler_config,
+    )
```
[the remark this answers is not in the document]
xufang-lisa: Thanks for catching this. I updated the tests to match currently supported behavior. In this PR, `num_assistant_tokens=0` main-only semantics are validated on the CB Eagle3 path: `test_eagle3_cb_generate_zero_assistant_tokens_matches_main_only` and `test_eagle3_cb_add_request_mixed_batch_zero_assistant_tokens_matches_main_only`. I also kept one LLMPipeline generate-entry parity check to validate user-facing generate-entry behavior: `test_eagle3_llm_generate_zero_assistant_tokens_matches_main_only`.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
apaniukov, line 1232: Stateful and CB pipelines have diverge for ` config.num_assistant_tokens == 0`. Might be an issue for users/validation.

sunxiaoxia2022: Update the logic in StatefulEagle3LLMPipeline, please take a look. 
If user set num_assistant_tokens=0, the draft model is disabled.

src/cpp/src/speculative_decoding/stateful/fast_draft_strategy.cpp
```diff
@@ -383,7 +383,7 @@ StatefulSpeculativeLLMPipeline::StatefulSpeculativeLLMPipeline(const ov::genai::
 
     // Specifying number candidates to generate
     ensure_num_assistant_tokens_is_set(m_generation_config);
-    m_candidates_num = m_generation_config.num_assistant_tokens;
+    m_candidates_num = m_generation_config.num_assistant_tokens.value();
     // We set the upper limit for candidates number as two times the number requested
```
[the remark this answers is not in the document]
sunxiaoxia2022: updated, num_assistant_tokens=0 is only supported for eagle3, not supported for fast draft and others.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1190,9 +1203,13 @@ StatefulEagle3LLMPipeline::~StatefulEagle3LLMPipeline() {
 }
 
 void StatefulEagle3LLMPipeline::ensure_tree_params_is_set(GenerationConfig& config) {
+    if (config.num_assistant_tokens.has_value() && config.num_assistant_tokens.value() == 0) {
+        config.tree_depth = 0;
+        config.branching_factor = 0;
+        return;
+    }
```
GuoliangShiIntel: As discussed offline, `ensure_tree_params_is_set()` is used during the pipeline construction phase to provide parameters required for static shape model compilation on the NPU. In actual user scenarios, these parameters are not expected to be passed explicitly. However, if a user sets `num_assistant_tokens = 0` during the pipeline construction stage, it can cause the NPU model compilation to crash (All tree_depth, branching_factor and num_assistant_tokens can not be zero at compile stage)

```
    config = openvino_genai.GenerationConfig()
    config.num_assistant_tokens = 0
    pipe = openvino_genai.LLMPipeline(args.model_dir, main_device, draft_model=draft_model, config)
```

Therefore, I would recommend not modifying the logic in `ensure_tree_params_is_set()` and instead reverting it to the original implementation.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1357,6 +1381,26 @@ EncodedResults StatefulEagle3LLMPipeline::generate_tokens(const EncodedInputs& i
...
+            streaming_status = stream_generated_tokens(streamer_ptr, {token});
+            ++generated_tokens;
+            eos_reached = is_stop_token_id_hit(token, config.stop_token_ids);
+        }
+
+        m_streaming_was_cancelled = (streaming_status == ov::genai::StreamingStatus::CANCEL);
+        if (streamer_ptr)
+            streamer_ptr->end();
+        return build_results(generate_timer, generated_tokens, 0, 0);
+    }
```
GuoliangShiIntel: is it better change to:

```
if (!m_draft_enabled) {
   while (...
} else {
   while (...
}
```

and these code can be reused:
```
        m_streaming_was_cancelled = (streaming_status == ov::genai::StreamingStatus::CANCEL);
        if (streamer_ptr)
            streamer_ptr->end();
        return build_results(generate_timer, generated_tokens, 0, 0);
```

GuoliangShiIntel: please also update the code comment:
`// --- Phase 2: Speculative Decoding Loop ---`

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1109,13 +1117,18 @@ Eagle3CompileConfig StatefulEagle3LLMPipeline::build_compile_config(const ModelD
...
 }
 
 void StatefulEagle3LLMPipeline::validate_construction_params(const ModelDesc& target_model_desc,
                                                              const ModelDesc& draft_model_desc) {
-    OPENVINO_ASSERT(m_generation_config.is_tree_search(), "Eagle3 pipeline requires tree_depth > 0.");
+    const bool draft_disabled =
+        m_generation_config.num_assistant_tokens.has_value() && m_generation_config.num_assistant_tokens.value() == 0;
+
+    OPENVINO_ASSERT(draft_disabled || m_generation_config.is_tree_search(),
+                    "Eagle3 pipeline requires tree search or num_assistant_tokens == 0.");
```
GuoliangShiIntel: same reason for `ensure_tree_params_is_set`, suggest reverting this change.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1109,13 +1117,18 @@ Eagle3CompileConfig StatefulEagle3LLMPipeline::build_compile_config(const ModelD
     Eagle3CompileConfig cfg;
     cfg.max_tree_depth = get_or("MAX_TREE_DEPTH", m_generation_config.tree_depth);
     cfg.max_branching_factor = get_or("MAX_BRANCHING_FACTOR", m_generation_config.branching_factor);
-    cfg.max_assistant_tokens = get_or("MAX_ASSISTANT_TOKENS", m_generation_config.num_assistant_tokens);
+    cfg.max_assistant_tokens =
+        get_or("MAX_ASSISTANT_TOKENS", m_generation_config.num_assistant_tokens.value_or(DEFAULT_EAGLE_NUM_ASSISTANT_TOKENS));
```
GuoliangShiIntel: after reverting `ensure_tree_params_is_set`, suggest reverting this change.

src/cpp/include/openvino/genai/generation_config.hpp
```diff
@@ -688,7 +688,7 @@ class OPENVINO_GENAI_EXPORTS GenerationConfig {
 
     // Assisting generation parameters
     float assistant_confidence_threshold = 0.f;
-    size_t num_assistant_tokens = 0;
+    std::optional<size_t> num_assistant_tokens = std::nullopt;
```
GuoliangShiIntel, line 691: why we need this change?

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -156,19 +156,27 @@ InputTensors Eagle3TargetWrapper::build_prefill_inputs() const {
     const size_t prompt_len = prompt_ids.size();
 
     OPENVINO_ASSERT(prompt_len > 0, "Empty prompt");
-    OPENVINO_ASSERT(generated_ids.empty(), "TARGET_PREFILL expects no generated tokens, got ", generated_ids.size());
+    const bool is_prefill = generated_ids.empty();
+    const size_t input_len = is_prefill ? prompt_len : 1;
+    const size_t start_pos = is_prefill ? 0 : prompt_len + generated_ids.size() - 1;
```
GuoliangShiIntel: please create a new function named `Eagle3TargetWrapper::build_generate_inputs()`

GuoliangShiIntel (commented):
Leave comments for the stateful Eagle3 pipeline only.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1340,7 +1361,10 @@ EncodedResults StatefulEagle3LLMPipeline::generate_tokens(const EncodedInputs& i
     m_target->reset_state();
     m_draft->reset_state();
 
-    m_draft->initialize_sequence(input_ids, config);
+    m_draft_enabled = config.num_assistant_tokens.value() > 0;
```
GuoliangShiIntel: is it better to remove this line into `resolve_generation_config()`?

src/cpp/include/openvino/genai/generation_config.hpp
sunxiaoxia2022: This change distinguishes “not specified” from an explicit `num_assistant_tokens = 0`.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
sunxiaoxia2022: moved to `resolve_generation_config()`.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
sunxiaoxia2022: added `build_generate_inputs()`.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
sunxiaoxia2022: Ok, reverted.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
sunxiaoxia2022: Ok, updated.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
sunxiaoxia2022: Ok, reverted.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
GuoliangShiIntel: Could you please explain the change here?

GuoliangShiIntel (approved):
LGTM for the stateful eagle3 pipeline

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
sunxiaoxia2022: Yes, m_generation_config.num_assistant_tokens has already been assigned a value here, reverted it now.

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
```diff
@@ -1204,8 +1235,10 @@ void StatefulEagle3LLMPipeline::ensure_tree_params_is_set(GenerationConfig& conf
 GenerationConfig StatefulEagle3LLMPipeline::resolve_generation_config(OptionalGenerationConfig generation_config) {
     GenerationConfig config = StatefulSpeculativePipelineBase::resolve_generation_config(std::move(generation_config));
     OPENVINO_ASSERT(!config.do_sample, "Eagle3 speculative decoding requires greedy sampling (do_sample=false)");
-    ensure_tree_params_is_set(config);
-
+    if (!(config.num_assistant_tokens.has_value() && config.num_assistant_tokens.value() == 0)) {
+        ensure_tree_params_is_set(config);
```
apaniukov, line 1241: Any `num_assistant_tokens` param supplied by the user swapped to `DEFAULT_EAGLE_NUM_ASSISTANT_TOKENS` here if  `tree_depth == 0`. Is it expected behavior?

sbalandi (approved):

src/cpp/src/speculative_decoding/stateful/eagle3_strategy.cpp
apaniukov: Also, if user sets `assistant_confidence_threshold`, we can end up with both it and `num_assistant_tokens`, but they are mutually exclusive. It is validated earlier here: `StatefulSpeculativePipelineBase::resolve_generation_config(std::move(generation_config))`. Maybe we could swap this lines?

sunxiaoxia2022: > Any `num_assistant_tokens` param supplied by the user swapped to `DEFAULT_EAGLE_NUM_ASSISTANT_TOKENS` here if `tree_depth == 0`. Is it expected behavior?

Yes, it is expected behavior. stateful eagle3 pipeline only support tree search now. What do you think? @GuoliangShiIntel

sunxiaoxia2022: > Also, if user sets `assistant_confidence_threshold`, we can end up with both it and `num_assistant_tokens`, but they are mutually exclusive. It is validated earlier here: `StatefulSpeculativePipelineBase::resolve_generation_config(std::move(generation_config))`. Maybe we could swap this lines?

Now stateful eagle3 pipeline does not support user sets `assistant_confidence_threshold`. So assert for `assistant_confidence_threshold`  is added here.

GuoliangShiIntel: > Any num_assistant_tokens param supplied by the user swapped to DEFAULT_EAGLE_NUM_ASSISTANT_TOKENS here if tree_depth == 0. Is it expected behavior?

Yes, the stateful pipeline does not support parameters such as tree_depth == 0 and silently falling back to default values. Users should be responsible for providing valid parameters. Although setting tree_depth to 0 is unusual, an explicit assertion with a clear error message would be preferable.

apaniukov (approved):

PR #4280, opened 2026-08-06
Update optimum-intel pin to unblock diffusers 0.39.0, drop Flux2Klein and Qwen-image workaround

goyaladitya05:
## Description
optimum-intel main dropped its safetensors<0.8.0 cap (huggingface/optimum-intel#1918), which conflicted with GenAI diffusers==0.39.0 pin. Bumps the pin in `tests/python_tests/requirements.txt` and `samples/export-requirements.txt`, and drops the flux.2-klein and Qwen-Image workaround.

Test/CI changes needed on the new pin:
- transformers 4.55.4 -> 4.57.6 in the under 5.0 jobs, since optimum-intel raised `MIN_TRANSFORMERS_VERSION` to 4.57 (huggingface/optimum-intel#1684). This also makes qwen3-next and PA run there (gates flip at 4.57); slowest job is 52 min against a 90 min timeout.
- gemma3n: export floor moved to transformers 5.0, skip gate retargeted.
- Group beam search moved to a custom_generate Hub repo in 4.57. The samples reference test now passes trust_remote_code from 4.51, and these tests are skipped on Windows where the Hub fetch 404s. Still covered on Linux and manylinux.
- Stop-string xfail widened to `(AssertionError, ValueError)`.
- nanollava WWB jobs pinned to the old optimum-intel: they run transformers 4.48, which predates `transformers.models.qwen2_5_vl` that the new pin imports at module level.

#### Known issue
qwen3-vl regresses with SDPA (PA is fine): the vision tower emits fewer embeddings than the prompt has image-pad tokens, tripping the assert in `qwen2vl/classes.cpp`. xfailed in the `ov_pipe_model` fixture so every other model still tests against the new pin. Follow-up.

## Checklist:
- NA Tests have been updated or added to cover the new code. 
- NA I have made corresponding changes to the documentation.


rkazants (approved):

avasenin-14 (approved):
Thank you, I think it is very timely PR! Looks good also.

as-suvorov (approved):

as-suvorov:
@avasenin-14 We have lot's of custom optimum-intel installs in our CI workflows. The follow up for this PR would be to remove those installs from CI.

goyaladitya05:
@as-suvorov @avasenin-14 I had to do 2 more changes, due to the CI faillures.

1. Bumped 4.55.4 -> 4.57.6 since optimum-intel raised MIN_TRANSFORMERS_VERSION to 4.57 for several models (huggingface/optimum-intel#1684). 4.57.6 is the latest 4.x release (still <5.0) and matches what mac.yml already uses for these jobs. 
2. Text2VideoPipeline hardcoded a rank-1 timestep tensor, which broke once the export started producing rank-2 unconditionally (huggingface/optimum-intel#1762). Added rank detection so it builds the tensor at whatever shape the export declares. I've tried to keep the changes minimal. This would mainly land with #3907. 

Could we please re-run CI?

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -609,15 +610,18 @@ class Text2VideoPipeline::LTXPipeline {
...
-        float* timestep_data = timestep.data<float>();
-
         ov::Shape latent_shape_cfg = latent.get_shape();
         latent_shape_cfg[0] *= batch_size_multiplier;
         ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
 
+        // Rank-1 [B] (legacy export) or rank-2 [B, S] (current export, per-token conditioning).
+        // T2V has no per-token conditioning, so every token gets the same timestep value either way.
+        const bool timestep_is_rank2 = m_transformer->get_timestep_partial_shape().size() == 2;
+        ov::Tensor timestep(ov::element::f32,
```
sgonorov: `reshape()` pins `timestep` batch to 1 while `hidden_states` gets 2 under CFG, so `set_tensor` throws and `test_reshape` goes red. I'd derive the batch from the model's declared shape.

.github/workflows/linux.yml
```diff
@@ -755,19 +755,19 @@ jobs:
...
             cmd: |
-              python -m pip install transformers==4.55.4 diffusers==0.37.1
+              python -m pip install transformers==4.57.6 diffusers==0.37.1
               python -m pytest -v ./tests/python_tests/test_vlm_pipeline.py -m "transformers_lower_v5 or transformers_dependent" --override-ini cache_dir=/mount/caches/pytest/
             run_condition: ${{ fromJSON(needs.smart_ci.outputs.affected_components).visual_language.test }}
             timeout: 90
           - name: 'API tests transformers under 5.0'
             cmd: |
-              python -m pip install transformers==4.55.4 diffusers==0.37.1
+              python -m pip install transformers==4.57.6 diffusers==0.37.1
```
sgonorov: 4.57.6 changes the HF reference for beam search and stop strings, so the the under 5.0 jobs go red. Might be worth updating expectations or xfailing with a ticket.

tests/python_tests/requirements.txt
```diff
@@ -1,6 +1,6 @@
 --extra-index-url https://download.pytorch.org/whl/cpu
 diffusers==0.39.0
-optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@140e49a68718ab0025a3b5f02864b9ca9f009558
+optimum-intel @ git+https://github.com/huggingface/optimum-intel.git@b35fbd7284d22f170168ff97112715530eadfd0f
```
sgonorov: This pin regresses qwen3-vl export on all three OSes and trips the image embed assert in `qwen2vl/classes.cpp`. We could fix it GenAI side or xfail with a ticket.

.github/workflows/linux.yml
```diff
@@ -755,19 +755,19 @@ jobs:
             timeout: 60
           - name: 'LLM with transformers under 5.0'
             cmd: |
-              python -m pip install transformers==4.55.4 diffusers==0.37.1
+              python -m pip install transformers==4.57.6 diffusers==0.37.1
```
sgonorov: 4.57.6 flips version gates in `models.py` and `ov_genai_pipelines.py`, so qwen3-next and paged attention newly run here. It's good to mention that in the description and recheck teh timeouts.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -646,7 +650,7 @@ class Text2VideoPipeline::LTXPipeline {
                 latent_cfg = numpy_utils::repeat(latent_cfg, request_input_batch / latent_cfg.get_shape()[0]);
             }
 
-            timestep_data[0] = timesteps[inference_step];
+            std::fill_n(timestep_data, timestep.get_size(), timesteps[inference_step]);
```
sgonorov: `latent_cfg` gets repeated to the request batch just above, but `timestep` keeps its pre loop batch, so the two disagree. I suggest to build both from one source.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -609,15 +610,18 @@ class Text2VideoPipeline::LTXPipeline {
...
-
         ov::Shape latent_shape_cfg = latent.get_shape();
         latent_shape_cfg[0] *= batch_size_multiplier;
         ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
 
+        // Rank-1 [B] (legacy export) or rank-2 [B, S] (current export, per-token conditioning).
+        // T2V has no per-token conditioning, so every token gets the same timestep value either way.
+        const bool timestep_is_rank2 = m_transformer->get_timestep_partial_shape().size() == 2;
+        ov::Tensor timestep(ov::element::f32,
+                           timestep_is_rank2 ? ov::Shape{latent_shape_cfg[0], video_sequence_length}
```
sgonorov: `video_sequence_length` is the unpacked latent count, which drifts from the packed token count once patch sizes stop being 1. Probably better to use `latent_cfg.get_shape()[1]`.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -130,6 +130,27 @@ size_t LTXVideoTransformer3DModel::get_request_input_batch() {
     return shape[0];
 }
 
+ov::PartialShape LTXVideoTransformer3DModel::get_timestep_partial_shape() {
+    if (m_model) {
```
sgonorov: `compile()` resets `m_model` and generate needs a compiled model, so this branch never fires. I think we can drop it or note which pre compile caller needs it.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -130,6 +130,27 @@ size_t LTXVideoTransformer3DModel::get_request_input_batch() {
...
+    }
+    if (m_request) {
+        ov::CompiledModel compiled = m_request.get_compiled_model();
+        for (const auto& input : compiled.inputs()) {
+            if (input.get_any_name() == "timestep") {
+                return input.get_partial_shape();
+            }
+        }
+    }
+    OPENVINO_ASSERT(false,
```
sgonorov: `OPENVINO_ASSERT(false, ...)` leaves a non void function with no return, and the rest of the file uses `OPENVINO_THROW`. Maybe switch it over for consistency.

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
```diff
@@ -55,6 +55,7 @@ class OPENVINO_GENAI_EXPORTS LTXVideoTransformer3DModel {
 
     size_t get_expected_batch_size() const;
     size_t get_request_input_batch();
+    ov::PartialShape get_timestep_partial_shape();
```
sgonorov: New public API with no docstring, and it's a pure query so it should be `const` like `get_expected_batch_size()`. Might be worth putting it on `Inference` instead.

sgonorov (changes requested):

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: Fixed. reshape() now derives both from a single `video_sequence_length` local, and sets the rank-2 timestep to {batch_size, video_sequence_length} so it matches hidden_states.

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: okay, fixed

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: agreed, I have changed it per your suggestion

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
goyaladitya05: removed it

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
goyaladitya05: I added a docstring and made it const. I don't think it can be moved to Inference, as that class is currently a forward only declaration.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
goyaladitya05: removed it, it's not needed now. The shape is cached during compile(), so the getter is a return with no fallthrough. The "timestep input not found" check moved into compile(), so a bad export now fails at compile time instead of at first generate.

.github/workflows/linux.yml
goyaladitya05: I extended the xfail marker, to now (AssertionError, ValueError).

tests/python_tests/requirements.txt
goyaladitya05: I've added xfail markers right now, but will attempt to fix it in the coming weeks.

goyaladitya05:
@as-suvorov @avasenin-14 Could you please approve the CI workflows? I've introduced fixes for the previous failures.

goyaladitya05:
@as-suvorov @avasenin-14 could you please help with a CI issue I can't get past? The transformers/optimum-intel failures are mostly fixed now, the 9 remaining failures are one and the same thing, and it's hit two runs in a row.

Every failing job shows this from `utils atomic_download.py`:

    Rename failed, falling back to shutil.move
      OSError: [Errno 5] Input/output error       (temp_path.rename)
      FileExistsError: [Errno 17] File exists     (shutil.move -> copytree -> makedirs)
    Error during move - assuming it was created successfully by another process

and then, later, `RuntimeError: utils.cpp:488 Could not find a model in the directory`. 12 models are affected, all under
`.ov_cache/{lin,win}/b53e4e21/<date>/optimum-intel-2.2.0.dev0+b35fbd7_transformers-5.0.0/`.

I think it's specific to this PR: the cache path embeds the package versions, so bumping the optimum-intel pin creates a partition nothing else pre-warms, and all the jobs convert into it at once. How do we resolve it?

goyaladitya05:
@Wovchena could you please trigger another re run for the tests? the failures have reduced in the last re run, they might pass on another run.

.github/workflows/linux.yml
goyaladitya05: Good catch, thanks. both are newly run in these jobs. Added a note to the description. I checked the timeouts as well, but I don't think they need any changes

goyaladitya05:
Thank you, all tests passed. Please ignore the changes releted to video generation in this PR, they are solely to pass CI, and I will remove them once #4298 is merged.

avasenin-14 (approved):

as-suvorov:
build_jenkins

likholat:
build_jenkins

likholat:
@sgonorov please take a look

as-suvorov (approved):

as-suvorov:
@avasenin-14 connected to your work on optimum-intel/transformers updates, so please take a look as well

sgonorov (approved):

goyaladitya05:
by the way, last time it took 4 re runs for the failing tests in this PR to pass all the tests, mainly due to cache invalidation issues. It might require similar re runs this time as well👀

avasenin-14 (approved):

avasenin-14:
I think we can add to merge queue, if no one has any objections?

goyaladitya05:
could someone please re run the failing tests?

goyaladitya05:
test failures are down to 1, @avasenin-14 could you please re run that test again?

avasenin-14:
Ok, I'm adding it to the merge queue, I think no one has any objections.

goyaladitya05:
@avasenin-14 could you please trigger the jenkins build as well?

avasenin-14:
build_jenkins

avasenin-14:
@goyaladitya05 pls resolve the merge conflicts, I'll resubmit

goyaladitya05:
> @goyaladitya05 pls resolve the merge conflicts, I'll resubmit

@avasenin-14 done

avasenin-14:
build_jenkins

goyaladitya05:
@avasenin-14 Thank you for your help on this!

PR #4283, opened 2026-08-07
Fix DeepStack visual_pos_masks layout mismatch in PagedAttention

liubo-intel:
## Description
Fixes PagedAttention backend crash (ScatterNDUpdate out-of-bounds assertion) when running Qwen3-VL with 
`ATTENTION_BACKEND="PA"` by correcting the axis order mismatch introduced when `SDPAToPagedAttention` 
reshapes `hidden_states` but leaves `visual_pos_masks`-derived index tuples un-updated.

The fix inserts a single `Transpose(perm=[1,0])` between the `visual_pos_masks` Parameter and its 
`NonZero` consumers, converting mask shape from `[1, total_tokens]` to `[total_tokens, 1]` so that 
index tuple order matches the post-PA layout. Activation is conservative: only applies when the 
Parameter is found, its rank is statically 2, and all direct consumers are `NonZero` ops; 
otherwise silently no-ops with optional warnings.

CVS-192302

### DIFF
--- a/src/cpp/src/continuous_batching/pipeline_impl.cpp
+++ b/src/cpp/src/continuous_batching/pipeline_impl.cpp
@@ -89,6 +89,7 @@ void prepare_model_for_paged_attention(const std::shared_ptr<ov::Model>& model,
         .run_on_model(model);
     model->validate_nodes_and_infer_types();
     utils::apply_gather_before_matmul_transformation(model);
+    utils::fix_deepstack_visual_pos_masks_layout_for_paged_attention(model);
 }
 
 ContinuousBatchingPipeline::ContinuousBatchingImpl::ContinuousBatchingImpl(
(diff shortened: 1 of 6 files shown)


rkazants (changes requested):
please provide tests

liubo-intel:
> please provide tests

Done

avasenin-14 (approved):
I've checked that it fixes the issue.

avasenin-14:
Should we close it? We went different way about the fix: https://github.com/openvinotoolkit/openvino.genai/pull/4381. Or should we keep it just in case? :)

liubo-intel:
> Should we close it? We went different way about the fix: #4381. Or should we keep it just in case? :)

since we have fixed the issue in #4381, I will close this pr

PR #4288, opened 2026-08-09
cmake: follow OpenVINO install layouts for native packages

stevegrubb:
OpenVINO GenAI currently installs every configuration using the relocatable archive layout. Libraries go under runtime/lib/<arch>, headers under runtime/include, CMake metadata under runtime/cmake, and Python modules under python.

That layout is correct for the official OpenVINO archive, but it produces incorrect paths for native distribution packages. For example, installing with CMAKE_INSTALL_PREFIX=/usr would place files below /usr/runtime and /usr/python rather than the platform's normal library, include, CMake, and Python directories. The exported OpenVINOGenAI targets also retain the archive-only runtime/include path.

Select GenAI's install destinations using the same CPACK_GENERATOR policy as OpenVINO:

* no generator, or an archive generator such as TGZ or ZIP, keeps the existing runtime/, python/, and docs/licensing layout;
* NPM keeps its existing NPM-specific layout;
* RPM, DEB, and other native package generators consume the OV_CPACK_* paths supplied by OpenVINODeveloperPackage;
* when only the public OpenVINO package is available, use the corresponding GNUInstallDirs values as a portable fallback.

Apply the selected paths consistently to libraries, headers, Python modules, licenses, generated CMake package files, and installed target interfaces. The destinations remain relative, preserving CMAKE_INSTALL_PREFIX, DESTDIR, and cmake --install --prefix behavior.

OpenVINO's RPM mapping derives its locations from GNUInstallDirs, while its DEB mapping supplies the Debian library and lib/python3/dist-packages locations. GenAI now follows whichever mapping OpenVINO selected rather than defining a second, GenAI-specific installation interface.

An RPM-oriented configuration can use:

    cmake -S . -B build-rpm \
        -DCMAKE_INSTALL_PREFIX=/usr \
        -DCMAKE_INSTALL_LIBDIR=lib64 \
        -DCPACK_GENERATOR=RPM

On Debian or Ubuntu, the corresponding configuration is:

    multiarch=$(dpkg-architecture -qDEB_HOST_MULTIARCH)
    cmake -S . -B build-deb \
        -DCMAKE_INSTALL_PREFIX=/usr \
        -DCMAKE_INSTALL_LIBDIR="lib/${multiarch}" \
        -DCPACK_GENERATOR=DEB

A normal source, Dockerfile, or SDK build does not need to change:

      cmake -S . -B build-sdk

With no native package generator selected, that configuration retains the existing archive-compatible runtime/, python/, and docs/licensing paths.

The configured inputs can be checked before compiling with:

    cmake -N -LA build-rpm |
        grep -E '^(CPACK_GENERATOR|CMAKE_INSTALL_(PREFIX|LIBDIR|BINDIR|INCLUDEDIR|DATADIR)):'


artanokhov:
There are currently no official APT/YUM distributions for Tokenizers or GenAI.

It will be nice for long-term if mechanism will be implemented as RPM CPack Generator instructions as we have for OV RT
https://github.com/openvinotoolkit/openvino/blob/master/cmake/packaging/rpm.cmake
https://github.com/openvinotoolkit/openvino/tree/master/cmake/developer_package/packaging/rpm

However, to unlock community with necessary packages - current approach LGTM as a middle-term, if GenAI team will review it and approve.

stevegrubb:
I have updated the PR based on the feedback from copilot. The new commit has updated instructions for how to work with this on Debian when openvino is not available. This is likely a paranoid analyzer issue because you cannot build this library without openvino being on the system. In any event, the fallback path is now available.

PR #4298, opened 2026-08-11
[Video Generation] Support rank-1 and rank-2 timestep inputs in LTX-Video pipeline

goyaladitya05:
## Description
After https://github.com/huggingface/optimum-intel/pull/1762, optimum-intel always produces a 2D timestep model, regardless of the task, whereas the GenAI pipeline only handled 1D timestep. 

This breaks text-to-video against any LTX-Video model exported with a current optimum-intel: the pipeline allocated a [1] timestep and reshaped the input to [1], which no longer matches the compiled model.

`LTXVideoTransformer3DModel` now gets an `infer(latent, float)` overload that builds the timestep input to match whatever the compiled model expects:

- rank-1 [B] legacy exports get a single value, as before
- rank-2 [B, S] exports get one entry per packed token, sized from the latents passed in, so batch and token count can't drift from `hidden_states`
- `reshape()` sizes a rank-2 timestep to `{batch_size, video_sequence_length}` and keeps [1] for rank-1

Both sites assert the rank is 1 or 2. Text-to-video conditions every token identically, so the rank-2 tensor is filled with a single value. 1D exports are unaffected.

The Python binding now names the tensor overload explicitly (`py::overload_cast`), since `infer` is overloaded; the exposed Python API is unchanged.

## Checklist:
- [ ]  Tests have been updated or added to cover the new code.  No, because adding tests for this, would also require doing a version bump for optimum, which causes a lot of failures. Tests will be added in #3907.
- NA  I have made corresponding changes to the documentation.


goyaladitya05:
The CI failures are not releted to these changes, so I'm addressing them in #4302.

likholat (approved):

likholat:
@sgonorov please take a look

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -646,7 +647,13 @@ class Text2VideoPipeline::LTXPipeline {
                 latent_cfg = numpy_utils::repeat(latent_cfg, request_input_batch / latent_cfg.get_shape()[0]);
             }
 
-            timestep_data[0] = timesteps[inference_step];
+            // Derive the timestep shape from the final latent_cfg (after the repeat above) so
+            // batch and token count always match. T2V conditions every token identically.
+            if (timestep_is_rank2) {
```
sgonorov: Fixture is a rank-1 export, so green CI only covers the legacy path. Might be worth gating on #3907.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -130,6 +143,10 @@ size_t LTXVideoTransformer3DModel::get_request_input_batch() {
     return shape[0];
 }
 
+const ov::PartialShape& LTXVideoTransformer3DModel::get_timestep_partial_shape() const {
```
sgonorov: Before `compile()` this returns rank-0 and we silently take the rank-1 path. I'd add `OPENVINO_ASSERT(m_request, ...)` like the siblings.

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
```diff
@@ -56,6 +56,10 @@ class OPENVINO_GENAI_EXPORTS LTXVideoTransformer3DModel {
     size_t get_expected_batch_size() const;
     size_t get_request_input_batch();
 
+    /// @brief Shape of the model's 'timestep' input, cached at compile time.
+    /// Rank-1 [B] for legacy exports, rank-2 [B, S] for per-token conditioning.
+    const ov::PartialShape& get_timestep_partial_shape() const;
```
sgonorov: Getter exists for one internal caller and breaks ABI. We could overload `infer()` with a float instead.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -609,15 +610,15 @@ class Text2VideoPipeline::LTXPipeline {
...
-        ov::Tensor timestep(ov::element::f32, {1});
-        float* timestep_data = timestep.data<float>();
-
         ov::Shape latent_shape_cfg = latent.get_shape();
         latent_shape_cfg[0] *= batch_size_multiplier;
         ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
 
+        // Rank-1 [B] (legacy export) or rank-2 [B, S] (current export, per-token conditioning).
+        // Sized from latent_cfg inside the denoising loop so the two always agree.
+        const bool timestep_is_rank2 = m_transformer->get_timestep_partial_shape().size() == 2;
```
sgonorov: `size()` on `PartialShape` reads like a container length. I think we can use `.rank()` to show intent.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -81,6 +81,19 @@ LTXVideoTransformer3DModel& LTXVideoTransformer3DModel::compile(const std::strin
     m_request = compiled_model.create_infer_request();
     const auto& input_shape = compiled_model.input(0).get_partial_shape();
     m_expected_batch_size = input_shape.is_static() ? input_shape[0].get_length() : 0;
+
+    bool timestep_found = false;
```
sgonorov: Eleven lines and sentinel bool for one named lookup. Probably better to call `compiled_model.input("timestep")` directly. Assert is nice.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -646,7 +647,13 @@ class Text2VideoPipeline::LTXPipeline {
                 latent_cfg = numpy_utils::repeat(latent_cfg, request_input_batch / latent_cfg.get_shape()[0]);
             }
 
-            timestep_data[0] = timesteps[inference_step];
+            // Derive the timestep shape from the final latent_cfg (after the repeat above) so
+            // batch and token count always match. T2V conditions every token identically.
+            if (timestep_is_rank2) {
+                const ov::Shape& latent_cfg_shape = latent_cfg.get_shape();
+                timestep.set_shape({latent_cfg_shape[0], latent_cfg_shape[1]});
```
sgonorov: I think latents stay rank-3. It's good to assert that here, `unpack_latents` already does.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
```diff
@@ -151,15 +168,24 @@ LTXVideoTransformer3DModel& LTXVideoTransformer3DModel::reshape(int64_t batch_si
...
+    const int64_t video_sequence_length = num_frames * height * width;
+
     for (auto&& input : m_model->inputs()) {
         std::string input_name = input.get_any_name();
         name_to_shape[input_name] = input.get_partial_shape();
         if (input_name == "timestep") {
-            name_to_shape[input_name][0] = 1;
+            // Rank-1 legacy export takes a single scalar regardless of batch. Rank-2 export
+            // conditions per token, so it must match hidden_states' batch and token count.
+            if (name_to_shape[input_name].size() >= 2) {
```
sgonorov: `>= 2` here, `== 2` in the pipeline. Maybe align on `== 2` and assert on other ranks.

src/cpp/src/video_generation/ltx_pipeline.hpp
```diff
@@ -609,15 +610,15 @@ class Text2VideoPipeline::LTXPipeline {
...
-        // // Prepare timesteps
-        // TODO: ov::Tensor timestep(ov::element::f32, {1}); is enough
-        ov::Tensor timestep(ov::element::f32, {1});
-        float* timestep_data = timestep.data<float>();
-
         ov::Shape latent_shape_cfg = latent.get_shape();
         latent_shape_cfg[0] *= batch_size_multiplier;
         ov::Tensor latent_cfg(ov::element::f32, latent_shape_cfg);
 
+        // Rank-1 [B] (legacy export) or rank-2 [B, S] (current export, per-token conditioning).
```
sgonorov: Same rank rationale sits in teh header, `reshape()`, and here. I suggest to keep one copy and link it.

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
```diff
@@ -65,6 +69,7 @@ class OPENVINO_GENAI_EXPORTS LTXVideoTransformer3DModel {
     std::string m_lora_prefix;
     ov::InferRequest m_request;
     std::shared_ptr<ov::Model> m_model;
+    ov::PartialShape m_timestep_partial_shape;
```
sgonorov: New member shifts object layout on an exported class, so it may be an ABI break.

sgonorov (approved):

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: yes, will do. I avoided in this PR because it required an optimum-intel version bump, causing a lot of failures.

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
goyaladitya05: removed entirely, the new infer() overload asserts m_request like the siblings.

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
goyaladitya05: done

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: switched to .rank().get_length()

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
goyaladitya05: removed it entirely, lookup is now input("timestep"), which throws on its own

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: added

src/cpp/src/video_generation/models/ltx_video_transformer_3d_model.cpp
goyaladitya05: aligned on ==2

src/cpp/src/video_generation/ltx_pipeline.hpp
goyaladitya05: one copy now, in infer()

src/cpp/include/openvino/genai/video_generation/ltx_video_transformer_3d_model.hpp
goyaladitya05: Resolved it instead of flagging, infer(float) makes the member unnecessary, so there's no layout change.

goyaladitya05:
@sgonorov @likholat I've addressed all the review comments, and re verified with both 1D and 2D timestep. could you please approve the tests again?

likholat (approved):

likholat:
build_jenkins

goyaladitya05:
@likholat could you please re run the failing test?

likholat:
build_jenkins

likholat:
build_jenkins

likholat:
build_jenkins

likholat:
build_jenkins

PR #4300, opened 2026-08-11
Reorder GenerationHandle notification

mzegla:
## Description
The main purpose of this PR is to remove race condition happening in OVMS execution scenario where one thread runs pipeline execution and another reads from GenerationHandle. Currently holder of the handle has two synchronization points - first when tokens arrive and second when metrics are updated. The dual synchronization seems awkward and indicates architectural issue in my opinion. 
In this PR I change order in ContinuousBatching::step so that when GenerationHandle is notified, metrics are already set. User can we sure that calling get_perf_metrics will not fail due to timing issue (calling it between tokens being ready + generation status being switched and perf metrics being set).

Also I extracted notify handle call from sampler as it seems clearer logically to limit sampler logic to sampling and wrapped entire notification logic in a single method called directly in `step`.

## Checklist:

### DIFF
--- a/src/cpp/src/continuous_batching/pipeline_impl.cpp
+++ b/src/cpp/src/continuous_batching/pipeline_impl.cpp
@@ -484,22 +484,6 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::step() {
         free_fork_timer.end();
     }
 
-    {
-        static ManualTimer candidates_timer("generate_candidates_for_prompt_lookup()");
-        candidates_timer.start();
-        generate_candidates_for_prompt_lookup();
-        candidates_timer.end();
-    }
-
-    // Append embeddings for tokens produced in this step.
-    // Validation mode usually skips this because speculative validation reuses/rewinds
-    // candidate tokens instead of committing them here. prompt_lookup is the exception:
-    // it appends validation candidates after sampling and must keep embeddings in sync
-    // before the next scheduling/hash step.
-    if (m_model_input_type == ModelInputType::EMBEDDINGS && sync_embeddings_after_candidates()) {
-        m_model_runner->append_embeddings(m_requests, scheduler_output);
-    }
-
     const auto step_end_time = std::chrono::steady_clock::now();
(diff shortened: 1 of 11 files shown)


rkazants (commented):
I think it is possible to implement such test to cover this case Please add.

src/cpp/src/continuous_batching/pipeline_impl.cpp
```diff
@@ -412,8 +412,11 @@ void ContinuousBatchingPipeline::ContinuousBatchingImpl::step() {
         for (size_t i = 0; i < m_requests.size(); ++i) {
             SequenceGroup::Ptr sequence_group = m_requests[i];
             if (!sequence_group->is_waiting()) {
+                auto perf_metrics = sequence_group->get_perf_metrics();
+                perf_metrics.load_time = m_load_time_ms;
+                sequence_group->get_generation_stream()->set_perf_metrics(std::move(perf_metrics));
                 sequence_group->set_out_of_memory();
```
mzegla, line 418:
```suggestion
```
status set in the wrapper method `notify_handle_oom` so it's redundant here

mzegla: invalid, I confused statuses of two different entities here, resolving

pavel-esir (approved):
I checked the core part with notify_handles, `m_terminal_notification_sent`, etc. New logic is clear. Looks good to me.

But please adress the copilots `High` comment and resolve conflicts before merge.

src/cpp/src/whisper/whisper.cpp
```diff
@@ -114,6 +114,10 @@ std::pair<ov::genai::EncodedResults, bool> decode(std::shared_ptr<ov::genai::Whi
         }
 
         std::unordered_map<uint64_t, ov::genai::GenerationOutput> token = handle->read();
+        if (token.empty()) {
+            // Empty terminator pushed by notify_handle_final()/notify_handle_oom() to unblock readers.
+            return;
+        }
```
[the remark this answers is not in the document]
mzegla: False positive. Return is located in void-deduced lambda.

PR #4303, opened 2026-08-12
[GHA] Clean Stale Dependabot Branches cronjob

jkoniusz:
[GHA] Clean Stale Dependabot Branches cronjob

CVS-192415

Doing same as in https://github.com/openvinotoolkit/openvino/pull/37224 & https://github.com/openvinotoolkit/openvino/pull/36928


.github/workflows/cleanup_dependabot_branches.yml
```diff
@@ -0,0 +1,37 @@
...
+permissions: read-all
+
+jobs:
+  Cleanup_Dependabot_Branches:
+    name: Cleanup stale dependabot branches
+    # The cleanup logic and its script are maintained once in the openvino
+    # repository and reused here, so this workflow only wires up the trigger,
+    # the dry-run choice and the token.
+    if: ${{ github.repository_owner == 'openvinotoolkit' }}
+    uses: openvinotoolkit/openvino/.github/workflows/cleanup_dependabot_branches.yml@master
```
akashchi: The convention is to use the commit hash after the `@` instead of everchanging `master`.

akashchi (changes requested):

.github/workflows/cleanup_dependabot_branches.yml
jkoniusz: i have changed it - now we use commit hash (commit associated with recently merged https://github.com/openvinotoolkit/openvino/pull/37399) (have tested a dry run: https://github.com/openvinotoolkit/openvino.genai/actions/runs/32132439475/job/95696319842) @akladiev @akashchi

PR #4310, opened 2026-08-12
Optimize GPU LoRA switching with precision-aware tensor caching

ahnyoung-paul:
## Description

Runtime LoRA switching previously ran the concat evaluator and then passed its outputs to `VariableState::set_state()`. When the evaluator output type did not match the GPU variable-state type, the GPU plugin fell back to element-wise CPU conversion and copy. Processing approximately 250 MB x 2 of LoRA tensor data made a single adapter switch take close to 500 ms.

This change prepares and caches the concat evaluator outputs per `AdapterConfig` during VLM construction. On GPU, when the compiled model's `ov::hint::inference_precision` is `f16`, the evaluator produces `f16` LoRA state tensors. Every other case preserves the declared state types. Matching the GPU state type avoids the CPU conversion loop and enables the direct GPU copy path.

The cache holds the large `A` and `B` tensors. On a cache hit, switching reuses them instead of running the concat evaluator for `A` and `B`. Alpha is not part of the cache key, so changing only alpha still reuses the cached `A` and `B`, and alpha alone is re-evaluated, which is cheap. A cache miss evaluates the configuration once and stores the result in a bounded LRU cache. Reapplying the same adapter remains a no-op. With cache hits and matching `f16` tensors, measured switching latency was reduced from close to 500 ms to approximately 30 ms.

Validation:

- `pre-commit run --from-ref upstream/master --to-ref HEAD`: passed
- Windows Release build: passed
- End-to-end VLM LoRA switching test: passed
- WWB accuracy comparison against the code before this change: no regression observed. Detailed results are available in CVS-187607.

CVS-187607

## Checklist:
- [x] Tests have been added: `test_visual_language_lora_switch_prepared_tensor_cache` (LoRA A→B→A switching + alpha-zero equivalence assertion).

## AI Assistance
- AI assistance used: yes

### DIFF
--- a/src/cpp/src/lora/adapter.cpp
+++ b/src/cpp/src/lora/adapter.cpp
@@ -1856,13 +1856,10 @@ struct AdapterControllerImpl {
         ov::OutputVector concat_inputs;
         concat_inputs.reserve(inputs.size());
         const auto output_type = output.get_element_type();
-        ov::element::Type concat_type = ov::element::dynamic;
         for(size_t i = 0; i < inputs.size(); ++i) {
             NodePtr input = parameters[(alpha_only ? 1 : 3)*i + offset] = input_accessor(inputs[i]);
-            if(concat_type == ov::element::dynamic) {
-                concat_type = input->get_output_element_type(0);
-            } else if(input->get_output_element_type(0) != concat_type) {
-                input = std::make_shared<v0::Convert>(input, concat_type);
+            if(input->get_output_element_type(0) != output_type) {
+                input = std::make_shared<v0::Convert>(input, output_type);
             }
             if(input->get_output_partial_shape(0).rank().get_length() > 2) {
                 input = squeeze_2d(input);
--- a/src/cpp/src/lora/adapter.cpp
+++ b/src/cpp/src/lora/adapter.cpp
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <algorithm>
+#include <cstdlib>
 #include <set>
 #include <map>
 #include <string>
@@ -13,6 +14,7 @@
 #include <unordered_map>
 #include <unordered_set>
 #include <functional>
+#include <list>
 #include <memory>
 #include <cmath>
 
@@ -36,6 +38,7 @@
 #include "openvino/pass/pattern/op/wrap_type.hpp"
 #include "openvino/pass/graph_rewrite.hpp"
 #include "openvino/pass/manager.hpp"
+#include "openvino/runtime/properties.hpp"
 
 #include "openvino/genai/lora_adapter.hpp"
 
@@ -746,10 +749,16 @@ class InferRequestSignatureCache {
 public:
     using Signature = std::string;
 
-    InferRequestSignatureCache (const std::string& device) : device(device) {}
+    InferRequestSignatureCache(const std::string& device, const std::string& infer_device = {}) :
+        device(device),
+        infer_device(infer_device.empty() ? device : infer_device) {}
 
     bool exist (const Signature& signature) {
         return requests.count(signature);
+    }
+
+    void clear() {
+        requests.clear();
     }
 
     void insert (const Signature& signature, ov::ResultVector& results, ov::ParameterVector& parameters) {
@@ -817,6 +826,7 @@ public:
             outputs[bypass.second] = inputs[bypass.first];
         }
         request.infer();    // TODO: Consider using async to increase throughput, requires more complicated archestration
+
     }
 
 private:
@@ -827,6 +837,7 @@ private:
 
     std::unordered_map<Signature, RequestWithBypass> requests;
     std::string device;
+    std::string infer_device;
 };
 
 
@@ -1285,15 +1296,47 @@ struct AdapterControllerImpl {
     std::unordered_set<std::string> variable_names;
     AdapterConfig current_config;
     bool need_full_apply = true;
+    bool infer_device_is_gpu = false;
+    bool output_type_initialized = false;
+    std::optional<ov::element::Type> state_output_type_override;
     InferRequestSignatureCache lora_state_evaluators;
+
+    struct PreparedTensorCacheEntry {
+        AdapterConfig config;
+        std::vector<LoRAParts<ov::Tensor>> tensors;
+        size_t byte_size = 0;
+    };
+
+    // The cache owns evaluator output tensors for the controller lifetime. Keep a small
+    // LRU bound because every entry contains alpha/A/B outputs for every transformed layer.
+    static constexpr size_t prepared_tensor_cache_capacity = 8;
+    static constexpr size_t prepared_tensor_cache_max_bytes = 512 * 1024 * 1024;
+    std::list<PreparedTensorCacheEntry> prepared_tensor_cache;
+    size_t prepared_tensor_cache_byte_size = 0;
 
     // Stores the actual LoRA weight getter used for Constant tensor replacement
     // Needed to track which LoRA tensors were actually applied to suppress unused tensor warnings
     std::shared_ptr<LoRAWeightGetterDefault<NodePtr, NodePtr>> const_getter_impl;
 
-    AdapterControllerImpl(std::shared_ptr<ov::Model> model, const AdapterConfig& config) :
+    static std::string lora_evaluator_device(const std::string& infer_device) {
+        if (const char* env = std::getenv("OV_LORA_EVALUATOR_DEVICE")) {
+            if (env[0] != '\0') {
+                std::string requested_device(env);
+                if (requested_device == "INFER" || requested_device == "infer") {
+                    return infer_device;
+                }
+                return requested_device;
+            }
+        }
+        return "CPU";
+    }
+
+    AdapterControllerImpl(std::shared_ptr<ov::Model> model,
+                          const AdapterConfig& config,
+                          const std::string& device = "CPU") :
         current_config(config),  // FIXME: Compare current and passed configs and change incrementally
-        lora_state_evaluators("CPU")    // FIXME: Try to run on the same device that is used for model inference
+        infer_device_is_gpu(device.find("GPU") != std::string::npos),
+        lora_state_evaluators(lora_evaluator_device(device), device)
     {
         LoRAConstantGetter const_getter;
         LoRAParametersByWeightGetter params_getter;
@@ -1337,7 +1380,7 @@ struct AdapterControllerImpl {
                 ov::Tensor(params_getter.type, ov::Shape{0})
             };
             auto name = node->get_friendly_name();
-            auto lora_weight = prepare_lora_tensors(name, params_getter.weight_getter, lora_placeholder, /*set_empty_tensors=*/false, /*alpha_only=*/false);
+            auto lora_weight = prepare_lora_tensors(name, params_getter.weight_getter, lora_placeholder, /*set_empty_tensors=*/false, /*alpha_only=*/false, current_config);
             if(lora_weight.alpha) {
                 return LoRANode(
                     // TODO: Make sure that tensors will not be disposed during constant life time
@@ -1402,15 +1445,17 @@ struct AdapterControllerImpl {
         bool mode = false;
         bool alpha = false;
         bool adapter = false;
+        bool tensor_name_prefix = false;
 
         operator bool() const {
-            return mode || alpha || adapter;
+            return mode || alpha || adapter || tensor_name_prefix;
         }
     };
 
     ConfigChanged compare_configs(const AdapterConfig& config1, const AdapterConfig& config2) {
         ConfigChanged diff;
         diff.mode = config1.get_mode() != config2.get_mode();
+        diff.tensor_name_prefix = config1.get_tensor_name_prefix() != config2.get_tensor_name_prefix();
         // TODO: Use `set` from this commented block when the config change tracking is implemented at adapter granularity and will track order of adapters correctly
         // std::set<Adapter>
         //     adapters1(config1.adapters.begin(), config1.adapters.end()),
@@ -1432,20 +1477,23 @@ struct AdapterControllerImpl {
         // FIXME: If a part of LoRA state tensors are not set here, then need to carefully reset state in LLMPipeline where global reset is called after the generation
         ConfigChanged diff;
         if(config) {
-            diff = compare_configs(current_config, *config);
+            AdapterConfig updated_config = current_config;
+            updated_config.update(*config);
+            diff = compare_configs(current_config, updated_config);
             OPENVINO_ASSERT(
                 !diff.mode || config->get_mode() == AdapterConfig::MODE_AUTO,  // MODE_AUTO in this call means that mode is not changed
                 "AdapterConfig::mode cannot be changed and should be configured once for a model at the initialization");
             OPENVINO_ASSERT(
                 config->get_mode() == AdapterConfig::MODE_AUTO || config->get_mode() == AdapterConfig::MODE_DYNAMIC || config->get_mode() == AdapterConfig::MODE_STATIC_RANK || (!diff.alpha && !diff.adapter),
                 "Cannot change adapters and/or the alphas when not one of the dynamic modes are used.");
-            current_config.update(*config);
+            current_config = std::move(updated_config);
         }
+        prepare(infer_request);
         if(need_full_apply) {
             need_full_apply = false;
             set_new_adapter_tensors(infer_request);
         } else if(diff) {
-            if(diff.adapter) {
+            if(diff.adapter || diff.tensor_name_prefix) {
                 set_new_adapter_tensors(infer_request);
             } else if(diff.alpha)  {
                 set_new_adapter_alphas(infer_request);
@@ -1459,6 +1507,172 @@ struct AdapterControllerImpl {
 
     void set_new_adapter_alphas (ov::InferRequest& infer_request) {
         set_new_adapter_tensors(infer_request, /*alpha_only=*/true);
+    }
+
+    bool same_prepared_tensor_cache_key(const AdapterConfig& lhs, const AdapterConfig& rhs) const {
+        return lhs.get_mode() == rhs.get_mode() &&
+               lhs.get_tensor_name_prefix() == rhs.get_tensor_name_prefix() &&
+               lhs.get_adapters_and_alphas() == rhs.get_adapters_and_alphas();
+    }
+
+    void prepare(ov::InferRequest& infer_request) {
+        std::optional<ov::element::Type> new_output_type;
+        const auto compiled_model = infer_request.get_compiled_model();
+        ov::element::Type inference_precision = ov::element::dynamic;
+        if (infer_device_is_gpu) {
+            inference_precision = compiled_model.get_property(ov::hint::inference_precision);
+            OPENVINO_ASSERT(inference_precision == ov::element::f16 ||
+                                inference_precision == ov::element::f32 ||
+                                inference_precision == ov::element::dynamic,
+                            "Unsupported GPU inference precision hint for LoRA state preparation: ",
+                            inference_precision);
+            if (inference_precision == ov::element::f16) {
+                new_output_type = ov::element::f16;
+            }
+        }
+
+        if (output_type_initialized && state_output_type_override == new_output_type) {
+            return;
+        }
+
+        state_output_type_override = new_output_type;
+        output_type_initialized = true;
+        prepared_tensor_cache.clear();
+        prepared_tensor_cache_byte_size = 0;
+        lora_state_evaluators.clear();
+
+        prepare_initial_configs();
+    }
+
+    ov::element::Type state_output_type(const ov::op::util::VariableInfo& variable_info) const {
+        OPENVINO_ASSERT(output_type_initialized,
+                        "LoRA output type must be prepared after model compilation");
+        return state_output_type_override.value_or(variable_info.data_type);
+    }
+
+    LoRAParts<ov::Tensor> allocate_lora_state_tensors(const LoRAVarIDs& lora_var_ids) const {
+        return {
+            ov::Tensor(state_output_type(lora_var_ids.alpha),
+                       dynamic_to_static(lora_var_ids.alpha.data_shape)),
+            ov::Tensor(state_output_type(lora_var_ids.A),
+                       dynamic_to_static(lora_var_ids.A.data_shape)),
+            ov::Tensor(state_output_type(lora_var_ids.B),
+                       dynamic_to_static(lora_var_ids.B.data_shape))
+        };
+    }
+
+    std::vector<LoRAWeightGetter> make_weight_getters(const AdapterConfig& config) const {
+        std::vector<LoRAWeightGetter> weight_getters;
+        const auto& adapters = config.get_adapters();
+        weight_getters.reserve(adapters.size());
+        for (const auto& adapter : adapters) {
+            auto adapter_impl = get_adapter_impl(adapter);
+            weight_getters.emplace_back(
+                LoRAWeightGetterDefault<LoRAWeight, LoRANode>(&adapter_impl->get_tensors(),
+                                                              config.get_tensor_name_prefix().value_or(std::string())));
+        }
+        return weight_getters;
+    }
+
+    void validate_prepared_tensor(const ov::Tensor& tensor,
+                                  const ov::op::util::VariableInfo& variable_info,
+                                  const ov::element::Type& expected_type) const {
+        OPENVINO_ASSERT(tensor);
+        OPENVINO_ASSERT(tensor.get_element_type() == expected_type);
+        OPENVINO_ASSERT(variable_info.data_shape.compatible(ov::PartialShape(tensor.get_shape())));
+    }
+
+    std::vector<LoRAParts<ov::Tensor>> prepare_config_tensors(
+        const AdapterConfig& config,
+        const std::vector<LoRAWeightGetter>& weight_getters) {
+        std::vector<LoRAParts<ov::Tensor>> prepared_tensors;
+        prepared_tensors.reserve(variable_ids.size());
+        for (const auto& lora_var_ids : variable_ids) {
+            auto output_tensors = allocate_lora_state_tensors(lora_var_ids.second);
+            auto tensors = prepare_lora_tensors(lora_var_ids.first,
+                                                weight_getters,
+                                                output_tensors,
+                                                /*set_empty_adapters=*/true,
+                                                /*alpha_only=*/false,
+                                                config);
+            validate_prepared_tensor(tensors.alpha,
+                                     lora_var_ids.second.alpha,
+                                     state_output_type(lora_var_ids.second.alpha));
+            validate_prepared_tensor(tensors.A,
+                                     lora_var_ids.second.A,
+                                     state_output_type(lora_var_ids.second.A));
+            validate_prepared_tensor(tensors.B,
+                                     lora_var_ids.second.B,
+                                     state_output_type(lora_var_ids.second.B));
+            prepared_tensors.push_back(std::move(tensors));
+        }
+        return prepared_tensors;
+    }
+
+    size_t prepared_tensors_byte_size(const std::vector<LoRAParts<ov::Tensor>>& tensors) const {
+        size_t result = 0;
+        for (const auto& tensor_parts : tensors) {
+            result += tensor_parts.alpha.get_byte_size();
+            result += tensor_parts.A.get_byte_size();
+            result += tensor_parts.B.get_byte_size();
+        }
+        return result;
+    }
+
+    const std::vector<LoRAParts<ov::Tensor>>& get_or_prepare_config_tensors(
+        const AdapterConfig& config,
+        const std::vector<LoRAWeightGetter>& weight_getters,
+        bool& cache_hit) {
+        for (auto it = prepared_tensor_cache.begin(); it != prepared_tensor_cache.end(); ++it) {
+            if (same_prepared_tensor_cache_key(it->config, config)) {
+                cache_hit = true;
+                prepared_tensor_cache.splice(prepared_tensor_cache.begin(), prepared_tensor_cache, it);
+                return prepared_tensor_cache.front().tensors;
+            }
+        }
+
+        cache_hit = false;
+        auto tensors = prepare_config_tensors(config, weight_getters);
+        auto byte_size = prepared_tensors_byte_size(tensors);
+
+        // Retain a single oversized entry so a valid config still benefits from caching.
+        // Otherwise evict least-recently-used entries until both limits are satisfied.
+        while (!prepared_tensor_cache.empty() &&
+               (prepared_tensor_cache.size() >= prepared_tensor_cache_capacity ||
+                byte_size > prepared_tensor_cache_max_bytes ||
+                prepared_tensor_cache_byte_size > prepared_tensor_cache_max_bytes - byte_size)) {
+            prepared_tensor_cache_byte_size -= prepared_tensor_cache.back().byte_size;
+            prepared_tensor_cache.pop_back();
+        }
+        prepared_tensor_cache.push_front({config, std::move(tensors), byte_size});
+        prepared_tensor_cache_byte_size += byte_size;
+        return prepared_tensor_cache.front().tensors;
+    }
+
+    void prepare_initial_configs() {
+        if (variable_ids.empty()) {
+            return;
+        }
+
+        std::vector<AdapterConfig> configs;
+        AdapterConfig empty_config(current_config.get_mode());
+        empty_config.set_tensor_name_prefix(current_config.get_tensor_name_prefix());
+        configs.push_back(std::move(empty_config));
+
+        for (const auto& adapter_and_alpha : current_config.get_adapters_and_alphas()) {
+            AdapterConfig single_config(adapter_and_alpha.first,
+                                        adapter_and_alpha.second,
+                                        current_config.get_mode());
+            single_config.set_tensor_name_prefix(current_config.get_tensor_name_prefix());
+            configs.push_back(std::move(single_config));
+        }
+        configs.push_back(current_config);
+
+        for (const auto& config : configs) {
+            auto weight_getters = make_weight_getters(config);
+            bool cache_hit = false;
+            get_or_prepare_config_tensors(config, weight_getters, cache_hit);
+        }
     }
 
     void set_new_adapter_tensors(ov::InferRequest& infer_request, bool alpha_only = false) {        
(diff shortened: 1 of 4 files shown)


src/cpp/src/lora/adapter.cpp
```diff
@@ -1632,10 +1855,14 @@ struct AdapterControllerImpl {
...
+        ov::element::Type concat_type = ov::element::dynamic;
         for(size_t i = 0; i < inputs.size(); ++i) {
             NodePtr input = parameters[(alpha_only ? 1 : 3)*i + offset] = input_accessor(inputs[i]);
-            if(input->get_output_element_type(0) != output.get_element_type()) {
-                input = std::make_shared<v0::Convert>(input, output.get_element_type());
+            if(concat_type == ov::element::dynamic) {
+                concat_type = input->get_output_element_type(0);
+            } else if(input->get_output_element_type(0) != concat_type) {
+                input = std::make_shared<v0::Convert>(input, concat_type);
             }
```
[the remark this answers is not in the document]
ahnyoung-paul: Selecting the `concat` type from the first input was outside the intended scope of this change and made mixed-precision behavior dependent on adapter order. I restored the previous input-side conversion so every concat input is converted to `output_type` before the `Concat` operation. The post-concat conversion remains unchanged. This  preserves the existing mixed-precision behavior while retaining the precision-aware output tensor allocation and prepared-tensor caching added by this PR.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1509,172 @@ struct AdapterControllerImpl {
...
+               lhs.get_tensor_name_prefix() == rhs.get_tensor_name_prefix() &&
+               lhs.get_adapters_and_alphas() == rhs.get_adapters_and_alphas();
+    }
+
+    void prepare(ov::InferRequest& infer_request) {
+        std::optional<ov::element::Type> new_output_type;
+        const auto compiled_model = infer_request.get_compiled_model();
+        ov::element::Type inference_precision = ov::element::dynamic;
+        if (infer_device_is_gpu) {
+            inference_precision = compiled_model.get_property(ov::hint::inference_precision);
```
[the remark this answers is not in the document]
ahnyoung-paul: I removed the requested-device-based `infer_device_is_gpu` member and now query `ov::execution_devices` from the  compiled model inside `prepare()`. The f16 output override is enabled only when all reported execution devices are GPU, so mixed CPU/GPU execution does not receive a global GPU-specific override. For direct device compiled  models, the inference precision is queried directly. For meta-device configurations such as `AUTO` or `HETERO`, where the compiled model may not expose `ov::hint::inference_precision` directly, it is resolved from  `ov::device::properties`. If the effective precision cannot be resolved consistently, the declared VariableInfo types are preserved.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1504,214 @@ struct AdapterControllerImpl {
...
+                get_inference_precision(compiled_model, execution_devices).value_or(ov::element::dynamic);
+            OPENVINO_ASSERT(inference_precision == ov::element::f16 ||
+                                inference_precision == ov::element::f32 ||
+                                inference_precision == ov::element::dynamic,
+                            "Unsupported GPU inference precision hint for LoRA state preparation: ",
+                            inference_precision);
+            if (inference_precision == ov::element::f16) {
+                new_output_type = ov::element::f16;
+            }
+        }
```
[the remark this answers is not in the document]
ahnyoung-paul: I removed the hard assertion so only f16 enables the output override, while all other precisions preserve the declared `VariableInfo` types.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1504,214 @@ struct AdapterControllerImpl {
...
+
+        for (const auto& adapter_and_alpha : current_config.get_adapters_and_alphas()) {
+            AdapterConfig single_config(adapter_and_alpha.first,
+                                        adapter_and_alpha.second,
+                                        current_config.get_mode());
+            single_config.set_tensor_name_prefix(current_config.get_tensor_name_prefix());
+            configs.push_back(std::move(single_config));
+        }
+        configs.push_back(current_config);
+
```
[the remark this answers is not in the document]
ahnyoung-paul: `MODE_STATIC_RANK` now warms up only `current_config`, avoiding incompatible empty or per-adapter ranks while preserving lazy preparation for valid configurations.

ahnyoung-paul:
@likholat Could you please review this PR? thanks.

MaximProshin:
@ahnyoung-paul , how did you confirm that it doesn't impact the accuracy? Do you have any numbers?

ahnyoung-paul:
> @ahnyoung-paul , how did you confirm that it doesn't impact the accuracy? Do you have any numbers?

Not yet. I'm currently checking the accuracy to see if there is any regression. I'll share the results once they're available. thanks.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1285,15 +1293,45 @@ struct AdapterControllerImpl {
...
 
+    struct PreparedTensorCacheEntry {
+        AdapterConfig config;
+        std::vector<LoRAParts<ov::Tensor>> tensors;
+        size_t byte_size = 0;
+    };
+
+    // The cache owns evaluator output tensors for the controller lifetime. Keep a small
+    // LRU bound because every entry contains alpha/A/B outputs for every transformed layer.
+    static constexpr size_t prepared_tensor_cache_capacity = 8;
```
likholat, line 1315: Any specific reason for these values? `prepared_tensor_cache_capacity` and `prepared_tensor_cache_max_bytes` are hardcoded without justification.
Please either document the reasoning or expose them as configurable properties

src/cpp/src/lora/adapter.cpp
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <algorithm>
+#include <cstdlib>
```
likholat, line 5: Please add CI tests covering the new logic

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1504,215 @@ struct AdapterControllerImpl {
...
+            }
+            inference_precision = device_precision;
+        }
+        return inference_precision;
+    }
+
+    void prepare(ov::InferRequest& infer_request) {
+        std::optional<ov::element::Type> new_output_type;
+        const auto compiled_model = infer_request.get_compiled_model();
+        ov::element::Type inference_precision = ov::element::dynamic;
```
likholat: `inference_precision` is declared before `if`, but only read inside it. Declare it locally inside the if block instead

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1504,215 @@ struct AdapterControllerImpl {
         set_new_adapter_tensors(infer_request, /*alpha_only=*/true);
     }
 
+    bool same_prepared_tensor_cache_key(const AdapterConfig& lhs, const AdapterConfig& rhs) const {
```
likholat, line 1515: Please add brief doc comments above the new methods. A one-line summary of purpose, parameters, and return value would help future maintainers

src/cpp/src/lora/adapter.cpp
likholat, line 5: Please check the wwb numbers, you can run wwb with LoRA like that:
```
wwb --base-model <model> \
    ...
    --adapters <hf_lora_repo_or_local_path> \
    --alphas <alpha> \
   
wwb --target-model <model> \
    ...
    --adapters <path/to/adapter.safetensors> \
    --alphas <alpha> \
    --genai
```

Please note that  GenAI implementation merges the traditional LoRA parameters into a single effective scaling factor used during inference:
https://github.com/openvinotoolkit/openvino.genai/tree/master/samples/cpp/image_generation#lora-alpha-interpretation-in-openvino-genai

ahnyoung-paul: @likholat Thanks for the suggestion — I'll run WWB to confirm the precision change (bf16 -> f16) doesn't affect output.

However, WWB only applies the adapter once at pipeline construction, so it doesn't cover runtime switching. Is there an existing test for that?

`test_visual_language_lora_pa_backend` does switch adapters, but only asserts the output is non-empty and never switches back to the original adapter — so the cache-hit path isn't covered.

Would extending it with an A -> B -> A sequence, asserting the reapplied adapter reproduces the original output, be the right way to add the CI tests you asked for?

likholat: Hi @ahnyoung-paul, no, WWB doesn't cover adapter switching, but we should at least validate the basic use case.
If you'd like, it would be great to add support for the adapter switching use case in WWB as well

> Would extending it with an A -> B -> A sequence, asserting the reapplied adapter reproduces the original output, be the right way to add the CI tests you asked for?

Yes, that sounds like the right approach

src/cpp/src/lora/adapter.cpp
```diff
@@ -1432,20 +1479,23 @@ struct AdapterControllerImpl {
...
                 config->get_mode() == AdapterConfig::MODE_AUTO || config->get_mode() == AdapterConfig::MODE_DYNAMIC || config->get_mode() == AdapterConfig::MODE_STATIC_RANK || (!diff.alpha && !diff.adapter),
                 "Cannot change adapters and/or the alphas when not one of the dynamic modes are used.");
-            current_config.update(*config);
+            current_config = std::move(updated_config);
         }
+        prepare(infer_request);
         if(need_full_apply) {
             need_full_apply = false;
             set_new_adapter_tensors(infer_request);
         } else if(diff) {
```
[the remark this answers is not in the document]
ahnyoung-paul: fixed. need_full_apply was tracked once per AdapterControllerImpl and didn't distinguish between infer request instances, so a pipeline recreating an infer request with an unchanged config would skip set_new_adapter_tensors() and leave that new request's LoRA state uninitialized.

Added a std::optional<ov::InferRequest> last_applied_infer_request member that tracks which request instance last received a full apply. apply() now forces set_new_adapter_tensors() whenever the incoming request differs from the last one that was fully applied (via ov::InferRequest::operator!=), regardless of whether the config changed.

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,59 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+    def test_visual_language_lora_switch_gpu_precision_cache(self, convert_model, download_test_content, prompt):
+        # Covers the GPU inference-precision-aware prepared tensor cache in AdapterControllerImpl:
+        # switching configs must invalidate/reprepare cached tensors as needed, and switching back
+        # to a previously used config must still produce the same output as the first time it was used.
+        gpu_devices = [device for device in ov.Core().get_available_devices() if device.startswith("GPU")]
+        if not gpu_devices:
+            pytest.skip("GPU device not available")
+
+        target_device = gpu_devices[0] if len(gpu_devices) == 1 else "GPU.1"
+
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed it.

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,57 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+    def test_visual_language_lora_switch_gpu_precision_cache(self, convert_model, download_test_content, prompt):
+        # Covers the GPU inference-precision-aware prepared tensor cache in AdapterControllerImpl:
+        # switching configs must invalidate/reprepare cached tensors as needed, and switching back
+        # to a previously used config must still produce the same output as the first time it was used.
+        gpu_devices = [device for device in ov.Core().get_available_devices() if device.startswith("GPU")]
+        if not gpu_devices:
+            pytest.skip("GPU device not available")
+
+        target_device = gpu_devices[0] if len(gpu_devices) == 1 else gpu_devices[1]
+
```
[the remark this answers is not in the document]
ahnyoung-paul: I considered always preferring GPU.0, but that's not safe either: some test environments can't be pinned to always-GPU.0-is-fine — in particular, older/weaker CPU (iGPU) + newer discrete GPU combos, where GPU.0 (iGPU) has actually hit CL_OUT_OF_RESOURCES on this VLM+LoRA workload in our testing (i.e. Raptor Lake iGPU + BMG discrete). Always picking GPU.0 would make the test fail precisely in that configuration.

Given that known failure mode, I'm keeping GPU.1 as the fallback when more than one GPU is enumerated. Open to a more robust device-capability check if one exists, but a fixed "always index 0" or "always discrete" rule doesn't hold across all HW generations either.

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,57 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+        image = Image.open(image_path).convert("RGB")
+        image_tensor = ov.Tensor(np.array(image))
+
+        adapter = ov_genai.Adapter(adapter_path)
+        config_a = ov_genai.AdapterConfig()
+        config_a.add(adapter, 2.0)
+        config_b = ov_genai.AdapterConfig()
+        config_b.add(adapter, 0.5)
+
+        pipe = ov_genai.VLMPipeline(convert_model, target_device, ATTENTION_BACKEND="PA", adapters=config_a)
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed it

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,80 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+        [
+            pytest.param(
+                "Qwen2-VL-2B-Instruct",
+                ("qwen2b_lora_100_adapter_model.safetensors", "monalisa.jpg"),
+                "Who drew this painting?",
+            ),
+        ],
+        indirect=["convert_model", "download_test_content"],
+    )
+    def test_visual_language_lora_switch_gpu_precision_cache(self, convert_model, download_test_content, prompt):
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed it.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1518,259 @@ struct AdapterControllerImpl {
...
+    // Evaluates and validates the alpha (and, unless alpha_only, A/B) tensors for every
+    // LoRA-applicable layer for a given config.
+    std::vector<LoRAParts<ov::Tensor>> prepare_config_tensors(
+        const AdapterConfig& config,
+        const std::vector<LoRAWeightGetter>& weight_getters,
+        bool alpha_only = false) {
+        std::vector<LoRAParts<ov::Tensor>> prepared_tensors;
+        prepared_tensors.reserve(variable_ids.size());
+        for (const auto& lora_var_ids : variable_ids) {
+            auto output_tensors = allocate_lora_state_tensors(lora_var_ids.second);
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed it.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1497,14 +1772,21 @@ struct AdapterControllerImpl {
...
         for(const auto& lora_var_ids : variable_ids) {
             // FIXME: Remove this mapping when the order of state will be the same as the order of variables
             LoRAIndices lora_indices;
             lora_indices.alpha = state_name_to_index.at(lora_var_ids.second.alpha.variable_id);
             lora_indices.A = state_name_to_index.at(lora_var_ids.second.A.variable_id);
             lora_indices.B = state_name_to_index.at(lora_var_ids.second.B.variable_id);
-            set_lora_tensors(state, lora_var_ids.first, lora_var_ids.second, lora_indices, weight_getters, alpha_only);
+            OPENVINO_ASSERT(prepared_tensor_it != prepared_tensors.end());
+            set_lora_tensors(state, lora_indices, *prepared_tensor_it, alpha_only);
+            ++prepared_tensor_it;
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed it.

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,71 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+        )
+
+        generation_config = ov_genai.GenerationConfig()
+        generation_config.max_new_tokens = 100
+
+        result_a_first = pipe.generate(prompt, images=[image_tensor], generation_config=generation_config)
+        assert len(result_a_first.texts[0]) > 0, "Generation with config A should produce output"
+
+        result_b = pipe.generate(prompt, images=[image_tensor], generation_config=generation_config, adapters=config_b)
+        assert len(result_b.texts[0]) > 0, "Generation with config B should produce output"
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed it.

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,82 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+
+        # A and B share the same adapter and differ only by alpha. The prepared-tensor cache is keyed
+        # on adapter identity alone, so a stale cached alpha would make B silently reuse A's scaling.
+        # Asserting the outputs differ is what makes the A -> B -> A check below meaningful: without
+        # it, ignoring alpha entirely would still satisfy every other assertion in this test.
+        assert result_b.texts[0] != result_a_first.texts[0], (
+            "Config B (alpha=0.0) should not reproduce config A (alpha=2.0) output; "
+            "identical output suggests the alpha update was ignored"
+        )
```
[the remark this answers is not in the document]
ahnyoung-paul: fixed it.

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Move `inference_precision` into the block

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Add CI test. thanks.

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Add the document for reasoning.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1285,15 +1293,57 @@ struct AdapterControllerImpl {
...
+    std::list<PreparedTensorCacheEntry> prepared_tensor_cache;
+    size_t prepared_tensor_cache_byte_size = 0;
+
     // Stores the actual LoRA weight getter used for Constant tensor replacement
     // Needed to track which LoRA tensors were actually applied to suppress unused tensor warnings
     std::shared_ptr<LoRAWeightGetterDefault<NodePtr, NodePtr>> const_getter_impl;
 
-    AdapterControllerImpl(std::shared_ptr<ov::Model> model, const AdapterConfig& config) :
+    static std::string lora_evaluator_device(const std::string& infer_device) {
+        if (const char* env = std::getenv("OV_LORA_EVALUATOR_DEVICE")) {
```
Wovchena, line 1330: `OV_LORA_EVALUATOR_DEVICE` isn't used anywhere nor it's documented. Remove `lora_evaluator_device()`

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,93 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+        ],
+        indirect=["convert_model", "download_test_content"],
+    )
+    def test_visual_language_lora_switch_gpu_precision_cache(self, convert_model, download_test_content, prompt):
+        # Covers the GPU inference-precision-aware prepared tensor cache in AdapterControllerImpl:
+        # switching configs must invalidate/reprepare cached tensors as needed, and switching back
+        # to a previously used config must still produce the same output as the first time it was used.
+        gpu_devices = [device for device in ov.Core().get_available_devices() if device.startswith("GPU")]
+        if not gpu_devices:
+            pytest.skip("GPU device not available")
```
Wovchena, line 211: There's no GPU on pre-commit so the test is always skipped. The test seem to make sense for CPU as well: it verifies cache's work. Can the device be set to CPU in the test if it's reasonable?

ahnyoung-paul:
> > @ahnyoung-paul , how did you confirm that it doesn't impact the accuracy? Do you have any numbers?
> 
> Not yet. I'm currently checking the accuracy to see if there is any regression. I'll share the results once they're available. thanks.

@MaximProshin, I ran WWB across precision (int4/FP16) and LoRA on/off combinations, comparing `baseline` vs `fix_lora_switching` — no accuracy regression was observed in any combination. Full numbers (tables, per-run data, and notes on measurement variation) are posted on the ticket, CVS-187607 — please check there for details.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1518,270 @@ struct AdapterControllerImpl {
...
+    // Returns the inference precision shared by all execution devices, or nullopt if it can't be determined or differs across devices.
+    std::optional<ov::element::Type> get_inference_precision(
+        const ov::CompiledModel& compiled_model,
+        const std::vector<std::string>& execution_devices) const {
+        const auto supported_properties = compiled_model.get_property(ov::supported_properties);
+        if (std::find(supported_properties.begin(),
+                      supported_properties.end(),
+                      ov::hint::inference_precision) != supported_properties.end()) {
+            return compiled_model.get_property(ov::hint::inference_precision);
+        }
```
Wovchena, line 1542: Are there actually cases when `ov::hint::inference_precision` isn't supported? If not, the rest of `get_inference_precision()` is a dead code which should be removed. If such cases exist, add a comment describing them

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Hi @likholat, I added a unit test covering the `A → B → A` adapter-switching sequence, asserting the reapplied adapter reproduces the original output.

I initially tried using the `Qwen3-VL-4B` model for this test, but locally it took too long to run, so I switched the CI test to use the existing `Qwen2-VL-2B` model instead.

Could you take a look and review the unit test? thanks.

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Added doc comments.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1518,270 @@ struct AdapterControllerImpl {
...
+        const bool infer_device_is_gpu =
+            !execution_devices.empty() &&
+            std::all_of(execution_devices.begin(), execution_devices.end(), [](const std::string& device) {
+                return device.find("GPU") != std::string::npos;
+            });
+        if (infer_device_is_gpu) {
+            const ov::element::Type inference_precision =
+                get_inference_precision(compiled_model, execution_devices).value_or(ov::element::dynamic);
+            if (inference_precision == ov::element::f16) {
+                new_output_type = ov::element::f16;
```
Wovchena, line 1587: Why do you only care about `f16`? Should `new_output_type` always match model's inference precision?

src/cpp/src/lora/adapter.cpp
```diff
@@ -1402,15 +1440,17 @@ struct AdapterControllerImpl {
...
         }
     };
 
     ConfigChanged compare_configs(const AdapterConfig& config1, const AdapterConfig& config2) {
         ConfigChanged diff;
         diff.mode = config1.get_mode() != config2.get_mode();
+        diff.tensor_name_prefix = config1.get_tensor_name_prefix() != config2.get_tensor_name_prefix();
         // TODO: Use `set` from this commented block when the config change tracking is implemented at adapter granularity and will track order of adapters correctly
         // std::set<Adapter>
         //     adapters1(config1.adapters.begin(), config1.adapters.end()),
```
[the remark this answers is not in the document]
ahnyoung-paul: fixed it.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1511,221 @@ struct AdapterControllerImpl {
         set_new_adapter_tensors(infer_request, /*alpha_only=*/true);
     }
 
+    // Checks whether two configs would produce the same prepared tensors, making one reusable for the other.
+    bool same_prepared_tensor_cache_key(const AdapterConfig& lhs, const AdapterConfig& rhs) const {
+        return lhs.get_mode() == rhs.get_mode() &&
+               lhs.get_tensor_name_prefix() == rhs.get_tensor_name_prefix() &&
+               lhs.get_adapters_and_alphas() == rhs.get_adapters_and_alphas();
+    }
```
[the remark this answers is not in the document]
ahnyoung-paul: fixed it.

ahnyoung-paul:
@likholat could you please review it again. thanks.

Wovchena (changes requested):

src/cpp/src/lora/adapter.cpp
```diff
@@ -1285,15 +1293,57 @@ struct AdapterControllerImpl {
...
+    // are ~252 MB, so 512 MiB covers roughly two full adapter configs before eviction kicks
+    // in; 8 entries bounds the config count for setups with many small/lightweight adapters,
+    // where the byte cap alone wouldn't trigger eviction soon enough. These numbers are
+    // model-dependent (rank, hidden size, and layer count all scale adapter size) and were
+    // not tuned per model; if a use case needs a different bound, exposing these as
+    // configurable properties is the natural next step.
+    static constexpr size_t prepared_tensor_cache_capacity = 8;
+    static constexpr size_t prepared_tensor_cache_max_bytes = 512 * 1024 * 1024;
+    std::list<PreparedTensorCacheEntry> prepared_tensor_cache;
+    size_t prepared_tensor_cache_byte_size = 0;
```
Wovchena, line 1323: Factor out cache implementation into a separate class so it's mixed less with the rest of adapter functitonality

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: `lora_evaluator_device()` was left from an experiment to run the evaluator on `GPU`. Since the evaluator now runs only on CPU, I removed `lora_evaluator_device()` and `OV_LORA_EVALUATOR_DEVICE`.

tests/python_tests/samples/test_lora.py
ahnyoung-paul: Changed the test to CPU, so it now runs in pre-commit. It verifies adapter switching and prepared-tensor cache reuse. However, I am concerned that the GPU-specific f16 and state upload paths are not covered because pre-commit has no GPU runner.

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Yes. AUTO in `CUMULATIVE_THROUGHPUT` mode and HETERO do not report `inference_precision` directly. In these cases, it is read from `ov::device::properties` of the execution devices. I added comments to explain this path.

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: The issue was found with `f16` inference when the adapter tensors were `bf16` or `f32`. This type mismatch caused `set_state()` to fall back to CPU conversion instead of direct GPU copy. I only handle `f16` because this is the case we tested and confirmed. For other precisions, the previous behavior is kept.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1497,14 +1818,31 @@ struct AdapterControllerImpl {
...
         }
 
+        // An alpha-only update needs neither the cached A/B nor a new cache entry, so evaluate
+        // just the alpha tensors directly. The full path goes through the cache, which refreshes
+        // stale alpha tensors on a hit (see get_or_prepare_config_tensors).
+        std::vector<LoRAParts<ov::Tensor>> alpha_only_tensors;
+        const std::vector<LoRAParts<ov::Tensor>>* prepared_tensors_ptr;
+        if (alpha_only) {
+            alpha_only_tensors = prepare_config_tensors(current_config, weight_getters, /*alpha_only=*/true);
+            prepared_tensors_ptr = &alpha_only_tensors;
```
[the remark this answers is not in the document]
ahnyoung-paul: This behavior is intentional. Alpha is not part of the cache key, so the large A/B tensors can be reused when only alpha changes. In this case, only the small alpha tensor is re-evaluated. I updated the PR description to clarify that the cache avoids evaluator execution for A/B, not for alpha.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1518,270 @@ struct AdapterControllerImpl {
...
+        // Otherwise evict least-recently-used entries until both limits are satisfied.
+        while (!prepared_tensor_cache.empty() &&
+               (prepared_tensor_cache.size() >= prepared_tensor_cache_capacity ||
+                byte_size > prepared_tensor_cache_max_bytes ||
+                prepared_tensor_cache_byte_size > prepared_tensor_cache_max_bytes - byte_size)) {
+            prepared_tensor_cache_byte_size -= prepared_tensor_cache.back().byte_size;
+            prepared_tensor_cache.pop_back();
+        }
+        prepared_tensor_cache.push_front({config, std::move(tensors), byte_size});
+        prepared_tensor_cache_byte_size += byte_size;
```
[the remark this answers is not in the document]
ahnyoung-paul: Updated the PR description with all required checklist items. I also ran WWB against the baseline with int4/FP16 and LoRA enabled/disabled. No accuracy regression was observed. The detailed results are available in CVS-187607.

src/cpp/src/lora/adapter.cpp
ahnyoung-paul: Moved the prepared tensor cache into a separate `PreparedTensorCache` class. Cache lookup, insertion, LRU eviction, and byte-size tracking are now handled by this class.

src/cpp/src/lora/adapter.cpp
```diff
@@ -1279,14 +1286,124 @@ bool operator== (const Adapter& a, const Adapter& b) {
...
+
+    Tensors& insert(const AdapterConfig& config, Tensors tensors) {
+        const size_t byte_size = total_byte_size(tensors);
+
+        // Retain a single oversized entry so a valid config still benefits from caching.
+        // Otherwise evict least-recently-used entries until both limits are satisfied.
+        while (!m_entries.empty() &&
+               (m_entries.size() >= capacity ||
+                byte_size > max_bytes ||
+                m_byte_size > max_bytes - byte_size)) {
```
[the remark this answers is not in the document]
ahnyoung-paul: Fixed. If the prepared tensors exceed `max_bytes`, they are used only for the current apply and are not inserted into the cache. I also removed `m_scratch`, so the cache does not keep an extra reference to them. `max_bytes` is now a hard bound for tensors retained by `PreparedTensorCache`.

tests/python_tests/samples/test_lora.py
```diff
@@ -189,3 +189,65 @@ def test_visual_language_lora_pa_backend(self, convert_model, download_test_cont
...
+        config_a.add(adapter, 2.0)
+        # A and B use the same adapter and differ only by alpha.
+        config_b = ov_genai.AdapterConfig()
+        config_b.add(adapter, 0.0)
+
+        pipe = ov_genai.VLMPipeline(
+            convert_model,
+            "CPU",
+            ATTENTION_BACKEND="PA",
+            adapters=config_a
```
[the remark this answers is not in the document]
ahnyoung-paul: This test was changed to CPU because pre-commit has no GPU runner and the previous GPU test was always skipped. It now covers the device-independent adapter switching and cache behavior. The GPU-specific f16 path cannot be covered by the current pre-commit environment, so it was tested separately on GPU.

ahnyoung-paul:
> ### 🟡 Changes recommended
> There is a runtime edge case in the evaluator-device selection (empty device with OV_LORA_EVALUATOR_DEVICE=INFER) and the new GPU test contains a potentially flaky assertion that should be made more robust.
> 
> _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> 
> Review details

`lora_evaluator_device()` and `OV_LORA_EVALUATOR_DEVICE` were removed, so this empty-device path no longer exists. The evaluator now uses CPU directly.

ahnyoung-paul:
> ### 🟡 Changes recommended
> The LoRA state-reset bug and unresolved caching and test issues must be addressed before approval.
> 
> _Once you've addressed the issues Copilot identified, you can request another Copilot review._
> 
> Review details

Fixed. I replaced `m_language.reset_state()` with `reset_language_state()`. It resets the non-LoRA states while preserving the LoRA states, so the following `apply()` does not leave the adapter states reset.

ahnyoung-paul:
> ### 🔵 Needs a closer look
> The new test retains a potentially flaky decoded-text inequality, and the PR description requires corrections.
> 
> Review details

Removed the non-deterministic inequality assertion and its related comment. The test now uses the alpha-zero equality with the no-adapter result and the A→B→A equality check.

ahnyoung-paul:
@Wovchena @likholat Could you please review this PR again?

src/cpp/src/lora/adapter.cpp
```diff
@@ -1461,6 +1629,239 @@ struct AdapterControllerImpl {
...
+        const std::vector<std::string>& execution_devices) const {
+        const auto supported_properties = compiled_model.get_property(ov::supported_properties);
+        // Direct devices such as CPU and GPU report the precision themselves.
+        if (std::find(supported_properties.begin(),
+                      supported_properties.end(),
+                      ov::hint::inference_precision) != supported_properties.end()) {
+            return compiled_model.get_property(ov::hint::inference_precision);
+        }
+
+        // AUTO in CUMULATIVE_THROUGHPUT mode and HETERO do not report it, because the real
```
Wovchena:
```suggestion
        // AUTO in CUMULATIVE_THROUGHPUT mode and HETERO do not report inference_precision, because the real
```

src/cpp/src/lora/adapter.cpp
Wovchena: This PR contains two independent changes:
1. **Caching evaluator outputs (A/B tensors)** — a proper GenAI optimization. Approved; please split it into its own PR so it can merge now.
2. **Forcing `f16` output when the device is GPU** — a workaround for a GPU plugin issue: the GPU plugin's `set_state()` converts on the CPU when tensor types differ, and the plugin does not expose the real state type. GenAI has to guess it from `inference_precision`, which only holds for GPU today and can silently break when the plugin changes.

I'm blocking part 2. The fix belongs in the GPU plugin (likely `convert_and_copy()`, the function already touched in https://github.com/openvinotoolkit/openvino/pull/37396). That would fix it for every OpenVINO user of `set_state()`, not just GenAI, and for all precisions.
Please create a GPU ticket for on-device conversion in `set_state()`. If the GPU team cannot schedule it before a deadline, GenAI will accept the workaround provided the new GPU ticket exists and the code is marked `// Workaround for CVS-###`.

PR #4311, opened 2026-08-12
[Whisper] Add multi-batch inference support

byekrang:
## Description

This PR addresses CVS-192316 by adding native multi-batch inference for distinct audio inputs to `WhisperPipeline`.

Previously, Whisper accepted only one audio input per `generate()` call. Applications therefore had to process multiple audios through separate pipeline calls and could not use batched encoder and decoder inference.

This change adds:

* C++ and Python `WhisperPipeline::generate()` overloads for multiple audio inputs.
* Batched encoder inference.
* Per-input language detection.
* Short-form and long-form batch scheduling.
* A shared decoder-generation loop that reuses the existing `SequenceGroup` and `Sampler` machinery.
* Fixed physical decoder rows for distinct audios, preserving their cross-attention state while finished rows remain inert.
* Packing of only live logits before sampling.
* Input-aligned results, including text, scores, detected languages, and timestamp chunks.
* Batch-size and device coverage in C++ and Python tests.

A one-element batch delegates to the existing scalar path, preserving scalar support for beam search, sampling, streaming, long-form generation, and word timestamps.

For batches with more than one input, the supported scope is:

* Greedy decoding.
* CPU and GPU.
* Short-form and long-form audio.
* Per-input automatic or forced language selection.
* Segment-level timestamps through `return_timestamps`.

The following are rejected for batches with more than one input:

* Beam search.
* Multinomial sampling.
* Streaming.
* Word-level timestamps.
* NPU and non-CPU/GPU device configurations.

Native batching is exposed through `WhisperPipeline`. The generic `ASRPipeline` API remains scalar.

Unsupported configurations fail explicitly instead of silently falling back to sequential generation.

CVS-192316

## Checklist:

* [x] This PR follows GenAI Contributing guidelines.


src/python/py_whisper_pipeline.cpp
```diff
@@ -403,6 +454,26 @@ void init_whisper_pipeline(py::module_& m) {
...
+        // Register after the scalar overload so flat float lists continue to bind as one audio.
+        .def(
+            "generate",
+            [](WhisperPipeline& pipe,
+               const std::vector<RawSpeechInput>& raw_speech_inputs,
+               const OptionalWhisperGenerationConfig& generation_config,
+               const pyutils::PyBindStreamerVariant& streamer,
+               const py::kwargs& kwargs) -> py::typing::List<ov::genai::WhisperDecodedResults> {
+                return call_whisper_batch_generate(pipe, raw_speech_inputs, generation_config, streamer, kwargs);
+            },
```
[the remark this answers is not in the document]
byekrang: Both Whisper overloads use the same established pattern: a `py::typing::*` return annotation around a helper that returns `py::object`. The scalar overload declares `py::typing::Union<WhisperDecodedResults>`, not bare `py::object`.

`call_whisper_batch_generate()` returns a genuine Python list, so conversion to `py::typing::List<WhisperDecodedResults>` does not copy the list. The explicit return type is also required to generate `-> list[WhisperDecodedResults`]; changing it to `py::object` would degrade the stub to `-> object`.

This matches existing batch bindings in `py_continuous_batching_pipeline.cpp`, so I am keeping it.

src/cpp/src/whisper/models/statefull_decoder.cpp
```diff
@@ -27,7 +27,8 @@ WhisperStatefullDecoder::WhisperStatefullDecoder(const std::filesystem::path& mo
                                                  const ov::AnyMap& properties,
                                                  const ov::PartialShape& lhs_shape,
                                                  const bool decompose_cross_attention_spda)
-    : m_decompose_cross_attention_spda_ops(decompose_cross_attention_spda) {
+    : m_decompose_cross_attention_spda_ops(decompose_cross_attention_spda),
+      m_recreate_request_per_round(device != "NPU") {
```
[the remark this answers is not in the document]
byekrang: This check intentionally matches the existing NPU-specific compile branch in the same constructor, which also uses `device == "NPU"`. With the current code, `NPU.0` consistently takes the generic compile and request-lifecycle paths.

Normalizing only the new recreation flag would decouple those decisions, making `NPU.0` use the generic compile path but the NPU-specific request-reuse policy. Supporting indexed NPU devices through the specialized path would require normalizing all existing NPU checks together, which is a broader pre-existing issue and outside this PR.

src/python/openvino_genai/py_openvino_genai.pyi
```diff
@@ -6365,6 +6365,7 @@ class WhisperPipeline:
                     models_path (os.PathLike): Path to the model file.
                     device (str): Device to run the model on (e.g., CPU, GPU).
         """
+    @typing.overload
     def generate(self, raw_speech_input: collections.abc.Sequence[typing.SupportsFloat], generation_config: openvino_genai.py_openvino_genai.WhisperGenerationConfig | None = None, streamer: collections.abc.Callable[[str], int | None] | openvino_genai.py_openvino_genai.StreamerBase | None = None, **kwargs) -> WhisperDecodedResults:
```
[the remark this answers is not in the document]
byekrang: Thanks, confirmed. The runtime and docstring accept callbacks returning `StreamingStatus | int | None`, while the generated annotation currently says `int | None`. This is a repository-wide pybind11-stubgen limitation affecting other pipeline overloads as well. I’m checking whether it can be corrected at the generator/binding source rather than manually editing the generated `.pyi`, which would be overwritten.

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -1131,3 +1131,582 @@ def test_streamers(sample_from_dataset, pipelines_fixture, streamer_for_test):
...
+    scalar_keyword = genai_pipe.generate(raw_speech_input=sample_from_dataset, max_new_tokens=10)
+    batch_keyword = genai_pipe.generate(raw_speech_inputs=[sample_from_dataset], max_new_tokens=10)
+
+    assert not isinstance(scalar_keyword, list)
+    assert isinstance(batch_keyword, list)
+
+    # An empty list binds to the scalar overload, not the C++ empty-batch overload.
+    with pytest.raises(RuntimeError, match="raw_speech too short"):
+        genai_pipe.generate([], max_new_tokens=10)
+
```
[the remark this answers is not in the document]
byekrang: Thanks. I added coverage for the batched overload using `raw_speech_inputs=[]` and verified that it raises the expected “empty batch” error. The existing `generate([])` assertion remains to verify that an empty positional list still resolves to the scalar overload.

src/cpp/src/whisper/whisper.hpp
```diff
@@ -48,5 +48,19 @@ WhisperGenerateResult whisper_generate(const ov::genai::WhisperGenerationConfig&
                                        Sampler& sampler,
                                        Tokenizer& tokenizer);
 
+/**
+ * Generates one result per input for a B > 1 batch, preserving input order.
+ * Supports greedy short-form and long-form decoding. Performance metrics are aggregate for the entire batch.
+ */
+std::vector<WhisperGenerateResult> whisper_generate_batch(const ov::genai::WhisperGenerationConfig& config,
```
as-suvorov: Please explain why we need dedicated path for batched inference

src/cpp/src/whisper/whisper.cpp
```diff
@@ -81,407 +92,955 @@ void process_whisper_logits(ov::Tensor logits,
...
+    std::vector<ov::genai::Sequence::Ptr> live_sequences;
+    std::vector<size_t> live_physical_rows;
+    std::vector<int32_t> packed_beam_idx;
+    const size_t capacity = std::max<size_t>(sequence_groups.size(), config.num_beams);
+    active_groups.reserve(capacity);
+    live_sequences.reserve(capacity);
+    live_physical_rows.reserve(capacity);
+    packed_beam_idx.reserve(capacity);
+
+    while (true) {
```
as-suvorov, line 306: Please explain why we cannot "re-use" (copy paste with encoder hidden states alignment according to scheduled sequencies) existing lm_encoding implementation?

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -1131,3 +1132,609 @@ def test_streamers(sample_from_dataset, pipelines_fixture, streamer_for_test):
 
     assert expected == result_handler.decode(genai_pipe.get_tokenizer())
     result_handler.reset()
+
+
+#
+# Batched (multi-audio) generation through WhisperPipeline, which returns one WhisperDecodedResults per audio.
+
+
+def assert_batch_result_shape(batched, pipeline_type: PipelineType, expected_size: int):
```
as-suvorov, line 1141: Please review tests:
1. Why do we need to additionally test behavior for single batch?
2. Consider replacing appropriate tests with genai vs optimum-intel tests

src/cpp/src/whisper/whisper.cpp
```diff
@@ -81,407 +92,955 @@ void process_whisper_logits(ov::Tensor logits,
...
+                         sampler,
+                         tokenizer,
+                         streamer,
+                         perf_metrics,
+                         time_precision,
+                         frame_length_in_seconds};
+
+    // Long-form inputs require internal timestamps for seek advancement even when return_timestamps is false.
+    const auto wants_timestamps = [&](const AudioState& state) {
+        return config.return_timestamps || (state.features.n_frames > feature_extractor.nb_max_frames);
```
as-suvorov: Please double check how transformers and whisper.cpp handles such scenario.
I believe transformers would just throw if it's long form and no return_timestramps=True provided explicitly.
Also as I understand later we process with timestamps and without timestamps groups sequentially. This I believe can eliminate perf improvement of not forcing timestamps. Consider to enable timestamps generation for the whole batch.

src/cpp/src/whisper/whisper.cpp
```diff
@@ -81,407 +92,955 @@ void process_whisper_logits(ov::Tensor logits,
...
-                                                feature_extractor.feature_size,
-                                                feature_extractor.nb_max_frames,
-                                                raw_metrics,
-                                                result.perf_metrics.whisper_raw_metrics);
+// Encoder rows and sequence groups follow cohort order; request IDs preserve input indices.
+void run_cohort_round(SchedulerContext& ctx,
+                      std::vector<AudioState>& states,
+                      const std::vector<size_t>& cohort,
+                      const bool cohort_return_timestamps,
+                      std::vector<std::vector<float>>& mel_windows,
```
as-suvorov: Some of arguments looks duplicated/redundant, please double check if they can be reduced
```suggestion
                      std::vector<AudioState>& states,
                      const bool cohort_return_timestamps,
```

src/cpp/src/whisper/whisper.hpp
byekrang: There isn't a separate generation engine for batching. The scalar and batch entry points both go through the same scheduling and decoder-generation machinery; the batch entry mainly handles vector input/result cardinality and enables fixed physical decoder rows for B>1.

I kept a separate entry point because B>1 also has a different supported-feature contract (greedy only, no streaming/word timestamps, supported devices only), while B==1 delegates to the scalar path and keeps all existing scalar functionality.

src/cpp/src/whisper/whisper.cpp
byekrang: Agreed that we should reuse the `lm_encoding` generation pattern, and we already reuse the same `SequenceGroup`/`Sampler` flow: schedule active groups, infer, sample, and remove logically finished requests.

The part we cannot reuse is its physical-row compaction. I checked the exported stateful Whisper decoder: `beam_idx` gathers the self-attention KV state, but the cross-attention K/V initialized from `encoder_hidden_states` are separate persistent states and are not gathered. For example, after `[A,B,C]` -> `[A,C]`, `beam_idx=[0,2]` would correctly move C's self-attention state into row 1, but row 1's cached cross-attention would still belong to B. Re-aligning `encoder_hidden_states`=`[A,C]` does not fix this because cross-attention K/V are only initialized when the state is empty; resetting the state would also discard the surviving self-attention history.

That's why B>1 keeps the physical audio rows fixed and only adapts the `lm_encoding` pattern with filler tokens for finished rows and packing the live logits back into active `SequenceGroup` order before `Sampler::sample()`. This is specific to the current stateful Whisper export; if cross-attention state becomes gatherable in the future, normal compaction could be used. Please let me know if I am missing something.

src/cpp/src/whisper/whisper.cpp
byekrang: You're right. I checked Transformers and changed this to a batch-wide timestamp-generation decision.

The scheduler now uses:

generate_timestamps = config.return_timestamps || any_input_is_long_form

src/cpp/src/whisper/whisper.cpp
byekrang: Good catch. Thank you. I simplified this together with the timestamp scheduling change.

`cohort_return_timestamps` was effectively the same value throughout the generation call, so it is now stored once as the internal `generate_timestamps` state rather than threaded through `run_cohort_round` and finalization. `mel_windows` and `sequence_groups` were only scratch buffers passed for capacity reuse, so they are now local to the round.

This also removes the timestamp/no-timestamp cohort split and leaves the round interface focused on the actual active audio set.

tests/python_tests/test_whisper_pipeline.py
byekrang: Good points, addressed both.

1. For batch size 1, the behavior worth testing is the delegation contract: `generate([audio])` must delegate to the existing scalar path before the B>1 restrictions are applied. This is important because features such as beam search, sampling, streaming, and word timestamps are unsupported for B>1 but must still work for a one-element batch.

I agree that retesting each scalar feature separately through `[audio]` was redundant, since they already have scalar coverage. I consolidated those tests into a small set that checks the delegation itself and that B>1-restricted configurations still reach the scalar path for B==1.

2. I also replaced GenAI-scalar references with the existing Optimum-Intel reference where the test is validating model output, including transcription and per-input language behavior.

For the mixed short/long cases, the tests are hybrid: text is checked against the Optimum timestamp-mode reference, while OV-specific behavior such as accepting `return_timestamps=false` and not exposing chunks remains asserted separately.

I kept GenAI scalar as the reference only where batch-vs-scalar equivalence is itself the contract, such as timestamp/chunk parity and B==1 delegation.

src/cpp/src/whisper/whisper.hpp
as-suvorov: Is it possible to have single path for B>=1 and just assert supported parameters?

src/cpp/src/whisper/whisper.cpp
as-suvorov: Thanks a lot for explanation! Now it's clear why such complexity required.
Empty batches inference is not optimal. Let's investigate how we can address initial issue:
1. Let's check how transformers implements mutli-batch support
2. Let's check if optimum-intel supports it. If yes, how it's implemented
3. `encoder_hidden_states` are static during specific window autoregression, right?
`encoder_hidden_states=[A,B,C]`, `input_ids=[A,B,C]` - cross-attention K/V are initialized
sequence B is ejected
`encoder_hidden_states=[A,C]`, `input_ids=[A,C]` - this doesn't work as cross-attention K/V are initialized when state is empty. But we still can access cross-attention K/V states directly via ov api and make slicing of a ejected B sequence.
4. Let's check if `encoder_hidden_states` can be re-initialized in input change inside ov graph

byekrang: Thank you for breaking it down.

Transformers Whisper implementation has two levels of batching. In the outer long-form `seek` loop, `_maybe_reduce_batch()` physically removes audios that have completed all their windows, so `[A,B,C]` can become `[A,C]` before the next 30s segment is processed.

Inside one autoregressive `generate()` call, however, the batch is not compacted when one sequence reaches EOS. `GenerationMixin::_sample()` keeps a fixed-size `unfinished_sequences` mask and replaces tokens for finished rows with `pad_token_id`, so the decoder continues effectively as `[A,pad,C]` until all rows in that generation call finish.

So Transformers shrinks between long-form windows, but does not solve intra-decoder-round empty-row compute.

In the Optimum-Intel implementation, `_OVModelForWhisper` explicitly uses `WhisperForConditionalGeneration.generate`, so the high-level Whisper batching/long-form scheduling is inherited from Transformers.

For the stateful OV decoder, `_reorder_cache()` stores the new `beam_idx` and lets the graph reorder cached state. In the non-stateful equivalent, only the first two cache entries per layer (self-attention K/V) are reordered; cross-attention K/V are intentionally kept unchanged because normal beam search assumes all beams belong to the same encoder input. That assumption is valid for several beams of one audio, but not for distinct audios `[A,B,C]`. So Optimum does not contain special logic to remove B's cross-attention state when one distinct audio finishes during autoregression.

I also tested the exact `[A,B,C] -> [A,C]` case on the stateful Whisper decoder. After prefill, I sliced the cross-attention `VariableState`s to rows `[0,2]` using `query_state()` / `get_state()` / `set_state()`, kept self-attention state unchanged, and passed `beam_idx=[0,2]` so the graph compacted self-KV. The next inference succeeded and all decoder states became width 2. Survivor tokens matched the current fixed-row implementation.

So direct cross-KV state slicing works functionally. My current prototype does introduce a noticeable compaction cost because it copies the cached cross-attention state, but I am still investigating whether that cost is inherent or can be reduced with a better C++/OpenVINO implementation before drawing a performance conclusion.

Summary:

1. Transformers

   * outer long-form loop: `[A,B,C] -> [A,C]`
   * inside one decoder call: `[A,B,C] -> [A,PAD,C]`
   * no intra-generation physical compaction

2. Optimum-Intel

   * inherits HF Whisper `generate()`
   * self-KV follows `beam_idx`
   * cross-KV does not
   * no distinct-audio intra-round compaction

3. Static encoder state / direct OV state slicing

   * `encoder_hidden_states` are static within a decoder window
   * manual cross-KV `[A,B,C] -> [A,C]` works
   * `query_state()` / `get_state()` / `set_state()` path is verified
   * performance of an optimized implementation is still under investigation

4. Reinitialization on input change

   * changing `encoder_hidden_states` to `[A,C]` mid-round does not recreate already-initialized cross-KV
   * after `reset_state()`, fresh cross-KV is initialized from the new encoder input
   * another candidate for mid-round compaction is to make cross-KV follow `beam_idx` inside the graph as self-KV already does

src/cpp/src/whisper/whisper.hpp
byekrang: Yes, I will change it to have a single internal path for B >= 1.
That should make the control flow simpler and better match the style you prefer.

src/cpp/src/whisper/whisper.cpp
byekrang: I investigated physical batch compaction further. I tried both direct cross-KV state compaction and making cross-KV follow `beam_idx` in the graph.

The first is functionally promising, but it exposed an OpenVINO CPU-plugin issue around shape-changing `VariableState::set_state()`. The second works functionally but adds a large per-token regression because the cross-KV Gather runs every decoder step.

Given these results, I think keeping filler rows is the better alternative for this PR. The batch still shrinks normally between long-form rounds.

PR #4318, opened 2026-08-13
[GGUF] Use the OpenVINO GGUF frontend: tokenizer from rt_info, stateful graph

mvafin:
## Description
Adopt the native `.gguf` builder from the OpenVINO GGUF frontend, so a `.gguf` is converted by the frontend instead of GenAI's own reader, and build the tokenizer from the converted model.

- **Tokenizer from the frontend model's `rt_info`** instead of re-parsing the file: SPM and BPE flavours, special-token ids, `add_space_prefix`, plus `gemma4` and `plamo2`.
- **BOS/EOS via `CombineSegments` on every path.** Two defects made `encode()` disagree with llama.cpp: the BPE paths never read `tokenizer.ggml.add_bos_token` (BOS-sensitive models such as muse-glimmer and llama3 were silently fed a BOS-less prompt and degenerated), and `add_special_tokens=false` was a **no-op on both paths**, because `MakeAddSpecialTokensSatateful` needs a `CombineSegments` node that the GGUF tokenizer never built. Defaults follow `llama-vocab.cpp` when the key is absent.
- **Make the model stateful here, not in the frontend.** The frontend is universal and always converts to a stateless graph, so GenAI registers `GGUFMakeStateful` as a `DecoderTransformationExtension`, scoped to this one conversion rather than globally via `ov::Core::add_extension`.
- GGUF models only run on the SDPA backend: the frontend's graph is not accepted by `SDPAToPagedAttention`, so continuous batching / PagedAttention is not supported for them and a `scheduler_config` passed alongside one is dropped with a warning instead of crashing.
- New `ov::genai::gguf_reader` property to select between the frontend and the legacy hand-written reader (`llama`/`qwen2`/`qwen3` only). **The legacy reader is the default for now**; the frontend must be selected explicitly (`gguf_reader=FRONTEND`) until it matures.
- WWB: `test_cli_text_gguf.py` plus `model_loaders` support for `.gguf` inputs.

CVS-194185

## Checklist:

### DIFF
--- a/src/cpp/include/openvino/genai/llm_pipeline.hpp
+++ b/src/cpp/include/openvino/genai/llm_pipeline.hpp
@@ -411,6 +411,20 @@ static constexpr ov::Property<bool> prompt_lookup{"prompt_lookup"};
 */
 static constexpr ov::Property<bool> enable_save_ov_model{"enable_save_ov_model"};
 
+/**
+* @brief Selects which reader converts a `.gguf` file.
+*
+* Accepted values:
+*  - `"FRONTEND"` - the OpenVINO GGUF frontend, supporting the full set of architectures.
+*  - `"LEGACY"` (default, for now) - the pre-frontend, hand-written GGUF reader. Only handles
+*    llama/qwen2/qwen3 and ignores `rope_freqs.weight` (no llama-3 RoPE scaling). The frontend
+*    reader is not yet the default because continuous batching / PagedAttention on its converted
+*    graph requires an OpenVINO-side fix (see openvinotoolkit/openvino#37606).
+*
+* Example: `LLMPipeline(model_path, "CPU", ov::genai::gguf_reader("FRONTEND"))`.
+*/
+static constexpr ov::Property<std::string> gguf_reader{"GGUF_READER"};
+
 
 }  // namespace genai
 }  // namespace ov
--- a/src/cpp/src/gguf_utils/gguf_tokenizer.cpp
+++ b/src/cpp/src/gguf_utils/gguf_tokenizer.cpp
@@ -1,22 +1,29 @@
 // Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: Apache-2.0
 
-#include <limits>
+#include <algorithm>
 #include <cstdint>
+#include <cstring>
+#include <limits>
+#include <set>
 
 #include "gguf_tokenizer.hpp"
 
+#include "openvino/frontend/gguf/tokenizer_metadata.hpp"
 #include "openvino/op/add.hpp"
 #include "openvino/op/constant.hpp"
 #include "openvino/op/convert.hpp"
 #include "openvino/op/gather.hpp"
+#include "openvino/op/less.hpp"
 #include "openvino/op/minimum.hpp"
 #include "openvino/op/parameter.hpp"
 #include "openvino/op/range.hpp"
 #include "openvino/op/reduce_max.hpp"
+#include "openvino/op/reduce_sum.hpp"
 #include "openvino/op/shape_of.hpp"
 #include "openvino/op/slice.hpp"
 #include "openvino/op/subtract.hpp"
+#include "openvino/op/unsqueeze.hpp"
 
 #ifdef _WIN32
 #    define NOMINMAX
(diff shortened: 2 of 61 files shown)


src/cpp/src/gguf_utils/gguf_tokenizer.cpp
```diff
@@ -547,43 +964,146 @@ create_tokenizer_from_config(const std::shared_ptr<void>& shared_object_ov_token
...
+            cfg[key] = value.as<ov::Tensor>();
+        } else if (value.is<int>()) {
+            cfg[key] = value.as<int>();
+        } else if (value.is<float>()) {
+            cfg[key] = value.as<float>();
+        }
+        // other types are not part of the tokenizer config and are ignored
+    }
+    return cfg;
+}
```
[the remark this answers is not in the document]
mvafin: Checked this one against the frontend: both GGUF readers store **every** scalar as a shape-`{}` `ov::Tensor` of the file's own element type (`GGUF_VALUE_TYPE_BOOL` -> `ov::element::boolean`, `UINT32` -> `u32`; see `read_metadata_value` in the frontend's `gguf.cpp`), and the frontend copies the variant alternative as-is into `rt_info`. So `bos_token_id`/`add_bos_token` do arrive as `ov::Tensor` and are not dropped today — the `int`/`float` branches in `tokenizer_config_from_rt_info` are unreachable for frontend-produced configs.

That said, the underlying fragility is real: the readers returned the default for any non-`Tensor` representation, which would fail silently. Fixed in 55f03d75 by having `read_tokenizer_flag`/`read_tokenizer_id` also accept a bare `int`.

rkazants:
@mvafin, please do not merge without our explicit review with Artur

MaximProshin:
As it's preview support, please clearly state it everywhere, so that we avoid a situation when GenAI users start using it and get bad performance or accuracy due to limited support in run-time and validation.

mvafin:
> As it's preview support, please clearly state it everywhere, so that we avoid a situation when GenAI users start using it and get bad performance or accuracy due to limited support in run-time and validation.

Currently it is disabled by default. Need to explicitly select frontend option to use it.

mvafin:
@rkazants @apaniukov PR is ready for review

src/cpp/include/openvino/genai/llm_pipeline.hpp
```diff
@@ -411,6 +411,20 @@ static constexpr ov::Property<bool> prompt_lookup{"prompt_lookup"};
 */
 static constexpr ov::Property<bool> enable_save_ov_model{"enable_save_ov_model"};
 
+/**
+* @brief Selects which reader converts a `.gguf` file.
+*
+* Accepted values:
+*  - `"FRONTEND"` - the OpenVINO GGUF frontend, supporting the full set of architectures.
```
rkazants: what does it mean full set of archs? Can you document it in OpenVINO GenAI? Any perf and accuracy results?

rkazants: I would rather say that this is preview feature for which we can expect different limitations.

mvafin: Agree, I changed wording

tests/python_tests/data/models.py
```diff
@@ -32,7 +32,8 @@ def get_models_list() -> tuple[str, ...]:
     {
         "gguf_model_id": "Qwen/Qwen2.5-0.5B-Instruct-GGUF",
         "gguf_filename": "qwen2.5-0.5b-instruct-q4_0.gguf",
-        "dynamic_quantization_group_size": None,
+        # Compare with HF's floating-point activations, independently of the CPU's VNNI support.
+        "dynamic_quantization_group_size": "0",
```
apaniukov, line 36: Is it a regression? Should we document it?

tools/who_what_benchmark/tests/test_cli_text_gguf.py
```diff
@@ -0,0 +1,228 @@
...
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+# Minimum WWB text similarity (cosine over sentence embeddings) between the llama.cpp reference
+# and the GenAI-frontend target. Similarity falls with generated length even for a correct
+# conversion (measured for qwen3-0.6B: ~0.89 @ 16 tokens, ~0.85 @ 32, ~0.74 @ 100), so the
+# threshold is set to clear a correct conversion comfortably while failing hard on a broken one.
+SIMILARITY_THRESHOLD = 0.8
```
apaniukov, line 37: Is it connected to a `dynamic_quantization_group_size` parameter that is no set for some models in tests?

apaniukov (approved):

rkazants:
@mvafin, @apaniukov, please hold on with merging. I will have additional review asap

tests/python_tests/data/models.py
mvafin: This changes the reference-test configuration, not GenAI runtime defaults. The Qwen2.5 Q4_0 exact-token comparisons exposed sensitivity to CPU dynamic activation quantization when exercising the frontend; setting the group size to 0 aligns activations with the floating-point HF reference. The same fixture setting applies to both readers. This observation alone does not establish a regression in the CPU quantization implementation, and it is not a Qwen-only restriction.

Documented the accuracy/performance tradeoff in the public GGUF API documentation and expanded the fixture comment in 62f9ed16. The documentation explicitly notes that FP32 inference alone does not disable activation quantization and that matching precision settings does not guarantee identical output across readers/runtimes. Existing Qwen reference tests pass with the explicit setting; this follow-up only changes documentation/comments.

tools/who_what_benchmark/tests/test_cli_text_gguf.py
mvafin: Yes, leaving dynamic activation quantization at the CPU plugin default can contribute to the difference; INFERENCE_PRECISION_HINT=f32 does not disable it. The WWB test config is separate from tests/python_tests/data/models.py: cases marked reference_precision explicitly set DYNAMIC_QUANTIZATION_GROUP_SIZE=0 and FP16 KV cache, whereas the original cases retain their previous plugin settings.

It is not the only possible source of drift: weight decoding, KV-cache precision and floating-point arithmetic can also differ, and autoregressive decoding amplifies token differences. In 62f9ed16 I replaced the overly broad "correct conversion" explanation and unqualified historical scores with that distinction. The 0.8 threshold is described as an end-to-end regression check, not proof of conversion correctness; the threshold and test behavior are unchanged.

PR #4327, opened 2026-08-14
Support FunAudioLLM/Fun-ASR Model

as-suvorov:
## Description

`optimum-intel` pr: https://github.com/huggingface/optimum-intel/pull/1801

CVS-183086

#### Implementation Notes

- **Original FunASR:** an unset language uses the neutral `语音转写：` ("speech transcription: ") prompt; a supplied language uses `语音转写成{language}：` ("speech transcription in {language}: ").
- **Optimum Intel difference:** its preprocessing defaults an unset language to Chinese and uses `语音转写成中文：` ("speech transcription in Chinese: "). OpenVINO GenAI follows the original implementation and does not force a language by default.
- **Language result contract:** neutral FunASR transcription does not identify the spoken language. When no language is supplied, `ASRDecodedResults.languages` contains an empty string; when supplied, it contains the requested language. Automatic language detection remains model-dependent.

## Checklist:
- [x]  I have made corresponding changes to the documentation.


src/cpp/src/automatic_speech_recognition/models/fun-asr/pipeline.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+namespace ov::genai {
+
+FunASR::FunASR(const std::filesystem::path& models_path, const std::string& device, const ov::AnyMap& properties)
+    : ASRPipelineImplBase(models_path) {
+    ov::AnyMap properties_copy = properties;
+    erase_allowed_asr_ctor_properties(properties_copy);
+    m_encoder = std::make_unique<FunASREncoder>(models_path, device, properties_copy);
+    m_decoder = std::make_unique<Qwen3ASRDecoder>(models_path, device, properties_copy);
+    m_generation_config.set_eos_token_id(151645);
```
byekrang: Could you please check whether `151643` (`<|endoftext|>`) should also be included as a stop token here?

tools/llm_bench/llm_bench_utils/ov_utils.py
```diff
@@ -664,6 +668,8 @@ def create_speech_2_txt_model(model_path, device, memory_data_collector, **kwarg
 
     if use_case.model_type == "qwen3-asr":
         pipe = Qwen3ASROptimumPipeline(model=ov_model, processor=processor)
+    elif use_case.model_type == "fun-asr":
```
byekrang, line 671: Should "fun-asr" also be added to `UseCaseSpeech2Text` in `config_class.py`?

Without it, it seems that automatic model-type detection may not reach this branch. I also noticed that `speech_to_text_generation.py` would give Fun-ASR the Whisper default "<|en|>", which may not match Fun-ASR’s expected language format. Could you please check these two paths?

src/cpp/src/automatic_speech_recognition/models/fun-asr/pipeline.cpp
```diff
@@ -0,0 +1,143 @@
...
+                                                     streamer);
+    if (streamer) {
+        streamer->end();
+    }
+
+    const auto detokenization_start_time = std::chrono::steady_clock::now();
+    results.texts.push_back(m_tokenizer.decode(encoded_results.tokens[0]));
+    const auto detokenization_stop_time = std::chrono::steady_clock::now();
+    results.scores.push_back(encoded_results.scores[0]);
+    results.languages.push_back(config.language.value_or(""));
```
byekrang: The prompt appears to default to something else IIUC.
Would it be better to use the same default language here as during prompt construction? The prompt appears to default to "中文", while the result reports an empty language. Perhaps the effective language could be resolved once and reused in both places.

byekrang (approved):
Thank you for adding Fun-ASR support. The feature-extractor implementation and its tests look carefully done. I went through the integration and left a few questions inline about `llm_bench` registration, language reporting, and handling of the Qwen stop tokens.

I may have overlooked an existing job, but I could not find a CI configuration that installs the Fun-ASR dependencies and runs the new Python tests. Since the tests are skipped when `funasr` is unavailable, the existing ASR job filters for qwen3-asr, and the pinned optimum-intel revision predates Fun-ASR support, it looks like these tests may currently be skipped in CI.

Would it be possible to add a Fun-ASR job similar to the Qwen3-ASR one, update the optimum-intel dependency to an upstream revision that includes the recently merged Fun-ASR support, and confirm in the CI log that the tests run rather than skip?

I may have missed something in the surrounding CI setup, so please correct me if this is already handled elsewhere.

tools/llm_bench/llm_bench_utils/ov_utils.py
sbalandi: no, automatic model-type detection is outdated, it does not need to be updated now, the user must specify a --task in order to select the model-type

sbalandi: agree with point about `language`, looks like it should be another default value for this model

as-suvorov: `language` addressed

src/cpp/src/automatic_speech_recognition/models/fun-asr/pipeline.cpp
as-suvorov: token added, aligned with Qwen3-ASR

src/cpp/src/automatic_speech_recognition/models/fun-asr/pipeline.cpp
as-suvorov: Language resolved once. Fun-asr doesn't support language autodetection by default. GenAI documentation is updated to reflect this.

as-suvorov:
> Thank you for adding Fun-ASR support. The feature-extractor implementation and its tests look carefully done. I went through the integration and left a few questions inline about `llm_bench` registration, language reporting, and handling of the Qwen stop tokens.
> 
> I may have overlooked an existing job, but I could not find a CI configuration that installs the Fun-ASR dependencies and runs the new Python tests. Since the tests are skipped when `funasr` is unavailable, the existing ASR job filters for qwen3-asr, and the pinned optimum-intel revision predates Fun-ASR support, it looks like these tests may currently be skipped in CI.
> 
> Would it be possible to add a Fun-ASR job similar to the Qwen3-ASR one, update the optimum-intel dependency to an upstream revision that includes the recently merged Fun-ASR support, and confirm in the CI log that the tests run rather than skip?
> 
> I may have missed something in the surrounding CI setup, so please correct me if this is already handled elsewhere.

CI tests enabled.

avasenin-14 (approved):

tests/python_tests/utils/asr_utils/fun_asr.py
```diff
@@ -0,0 +1,29 @@
+import importlib.util
+
+import pytest
+
+
+# todo for tests: install pip install -U git+https://github.com/openvino-agent/optimum-intel.git@b841dde559e306a0535f55a5cd4432239ca18c09
```
Wovchena:
```suggestion
```
Seem to be completed

.github/workflows/test_matrices/linux/wheel_tests.yml
```diff
@@ -21,7 +21,17 @@
...
-    python -m pytest -s -v ./tests/python_tests/test_whisper_pipeline.py -k "qwen3-asr"
+    python -m pytest -v ./tests/python_tests/test_whisper_pipeline.py -k "qwen3-asr"
+  timeout: 20
+  components:
+    - automatic_speech_recognition
+
+- name: ASR (Fun-ASR)
+  cmd: |-
+    # optimum intel master commit with Fun-ASR support
+    python -m pip install --no-deps --force-reinstall "optimum-intel @ git+https://github.com/openvino-agent/optimum-intel.git@b841dde559e306a0535f55a5cd4432239ca18c09"
```
Wovchena: https://github.com/huggingface/optimum-intel/pull/1801 is merged. Move to `huggingface` upstream

byekrang (approved):

.github/workflows/test_matrices/linux/wheel_tests.yml
```diff
@@ -21,7 +21,17 @@
...
+    python -m pip install transformers==4.57.6 funasr==1.4.1
```
Wovchena, line 33: Why isn't `funasr` in `tests/python_tests/requirements.txt` and `samples/export-requirements.txt`?

src/cpp/src/automatic_speech_recognition/models/fun-asr/pipeline.cpp
```diff
@@ -0,0 +1,152 @@
...
+    std::fill_n(input_ids_data, num_audio_tokens, 0);
+    input_ids_data += num_audio_tokens;
+    std::memcpy(input_ids_data, suffix_ids.data<const int64_t>(), suffix_ids.get_byte_size());
+
+    return {input_ids, std::move(text_prompt.language)};
+}
+
+ASRGenerationConfig FunASR::resolve_generation_config(
+    const std::optional<ASRGenerationConfig>& generation_config) const {
+    ASRGenerationConfig config = generation_config.value_or(m_generation_config);
```
Wovchena, line 137:
```suggestion
    ASRGenerationConfig config = generation_config.value_or(m_generation_config);

    if (config.stop_token_ids.empty()) {
        config.stop_token_ids = m_generation_config.stop_token_ids;
    }
```
Taken from Qwen3ASR::resolve_generation_config()

Wovchena (changes requested):

tests/python_tests/utils/asr_utils/fun_asr.py
as-suvorov: Thanks for catch, I missed it. Removed

src/cpp/src/automatic_speech_recognition/models/fun-asr/pipeline.cpp
as-suvorov: Applied

.github/workflows/test_matrices/linux/wheel_tests.yml
as-suvorov: Moved to upstream

.github/workflows/test_matrices/linux/wheel_tests.yml
as-suvorov: `optimum-intel` states `transformers==4.57.6` version required. Our tests and samples has `transformers==5.0`. So export flow require additional steps. We have same flow with qwen3-asr.
Both qwen3-asr and fun-asr are going to be added or already added to huggingface transformers. This might simplify export for users. As I understand there are no plans to align `optimum-intel` to pure transformers path.

Wovchena (approved):

PR #3827, opened 2026-05-08
LFM2 and Qwen3-Next CB Pipeline Inference for Copilot Review

apaniukov:
## Description

CVS-183798
CVS-183799

## Checklist:
- [x] This PR follows [GenAI Contributing guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing). 
- [x] Tests have been updated or added to cover the new code. 
- [x] This PR fully addresses the ticket. 
- [x] I have made corresponding changes to the documentation.

PR #4339, opened 2026-08-18
[CB][VLM] Fix Coverity uninitialized reads and destructor throws

sgonorov:
## Description

Two Coverity fixes.

**CB destructors (CID 6465822, CID 6465814).** `GENAI_ERR` can throw, and an implicitly-`noexcept` destructor turns that into `std::terminate`. Now caught, with an `std::fprintf(stderr, ...)` fallback. Regression from #3827.

**`EncodedVideo` (CID 6465765, CID 6465769, CID 6923861, CID 6465768).** `num_video_tokens` and `frame_num` had no default initializers, so encoders copy one out indeterminate. `= 0` also covers `qwen3_vl/classes.cpp:397` and `qwen3_5`.

CID 6922826 and CID 6267146 are false positives (libstdc++ `std::variant` `_M_u`), triaged in Coverity.

CVS-192418

## Checklist:
- [ ] N/A Tests have been updated or added to cover the new code. Nothing observable changes: the destructor diagnostics only run during a teardown that has already leaked, and no consumer reads the values the initializers replace.
- [x] This PR fully addresses the ticket. CVS-192418 covers the two destructor CIDs. No Jira ticket is filed for the four `EncodedVideo` CIDs.

### DIFF
--- a/src/cpp/src/continuous_batching/cache/block_manager.hpp
+++ b/src/cpp/src/continuous_batching/cache/block_manager.hpp
@@ -10,6 +10,7 @@
 #include <set>
 #include <algorithm>
 #include <chrono>
+#include <cstdio>
 #include <limits>
 #include <utility>
 #include <vector>
@@ -268,9 +269,16 @@ public:
         for (auto& free_block : m_free_blocks_num) {
             const size_t free_and_overwritable_block_cnt = free_block + num_overwriteable_blocks();
             if (m_total_num_blocks != free_and_overwritable_block_cnt) {
-                GENAI_ERR("BlockAllocator leaked blocks. Expected num free blocks: %zu, actual: %zu",
-                          m_total_num_blocks,
-                          free_and_overwritable_block_cnt);
+                try {
+                    GENAI_ERR("BlockAllocator leaked blocks. Expected num free blocks: %zu, actual: %zu",
+                              m_total_num_blocks,
+                              free_and_overwritable_block_cnt);
+                } catch (...) {
+                    std::fprintf(stderr,
+                                 "BlockAllocator leaked blocks. Expected: %zu, actual: %zu\n",
+                                 m_total_num_blocks,
+                                 free_and_overwritable_block_cnt);
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/cpp/src/continuous_batching/cache/block_manager.hpp
```diff
@@ -264,14 +266,22 @@ class BlockAllocator {
...
+                    GENAI_ERR("BlockAllocator leaked blocks. Expected num free blocks: %zu, actual: %zu",
+                              m_total_num_blocks,
+                              free_and_overwritable_block_cnt);
+                }
             }
+        } catch (const std::exception& ex) {
+            std::fprintf(stderr, "BlockAllocator: failed to report leaked blocks: %s\n", ex.what());
+        } catch (...) {
+            std::fprintf(stderr, "BlockAllocator: failed to report leaked blocks: unknown error\n");
         }
```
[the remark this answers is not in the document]
sgonorov: Good catch. The `try` wrapped the whole loop, so the counts were out of scope in the handler. Moved it inside the `if` — the fallback now prints expected vs actual, matching what `~BlockManager` already did.

src/cpp/src/continuous_batching/cache/block_manager.hpp
```diff
@@ -268,9 +269,16 @@ class BlockAllocator {
         for (auto& free_block : m_free_blocks_num) {
             const size_t free_and_overwritable_block_cnt = free_block + num_overwriteable_blocks();
             if (m_total_num_blocks != free_and_overwritable_block_cnt) {
-                GENAI_ERR("BlockAllocator leaked blocks. Expected num free blocks: %zu, actual: %zu",
-                          m_total_num_blocks,
-                          free_and_overwritable_block_cnt);
```
popovaan, line 273: Will we not miss the memory leak if we remove this exception?

yatarkan (approved):

src/cpp/src/continuous_batching/cache/block_manager.hpp
as-suvorov, line 273: Is it clear what is the root cause of throw inside `GENAI_ERR` ?
Should we assert arguments instead?

as-suvorov: Assert was removed in @apaniukov PR: https://github.com/openvinotoolkit/openvino.genai/pull/3827

sgonorov: Yes. `Logger::log_format` rethrows as `std::runtime_error` (logger.cpp:56-64); causes: `vsnprintf < 0`, `bad_alloc`, stream/mutex state. Args are checked at compile time via `format(printf, 5, 6)`. An assert throws too, and this is a noexcept destructor, exactly the defect #3827 fixed here. Can make the logger noexcept instead if preferred.

PR #4341, opened 2026-08-18
Implement OmniPipeline preview

Retribution98:
## Description
- The Node.js bindings have been implemented for OmniPipeline.
- JS tests were added using tiny-random-qwen3-omni. This model has a problem by default, so a workaround has been added to patch this model
- The sample was not added to this PR because it could not be tested with Qwen3-Omni-30B-A3B-Instruct.
- The documentation has not been updated as this is a preview version.

CVS-190680

## Checklist:


Retribution98:
@copilot resolve the merge conflicts in this pull request

.github/workflows/linux.yml
```diff
@@ -944,6 +944,9 @@ jobs:
           requirements_files: "${{ env.SRC_DIR }}/tests/python_tests/requirements.txt"
           local_wheel_dir: ${{ env.PY_INSTALL_DIR }}/wheels
 
+      - name: Install latest Optimum Intel
+        run: python -m pip install --upgrade "optimum-intel @ git+https://github.com/huggingface/optimum-intel.git"
```
almilosz, line 948: Is latest `optimum_intel` necessary also for python api? 
I share the copilot concern -> this will make the workflow undeterministic

src/js/lib/addon.ts
```diff
@@ -252,6 +258,42 @@ export interface VLMPipeline {
   getGenerationConfig(): GenerationConfig;
 }
 
+/**
+ * Native binding interface used by the high-level Omni pipeline wrapper.
+ *
+ * @note This is a preview API and is subject to change in future releases.
```
almilosz, line 264: Shouldn't this note be in `index.ts`? `addon.ts` is just our intermediate layer so we can change it by default

src/js/lib/index.ts
```diff
@@ -1,8 +1,9 @@
-// Copyright (C) 2025-2026 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
```
almilosz, line 1:
```suggestion
// Copyright (C) 2025-2026 Intel Corporation
```
what is our policy here?
Some other files have 2025-2026

src/js/tests/setup.py
```diff
@@ -61,6 +67,62 @@
     },
 }
 
+
+def _patch_tiny_qwen3_omni(model_path: Path, hf_tokenizer) -> None:
```
almilosz, line 71: Do python api unit test use `Qwen3-Omni-30B-A3B-Instruct`?

src/js/tests/omniPipeline.test.js
```diff
@@ -0,0 +1,132 @@
...
+  it("rejects when the model path does not exist", async () => {
+    await assert.rejects(OmniPipeline("./nonexistent-omni-model", "CPU"));
+  });
+});
+
+// Functional tests require a Qwen3-Omni model exported with audio output. Speech output
+// uses the continuous-batching backend, which OmniPipeline enables by default. When
+// OMNI_PATH is not provided the suite is skipped so the shared JS test run is not blocked
+// by the large checkpoint.
+describe("OmniPipeline", { skip: !OMNI_PATH || process.platform === "darwin" }, () => {
```
almilosz, line 46:
```suggestion
describe("OmniPipeline", {  process.platform === "darwin" }, () => {
```
Not sure why this skip is needed if path to tiny qwen is provided in _setup.py_
Plus, what is the issue with darwin? Does the model patch not works on darwin? 
If sth is skipped, I would add a ticket number to track

src/js/tests/omniPipeline.test.js
```diff
@@ -0,0 +1,132 @@
...
+  return_audio: true,
+  max_new_tokens: 4,
+  talker_top_k: 1,
+  cp_top_k: 1,
+};
+
+// Smoke tests: exercise the binding wiring without loading a multi-GB Qwen3-Omni
+// checkpoint. Generation that requires a loaded pipeline lives in the model-gated
+// suite below.
+describe("OmniPipeline smoke", { skip: process.platform === "darwin" }, () => {
```
almilosz, line 28:
```suggestion
describe("OmniPipeline smoke", () => {
```
? it's just api check

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+        auto inputs = js_to_cpp<VLMGenerateInputs>(env, info[0]);
+        auto images = js_to_cpp<std::vector<ov::Tensor>>(env, info[1]);
+        auto videos = js_to_cpp<std::vector<ov::Tensor>>(env, info[2]);
+        auto audios = js_to_cpp<std::vector<ov::Tensor>>(env, info[3]);
+        auto streamer = info[4];
+        OPENVINO_ASSERT(streamer.IsFunction() || streamer.IsUndefined(), "streamer must be a function or undefined");
+        auto speech_streamer = info[5];
+        OPENVINO_ASSERT(speech_streamer.IsFunction() || speech_streamer.IsUndefined(),
+                        "speechStreamer must be a function or undefined");
+        auto text_config = js_to_cpp<ov::genai::GenerationConfig>(env, info[6]);
```
sgonorov, line 245: `js_to_cpp` starts from a default config, so JS ignores `generation_config.json` and samples differently from Python. I think we can route through the `AnyMap` overload which already merges it.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+        Napi::Error::New(env, ex.what()).ThrowAsJavaScriptException();
+    }
+    return env.Undefined();
+}
+
+Napi::Value OmniPipelineWrapper::generate(const Napi::CallbackInfo& info) {
+    auto env = info.Env();
+    try {
+        OPENVINO_ASSERT(this->pipe, "OmniPipeline is not initialized");
+        OPENVINO_ASSERT(!*this->is_generating, "Another generation is already in progress");
```
sgonorov, line 230: The catch clears `is_generating` even when the throw was this guard, so a later call can enter `generate` concurently. I'd claim the flag with `compare_exchange_strong`.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+    try {
+        if (!streamer_exceptions.empty()) {
+            std::string combined_error = "Streamer exceptions occurred:\n";
+            for (size_t i = 0; i < streamer_exceptions.size(); ++i) {
+                combined_error += "[" + std::to_string(i + 1) + "] " + streamer_exceptions[i] + "\n";
+            }
+            report_error(combined_error);
+        } else {
+            napi_status status =
+                context->callback.BlockingCall([result, &report_error](Napi::Env env, Napi::Function jsCallback) {
```
sgonorov, line 170: `BlockingCall` returns immediately with an unbounded queue, so the frame owning `report_error` dies before the callback runs. It's good to copy the whisper wrapper and use a `shared_ptr`.

.github/workflows/linux.yml
```diff
@@ -944,6 +944,9 @@ jobs:
           requirements_files: "${{ env.SRC_DIR }}/tests/python_tests/requirements.txt"
           local_wheel_dir: ${{ env.PY_INSTALL_DIR }}/wheels
 
+      - name: Install latest Optimum Intel
```
sgonorov, line 947: This upgrades past the pin in `requirements.txt` for all four platforms. Might be worth mirroring `linux.yml:871`, which uses `--no-deps` with a fixed SHA.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+                                resultPromise.set_value(ov::genai::StreamingStatus::RUNNING);
+                            }
+                        } catch (const std::exception& err) {
+                            streamer_exceptions.push_back(err.what());
+                            resultPromise.set_value(ov::genai::StreamingStatus::CANCEL);
+                        }
+                    });
+
+                if (status != napi_ok) {
+                    streamer_exceptions.push_back("The streamer callback BlockingCall failed with the status: " +
```
sgonorov, line 85: `napi_status` promotes to int, so this offsets the literal instead of concatenating and the message comes out truncated. We could use `std::to_string(static_cast<int>(status))`, same at 116 and 182.

src/js/src/omni_pipeline/init_worker.cpp
```diff
@@ -0,0 +1,32 @@
...
+                               ov::AnyMap properties)
+    : AsyncWorker(callback),
+      pipe(pipe),
+      is_initializing(is_initializing),
+      model_path(std::move(model_path)),
+      device(std::move(device)),
+      properties(std::move(properties)) {};
+
+void OmniInitWorker::Execute() {
+    *this->is_initializing = true;
```
sgonorov, line 20: Setting this in `Execute()` happens after `init()` already returned, so two back to back `init()` calls both pass the guard. Probably better to set it before queueing the worker.

src/js/tests/setup.py
```diff
@@ -79,6 +141,8 @@
     for model_name, model_info in TEST_MODELS.items():
         try:
             result = hugging_face.download_and_convert_model_class(**model_info)
+            if model_name == "OMNI":
+                _patch_tiny_qwen3_omni(result.models_path, result.hf_tokenizer)
```
sgonorov, line 145: This writes into the dir after `AtomicDownloadManager` marked it complete, so a cancelled job leaves a half patched model that `is_complete()` still trusts. Maybe move it inside teh conversion callback.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+                    streamer_exceptions.push_back(
+                        "The speech streamer callback BlockingCall failed with the status: " + status);
+                    return ov::genai::StreamingStatus::CANCEL;
+                }
+
+                return resultPromise.get_future().get();
+            };
+        }
+
+        const std::vector<ov::genai::VideoMetadata> videos_metadata;
```
sgonorov, line 125: Always empty means JS video input cannot carry `fps` or `frames_indices`, and nothing warns the user. I'd add a `videosMetadata` option or note the limit in the JSDoc.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+                                          callback,                     // JavaScript function called asynchronously
+                                          "Omni_generate_callback",     // Name
+                                          0,                            // Unlimited queue
+                                          1,                            // Only one thread will use this initially
+                                          [context, this](Napi::Env) {  // Finalizer used to clean threads up
+                                              context->native_thread.join();
+                                              delete context;
+                                          });
+        if (!streamer.IsUndefined()) {
+            context->streamer = Napi::ThreadSafeFunction::New(env,
```
sgonorov, line 269: If this or the thread constructor throws, `context` leaks and the callback TSFN is never released, so the loop never drains. I suggest to hold it in a `unique_ptr`.

src/js/tests/omniPipeline.test.js
```diff
@@ -0,0 +1,132 @@
...
+        max_new_tokens: 4,
+        ignore_eos: true,
+        structured_output_config: { regex: "a+" },
+      },
+      talkerSpeechConfig: deterministicSpeechConfig,
+      streamer: () => {
+        textChunks++;
+        return StreamingStatus.RUNNING;
+      },
+      speechStreamer: () => {
```
sgonorov, line 112: Every streamer here returns `RUNNING`, so STOP, CANCEL and the exception aggregation path never run. Might be worth a `STOP` case and one with two concurrent `generate()` calls.

src/js/lib/index.ts
sgonorov, line 1: The file did not exist in 2023 and nothing else here needs this, so I'd revert the line to 2025-2026.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+
+#include "include/addon.hpp"
+#include "include/helper.hpp"
+#include "include/omni_pipeline/init_worker.hpp"
+
+struct OmniTsfnContext {
+    OmniTsfnContext(VLMGenerateInputs inputs, std::shared_ptr<std::atomic<bool>> is_generating)
+        : inputs(std::move(inputs)),
+          is_generating(is_generating) {};
+    ~OmniTsfnContext() {};
```
sgonorov, line 16: The empty destructor suppresses implicit moves for no gain, and the the finalizer captures `this` without using it. I think we can drop both, plus the stray semicolons.

src/js/tests/setup.py
```diff
@@ -35,6 +37,10 @@
         "model_id": "optimum-intel-internal-testing/tiny-random-qwen2vl",
         "model_class": OVModelForVisualCausalLM,
     },
+    "OMNI": {
```
sgonorov, line 40: Adding OMNI to this loop makes the most fragile entry fatal, since `raise` takes down the LLM, VLM and Whisper models with it. Probably better to let it fail soft.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
```diff
@@ -0,0 +1,288 @@
...
+    std::vector<ov::Tensor> images;
+    std::vector<ov::Tensor> videos;
+    std::vector<ov::Tensor> audios;
+    ov::genai::GenerationConfig text_config;
+    ov::genai::OmniTalkerSpeechConfig talker_speech_config;
+    std::shared_ptr<std::atomic<bool>> is_generating;
+    std::shared_ptr<ov::genai::OmniPipeline> pipe = nullptr;
+};
+
+void omniPerformInferenceThread(OmniTsfnContext* context) {
```
sgonorov, line 33: This is nearly a verbatim copy of `vlmPerformInferenceThread`, which is why the same bugs apply to both. Not for this PR, but worth extracting at some point.

src/js/tests/omniPipeline.test.js
sgonorov, line 28: The skip is right, but `TEST_MODELS` is unconditional, so macOS still downloads and converts this checkpoint and then skips every test using it. We could gate the entry by platform.

src/js/src/helper.cpp
```diff
@@ -1736,6 +1784,33 @@ Napi::Object to_vlm_decoded_result(const Napi::Env& env, const ov::genai::VLMDec
...
+    obj.Set("parsed", cpp_to_js<std::vector<ov::genai::JsonContainer>, Napi::Value>(env, results.parsed));
+    obj.Set("finishReasons",
+            cpp_to_js<std::vector<ov::genai::GenerationFinishReason>, Napi::Value>(env, results.finish_reasons));
+
+    const ov::genai::TalkerResults& speech_result = results.speech_result;
+    Napi::Array waveforms = Napi::Array::New(env, speech_result.waveforms.size());
+    for (size_t i = 0; i < speech_result.waveforms.size(); ++i) {
+        waveforms[i] = cpp_to_js<ov::Tensor, Napi::Value>(env, speech_result.waveforms[i]);
+    }
+    Napi::Object speech_perf_metrics = Napi::Object::New(env);
```
sgonorov, line 1801: The text metrics at 1791 go through a wrapper, so this one gets no methods and will drift from `TalkerPerfMetrics` if fields are added. Fine for a preview.

sgonorov (changes requested):

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
Retribution98: No, it doesn't work. The AI assistant is wrong here because you get a number code either way.

.github/workflows/linux.yml
Retribution98: I pinned the same commit that is used in the Python tests.

src/js/lib/addon.ts
Retribution98: yes, it could be excess

src/js/tests/omniPipeline.test.js
Retribution98: We only test Darwin with ARM, but it doesn't support the continuous batching backend. I added a comment near the test. This is a limitation of the core, so I didn't create an issue for it.

src/js/tests/omniPipeline.test.js
Retribution98: Yes, it was the first version and I removed it.

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
Retribution98: I tried, but ran into problems converting the speech streamer via any map. We usually use an implementation with explicit parameters to be more rigorous semantically and avoid hidden issues. Could you please explain why different implementations behave differently when a parameter value is empty?

src/js/src/omni_pipeline/omni_pipeline_wrapper.cpp
Retribution98: I added the videosMetadata

PR #4348, opened 2026-08-19
[CI] Make test content download atomic to avoid corrupted shared cache

goyaladitya05:
## Description

CVS-193070

`test_cpp_tool_benchmark` (Tools tests, Windows) failed on every PR on 2026-08-19 with: `[json.exception.parse_error.101] parse error at line 1594473, column 491: ... missing closing quote`

#### Cause

download_test_content in `tests/python_tests/samples/conftest.py` streams the download directly to the final cache path. If the transfer is interrupted, a truncated file is left behind, and the `os.path.exists()` check treats it as complete on all subsequent runs.

That is what happened in [run 32210833659, attempt 1](https://github.com/openvinotoolkit/openvino.genai/actions/runs/32210833659/attempts/1) (Tools tests, 03:55 UTC, the first job to populate the day's `.ov_cache/win/.../20260819` folder). The ShareGPT download died mid-stream: `urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(327238867 bytes read, 345599075 more expected)')`

327,238,867 + 345,599,075 = 672,837,942 bytes, which is the file's Content-Length on hf. The truncated 327MB file stayed at the final path. Since the OV_CACHE day folder is shared by all Windows runners, every later job that day (e.g. [32236261371](https://github.com/openvinotoolkit/openvino.genai/actions/runs/32236261371), [32218248194](https://github.com/openvinotoolkit/openvino.genai/actions/runs/32218248194), [32242846268](https://github.com/openvinotoolkit/openvino.genai/actions/runs/32242846268/job/96043848909), [32246735815](https://github.com/openvinotoolkit/openvino.genai/actions/runs/32246735815/job/96052993722?pr=4247)) logged `Test content already exists` and failed at the identical byte offset.

This went unnoticed until now because only the first Tools job of each day downloads the file, and the date-partitioned cache self-clears at midnight.

#### Fix

Download to a unique temp file and `os.replace()` it into place only after the stream completes; remove the temp file on failure. This is the same temp-then-rename pattern `AtomicDownloadManager` already uses for model downloads in this file. An interrupted download now leaves nothing behind, so the next session retries instead of consuming a partial file.

## Checklist:

### DIFF
--- a/tests/python_tests/samples/conftest.py
+++ b/tests/python_tests/samples/conftest.py
@@ -7,6 +7,7 @@ import json
 import pytest
 import shutil
 import logging
+import uuid
 import requests
 from importlib import metadata
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


likholat:
build_jenkins

goyaladitya05:
@sgonorov @as-suvorov Could you please take a look? thanks!

sgonorov (approved):

PR #4349, opened 2026-08-19
llm_bench: timestamp-based marker assignment for process memory monitor

krzyczar:
The process memory monitor sent bare marker strings over the queue and the background sampler tagged every sample with whatever marker it had read last, applying a phase change whenever the loop next drained the queue rather than when the phase actually changed. Samples straddling a transition were mislabeled, and intermediate markers arriving between two polls were dropped (check_for_markers kept only the latest).

Now each marker carries a wall-clock timestamp (time.time_ns(), comparable across processes with the sampler's own timestamps). The sampler keeps a marker history and assigns each sample the marker whose interval covers its timestamp.

Markers arrive in non-decreasing timestamp order, so once the latest marker has timestamp T_m no future marker can affect a sample with ts < T_m. Such samples are finalized and flushed; samples at/after this watermark stay buffered until the next marker advances it (or "stop" finalizes all). This removes any need to rewrite already-dumped files, and lets per-marker statistics be aggregated at flush time — when a sample's label is final — instead of at collection time (which made retroactive min/max correction impossible).

- update_marker sends (marker, time.time_ns())
- check_for_markers drains the whole queue into the marker history
- collect_samples buffers raw (ts, *vals); marker_for() resolves labels
- write_chunk flushes only the finalized prefix and aggregates there
- samplers gain read_raw() split out from collect()

## Description

CVS-192947

Fixes CVS-192947

## Checklist:

### DIFF
--- a/tools/llm_bench/llm_bench_utils/memory_monitor.py
+++ b/tools/llm_bench/llm_bench_utils/memory_monitor.py
@@ -9,6 +9,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import atexit
+import bisect
 import queue
 import threading
 import multiprocessing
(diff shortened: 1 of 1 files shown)


sbalandi (approved):

sbalandi:
build_jenkins

avasenin-14 (approved):

PR #4354, opened 2026-08-20
[llm_bench] Fixed running the video_gen task with prompt from cmd and alignment def parameters

sbalandi:
## Description
cmd:
python benchmark.py -d CPU -n 3  -mc 0 -m /home/labuser/work/mnt_13/WW33_llm-optimum_2026.4.0-22741/ltx-video/pytorch/ov/FP16/ --task text-to-video 
cause to error since num_inference_steps is not define, fixed in this PR.
+ set default parameters so they are clear and the same for both backends.

## Checklist:


avasenin-14 (approved):
LGTM, pls check if copilot's comment make sense?

PR #4361, opened 2026-08-21
[tools] Upgrade transformers version in tools to 5.5

avasenin-14:
## Description

[CVS-191273](https://jira.devtools.intel.com/browse/CVS-191273)
[CVS-193005](https://jira.devtools.intel.com/browse/CVS-193005)
[CVS-193006](https://jira.devtools.intel.com/browse/CVS-193006)

Pip resolves optimum-intel >=1.25.0 to 1.27.0 by default, which requires optimum-onnx~=0.1.0, which requires transformers<4.58.0.
With optimum-intel 2.1.0 default transformers installed in tools is 5.5.4 now.

## Checklist:
- N/A  Tests have been updated or added to cover the new code. 
- N/A  I have made corresponding changes to the documentation.


avasenin-14:
Moving to 'Ready for review' to start tests.

tools/llm_bench/requirements.txt
```diff
@@ -10,8 +10,7 @@ torch>=2.1.0,<=2.12
 transformers[sentencepiece]>=4.40.0,<=5.13.0
 diffusers>=0.22.0,<=0.39.0
 # optimum is in dependency list of optimum-intel
-# last supported commit f50a0b326df82d4b05c461a9184a97f616bfa63a
-optimum-intel[nncf]>=1.25.0
+optimum-intel[nncf]>=2.1.0
```
sbalandi, line 13: please, keep the whole range for tools

tools/who_what_benchmark/setup.py
```diff
@@ -52,15 +52,15 @@ def set_version(base_version: str):
             "torchvision==0.28.0",
             "torchaudio==2.11.0",
             "transformers>=4.50,<=4.51.3",
-            "optimum-intel==1.27.0",
+            "optimum-intel==2.1.0",
```
sbalandi: is minicpm-o-2_6 supported by optimum-intel 2.0 ?

tools/llm_bench/requirements.txt
avasenin-14: I need to think how to do it. Pip resolves optimum-intel with the first match that satisfies deps, and it is 1.27.0. I can't pin lower transformers also, since I think we need earlier versions for some models...

avasenin-14: @sbalandi , maybe you have some thoughts on how to better do it? I'm not sure about it myself

What I've tried:
- bumping lower bound of transformers to 5.5 - decided against it, as I think we need lower transformers for minicpm, nanollava and videochat.
- bumping lower version of optimum-intel, like in this PR. the downside is that older versions of optimum won't be compatible with WWB.
- and I don't think I can just add additional requirements file, as people do want to install the WWB as a package, and it uses some requirements file by default in install_requires at the same time as having extras like minicpm that should be compatible with this file...

avasenin-14:
The pull request is not needed anymore. I've checked that Optimum-Intel release 2.2.0 now installed by default in WWB and LLM benchmarks, and with it transformers 5.5.4:

$ pip show transformers
Name: transformers
Version: 5.5.4

PR #4372, opened 2026-08-25
[WWB] Enable video input for Gemma4

yatarkan:
## Description
This PR enables video input for Gemma4 in WWB.
optimum-intel PR - https://github.com/huggingface/optimum-intel/pull/1951

CVS-190256

## Checklist:
- N/A - Tests have been updated or added to cover the new code. 
- N/A - This PR fully addresses the ticket. 
- N/A - I have made corresponding changes to the documentation.


yatarkan:
@avasenin-14 @sbalandi Could you please review from other tools and tests perspective, if anything else should be updated?

Also, running WWB for Gemma4 with video input requires setting `--video-frames-num` (e.g. to 32) as default `DEF_VIDEO_FRAMES_AMOUNT = 10` is too small for Gemma4 video preprocessing and results in error.

tools/who_what_benchmark/whowhatbench/inputs_preprocessors/gemma4.py
```diff
@@ -118,6 +124,55 @@ def align_inputs_with_cache(self, model: Any, inputs: dict, full_tokenized_chat:
...
+            conversation = [{"role": "user", "content": content}]
+
+        text_prompt = processor.apply_chat_template(conversation, add_generation_prompt=True, tokenize=False)
+
+        # switch off add_bos_token if chat template already includes it
+        orig_add_bos_token = processor.tokenizer.add_bos_token
+        if getattr(processor.tokenizer, "chat_template", None) and "bos_token" in processor.tokenizer.chat_template:
+            processor.tokenizer.add_bos_token = False
+
+        inputs = processor(images=self.images, videos=video, text=text_prompt, return_tensors="pt")
```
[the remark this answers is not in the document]
yatarkan: @sbalandi @avasenin-14 Do we need a separate method `update_videos` similar to `self.update_images(image)`? Or it is already planned or WIP?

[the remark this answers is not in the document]
sbalandi: @yatarkan yes, please, add it here, at the moment there are no clear plans whether and when it will be done, so it is better(and faster) to add it here

sbalandi:
> @avasenin-14 @sbalandi Could you please review from other tools and tests perspective, if anything else should be updated?
> 
> Also, running WWB for Gemma4 with video input requires setting `--video-frames-num` (e.g. to 32) as default `DEF_VIDEO_FRAMES_AMOUNT = 10` is too small for Gemma4 video preprocessing and results in error.

i think gemma's specific def number can be introduced for this model, smth like DEF_VIDEO_FRAMES_AMOUNT_GEMMA (https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py#L55)

tools/who_what_benchmark/whowhatbench/inputs_preprocessors/gemma4.py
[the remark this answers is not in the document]
sbalandi: thought about it again, and now I think there's no need to add it, since we don't have the infrastructure to run chat with video. so it will be a separate task, apologize for the confusion

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
```diff
@@ -52,7 +54,7 @@ def __init__(
         self.processor = processor
         self.is_image_input = task_type == "visual-text"
         self.is_text_only = task_type == "visual-text-only"
-        self.frames_num = frames_num or DEF_VIDEO_FRAMES_AMOUNT
+        self.frames_num = self._resolve_frames_num(frames_num, config)
```
yatarkan, line 57: Add frames num resolving based on model config so Gemma4 has specific frames num.
@sbalandi Could you please review agan?

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
```diff
@@ -32,6 +33,7 @@ def __init__(
         base_model: Any = None,
         tokenizer: Any = None,
         processor: Any = None,
+        config: Any = None,
```
sbalandi, line 36: no need to pass config here, base_model has it, so you can use `base_model.config.model_type`
for genai wwb has wrapper https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/who_what_benchmark/whowhatbench/model_loaders.py#L143

yatarkan: In `VisualTextEvaluator` constructor only base model exists. When running wwb with GT and `--genai`, base model is `None`. Model type is not defined here yet for this use case, but we need it to resolve `frames_num`, that is why I added config to constructor.

Another option instead of providing config is to resolve `frames_num` outside the `VisualTextEvaluator` constructor, e.g. in `score()` -> `_generate_data()` method where `GenAIModelWrapper` with defined model type is available.

sbalandi (approved):

tools/who_what_benchmark/whowhatbench/visualtext_evaluator.py
sbalandi: yes, you are right, it's okay to keep the config then, thank you

PR #4382, opened 2026-08-26
llm_bench: Fix exception when --speech_language isn't specified for Kokoro

RyanMetcalfeInt8:
## Description

This PR fixes an issue when running Kokoro model without `--speech_langauage` for GenAI. In that case, the language was getting set to empty string "". 

CVS-193147

## Checklist:
- [N/A]  Tests have been updated or added to cover the new code. 
- [X]  This PR fully addresses the ticket. 
- [N/A]  I have made corresponding changes to the documentation.

### DIFF
--- a/tools/llm_bench/task/text_to_speech_generation.py
+++ b/tools/llm_bench/task/text_to_speech_generation.py
@@ -176,7 +176,8 @@ def run_text_to_speech_generation_genai(
     )
 
     if is_kokoro_model:
-        additional_args["language"] = args.get("speech_language", "")
+        if (language := args.get("speech_language")) is not None:
+            additional_args["language"] = language
 
     start = time.perf_counter()
     generation_result = model.generate(input_text_list, **additional_args)
--- a/tools/llm_bench/task/text_to_speech_generation.py
+++ b/tools/llm_bench/task/text_to_speech_generation.py
@@ -176,8 +176,7 @@ def run_text_to_speech_generation_genai(
     )
 
     if is_kokoro_model:
-        if (language := args.get("speech_language")) is not None:
-            additional_args["language"] = language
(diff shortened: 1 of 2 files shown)


Maxim-Doronin (approved):

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -176,7 +176,8 @@ def run_text_to_speech_generation_genai(
     )
 
     if is_kokoro_model:
-        additional_args["language"] = args.get("speech_language", "")
+        if (language := args.get("speech_language")) is not None:
```
sbalandi: shouldn't it be corrected here as well https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/llm_bench_utils/tts_utils.py#L124 ?

RyanMetcalfeInt8: Thanks, I changed my approach with latest commit. Optimum didn't have this issue (this is why tts_utils.py didn't need patching) -- but ultimately because `normalize_kokoro_lang_code` is called in that path, which handled null strings:
```
    if not isinstance(language, str) or language.strip() == "":
        return "a"
```

So I think the right solution here is to do the same in GenAI path.

tools/llm_bench/task/text_to_speech_generation.py
```diff
@@ -176,7 +176,8 @@ def run_text_to_speech_generation_genai(
...
+            additional_args["language"] = language
```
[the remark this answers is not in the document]
RyanMetcalfeInt8: With latest changes, `normalize_kokoro_lang_code` is called, which checks `language.strip() == ""`

sbalandi (approved):

Issue #4368, opened 2026-08-25
VLMPipeline pad-filled prompt_ids vs. the reasoning-budget prompt scan — does not reproduce


blairducrayoppat:
### Summary

On `VLMPipeline`, the `prompt_ids` tensor handed to the continuous-batching sampler does not contain the request's prompt tokens. It is filled with `pad_token_id`, and only the tokenized KV-cache history is copied in; the prompt itself reaches the model as embeddings and never reaches the sampler as token ids. Any sampler-side component that inspects prompt content therefore sees padding on the VLM path.

This was found during hardware verification of PR #4139 (reasoning budget). @PlanteAmigor suggested tracking it as a separate issue, since the fix touches how `visual_language/pipeline.cpp` plumbs prompt tokens to the sampler and is beyond that PR's focused scope: https://github.com/openvinotoolkit/openvino.genai/pull/4139#issuecomment-5377990007

### Mechanism

At current master `a088679a`, `src/cpp/src/visual_language/pipeline.cpp` lines 892-895:

```cpp
ov::Tensor prompt_ids(ov::element::i64, { history_size + inputs_embeds_size });
OPENVINO_ASSERT(prompt_ids.get_size() >= tokenized_history.size(), "Prompt ids size is less than tokenized history size");
std::fill_n(prompt_ids.data<int64_t>(), prompt_ids.get_size(), m_tokenizer.get_pad_token_id());
std::copy(tokenized_history.begin(), tokenized_history.end(), prompt_ids.data<int64_t>());
```

### Observed consequence (PR #4139 branch, runtime verified)

`ThinkingBudgetTransform`'s constructor reverse-scans `prompt_ids` to decide its initial state (whether the prompt ends inside an open `<think>` block). On `VLMPipeline` the scan sees pads, stays IDLE, and mid-generation budget forcing never engages:

- LLM control (Qwen3-0.6B, raw prompt ending `<think>\n`, `apply_chat_template=false`, budget 12, greedy): forcing fires, `</think>` emitted roughly 12 tokens in.
- VLM (Qwen3-VL-8B-Instruct int4, identical prompt shape and settings): forcing never fires; reproduced twice with byte-identical output.
- Thinking-trained VLM (fresh Qwen3.5-0.8B export, think-token ids 248068/248069, single tokens through its OpenVINO tokenizer): the model visibly reasons past budget 12 and enforcement never engages.
[...]

PR #4139, opened 2026-07-11
Add ReasoningConfig to GenerationConfig for thinking suppression and budget control

PlanteAmigor:
### Motivation
Qwen3/3.5/3.6/3.8, DeepSeek and other reasoning models output `<think>`...`</think>` blocks by default, which adds latency and token cost. There is currently no way to disable or limit this reasoning behavior in OpenVINO GenAI (see issues #3937, #3946). Qwen's own serving platform exposes the same control (`thinking_budget`), so this PR aligns `openvino.genai` with the first-party serving API.

### Changes
This PR adds thinking suppression / reasoning budget control via a new `ReasoningConfig`:

**`ReasoningConfig`** (`generation_config.hpp`, `generation_config.cpp`)
- New struct `ReasoningConfig { budget, start_token_id, end_token_id }`, stored as `optional<ReasoningConfig>` on `GenerationConfig`
- `budget`: max tokens allowed in the thinking block before forcing `</think>` — `0` = disable thinking (force immediately), `N` = limit, `-1` = unlimited (no enforcement)
- No separate `enable_thinking` needed; `budget=0` covers it
- JSON/AnyMap parsing and validation

**ThinkingBudgetTransform** (`logit_transformers.hpp`)
- State machine: `IDLE → COUNTING → FORCING → DONE`
- Forces `end_token_id` when the budget is exhausted or `budget=0`
- Constructor reverse prompt-scan initializes state for Qwen 3.5/3.6/3.8 templates (which pre-fill `<think>\n`)
- Supports multi-think-block re-arm (DONE → COUNTING)

**LogitProcessor** (`logit_processor.hpp`)
- Integrates ThinkingBudgetTransform into the generation loop

**Python bindings** (`py_generation_config.cpp`, `py_llm_pipeline.cpp`, `py_vlm_pipeline.cpp`)
- Exposes `GenerationConfig.reasoning_config` and `ReasoningConfig { budget, start_token_id, end_token_id }` (with value equality)
- Auto-detect: when `reasoning_config` is set without token IDs, the bindings tokenize `<think>`/`</think>` and fill them in automatically

### Testing
Verified with Qwen3-14B, Qwen3-30B-MoE, DeepSeek-7B (LLM) and Qwen3-8B-VLM, Qwen3.5-4B-VLM, Qwen3.6-27B-VLM (VLM). `budget=0` suppresses `<think>` blocks in all models; `budget=N` force-closes at the budget point under greedy decoding.

### Related Issues
- #3937
- #3946

PR #4392, opened 2026-08-30
Fix VLM sampler prompt token propagation in continuous batching

MaxxxDong:
## Description

Propagates the prompt IDs produced by `InputsEmbedder` alongside VLM `inputs_embeds` on the continuous-batching path, so prompt-aware sampling sees the same chat-templated, multimodal-expanded, and pruned prompt that produced the embeddings.

- captures one cache delta per request for string, `ChatHistory`, and `add_request` entry points
- removes the separate raw-string tokenization previously used by prompt lookup, avoiding duplicate or misaligned entries for multi-prompt requests
- preserves internal negative visual-row markers for embedding/logprob alignment while exposing only real vocabulary IDs to repetition/no-repeat, prompt lookup, echo, full-token output, and speculative sampler setup
- adds C++ coverage through `SequenceGroup` and `LogitProcessor`, plus public string/`ChatHistory`/`add_request` echo regressions in the existing VLM Python fixture

The stateful SDPA path is intentionally unchanged: on that path the prompt tensor also participates in scheduling, and safely separating scheduling rows from sampler semantics requires broader validation than this bounded continuous-batching fix. The change is based directly on `master` and does not depend on #4139.

Related to #4368

## Validation

- clean Windows Release build with MSVC 19.44, CMake/Ninja, and `/utf-8`: 1264/1264 targets
- `tests_continuous_batching.exe --gtest_filter=VLMPromptIdsTest.*`: 4/4 passed
- locally built `py_openvino_genai` extension import: passed
- `pre-commit run --files ...`: passed for all changed files
- `git diff --check upstream/master`: passed

The public Python regression uses the repository's existing VLM model fixture and was not executed locally because no standalone cached VLM model was available outside an explicitly out-of-scope repository. It is intended for the upstream model-backed test matrix; no model was downloaded for this change.

## AI assistance disclosure

AI assistance used: yes.

How: Codex assisted with source-path analysis, implementation drafting, regression-test drafting, code review, and validation orchestration.

Contributor-side validation: the relevant VLM, `InputsEmbedder`, `SequenceGroup`, sampler, prompt-lookup, cache-trimming, CDPruner, and Omni paths were inspected; the complete final diff was reviewed; and the validation commands above were run.

## Checklist:

- [x] I have read the [contributing guide](https://github.com/openvinotoolkit/openvino.genai/blob/master/CONTRIBUTING.md)
- [x] Tests were added for the fixed behavior
- [x] The change covers the reproduced continuous-batching VLM entry points
- [x] No public API or separate tokenizer path was added


blairducrayoppat:
I filed #4368 and have since retracted its behavioural report: the symptom does not reproduce at #4139 head `a5fc4d5f` on my hardware. That retraction concerns my own measurements and does not bear on this PR.

The description notes that `test_vlm_prompt_ids_reach_sampler_for_chat_history_and_add_request` was not executed locally for want of a model. I have a working VLM bench on this machine (Intel Core Ultra 7 258V, Arc 140V, Windows 11) and can run that test against a real VLM export on CPU and on the GPU, and post the output here.

---

*AI assistance disclosure: this comment was drafted with AI assistance (Claude Code), which checked the tests in this PR and the current state of #4368. I reviewed and approved this text before posting. The offer above is for runs on my own hardware; I have not built this branch.*

MaxxxDong:
Thanks for clarifying #4368 — if convenient, a CPU/GPU run of `test_vlm_prompt_ids_reach_sampler_for_chat_history_and_add_request` against this PR branch would still be very helpful; please post the output here.

blairducrayoppat:
Ran it. Summary first: as the branch stands the test does not execute, and once it does, it shows
this PR removing a process crash rather than only a wrong echo.

**What I measured.** PR head `e6fced66` against its merge base `c9fff70b`, not against current
master. Master is now `26144e25`, four commits ahead; I read that compare and those commits touch
only `.github/`, `samples/` and `tests/python_tests/requirements.txt`, with no change under `src/`,
so the code paths below are byte-identical at master — but I did not run there. One of those four
commits moved the test pin from transformers 5.0.0 to 5.5.4 and optimum-intel `4f1a926` to
`dd4ed1a`; the environment I installed is the merge base's pin, not master's.

**Build.** Both refs built from source, identically (Ninja, Release, MSVC 19.44.35222,
`-DENABLE_PYTHON=ON -DENABLE_TESTS=ON`), against the OpenVINO nightly wheel
`2026.5.0.dev20260903`; `thirdparty/openvino_tokenizers` at `824033c3` in both trees, so the
compiled commit is the only difference between the two sides of every comparison below. Host:
Intel Core Ultra 7 258V, Arc 140V (driver 32.0.101.8991), Windows 11 26200, Python 3.11.9,
`tests/python_tests/requirements.txt` as pinned at `c9fff70b` (transformers 5.0.0, optimum-intel
`2.2.0.dev0+4f1a926`).

**1. In that environment the test never runs.** `MODEL_IDS[0]` resolves to
`tiny-random-phi3-vision`, and the export guard xfails it, so the fixture never builds a pipeline.
N=3 on the PR head, N=3 on the merge base, identical. From
`A_prhead_shipped_cpu_open_rep1_20260907T172432Z.log`, line 9 and then lines 27-28 (the intervening
lines are pytest's warnings summary):

```
test_vlm_pipeline.py::test_vlm_prompt_ids_reach_sampler_for_chat_history_and_add_request XFAIL [100%]
```
```
XFAIL test_vlm_pipeline.py::test_vlm_prompt_ids_reach_sampler_for_chat_history_and_add_request - AttributeError: 'DynamicCache' object has no attribute 'get_usable_length'. Ticket CVS-175110
======================= 1 xfailed, 3 warnings in 0.58s ========================
```

**2. Pointed at a model that does export, the first assertion raises.** With the fixture on
`tiny-random-gemma3` (N=3, both builds), contiguous from
`B_prhead_gemma3_cpu_rep1_20260907T172540Z.log`:

```
>       assert prompt in string_result.texts[0]
                         ^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'openvino_genai.py_openvino_genai.GenerationResult' object has no attribute 'texts'

test_vlm_pipeline.py:987: AttributeError
```

Both `generate` overloads the test calls are declared `-> py::typing::List<GenerationResult>`
(`src/python/py_continuous_batching_pipeline.cpp:609` and `:689`), and `texts` is on
`DecodedResults` / `VLMDecodedResults` (`py_openvino_genai.cpp:116`, `py_vlm_pipeline.cpp:309`), so
line 991 raises the same way. Reading `.m_generation_ids[0]` on both lines is what I ran below.

**3. With those two lines corrected, the test passes on this branch — and the merge base crashes.**
Same fixture change, same model, N=3 per cell:

| arm | build | device | outcome |
|---|---|---|---|
| corrected test | `e6fced66` | CPU | PASS 3/3 |
| corrected test | `c9fff70b` | CPU | pytest exits `3221225477` (0xC0000005) 3/3 |
| corrected test | `e6fced66` | GPU (Arc 140V) | PASS 3/3 |
| corrected test | `c9fff70b` | GPU (Arc 140V) | pytest exits `3221225477` 3/3 |

**4. The same thing outside pytest, sweeping `echo`.** One child process per call, so a crash is
attributed by which child dies. 108 units, N=3 per cell, `echo` at both levels, two models, two
devices. `generate(list[str])` is included as a comparison arm rather than an inert one: on a VLM
pipeline it delegates through `pipeline_base.cpp:69` into the same VLM string overload at `:313`
that this PR edits, so it is a path that happens to behave the same on both builds, not a path the
PR leaves alone.

| device | model | path | echo=false | echo=true (`c9fff70b`) | echo=true (`e6fced66`) |
|---|---|---|---|---|---|
| CPU | gemma3 | `string` | ok both builds | ok | ok |
| CPU | qwen2vl | `string` | ok both builds | ok | ok |
| GPU | gemma3 | `string` | ok both builds | ok | ok |
| CPU | gemma3 | `chat_history` | ok both builds | **crash** | ok |
| CPU | gemma3 | `add_request` | ok both builds | **crash** | ok |
| CPU | qwen2vl | `chat_history` | ok both builds | **crash** | ok |
| CPU | qwen2vl | `add_request` | ok both builds | **crash** | ok |
| GPU | gemma3 | `chat_history` | ok both builds | **crash** | ok |
| GPU | gemma3 | `add_request` | ok both builds | **crash** | ok |

Counts: on `c9fff70b`, 18/18 `echo=true` cells over `chat_history` and `add_request` crash, all with
exit `3221225477`, and 0/18 `echo=false` cells do; on `e6fced66` the same 36 cells crash 0/36; the
36 `string` cells complete on both builds. Bounding it: the same six configurations on a text-only
`ContinuousBatchingPipeline` (Qwen3-0.6B int8) complete on `c9fff70b` 18/18, so what I measured is
confined to the embeddings path this PR touches. Everything above was run twice, in two complete
passes; the second is what is reported. The passes agree on 108/108 exit codes in this sweep, and on
54/54 of the units in the pytest matrix that produce a comparable artifact — the other 6 are the
crashing runs, which write nothing by design, and they were the same 6 in both passes.

**5. C++ cases.** `VLMPromptIdsTest.*` 4/4 on Windows/MSVC as well, N=3.

**Not verified.**
- **No provably-zero control on the crashing configuration.** The repeated same-build arm bounds
  session drift, but it runs the test that xfails, so it never reaches the crash path; and the
  `string` arm is not an untouched path (above). The crash/no-crash split is a clean binary across
  36 cells, but I cannot put a resolution limit under it.
- Measured at `c9fff70b`, not at master `26144e25`, and at the transformers 5.0.0 pin, not 5.5.4.
- One OpenVINO build (`2026.5.0.dev20260903`) and one toolchain (MSVC 19.44, Release).
- Two tiny random VLMs, no full-size VLM; no image-bearing case, since the new test is text-only;
  NPU untouched.
- No GPU row for qwen2vl: that tiny model's `head_size` is 8 and the GPU PagedAttention kernel
  requires at least 16, so the configuration cannot be built. The gap is the model's, not the PR's.
- The environment is partly broken independently of this PR: running the neighbouring CB tests with
  the fixture on gemma3 fails on **both** builds, 3/3 each, with
  `Eltwise shape infer input shapes dim index: 1 mismatch` — the CVS-186059 condition your own test
  file records for gemma3 at `test_vlm_pipeline.py:176`. Identical on both sides, so it does not
  bear on the comparison, but it is in the runs.
- I have not localised the fault to a line: I have the exit code and a reproducer, not a stack.
- The `.texts` correction above is what I ran, not a review comment on how you should fix it.

I can attach the row-level table (204 units, one row per run, with build, driver, model, read-back
generation config, exit codes and log paths) if that is useful.

**What I checked before posting.** Docs:
`site/docs/use-cases/visual-processing/_sections/_run_model/index.mdx` documents `VLMPipeline` on
CPU and GPU ("Use CPU or GPU as devices without any other code change");
`ContinuousBatchingPipeline` constructed directly on a VLM directory is not in that device matrix.
Tests: `tests/python_tests/test_vlm_pipeline.py` has ten continuous-batching tests and every CB
fixture hardcodes `"CPU"` (fixtures at lines 683, 689, 695, 2895), so the GPU rows above are not a
configuration the suite covers today. Rules: `.github/CONTRIBUTING.md` and the AI Usage Policy it
pins at `c4f4325c`, both read in full. Duplicates: `gh issue list --state all --search` for
"GenerationResult texts" and "CVS-175110", and `gh pr list --state all --search
test_vlm_prompt_ids`, return nothing.

---

```text
AI assistance used: yes
If yes: Claude Code drafted this comment and the test harness behind it, built both refs, ran the
        sweeps and assembled the tables.
Human validation performed: I reviewed and approved this text before posting. I did not
        independently audit the C++ or the plugin internals. What I stand behind is that the runs
        described here were executed as stated on my hardware, and that the quoted output is
        reproduced from logs still on disk, verbatim, with the one elision marked.
```

PR #4397, opened 2026-09-01
[26.4] kokoro: Truncate phonemes segments that are > 510 in length (#4385)

RyanMetcalfeInt8:
## Description

This PR fixes some corner case issues, where exceptions are thrown because token_ids (from tokenized phonemes) exceed 512 (the max context length).

For Kokoro model, phonemized intermediate result is split into several chunks, where each chunk is very likely to have length <= 510.

Although, it is possible that there are corner cases where a size slightly larger is produced (e.g. 511). In python implementation of Kokoro, these cases are simply truncated:

https://github.com/hexgrad/kokoro/blob/dfb907a02bba8152ca444717ca5d78747ccb4bec/kokoro/pipeline.py#L390

```
...
                    elif len(ps) > 510:
                        logger.warning(f"Unexpected len(ps) == {len(ps)} > 510 and ps == '{ps}'")
                        ps = ps[:510]
...
```

This truncation is what was missing from our GenAI implementation. So, for corner cases where a phonemized chunk happen to overrun the 510 token limit, instead of truncating it, we hit this exception:
```
[2026-08-26T07:06:17.472Z] Kokoro tokenized length exceeds model context length: 513 > 512
```

Anyway, the fix is to introduce the same truncation logic for GenAI implementation.

 CVS-193390

## Checklist:
- [N/A]  Tests have been updated or added to cover the new code. 
- [X]  This PR fully addresses the ticket. 
- [N/A]  I have made corresponding changes to the documentation.

### DIFF
--- a/src/cpp/src/speech_generation/kokoro_tts_model.cpp
+++ b/src/cpp/src/speech_generation/kokoro_tts_model.cpp
@@ -999,16 +999,29 @@ Text2SpeechDecodedResults KokoroTTSImpl::synthesize_from_phoneme_chunks(
 
     const auto& vocab = m_runtime->vocab();
     const uint32_t context_length = m_runtime->context_length();
+    const size_t max_input_phoneme_length = std::min<size_t>(
+        generation_config.max_phoneme_length,
+        context_length > 2 ? static_cast<size_t>(context_length - 2) : static_cast<size_t>(0));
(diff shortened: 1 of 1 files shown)


sbalandi (approved):

Wovchena (approved):

MaximProshin:
@RyanMetcalfeInt8 , please rebase this PR as there were some CI issues fixed yesterday.

PR #4411, opened 2026-09-02
[SD Eagle3] fix kv cache mismatch of kv update model with main model

songbell:
This pull request enhances support for testing and configuring KV cache precision in the speculative decoding  for the EAGLE3 strategy. The main changes include improved handling of KV cache precision selection in the C++ implementation, and expanded Python test coverage to parameterize and validate behavior across different KV cache precisions (`f16`, `u8`, and unset). This ensures that the pipeline and plugin handle cache precision consistently and robustly.

**KV Cache Precision Handling Improvements:**

* The C++ implementation in `eagle3_strategy.cpp` now explicitly distinguishes between the runtime KV cache precision of the input ports and the property hint, ensuring the correct precision is used for model transformation and plugin configuration. It logs when the precision is unset and avoids mismatches between model properties and actual tensor types.

**Test Suite Enhancements:**

* Python tests in `test_continuous_batching.py` now parameterize over different KV cache precisions (`f16`, `u8`, and unset), allowing comprehensive validation of pipeline behavior for each setting. Helper functions were added to generate appropriate configuration dictionaries and test IDs. [[1]](diffhunk://#diff-46367752413258387bf181db2fc0a85c7d1f5ac3b62a743f0891c922139b4b54R604-R622) [[2]](diffhunk://#diff-46367752413258387bf181db2fc0a85c7d1f5ac3b62a743f0891c922139b4b54L611-R634)

## Description
Fix the regression of kv cache mismatch with runtime introduced by [openvinotoolkit/openvino.genai#4227](https://github.com/openvinotoolkit/openvino.genai/pull/4227).

The default kv precision for cpu is u8, gpu is u4
wwb data  (target model: [qwen3-8b/]  eagle3 model: [qwen3_8b_eagle3/]):
image

CVS-193994

One punch data covering GPU tests will be update once finished.

## Checklist:

### DIFF
--- a/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
+++ b/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
@@ -102,8 +102,8 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
     // Read the runtime KV cache precision from the compiled main model's key_cache input port: plugins may resolve the
     // actual cache precision (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and
     // the reorder model must match the precision of the tensors actually bound by the scheduler.
-    auto rt_kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
-    auto kv_cache_precision =
+    const auto rt_kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    const auto kv_cache_precision =
         m_main_pipeline->get_model_property(ov::hint::kv_cache_precision.name()).as<ov::element::Type>();
     // transformation for kv update model: u4 KV cache is stored as u8 internally,
     // so the reorder pass operates on u8 while the original precision is preserved in rt_info.
--- a/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
+++ b/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
@@ -91,26 +91,25 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
     ov::genai::ModelDesc kv_model_desc;
     kv_model_desc.model = kv_model;
     kv_model_desc.device = std::move(main_device);
+    // only kv cache information is needed for kv update model
+    if (main_model_desc.properties.count(ov::hint::kv_cache_precision.name()) > 0) {
+        kv_model_desc.properties[ov::hint::kv_cache_precision.name()] =
+            main_model_desc.properties.at(ov::hint::kv_cache_precision.name());
+    } else {
+        GENAI_INFO("kv cache precision not specified in main model properties. leave to plugin for default precision.");
+    }
 
-    // Read the KV cache precision from the compiled main model's key_cache input port rather than
-    // the ov::hint::kv_cache_precision property: plugins may resolve the actual cache precision
-    // (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and the
-    // reorder model must match the precision of the tensors actually bound by the scheduler.
-    auto kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    // Read the runtime KV cache precision from the compiled main model's key_cache input port: plugins may resolve the
+    // actual cache precision (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and
+    // the reorder model must match the precision of the tensors actually bound by the scheduler.
+    auto rt_kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    auto kv_cache_precision =
+        m_main_pipeline->get_model_property(ov::hint::kv_cache_precision.name()).as<ov::element::Type>();
     // transformation for kv update model: u4 KV cache is stored as u8 internally,
     // so the reorder pass operates on u8 while the original precision is preserved in rt_info.
     kv_model->set_rt_info(kv_cache_precision, "auxiliary_kv_cache_precision");
-    if (kv_cache_precision == ov::element::u4) {
-        kv_cache_precision = ov::element::u8;
-    }
-    ov::pass::PaKVReorderFusion(kv_cache_precision).run_on_model(kv_model);
-    // Pass the resolved (post u4->u8) precision explicitly instead of forwarding the raw
-    // ov::hint::kv_cache_precision value from main_model_desc.properties: for u4 caches that raw
-    // value would still say "u4" while PaKVReorderFusion above already rewrote the key_cache./
-    // value_cache. parameters to u8, and some plugins (e.g. GPU) give an explicitly user-set
-    // kv_cache_precision property priority over the auxiliary_kv_cache_precision rt_info, which
-    // would reintroduce a precision mismatch between the property and the actual parameter type.
-    kv_model_desc.properties[ov::hint::kv_cache_precision.name()] = kv_cache_precision;
+    ov::pass::PaKVReorderFusion(rt_kv_cache_precision).run_on_model(kv_model);
+
     m_kv_update_wrapper = std::make_shared<KVUpdateWrapper>(kv_model_desc);
 
     m_perf_metrics = ov::genai::SDPerModelsPerfMetrics();
--- a/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
+++ b/src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
@@ -91,17 +91,19 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
     ov::genai::ModelDesc kv_model_desc;
     kv_model_desc.model = kv_model;
     kv_model_desc.device = std::move(main_device);
-    // only kv cache information is needed for kv update model
+    // as of now, only kv cache information is needed for kv update model
     if (main_model_desc.properties.count(ov::hint::kv_cache_precision.name()) > 0) {
         kv_model_desc.properties[ov::hint::kv_cache_precision.name()] =
             main_model_desc.properties.at(ov::hint::kv_cache_precision.name());
     } else {
-        GENAI_INFO("kv cache precision not specified in main model properties. leave to plugin for default precision.");
+        GENAI_INFO("kv cache precision not specified in main model properties. Leave to the plugin for default precision.");
     }
 
     // Read the runtime KV cache precision from the compiled main model's key_cache input port: plugins may resolve the
-    // actual cache precision (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and
-    // the reorder model must match the precision of the tensors actually bound by the scheduler.
+    // actual cache precision independently of that hint (e.g. CPU promoting to bf16 when kv cache precision hint is
+    // fp16, GPU promoting to i8 when kv cache precision hint is u8, and for GPU, when the hint is u4, the kv data are
+    // packed as uint8 and be reinterpreted to u4 inside plugin), and the reorder model must match the precision of the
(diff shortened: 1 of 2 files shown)


src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
```diff
@@ -91,26 +91,25 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
...
-    // the ov::hint::kv_cache_precision property: plugins may resolve the actual cache precision
-    // (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and the
-    // reorder model must match the precision of the tensors actually bound by the scheduler.
-    auto kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    // Read the runtime KV cache precision from the compiled main model's key_cache input port: plugins may resolve the
+    // actual cache precision (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and
+    // the reorder model must match the precision of the tensors actually bound by the scheduler.
+    auto rt_kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    auto kv_cache_precision =
+        m_main_pipeline->get_model_property(ov::hint::kv_cache_precision.name()).as<ov::element::Type>();
```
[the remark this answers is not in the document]
songbell: you are wrong, main model_desc may not carry this property information

tests/python_tests/test_continuous_batching.py
```diff
@@ -600,6 +601,18 @@ def add(a, b):
     eagle_models_and_input[0],
 ]
 
+# u8 stores the KV cache quantized, which selects different attention kernels than f16.
+kv_cache_precisions = [ov.Type.f16, ov.Type.u8]
```
rkazants: why int4 is missed. Looks the fix is for int4, isn't?

rkazants (changes requested):
Please elaborate in PR description for which OV model the failure is reproduced? Is it int4 IR? And by default, what is the precision of KV-cache for int4 IR on CPU and GPU? Did you measure wwb accuracy after this fix and baseline without eagle3?

liubo-intel (approved):
have verified that this pr changes also can fix previous cpu "`ParameterMismatch` crash issue(CVS-191392)"

tests/python_tests/test_continuous_batching.py
songbell: can the test be run on GPU? for CPU, we don't use 4 bit kv cache due to accuracy issue ([CVS-142901](https://jira.devtools.intel.com/browse/CVS-142901)).

songbell: the fix is for GPU int4, and CPU fp16

peterchen-intel: GPU int4 KV cache is covered by weekly validation.

tests/python_tests/test_continuous_batching.py
```diff
@@ -600,6 +601,25 @@ def add(a, b):
     eagle_models_and_input[0],
 ]
 
+# u8 stores the KV cache quantized, which selects different attention kernels than f16.
+# None leaves the hint unset so the plugin resolves the cache type for the main and KV-update models.
+kv_cache_precisions = [ov.Type.f16, ov.Type.u8, None]
```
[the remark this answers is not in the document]
songbell: GPU tests will be covered in weekly

rkazants (commented):
I think we need to document this properly in GenAI user documentation,
I still don't understand - what shall I expect if I set some KV cache precision using hints and how shall I define which KV cache precision used in run-time or set by plugin?

We have other SD strategies (MTP, Dflash) for them this behavior be the same or it will be eagle3 specific stuff?

songbell:
> I think we need to document this properly in GenAI user documentation, I still don't understand - what shall I expect if I set some KV cache precision using hints and how shall I define which KV cache precision used in run-time or set by plugin?
> 
> We have other SD strategies (MTP, Dflash) for them this behavior be the same or it will be eagle3 specific stuff?

it's eagle3 specific, because MTP/DFLASH draft models only propose tokens linearly (draft token on position x, draft token on position x+1 ....), while eagle3 proposed draft tokens are in tree shape (draft token 1 on position x, draft token 2 on position x, draft token 3 on position x+1 ....), it has to update the kv cache base on the tree validation.

This has no change to the user interface, users still set kv cache precision using hints, plugin/runtime determine the kv cache precision used in runtime, and the eagle3 kv update model just need to make sure it is manipulating the kv cache in the correct precision.

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
```diff
@@ -91,26 +91,25 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
     ov::genai::ModelDesc kv_model_desc;
     kv_model_desc.model = kv_model;
     kv_model_desc.device = std::move(main_device);
+    // only kv cache information is needed for kv update model
+    if (main_model_desc.properties.count(ov::hint::kv_cache_precision.name()) > 0) {
+        kv_model_desc.properties[ov::hint::kv_cache_precision.name()] =
+            main_model_desc.properties.at(ov::hint::kv_cache_precision.name());
+    } else {
+        GENAI_INFO("kv cache precision not specified in main model properties. leave to plugin for default precision.");
```
Wovchena:
```suggestion
        GENAI_INFO("kv cache precision not specified in main model properties. Leave to the plugin for default precision.");
```

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
```diff
@@ -91,26 +91,25 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
...
+    }
 
-    // Read the KV cache precision from the compiled main model's key_cache input port rather than
-    // the ov::hint::kv_cache_precision property: plugins may resolve the actual cache precision
-    // (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and the
-    // reorder model must match the precision of the tensors actually bound by the scheduler.
-    auto kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    // Read the runtime KV cache precision from the compiled main model's key_cache input port: plugins may resolve the
+    // actual cache precision (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and
```
Wovchena: Can you be more specific what inference precision gets promoted to bf16 on CPU since this is an example?

Wovchena (commented):
Not blocking since the current state was approved in previous PR. But would be nice to fix

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
```diff
@@ -91,26 +91,26 @@ ContinuousBatchingPipeline::Eagle3DecodingImpl::Eagle3DecodingImpl(const ov::gen
...
-    // (e.g. CPU promoting to bf16 based on inference precision) independently of that hint, and the
-    // reorder model must match the precision of the tensors actually bound by the scheduler.
-    auto kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    // Read the runtime KV cache precision from the compiled main model's key_cache input port: plugins may resolve the
+    // actual cache precision (e.g. CPU promoting to bf16 when kv cache precision hint is fp16, GPU promoting to i8 when
+    // kv cache precision hint is u8) independently of that hint, and the reorder model must match the precision of the
+    // tensors actually bound by the runtime.
+    const auto rt_kv_cache_precision = m_main_pipeline->get_kv_cache_element_type();
+    const auto kv_cache_precision =
+        m_main_pipeline->get_model_property(ov::hint::kv_cache_precision.name()).as<ov::element::Type>();
```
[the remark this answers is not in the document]
songbell: auto will not be applied with eagle3

src/cpp/src/speculative_decoding/continuous_batching/eagle3_strategy.cpp
liubo-intel: On CPU, the promotion is: ov::hint::kv_cache_precision == f16 gets upgraded to bf16 whenever the CPU plugin's inference precision (ov::hint::inference_precision) resolves to bf16.:
https://github.com/openvinotoolkit/openvino/blob/11b9d3a355f29b5c103527da7c64554c908430f6/src/common/transformations/src/transformations/paged_attention/convert_pagedattn_inputs.cpp#L51-L54

ov::hint::inference_precision, the CPU plugin's compute/execution precision, unrelated to any KV-cache-specific hint) is resolved in :
Default path (no explicit inference_precision set by the user):
https://github.com/openvinotoolkit/openvino/blob/11b9d3a355f29b5c103527da7c64554c908430f6/src/plugins/intel_cpu/src/config.cpp#L510-L528
Explicit path (user sets ov::hint::inference_precision to bf16):
https://github.com/openvinotoolkit/openvino/blob/11b9d3a355f29b5c103527da7c64554c908430f6/src/plugins/intel_cpu/src/config.cpp#L248-L261

So concretely: on a Sapphire-Rapids-or-later Xeon, with default CPU properties (no explicit inference_precision), if a model/pipeline sets ov::hint::kv_cache_precision to f16, the KV cache tensors actually allocated/bound at runtime will be bf16, not f16 — because CPU attention kernels compute in bf16 there and format_cache_precision() re-writes the cache Parameter's element type to match. So for this issue on the CPU plugin, we don't need to care about the specific value of the kv_cache_precision precision hint — we just need the kv_cache_reorder_model's Cache Input precision to be read directly from the compiled Target Model's InferRequest, from the key_cache.* input port, to get the actual runtime precision.
This pr's changes also resolves the CPU issue from before: rt_kv_cache_precision is read via get_kv_cache_element_type() directly from the compiled main model's key_cache.*/value_cache.* input ports, so it already reflects the actual runtime precision after CPU's f16→bf16 promotion.

rkazants (dismissed):
In JIRA ticket, it is mentioned that this is a regression. What is a source of its regression? Guilty commit to GenAI?

Also, it is not transparent to user what KV-cache precision is really used when user explicitly set u4, u8, etc? Can you explicitly document it for this kv cache precision hint? Now this contract is hardcoded into the comment but it look like WA and not a productized solution.

I prefer to have transparent, long-term solution and avoid WAs (temporal solutions)

songbell:
> In JIRA ticket, it is mentioned that this is a regression. What is a source of its regression? Guilty commit to GenAI?

yes, the source of the regression is already in the desc of the PR.

> Also, it is not transparent to user what KV-cache precision is really used when user explicitly set u4, u8, etc? Can you explicitly document it for this kv cache precision hint? Now this contract is hardcoded into the comment but it look like WA and not a productized solution.
> 
> I prefer to have transparent, long-term solution and avoid WAs (temporal solutions)

No, it is a long-term solution and not a WA.  If user want to know the real kv cache precision it should be handled in a readable property exposed from underline plugin, e.g, ov::hint::kv_cache_precision, if it reports wrong or improper information, then the plugin should consider to fix it.
This PR resolves the issue of internal kv_update_wrapper model, which is composed as part of the eagle3 pipeline and invisible to users, for our case of u4 precision for GPU, while the tensors of kv cache dt_type is saved with 8bit. 
So, to make the kv update model work, it needs 2 types of kv precision information 1) the tensor port dt_type, as in my example, u8 2) the real data within the tensors , as in my example, u4,  because it needs to manipulate the kv cache data in a correct way. @rkazants , not sure if this makes sense to you now?

rkazants:
@songbell, please ask review from @as-suvorov to be aligned

as-suvorov (approved):

Wovchena (approved):

peterchen-intel:
Removed "Code Freeze" label from this PR since the back ported PR has the label. To avoid duplicate in the list.
https://github.com/openvinotoolkit/openvino.genai/pull/4428

PR #4420, opened 2026-09-03
[PoC] ASR-Streaming Support

RyanMetcalfeInt8:
## Description

This is an experimental / proof-of-concept branch for ASR streaming support. 

CVS-192300

Fixes #(issue)

## Checklist:


rkazants (commented):
I think we need changes in IR for Qwen3-ASR for proper streaming support or implement transformation to modify ov.Model. The current IR is not efficient for streaming scenario. I will check from IR perspective.
Also, I think that it requires continious batching support first so we can handle multi-user streaming.

RyanMetcalfeInt8:
> I think we need changes in IR for Qwen3-ASR for proper streaming support or implement transformation to modify ov.Model. The current IR is not efficient for streaming scenario. I will check from IR perspective. Also, I think that it requires continious batching support first so we can handle multi-user streaming.

Hi @rkazants -- Could you give more details when you get a chance -- mainly about "The current IR is not efficient for streaming scenario"? 

I understand the latter point about continuous batching / multi-user streaming.

src/python/py_asr_pipeline.cpp
```diff
@@ -368,7 +454,16 @@ void init_asr_pipeline(py::module_& m) {
             py::arg("streamer") = std::monostate(),
             "streamer",
             (asr_generate_docstring + std::string(" \n ") + asr_generation_config_docstring).c_str())
-
+        .def("create_streaming_session",
+             [](ASRPipeline& pipe,
+                const std::optional<ASRStreamingConfig>& streaming_config,
+                const std::optional<ASRGenerationConfig>& generation_config) {
+                 py::gil_scoped_release release;
```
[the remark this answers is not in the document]
byekrang: I think `ASRStreamingSession` keeps non-owning pointers into the pipeline implementation, but this binding does not keep the originating `ASRPipeline` alive. For example, `ASRPipeline(...).create_streaming_session()` can destroy the pipeline while the returned session is still alive, leaving a dangling pointer that is used by `push_chunk()/finish()`. Could we add `py::keep_alive<0, 1>()` here so the returned session keeps self alive?

src/cpp/src/whisper/whisper.cpp
```diff
@@ -398,6 +398,16 @@ WhisperGenerateResult whisper_generate(const ov::genai::WhisperGenerationConfig&
...
 
+        // Prefix tokens are forced as the first output tokens; appended after the SOT header so the
+        // decoder reproduces them verbatim before free generation begins.
+        if (!context_tokens.prefix.empty()) {
+            chunk_sot_tokens.insert(chunk_sot_tokens.end(),
+                                    context_tokens.prefix.begin(),
+                                    context_tokens.prefix.end());
+        }
+
+        SequenceGroup::Ptr sequence_group = std::make_shared<SequenceGroup>(0, chunk_sot_tokens, config);
```
byekrang, line 409: IIUC this `SequenceGroup` is unused: the following decode path consumes `chunk_sot_tokens` directly and never reads `sequence_group`. Since this is inside the regular Whisper chunk loop, it also adds an unnecessary allocation to every `generate()` call, not only streaming.

src/cpp/include/openvino/genai/automatic_speech_recognition/generation_config.hpp
```diff
@@ -117,6 +124,7 @@ static constexpr ov::Property<bool> word_timestamps{"word_timestamps"};
 static constexpr ov::Property<std::vector<std::pair<size_t, size_t>>> alignment_heads{"alignment_heads"};
 static constexpr ov::Property<std::string> initial_prompt{"initial_prompt"};
 static constexpr ov::Property<std::string> hotwords{"hotwords"};
+static constexpr ov::Property<std::string> prefix{"prefix"};
```
byekrang, line 127: I think the public `prefix` surface is only partially wired. `ASRGenerationConfig::update_generation_config` and `WhisperGenerationConfig::update_generation_config` don't read "prefix" from the AnyMap, so `generate(audio, ov::genai::prefix("x"))` is silently ignored, unlike the neighboring initial_`prompt`/`hotwords`/`context` properties. It's also not exposed on either generation config in Python, and the static/NPU path silently ignores it.

samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
```diff
@@ -47,8 +48,11 @@ int main(int argc, char* argv[]) try {
 
     // Pipeline expects normalized audio with Sample Rate of 16kHz
     ov::genai::RawSpeechInput raw_speech = utils::audio::read_wav(wav_file_path);
+    const auto wall_start = std::chrono::steady_clock::now();
     auto result = pipeline.generate(raw_speech, config);
-
+    const auto wall_sec = std::chrono::duration<float>(std::chrono::steady_clock::now() - wall_start).count();
```
byekrang, line 53: This wall-time output is currently breaking the `test_sample_automatic_speech_recognition` compares the Python and C++ sample stdout, and the CI diff is this additional `Total wall time: ...` line.

byekrang (changes requested):
Overall, I think the streaming direction looks reasonable. The session API, committed/partial text handling, prefix-relative delta calculation, and sliding-window bookkeeping form a coherent design, and I like that the model-specific behavior stays behind the streaming session implementation.

rkazants (changes requested):
Let us implement CB for Qwen3-ASR first that will lead to IR changes. Then on top of it, we will do streaming.

PR #4434, opened 2026-09-04
[Dependabot] Add transformers ignore version

avasenin-14:
## Description

Dependabot is opening PRs on security updates of transformers, despite transformers being in ignore list (we resolve them manually, as it is usually involves dependencies like optimum-intel and might be not trivial). Example of PR: https://github.com/openvinotoolkit/openvino.genai/pull/4433/.

The theory of why it is happening, though I can be wrong:

1. The repository rule has no versions field in .github/dependabot.yml. Security updates use this field to check ignore lists.

2. Dependabot then returns versions as [] for security alerts: https://github.com/dependabot/dependabot-core/blob/091038ad8f69ac6b2b60db5cf3aeede6cab2d9ba/common/lib/dependabot/config/ignore_condition.rb#L42. If it won't be the security update, it'd proceed to the next line and get his ">=0" string, which will block PR creation. 

3. It checks then if the upgraded version is in ignored versions, which it isn't. And checks then if it is == ALL_VERSIONS, which is a literal string ">=0", not comparison? I know it's weird, but I even run it locally to make sure:
```ruby
[].any?(">= 0")
# => false
```
(https://github.com/dependabot/dependabot-core/blob/091038ad8f69ac6b2b60db5cf3aeede6cab2d9ba/updater/lib/dependabot/job.rb#L554-L557),
(https://github.com/dependabot/dependabot-core/blob/091038ad8f69ac6b2b60db5cf3aeede6cab2d9ba/common/lib/dependabot/config/ignore_condition.rb#L16).

4. As it is not ignored, it then return allow_update true: https://github.com/dependabot/dependabot-core/blob/091038ad8f69ac6b2b60db5cf3aeede6cab2d9ba/updater/lib/dependabot/job.rb#L338-L370.

I'm adding the version then, it should block the automatic PRs creation, so we can continue upgrading transformers manually when we are ready for it. Tried to run dependabot gems locally and it seems like it helped.

## Checklist:

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -97,7 +97,7 @@ updates:
       # update transformers version is dependent on optimum-intel for now and required more precise changes in repo
       - dependency-name: "transformers"
         versions:
-          - ">=5.0"
+          - ">=5.6"
       # update required more precise changes in repo and made upon request
       - dependency-name: "optimum-intel"
     groups:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -96,6 +96,8 @@ updates:
       - dependency-name: "huggingface-hub"
       # update transformers version is dependent on optimum-intel for now and required more precise changes in repo
       - dependency-name: "transformers"
+        versions:
+          - ">=5.0"
       # update required more precise changes in repo and made upon request
       - dependency-name: "optimum-intel"
     groups:


.github/dependabot.yml
```diff
@@ -96,6 +96,8 @@ updates:
       - dependency-name: "huggingface-hub"
       # update transformers version is dependent on optimum-intel for now and required more precise changes in repo
       - dependency-name: "transformers"
+        versions:
+          - ">=5.6"
```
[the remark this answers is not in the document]
avasenin-14: I'm not sure which is better here.

as-suvorov, line 100: It's strange that this is required for transformers and not the rest of ignore list. So transformers should be a kind of special case. Let's check dependabot logs, maybe there is some clue there? Otherwise I'm ok with proposed solution

avasenin-14: Aye, I've found the line in code where security updates are treated differently, maybe that is the cause?
https://github.com/dependabot/dependabot-core/blob/091038ad8f69ac6b2b60db5cf3aeede6cab2d9ba/common/lib/dependabot/config/ignore_condition.rb#L42 (details see in PR description).

I don't see much useful info in logs, just checks for version and then update, they are not giving a lot of details there: https://github.com/openvinotoolkit/openvino.genai/actions/runs/33560726681/job/100032292904

as-suvorov (approved):

PR #4438, opened 2026-09-07
[Video] Add LTX2 Text2Video Support

goyaladitya05:
## Description
This PR adds text-to-video generation with audio for LTX-2 through the existing `Text2VideoPipeline` API. `Text2VideoPipeline` now dispatches on the model's `model_index.json` `_class_name` to select LTX-Video or LTX-2 internally, so existing LTX-Video users are unaffected.

### This PR introduces

- Internal `VideoPipeline` abstract base, `LTXPipeline` and the new `LTX2Pipeline` derive from it.
- LTX-2 follows the diffusers reference: x0-space classifier-free guidance per modality, constant mu shift, separate scheduler instances for video and audio since `step()` carries state.
- `VideoGenerationResult` carries `audio` and `audio_sample_rate` together (0 for video-only pipelines), following `Text2SpeechDecodedResults`.
- WhoWhatBenchmark loads text-to-video models through the diffusers/optimum auto-classes, so LTX-2 needs no dedicated handling. WWB scores video only; audio is not compared (at least for now).
- Changes to llm_bench, docs and samples will be in a seperate follow up PR.

### Export and Sample

```sh
optimum-cli export openvino --model Lightricks/LTX-2 --trust-remote-code --weight-format fp32 ltx2_fp32_ov
```

```
import os, shutil, subprocess, tempfile, time
import cv2, numpy as np, soundfile as sf
import openvino_genai

MODEL_DIR = r"C:\Users\devcloud\Desktop\genai\ltx2-fp32-ov"
OUT_DIR   = r"C:\Users\devcloud\Desktop\genai\outputs"
DEVICE, PRECISION = "CPU", "f32"

# LTX-2 defaults: 24 fps, cfg 4.0, 40 steps. num_frames must be 8k+1; H/W divisible by 32.
H, W, FRAMES, STEPS, CFG, FPS, SEED = 480, 704, 49, 30, 4.0, 24, 42

PROMPT = (
    "A golden retriever lies on a sunlit hardwood floor beside a tall window, then slowly raises its head "
    "and turns toward the camera, ears lifting, tail thumping once against the floor. Warm late-afternoon "
    "light streams across the room, dust motes drifting through the beam. The camera holds a steady "
    "medium close-up with a shallow depth of field, the background softly blurred. Photorealistic, "
    "natural fur detail, cinematic color grading."
)
NEG_PROMPT = (
    "static, motionless, frozen, still photograph, low quality, blurry, distorted, deformed limbs, "
    "extra paws, morphing, melting, choppy motion, watermark, text"
)

os.makedirs(OUT_DIR, exist_ok=True)
t0 = time.time()
log = lambda m: print(f"[{time.time()-t0:7.1f}s] {m}", flush=True)

def callback(step, num_steps, latent):
    a = np.array(latent.data, copy=False)
    log(f"  step {step+1:>3}/{num_steps}  latent std={a.std():.4f}")
    return False

log(f"{W}x{H} frames={FRAMES} steps={STEPS} cfg={CFG} fps={FPS} precision={PRECISION}")
pipe = openvino_genai.Text2VideoPipeline(MODEL_DIR, DEVICE, INFERENCE_PRECISION_HINT=PRECISION)
log("pipeline loaded")

r = pipe.generate(
    PROMPT, negative_prompt=NEG_PROMPT,
    height=H, width=W, num_frames=FRAMES, num_inference_steps=STEPS,
    guidance_scale=CFG, frame_rate=FPS,
    generator=openvino_genai.CppStdGenerator(SEED), callback=callback,
)
log("generate done")

v = np.array(r.video.data[0]); a = np.array(r.audio.data[0])
log(f"video {v.shape} mean={v.mean():.1f} std={v.std():.1f} | audio {a.shape} @ {r.audio_sample_rate} Hz max|x|={np.abs(a).max():.4f}")

with tempfile.TemporaryDirectory() as tmp:
    vf, af = os.path.join(tmp, "v.avi"), os.path.join(tmp, "a.wav")
    wr = cv2.VideoWriter(vf, cv2.VideoWriter_fourcc(*"MJPG"), FPS, (W, H))
    for f in v:
        wr.write(cv2.cvtColor(f, cv2.COLOR_RGB2BGR))
    wr.release()
    sf.write(af, a.T, samplerate=r.audio_sample_rate)
    out = os.path.join(OUT_DIR, f"ltx2_{W}x{H}_{FRAMES}f_{STEPS}s_bf16.mp4")
    if shutil.which("ffmpeg"):
        subprocess.run(["ffmpeg", "-y", "-i", vf, "-i", af, "-c:v", "libx264", "-pix_fmt", "yuv420p",
                        "-c:a", "aac", "-shortest", out], capture_output=True)
        log(f"saved {out}")
    else:
        shutil.copyfile(vf, out.replace(".mp4", ".avi")); shutil.copyfile(af, out.replace(".mp4", ".wav"))
        log("ffmpeg not found - saved separate .avi + .wav")

m = r.perf_metrics
log(f"load={m.get_load_time():.0f}ms generate={m.get_generate_duration():.0f}ms transformer_mean={m.get_transformer_infer_duration().mean:.0f}ms")

```

https://github.com/user-attachments/assets/0097af39-737d-4ac1-a302-8d19665279fc

### WWB similarity

LTX-Video (fp32, 10 samples, CPU)

| Mode | Optimum vs HF | GenAI vs HF | GenAI vs Optimum |
|---|---|---|---|
| text-to-video | 0.9868 | 0.9868 | 0.9999 |
| text-to-video + LoRA | NA | 0.9685 | NA |
| image-to-video | 0.981 | 0.983 | 0.986 |

LTX-2 (1 sample, fp32, CPU):

| Mode | Optimum vs HF | GenAI vs HF | GenAI vs Optimum |
|---|---|---|---|
| text-to-video | 0.977 | 0.984 | 0.986 |

## Checklist:
- N/A I have made corresponding changes to the documentation.  Will do them as a seperate follow up PR.


goyaladitya05:
current numbers for LTX-Video, no regressions!  (fp32, 10 samples, CPU)

| Comparison        | t2v    | i2v    |
|-------------------|--------|--------|
| Optimum vs HF     | 0.9870 | 0.9776 |
| GenAI vs HF       | 0.9872 | 0.9834 |
| GenAI vs Optimum  | 0.9999 | 0.9846 |


LTX2 Text2Video, fp32, 1 sample, CPU
HF vs GenAI 0.954

anatyrova:
@goyaladitya05 I see than you've got not very good numbers for ltx2 optimum/genai VS HF; it probably happens, because modality isolation for LTX2.0 is not supported in optimum-intel, and in 0.40.0 diffusers it is set with a default scale (for HF GT). please, when you measure wwb for ltx2 on >=0.40.0 diffusers, explicitly turn off modality isolation (video AND audio) by setting each of them to 1.0 for both HF GT and optimum-intel LTX2.0

this is a known limitation, that has to be hardcoded for LTX2.0 only. 

https://github.com/huggingface/optimum-intel/pull/1939/changes#diff-3108d4cfb605c9684b5b193d831fbe118eeef524a99d659577667438990c0b03:~:text=def%20_ltx2_modality_isolation_off(,%3A%201.0%7D

goyaladitya05:
> @goyaladitya05 I see than you've got not very good numbers for ltx2 optimum/genai VS HF; it probably happens, because modality isolation for LTX2.0 is not supported in optimum-intel, and in 0.40.0 diffusers it is set with a default scale (for HF GT). please, when you measure wwb for ltx2 on >=0.40.0 diffusers, explicitly turn off modality isolation (video AND audio) by setting each of them to 1.0 for both HF GT and optimum-intel LTX2.0

I used transformers==5.0.0 and diffusers=0.39.0, so that's probably not the issue. But the scores improved when I used transformers==4.57.6. I have updated the improved scores in the PR desc, and opened a seperate issue for this: huggingface/optimum-intel/issues/1998

goyaladitya05:
Even after the fixes done in OpenVINO for LTX-Video bf16 precision issue, LTX2 still produced low variance/bad quality outputs (bf16, CPU) , whereas LTX-Video works perfectly fine now. Opened a tracking issue: https://github.com/openvinotoolkit/openvino/issues/38137.

Also, fp16 on CPU is NaN for this model as well. We already have a tracking issue open for it- https://github.com/openvinotoolkit/openvino/issues/37039

bil-ash:
@goyaladitya05 Would be nice if you add image-to-video also for LTX2.

goyaladitya05:
> I used transformers==5.0.0 and diffusers=0.39.0, so that's probably not the issue. But the scores improved when I used transformers==4.57.6. I have updated the improved scores in the PR desc, and opened a seperate issue for this: huggingface/optimum-intel/issues/1998

The issue has been resolved now, it works fine with both transformers v4 and v5.

src/cpp/src/video_generation/video_generation_utils.hpp
```diff
@@ -0,0 +1,238 @@
...
+    size_t elems = 1;
+    for (auto d : t.get_shape())
+        elems *= d;
+
+    OPENVINO_ASSERT(elems == C, "latents_mean/std must contain exactly C elements (got ", elems, ", expected ", C, ")");
+
+    t.set_shape({1, C, 1, 1, 1});
+}
+
+inline void check_video_size(int64_t height, int64_t width, int64_t divisor) {
```
anatyrova, line 161: duplicate check_video_size logic: `ltx_pipeline.hpp:49` (`generate()`) was updated to use the new shared `video_generation_utils::check_video_size()`, but the sibling call at `ltx_pipeline.hpp:849` (`reshape()`) still has the old hand-rolled version of the same check. behavior matches today, but the two will silently drift if the shared logic changes later (e.g. new divisor, different error message) and only one call site gets updated. Worth consolidating `reshape()` onto the shared helper too

src/python/py_video_generation_models.cpp
```diff
@@ -250,7 +255,415 @@ void init_autoencoder_kl_ltx_video(py::module_& m) {
...
+             py::arg("config_path"))
+        .def_readonly("in_channels", &ov::genai::LTX2VideoTransformer3DModel::Config::in_channels)
+        .def_readonly("audio_in_channels", &ov::genai::LTX2VideoTransformer3DModel::Config::audio_in_channels)
+        .def_readonly("patch_size", &ov::genai::LTX2VideoTransformer3DModel::Config::patch_size)
+        .def_readonly("patch_size_t", &ov::genai::LTX2VideoTransformer3DModel::Config::patch_size_t)
+        .def_readonly("vae_scale_factors", &ov::genai::LTX2VideoTransformer3DModel::Config::vae_scale_factors)
+        .def_readonly("audio_scale_factor", &ov::genai::LTX2VideoTransformer3DModel::Config::audio_scale_factor)
+        .def_readonly("audio_sampling_rate", &ov::genai::LTX2VideoTransformer3DModel::Config::audio_sampling_rate)
+        .def_readonly("audio_hop_length", &ov::genai::LTX2VideoTransformer3DModel::Config::audio_hop_length);
+
```
anatyrova, line 311: causal_offset is one of four sibling fields (ltx2_video_transformer_3d_model.hpp:26-29) 

int64_t audio_scale_factor = 4;
int64_t causal_offset = 1;
int64_t audio_sampling_rate = 16000;
int64_t audio_hop_length = 160;

causal_offset Python binding is missing. please fix.

anatyrova (changes requested):
overall LGTM, but needs some changes

src/cpp/src/video_generation/video_generation_utils.hpp
goyaladitya05: done in [b5fe964](https://github.com/openvinotoolkit/openvino.genai/pull/4438/commits/b5fe964b71b43882bb6b401fe819a9c49c159175)

src/python/py_video_generation_models.cpp
goyaladitya05: fixed

anatyrova (approved):

tests/python_tests/test_video_generation.py
```diff
@@ -607,3 +624,162 @@ def test_lora_passthrough(self, video_generation_model):
         image = self._make_image()
         result = pipe.generate(image, "test prompt", **self.GENERATE_KWARGS, adapters=adapter_config)
         assert result.video is not None
+
+
+LTX2_GEN_KWARGS = dict(height=32, width=32, num_frames=9, num_inference_steps=2)
```
likholat: Could you please create a common `TestVideoGenerationPipelines` class for both LTX and LTX2, and specify the model for each test via test parameters, similar to how it is done for `ImageGeneration` here: https://github.com/openvinotoolkit/openvino.genai/blob/eaaf9a97638c3e9cf5817bfdec28d4df6f515c94/tests/python_tests/test_image_generation.py#L454

sbalandi (approved):
wwb lgtm

tools/who_what_benchmark/tests/test_cli_videos.py
sbalandi, line 1: shouldn't the classes also be updated for llm_bench https://github.com/openvinotoolkit/openvino.genai/blob/master/tools/llm_bench/llm_bench_utils/config_class.py#L68 ?

goyaladitya05: > Changes to llm_bench, docs and samples to be in a seperate follow up PR.

@sbalandi would it be fine if I include the llm_bench changes in a seperate PR, along with the docs and samples?

sbalandi: yes, it's fine

anatyrova:
build_jenkins

tools/who_what_benchmark/README.md
```diff
@@ -179,6 +179,7 @@ wwb --target-model ./qwen3-vl-embedding-2B --gt-data embed_test/gt.csv --model-t
 ### Compare Text-to-video models
 ```sh
 # Export model to OpenVINO, you can specify weight format with --weight-format option, for example --weight-format fp32/fp16/int8
+# Lightricks/LTX-2 is exported and evaluated the same way
```
likholat, line 182:
```suggestion
```

likholat, line 182: Please update docs: https://github.com/openvinotoolkit/openvino.genai/blob/master/site/docs/supported-models/_components/video-generation-models-table/models.ts

likholat (commented):
As far as I understand, our video generation samples don't currently support audio output (please correct me if I'm wrong). If that's the case, we need to update the samples as well

tools/who_what_benchmark/README.md
goyaladitya05: done

goyaladitya05:
@likholat I've included the docs and samples changes. @sbalandi I've included the llm_bench changes as well now, in this PR itself.
Please take a look, thanks!

PR #4460, opened 2026-09-10
Avoid setting PA_BACKEND as explicit for benchmark.py when none is provided

CuriousPanCake:
When benchmark.py is utilized without explicitly specifying backend like that:
`clear && python benchmark.py -d CPU -m models/LFM2.5-350M-fp16-ov` the PA backend is set as if explicit causing no possibility to fallback to the SDPA pipeline initialization as it would only fallback when no explicit backend is specified.

Fix this by removing from the arguments processing the explicit setting of `PA_ATTENTION_BACKEND` when none is provided.

### DIFF
--- a/tools/llm_bench/llm_bench_utils/model_utils.py
+++ b/tools/llm_bench/llm_bench_utils/model_utils.py
@@ -10,7 +10,6 @@ import numpy as np
 import urllib.request
 import logging as log
 from pathlib import Path
-from llm_bench_utils.config_class import PA_ATTENTION_BACKEND
 from llm_bench_utils.tts_utils import (
     SPEECHT5_SPEAKER_EMB_SHAPE,
     KOKORO_SPEAKER_EMB_SHAPE,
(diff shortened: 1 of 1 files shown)


CuriousPanCake:
The validation team needs to be notified to use `-lc "{\"ATTENTION_BACKEND\":\"PA\"}"` when testing the CB path to avoid silent fallbacks to SDPA.

avasenin-14:
@CuriousPanCake so GenAI tries PA backend first by default, and then falls back to SDPA if it's not possible, right? Just trying to understand how the stuff works outside of this PR :)
Also will you notify the validation team about the change?

CuriousPanCake:
Please, note that my PR concerns benchmark.py only. It does not affect pipelines in any way.

> @CuriousPanCake so GenAI tries PA backend first by default, and then falls back to SDPA if it's not possible, right? Just trying to understand how the stuff works outside of this PR :)
 
Yes, the pipeline logic is written that way, but when no backend is specified by the user, benchmark.py silently inserts PA_BACKEND as if the user has specified it. So we cannot really observe the fallback scenario when using benchmark.py because it's treated as an _explicit_ backend. I'm not insisting on my changes, this behavior needs to be analyzed and decided if we really want to see fallback happening for the benchmark tool.

avasenin-14:
Got it, thank you. I'll ask the other guys opinion first.
My thoughts are that for some other product fallbacks can be very convenient from the user perspective.
For benchmarks it may lead to some unexpected measurements, so sometimes to fail explicitly might be better...

avasenin-14:
Discussed it with Andrei and the guys. For now we've decided that fallback is a convenient functionality for the end users, but for this benchmark specifically we want to have explicit fails for now in case of potential measurement changes.

PR #4461, opened 2026-09-10
[SDL] Add numpy licenses Zlib AND CC0-1.0

avasenin-14:
## Description

Numpy has Zlib AND CC0-1.0 licenses (https://github.com/numpy/numpy/blob/dd88c0c19b54ad9ed3533224221285bf0873249a/pyproject.toml#L52), which are not in allowlist. This will fail CI for any bumps in numpy versions, example: https://github.com/openvinotoolkit/openvino.genai/actions/runs/34441155715/job/102756297481.
It doesn't fail CI without bumps, as it is scanning diffs only.

Found in dependabot bump originally (https://github.com/openvinotoolkit/openvino.genai/pull/4453) - I've added dependabot ignore rule there, but in the future we might want to upgrade numpy, and we probably won't migrate from it :) So I think it might be reasonable to add it's licenses.

## Checklist:
- N/A  Tests have been updated or added to cover the new code. 
- N/A  This PR fully addresses the ticket. 
- N/A  I have made corresponding changes to the documentation.

### DIFF
--- a/.github/dependency_review.yml
+++ b/.github/dependency_review.yml
@@ -16,6 +16,9 @@ allow-licenses:
   - "LGPL-2.0-or-later"
   - "LicenseRef-scancode-proprietary-license"
   - "LicenseRef-scancode-unicode"
+  # Numpy licenses
+  - "Zlib"
+  - "CC0-1.0"
 fail-on-scopes:
   - "runtime"
   - "development"


.github/dependency_review.yml
```diff
@@ -16,6 +16,9 @@ allow-licenses:
   - "LGPL-2.0-or-later"
   - "LicenseRef-scancode-proprietary-license"
   - "LicenseRef-scancode-unicode"
+  # Numpy licenses
+  - "Zlib"
+  - "CC0-1.0"
```
as-suvorov, line 21: Please double check if this are permissive licenses

avasenin-14: Yup, both are:
https://interoperable-europe.ec.europa.eu/licence/zliblibpng-license
https://interoperable-europe.ec.europa.eu/licence/cc0-10-universal-cc0-10-public-domain-dedication

as-suvorov (approved):

PR #4466, opened 2026-09-11
Add Qwen3-ASR forced aligner for word timestamps

byekrang:
## Description

Adds optional **word-level timestamps for Qwen3-ASR** using the separate Qwen3 Forced Aligner model. Whisper behavior is unchanged.

**Usage**

* Configure the aligner at pipeline construction:

  * C++: `ov::genai::ASRPipeline(asr_model_path, "CPU", ov::genai::forced_aligner(aligner_path))`
  * Python: `ov_genai.ASRPipeline(asr_model_path, "CPU", forced_aligner=aligner_path)`
* Request word timestamps per generation with `word_timestamps=true`.

**Semantics**

* The forced aligner is a separate model with its own feature extractor, tokenizer, encoder, and classifier; it does not reuse the base Qwen3-ASR model internals.
* Supplying an aligner alone does not enable word timestamps.
* Requesting `word_timestamps=true` without a configured aligner produces a validation error pointing to `ov::genai::forced_aligner(path)`.
* `forced_aligner` is rejected for non-Qwen ASR pipelines before backend construction.
* For long audio, the aligner returns chunk-local timestamps and `Qwen3ASR` applies the chunk offset exactly once, so returned word timestamps use full-audio time.

**Language support**

* Supported alignment languages: English, Chinese, Cantonese, German, Spanish, French, Italian, Portuguese, Russian.
* Japanese and Korean are supported by the upstream aligner model, but their segmentation is not currently implemented in OpenVINO GenAI.

**Config / model validation**

* `classify_num` is read from model config and checked against the decoder logits width rather than being hardcoded.
* Validation covers `classify_num`, `n_window`, `timestamp_token_id`, `timestamp_segment_time`, and the required decoder ports/model shape contract.

**Model export**

* The implementation targets `Qwen/Qwen3-ForcedAligner-0.6B` exported as a stateful ASR-with-past OpenVINO model.
* The export contract was verified with current Optimum-Intel.

```bash
optimum-cli export openvino \
    --model Qwen/Qwen3-ForcedAligner-0.6B \
    --task automatic-speech-recognition-with-past \
    --trust-remote-code \
    <output_dir>
```

The stateful export is required. The stateless `automatic-speech-recognition` export does not provide the `beam_idx`/state contract required by this implementation.

`Qwen/Qwen3-ForcedAligner-0.6B-hf` uses a different token-classification interface and is not supported by this PR.

**Samples & documentation**

* Updated the C++ and Python ASR samples to accept an explicit forced-aligner model.
* Updated speech-recognition documentation with usage, language support, model-export requirements, and long-audio timestamp behavior.

CVS-[191025](https://jira.devtools.intel.com/browse/CVS-191025)

## Checklist:

* [x] This PR follows [[GenAI Contributing guidelines](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing)](https://github.com/openvinotoolkit/openvino.genai?tab=contributing-ov-file#contributing).

  * C++ unit coverage includes timestamp repair, transcript segmentation / Unicode preprocessing, language normalization, and unsupported Japanese/Korean segmentation.
  * Python coverage includes constructor/error behavior, non-Qwen rejection, model/config validation, enabled/disabled alignment, language validation, real alignment, and long-audio global timestamp offsets.
  * Local real-model verification with a freshly built source tree passes **19/19 Python** forced-aligner tests and **13/13 C++** forced-aligner tests.
  * Real-aligner Python tests remain environment-gated in PR CI because a tiny forced-aligner CI model is not currently available.
* [ ] This PR fully addresses the ticket.

### DIFF
--- a/samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
+++ b/samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
@@ -11,13 +11,14 @@ auto get_config_for_cache() {
 }
 
 int main(int argc, char* argv[]) try {
-    if (argc < 3 || argc > 4) {
-        throw std::runtime_error(std::string{"Usage: "} + argv[0] + " <MODEL_DIR> \"<WAV_FILE_PATH>\" <DEVICE>");
+    if (argc < 3 || argc > 5) {
+        throw std::runtime_error(std::string{"Usage: "} + argv[0] +
+                                 " <MODEL_DIR> \"<WAV_FILE_PATH>\" [DEVICE] [FORCED_ALIGNER_DIR]");
     }
 
     std::filesystem::path models_path = argv[1];
     std::string wav_file_path = argv[2];
-    std::string device = (argc == 4) ? argv[3] : "CPU";  // Default to CPU if no device is provided
+    std::string device = (argc >= 4) ? argv[3] : "CPU";  // Default to CPU if no device is provided
 
     ov::AnyMap ov_config;
     if (device == "NPU" ||
@@ -27,23 +28,29 @@ int main(int argc, char* argv[]) try {
         ov_config = get_config_for_cache();
     }
 
-    // Word timestamps supported by Whisper models only
-    // Must be passed to ASRPipeline constructor as a property
+    // Word timestamps are supported by Whisper and by Qwen3-ASR for supported languages.
+    // For Whisper, word_timestamps must be passed to the ASRPipeline constructor.
     ov_config.insert(ov::genai::word_timestamps(true));
+
+    // Qwen3-ASR word timestamps require a separate forced-aligner model directory.
+    // The ASR model and forced aligner are loaded from independent model directories.
+    if (argc == 5) {
+        ov_config.insert(ov::genai::forced_aligner(argv[4]));
+    }
 
     ov::genai::ASRPipeline pipeline(models_path, device, ov_config);
 
     ov::genai::ASRGenerationConfig config = pipeline.get_generation_config();
 
-    // If language is known in advance it can be passed to the pipeline
-    // In the form of "<|en|>" for Whisper models. Supported by multilingual models only
-    // In the form of "English" for Qwen3-ASR models.
-    config.language = "<|en|>";
+    // If the language is known in advance, it can be set via config.language.
 
-    // Whisper models parameters. Ignored for Qwen3-ASR models
+    // Request word-level timestamps. Whisper also requires word_timestamps=true at construction (set above);
+    // Qwen3-ASR also requires a forced aligner.
+    config.word_timestamps = true;
+
+    // Whisper-only parameters, ignored for Qwen3-ASR models.
     config.task = "transcribe";  // Supported by multilingual models only
     config.return_timestamps = true;
-    config.word_timestamps = true;
 
     // Pipeline expects normalized audio with Sample Rate of 16kHz
     ov::genai::RawSpeechInput raw_speech = utils::audio::read_wav(wav_file_path);
--- a/site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
+++ b/site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
@@ -357,6 +357,55 @@ Whisper can predict timestamps for each word of speech, which provides more gran
     </TabItemJS>
 </LanguageTabs>
 
+For Qwen3-ASR word timestamps, OpenVINO GenAI uses the companion Qwen3 Forced Aligner model. Qwen3-ASR and the forced aligner are separate upstream models exported to two independent directories. The forced aligner is passed explicitly to `ASRPipeline` through the `forced_aligner` constructor property.
+
+When the property is supplied, `ASRPipeline` loads and compiles the forced aligner at construction using the same device and compile properties as the Qwen3-ASR pipeline. Word-level timestamps are then selected with the existing `word_timestamps` generation option. For long audio processed in multiple chunks, word timestamps are returned in full-audio (global) time.
+
+The forced aligner is optional. Qwen3-ASR transcription works normally without it; requesting `word_timestamps=true` without a configured forced aligner raises an error asking you to construct the pipeline with `ov::genai::forced_aligner(path)`.
+
+OpenVINO GenAI currently supports forced-alignment timestamps for Chinese, Cantonese, English, German, Spanish, French, Italian, Portuguese, and Russian. The Qwen3 Forced Aligner model also supports Japanese and Korean, but Japanese and Korean segmentation are not currently implemented in OpenVINO GenAI. The Unicode helpers used for preprocessing target reference-compatible segmentation of the supported languages. Forced-aligner word chunks currently have empty `token_ids`.
+
+Because Qwen3-ASR and its forced aligner are distributed separately, export them to two independent directories. Install `qwen-asr` before exporting:
+
+```sh
+pip install qwen-asr
+
+# Export Qwen3-ASR.
+optimum-cli export openvino --model Qwen/Qwen3-ASR-0.6B qwen3-asr-ov
+
+# Export its companion aligner to a separate directory.
+optimum-cli export openvino --model Qwen/Qwen3-ForcedAligner-0.6B qwen3-forced-aligner-ov
+```
+
+Pass the forced-aligner directory to `ASRPipeline` through the `forced_aligner` property:
+
+<LanguageTabs>
+    <TabItemPython>
+        ```python
+        pipe = openvino_genai.ASRPipeline(asr_model_path, "CPU", forced_aligner=forced_aligner_model_path)
+
+        result = pipe.generate(raw_speech, word_timestamps=True)
+
+        for word in result.words[0]:
+            print(f"[{word.start_ts:.2f}, {word.end_ts:.2f}]: {word.text}")
+        ```
+    </TabItemPython>
+    <TabItemCpp>
+        ```cpp
+        int main() {
+            ov::genai::ASRPipeline pipe(asr_model_path, "CPU", ov::genai::forced_aligner(forced_aligner_model_path));
+
+            auto result = pipe.generate(raw_speech, ov::genai::word_timestamps(true));
+
+            std::cout << std::fixed << std::setprecision(2);
+            for (const auto& word : (*result.words)[0]) {
+                std::cout << "[" << word.start_ts << ", " << word.end_ts << "]: " << word.text << "\n";
+            }
+        }
+        ```
+    </TabItemCpp>
+</LanguageTabs>
+
 #### Long-Form Audio Processing
 
 Whisper models are designed for audio segments up to 30 seconds in length.
--- a/src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
+++ b/src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
@@ -104,4 +104,12 @@ public:
     void set_generation_config(const ASRGenerationConfig& config);
 };
 
+/**
+ * @brief Path to a separate Qwen3 forced-aligner model used for word-level timestamps.
+ *
+ * Constructor-only property. When supplied for a Qwen3-ASR pipeline, the forced aligner is loaded eagerly
+ * and used when generation requests ov::genai::word_timestamps(true). Only supported by Qwen3-ASR models.
+ */
+static constexpr ov::Property<std::string> forced_aligner{"forced_aligner"};
+
 }  // namespace ov::genai
--- a/src/cpp/src/automatic_speech_recognition/models/qwen3-asr/pipeline.cpp
+++ b/src/cpp/src/automatic_speech_recognition/models/qwen3-asr/pipeline.cpp
@@ -5,6 +5,9 @@
 
 #include <algorithm>
 #include <chrono>
+#include <filesystem>
+#include <iterator>
+#include <optional>
 
 #include "audio_chunk.hpp"
 #include "streamer.hpp"
@@ -27,10 +30,19 @@ Qwen3ASR::Qwen3ASR(const std::filesystem::path& models_path, const std::string& 
     : ASRPipelineImplBase(models_path),
       m_feature_extractor{models_path / "preprocessor_config.json"},
       m_asr_text_token_id{get_required_token_id(m_tokenizer, "<asr_text>")} {
+    std::optional<std::string> forced_aligner_path;
+    if (const auto it = properties.find(ov::genai::forced_aligner.name()); it != properties.end()) {
+        forced_aligner_path = it->second.as<std::string>();
+    }
+
     ov::AnyMap properties_copy = properties;
     erase_allowed_asr_ctor_properties(properties_copy);
     m_encoder = std::make_unique<Qwen3ASREncoder>(models_path, device, properties_copy);
     m_decoder = std::make_unique<Qwen3ASRDecoder>(models_path, device, properties_copy);
+
+    if (forced_aligner_path.has_value()) {
+        m_forced_aligner = std::make_unique<Qwen3ForcedAligner>(*forced_aligner_path, device, properties_copy);
+    }
 
     // Qwen3-ASR EOS tokens: <|endoftext|>=151643, <|im_end|>=151645
     // The exported model has no generation_config.json. Qwen3-ASR original implementation hardcodes them as well.
@@ -45,6 +57,7 @@ ASRDecodedResults Qwen3ASR::generate(const AudioInputs& audio_inputs,
     auto start_time = std::chrono::steady_clock::now();
 
     const ASRGenerationConfig config = resolve_generation_config(generation_config);
+
     m_decoder->set_seed(config.rng_seed);
 
     ASRDecodedResults results;
@@ -59,19 +72,27 @@ ASRDecodedResults Qwen3ASR::generate(const AudioInputs& audio_inputs,
         },
         audio_inputs);
 
+    const size_t max_chunk_sec = config.word_timestamps ? MAX_TIMESTAMPED_ASR_INPUT_SECONDS : MAX_ASR_INPUT_SECONDS;
     const std::vector<AudioChunk> chunks =
-        split_audio_into_chunks({audio}, m_feature_extractor.sampling_rate, MAX_ASR_INPUT_SECONDS);
+        split_audio_into_chunks({audio}, m_feature_extractor.sampling_rate, max_chunk_sec);
 
     const auto infer_results = infer(chunks, config, results.perf_metrics, streamer);
     if (streamer) {
         streamer->end();
     }
 
-    auto [merged_texts, merged_languages] = merge_chunk_results(chunks, infer_results, config);
+    const std::vector<std::pair<std::string, std::string>> parsed_results =
+        parse_chunk_results(infer_results, config.language);
+
+    auto [merged_texts, merged_languages] = merge_chunk_results(chunks, parsed_results);
     results.languages = std::move(merged_languages);
     for (size_t i = 0; i < merged_texts.size(); ++i) {
         results.texts.push_back(merged_texts[i]);
         results.scores.push_back(0.0f);
+    }
+
+    if (config.word_timestamps) {
+        align_words(chunks, parsed_results, results);
     }
 
     auto stop_time = std::chrono::steady_clock::now();
@@ -132,10 +153,20 @@ std::pair<std::string, std::string> Qwen3ASR::parse_asr_output(const std::string
     return {language, text_part};
 }
 
+std::vector<std::pair<std::string, std::string>> Qwen3ASR::parse_chunk_results(
+    const std::vector<std::string>& infer_results,
+    const std::optional<std::string>& forced_language) {
+    std::vector<std::pair<std::string, std::string>> parsed_results;
+    parsed_results.reserve(infer_results.size());
+    for (const std::string& raw : infer_results) {
+        parsed_results.push_back(parse_asr_output(raw, forced_language));
+    }
+    return parsed_results;
+}
+
 std::pair<std::vector<std::string>, std::vector<std::string>> Qwen3ASR::merge_chunk_results(
     const std::vector<AudioChunk>& chunks,
-    const std::vector<std::string>& infer_results,
-    const ASRGenerationConfig& config) {
+    const std::vector<std::pair<std::string, std::string>>& parsed_results) {
     size_t num_samples = 0;
     for (const auto& chunk : chunks) {
         num_samples = std::max(num_samples, chunk.orig_batch + 1);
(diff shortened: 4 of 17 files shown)


src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner_encoder.cpp
```diff
@@ -0,0 +1,166 @@
...
+namespace {
+
+size_t read_n_window(const std::filesystem::path& models_path) {
+    size_t n_window = 100;
+    const std::filesystem::path config_path = models_path / "config.json";
+    if (std::filesystem::exists(config_path)) {
+        std::ifstream stream(config_path);
+        const nlohmann::json config = nlohmann::json::parse(stream);
+        if (config.contains("thinker_config") && config.at("thinker_config").contains("audio_config") &&
+            config.at("thinker_config").at("audio_config").contains("n_window")) {
```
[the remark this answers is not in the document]
byekrang: The fallback to 100 is intentional and mirrors the upstream `Qwen3ASRAudioEncoderConfig` default (`n_window`=100). The current forced-aligner checkpoint explicitly overrides it with `n_window`=50. When the field is present, GenAI validates that it is a positive integer before conversion to `size_t`; when absent, using 100 preserves the upstream config semantics. Deriving it from the encoder input shape is not reliable because the exported shape is dynamic. I’ll make the source of the default explicit in the code/comment.

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/pipeline.cpp
```diff
@@ -27,11 +30,20 @@ Qwen3ASR::Qwen3ASR(const std::filesystem::path& models_path, const std::string&
     : ASRPipelineImplBase(models_path),
       m_feature_extractor{models_path / "preprocessor_config.json"},
       m_asr_text_token_id{get_required_token_id(m_tokenizer, "<asr_text>")} {
+    std::optional<std::string> forced_aligner_path;
+    if (const auto it = properties.find(ov::genai::forced_aligner.name()); it != properties.end()) {
+        forced_aligner_path = it->second.as<std::string>();
+    }
```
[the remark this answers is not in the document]
byekrang: I checked this against the built Python module. `forced_aligner=pathlib.Path(...)` does not fail at `.as<std::string>();` it is converted successfully and reaches the forced-aligner filesystem validation. `ov::Any::as<std::string>()` handles the stored `std::filesystem::path` in this path, so the documented `os.PathLike | str` behavior is working.

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -1131,3 +1132,209 @@ def test_streamers(sample_from_dataset, pipelines_fixture, streamer_for_test):
...
+    model_id, pipeline_type = model_pipeline_pair[:2]
+    model_path = get_ov_cache_converted_models_dir() / model_id.split("/")[-1]
+    read_asr_model((model_id, model_path), pipeline_type=pipeline_type)
+    return model_path
+
+
+def forced_aligner_model_path():
+    path = os.environ.get("QWEN3_FORCED_ALIGNER_MODEL")
+    if not path or not pathlib.Path(path).exists():
+        pytest.skip("Set QWEN3_FORCED_ALIGNER_MODEL to a converted Qwen3-ForcedAligner directory.")
```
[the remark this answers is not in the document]
byekrang: Agreed. The positive-path integration tests currently require `QWEN3_FORCED_ALIGNER_MODEL`, so they are skipped in standard PR CI because there is no tiny forced-aligner test asset yet. The non-model-dependent constructor/config/error coverage does run in PR CI, and I verified the complete real-model subset locally against a stateful forced-aligner export (19/19 Python and 13/13 C++ tests pass). I plan to follow up with a tiny forced-aligner testing asset rather than add the full 0.6B model to regular PR CI.

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner_encoder.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    const ov::Tensor chunked_output = m_request.get_tensor("last_hidden_state");
+    ov::Tensor output = merge_chunked_encoder_output(chunked_output, remainder_frames);
+
+    m_request.set_tensor("input_features", ov::Tensor(ov::element::f32, {0, 0, 0}));
+
+    return output;
+}
+
+ov::Tensor Qwen3ForcedAlignerEncoder::chunk_mel_features(const WhisperFeatures& features) const {
```
as-suvorov: That's very similar to original Qwen3 asr encoder. Can we reuse the whole class or some functions?

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,531 @@
...
+const std::string AUDIO_PAD_TOKEN = "<|audio_pad|>";
+const std::string AUDIO_END_TOKEN = "<|audio_end|>";
+const std::string TIMESTAMP_TOKEN = "<timestamp>";
+
+struct Utf8Char {
+    uint32_t code;
+    std::string bytes;
+};
+
+std::vector<Utf8Char> decode_utf8(const std::string& text) {
```
as-suvorov: let's extract unicode/utf-8 helpers to a dedicated file under models/qwen3-asr, under unicode_helpers namespace.
Please also provide short docstrings with examples for functions

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,531 @@
...
+        } else {
+            segment.push_back(ch);
+        }
+    }
+    flush_segment();
+
+    return units;
+}
+
+std::vector<int64_t> fix_timestamp(const std::vector<int64_t>& data) {
```
as-suvorov, line 135: Please provide a docstring and link to original implementation

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,531 @@
...
+    OPENVINO_ASSERT(std::filesystem::exists(config_path),
+                    "Forced-aligner config.json not found at '", config_path.string(), "'.");
+    const char* required_files[] = {"openvino_encoder_model.xml",
+                                    "openvino_decoder_model.xml",
+                                    "openvino_tokenizer.xml"};
+    for (const char* required : required_files) {
+        OPENVINO_ASSERT(std::filesystem::exists(models_path / required),
+                        "Forced-aligner model is missing required file '", required, "' in '",
+                        models_path.string(), "'.");
+    }
```
as-suvorov:
```suggestion
```
No need to validate files existence before passing to ov. Let's rely on ov error handling

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,531 @@
...
+    return fixed;
+}
+
+Qwen3ForcedAligner::Qwen3ForcedAligner(const std::filesystem::path& models_path,
+                                       const std::string& device,
+                                       const ov::AnyMap& properties)
+    : m_feature_extractor{models_path / "preprocessor_config.json"},
+      m_tokenizer{models_path} {
+    const std::filesystem::path config_path = models_path / "config.json";
```
as-suvorov: Let's extract config handling into separate class. Check if we can reuse existing `Qwen3ASRConfig`.

as-suvorov (changes requested):

samples/cpp/automatic_speech_recognition/automatic_speech_recognition.cpp
```diff
@@ -27,23 +28,29 @@ int main(int argc, char* argv[]) try {
...
+
     ov::genai::ASRPipeline pipeline(models_path, device, ov_config);
 
     ov::genai::ASRGenerationConfig config = pipeline.get_generation_config();
 
-    // If language is known in advance it can be passed to the pipeline
-    // In the form of "<|en|>" for Whisper models. Supported by multilingual models only
-    // In the form of "English" for Qwen3-ASR models.
-    config.language = "<|en|>";
+    // If the language is known in advance, it can be set via config.language.
```
as-suvorov, line 45: Why language setting removed?

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
```diff
@@ -357,6 +357,94 @@ Whisper can predict timestamps for each word of speech, which provides more gran
...
+
+The forced aligner is optional. Qwen3-ASR transcription works normally without it; requesting `word_timestamps=true` without a configured forced aligner raises an error asking you to construct the pipeline with `ov::genai::forced_aligner(path)`.
+
+OpenVINO GenAI currently supports forced-alignment timestamps for Chinese, Cantonese, English, German, Spanish, French, Italian, Portuguese, and Russian. The Qwen3 Forced Aligner model also supports Japanese and Korean, but Japanese and Korean segmentation are not currently implemented in OpenVINO GenAI. The Unicode helpers used for preprocessing target reference-compatible segmentation of the supported languages. Forced-aligner word chunks currently have empty `token_ids`.
+
+Because Qwen3-ASR and its forced aligner are distributed separately, export them to two independent directories. Install `qwen-asr` before exporting. Both models must be exported as stateful `automatic-speech-recognition-with-past` models: the GenAI Qwen3-ASR and forced-aligner decoders require the stateful decoder with the `beam_idx` interface, which the default stateless export does not provide.
```
as-suvorov, line 368: Let's make code snippets self-documented

```suggestion
For Qwen3-ASR word timestamps, OpenVINO GenAI uses the companion Qwen3 Forced Aligner model. 

OpenVINO GenAI currently supports forced-alignment timestamps for Chinese, Cantonese, English, German, Spanish, French, Italian, Portuguese, and Russian. The Qwen3 Forced Aligner model also supports Japanese and Korean, but Japanese and Korean segmentation are not currently implemented in OpenVINO GenAI.
```

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
```diff
@@ -357,6 +357,94 @@ Whisper can predict timestamps for each word of speech, which provides more gran
...
+    --trust-remote-code \
+    qwen3-forced-aligner-ov
+```
+
+Pass the forced-aligner directory to `ASRPipeline` through the `forced_aligner` property:
+
+<LanguageTabs>
+    <TabItemPython>
+        ```python
+        pipe = openvino_genai.ASRPipeline(asr_model_path, "CPU", forced_aligner=forced_aligner_model_path)
```
as-suvorov, line 393: Do we support both - aligner instance and path property?

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
```diff
@@ -357,6 +357,94 @@ Whisper can predict timestamps for each word of speech, which provides more gran
...
+                std::cout << "[" << word.start_ts << ", " << word.end_ts << "]: " << word.text << "\n";
+            }
+        }
+        ```
+    </TabItemCpp>
+</LanguageTabs>
+
+The forced aligner is also available as a standalone `ASRForcedAligner` class, usable independently of `ASRPipeline`. It aligns a transcript you already have to its audio segment, so it can add word-level timestamps to transcripts produced elsewhere (for example by a separate transcription run). The transcript must correspond to the supplied audio segment, and returned timestamps are relative to the start of that audio.
+
+The API is currently scalar (one audio/transcript per `align` call). The same language support applies: English, Chinese, Cantonese, German, Spanish, French, Italian, Portuguese, and Russian; Japanese and Korean are present in the model configuration but are not currently implemented by OpenVINO GenAI.
```
as-suvorov, line 419:
```suggestion
The forced aligner is also available as a standalone `ASRForcedAligner` class, usable independently of `ASRPipeline`.
```

src/cpp/include/openvino/genai/automatic_speech_recognition/forced_aligner.hpp
```diff
@@ -0,0 +1,64 @@
...
+     * @brief Aligns a transcript to audio and returns word-level chunks with timestamps relative to the audio start.
+     * @param audio Raw mono float PCM at the model's sampling rate.
+     * @param transcript Transcript text to align to the audio.
+     * @param language Language of the transcript. Optional at the generic API boundary;
+     *                 the current Qwen3 forced-aligner implementation requires it.
+     * @return Word-level chunks with timestamps in seconds relative to the audio start.
+     */
+    std::vector<ASRDecodedResultChunk> align(const std::vector<float>& audio,
+                                             const std::string& transcript,
+                                             std::optional<std::string> language = std::nullopt);
```
as-suvorov, line 57: Let's reflect in API that language is optional at the generic API boundary.
We also need overload with `Properties&&... properties)`, so we can pass `ov::genai::language("English")`.

```suggestion
    std::vector<ASRDecodedResultChunk> align(const std::vector<float>& audio,
                                             const std::string& transcript,
                                             const ov::AnyMap properties = {});
```

src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
```diff
@@ -104,4 +104,12 @@ class OPENVINO_GENAI_EXPORTS ASRPipeline {
     void set_generation_config(const ASRGenerationConfig& config);
 };
 
+/**
+ * @brief Path to the Qwen3 forced-aligner model used for word-level timestamps.
+ *
+ * Constructor-only property. The model is loaded eagerly and used when word-level timestamps are requested.
+ * Supported only for Qwen3-ASR.
+ */
+static constexpr ov::Property<std::filesystem::path> forced_aligner{"forced_aligner"};
```
as-suvorov, line 113:
```suggestion
static constexpr ov::Property<ASRForcedAligner> forced_aligner{"forced_aligner"};
```

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/forced_aligner_config.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <fstream>
+#include <nlohmann/json.hpp>
+
+#include "openvino/core/except.hpp"
+
+namespace ov {
+namespace genai {
+
+Qwen3ForcedAlignerConfig::Qwen3ForcedAlignerConfig(const std::filesystem::path& json_path) {
```
as-suvorov, line 14: Let's use `read_json_param` from utils. Let's also extract config validation to dedicated `.validate` function as we do for other configs.

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.hpp
```diff
@@ -0,0 +1,57 @@
...
+    std::string build_marker_input(const std::vector<std::string>& units, size_t audio_frames) const;
+
+    WhisperFeatureExtractor m_feature_extractor;
+    Tokenizer m_tokenizer;
+    std::unique_ptr<Qwen3ASREncoder> m_encoder;
+    ov::InferRequest m_decoder;
+
+    int64_t m_timestamp_token_id;
+    float m_timestamp_segment_ms;
+    std::set<std::string> m_supported_languages;
```
as-suvorov, line 54: Let's use whole config instead of multile members
```suggestion
    ForcedAlignerConfig m_config
```

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.hpp
```diff
@@ -0,0 +1,57 @@
...
+namespace ov::genai {
+
+// Normalizes a language name or code to the lowercase name used by the forced aligner.
+std::string normalize_alignment_language(const std::string& language);
+
+// Segments transcript text into the units expected by the Qwen forced aligner.
+std::vector<std::string> segment_alignment_units(const std::string& text, const std::string& canonical_language);
+
+// Repairs non-monotonic timestamp predictions using the Qwen forced-aligner reference algorithm.
+std::vector<int64_t> fix_timestamp(const std::vector<int64_t>& data);
```
as-suvorov, line 30: Do we use this functions outside of `qwen3_forced_aligner.cpp`? If not, please move them into anonymous namespace in `.cpp`

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,375 @@
...
+    };
+    OPENVINO_ASSERT(has_input("encoder_hidden_states") && has_input("input_ids") && has_input("beam_idx"),
+                    "Forced-aligner decoder must expose inputs encoder_hidden_states, input_ids and "
+                    "beam_idx.");
+
+    const ov::PartialShape logits_shape = compiled_decoder.output("logits").get_partial_shape();
+    OPENVINO_ASSERT(logits_shape.rank().is_static() && logits_shape.size() == 3 &&
+                        logits_shape[2].is_static() &&
+                        static_cast<size_t>(logits_shape[2].get_length()) == classify_num,
+                    "Forced-aligner decoder logits last dimension must equal classify_num=", classify_num, ".");
```
as-suvorov, line 274: I believe almost always inputs will be fine and probably we can rely on `.set_tensor` validation. I guess the same applicable for output validation. Could you please check if we perform such checks in ctor somewhere else. Then we can do the same here, but we will need to move `m_decoder = compiled_decoder.create_infer_request();` after compiled model validation. And let's extract validation into a separate method. Otherwise let's skip this checks.

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,375 @@
...
+    std::string result = value;
+    std::transform(result.begin(), result.end(), result.begin(), [](unsigned char c) {
+        return static_cast<char>(std::tolower(c));
+    });
+    return result;
+}
+
+}  // namespace
+
+std::string normalize_alignment_language(const std::string& language) {
```
as-suvorov, line 60: I think we need to work on error handling here.
Do we have a use case for empty language? If not let's assert if it's not empty after trimming inside this function.
Next thing is return value. Do we have a use case where empty normalized language is used? If not let's also throw if no normalized language found.
Please also update callers code after asserts implemented, some of the check will be duplicated

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
```diff
@@ -0,0 +1,375 @@
...
+    OPENVINO_ASSERT(logits_shape.size() == 3 && logits_shape[0] == 1 && logits_shape[1] == input_len,
+                    "Forced-aligner decoder logits must be shaped [1, input_len, classify_num].");
+    const size_t class_count = logits_shape.at(2);
+    const float* logits_data = logits.data<const float>();
+
+    std::vector<int64_t> timestamp_ms;
+    timestamp_ms.reserve(timestamp_positions.size());
+    for (const size_t position : timestamp_positions) {
+        const float* row = logits_data + position * class_count;
+        const size_t class_id = static_cast<size_t>(std::max_element(row, row + class_count) - row);
```
as-suvorov, line 355: It looks like we can add model postprocessing to compute argmax of `classify_num`. Basically changing output `[batch_size, seq_len, num_classes] -> [batch_size, seq_len, timestamp_class]`.
It's not mandatory for this PR.

tests/cpp/test_asr_forced_aligner.cpp
```diff
@@ -0,0 +1,46 @@
+// Copyright (C) 2026 Intel Corporation
```
as-suvorov, line 1: I think we can remove this tests. They are e2e and preferably to be tested with python API

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -1131,3 +1134,292 @@ def test_streamers(sample_from_dataset, pipelines_fixture, streamer_for_test):
...
+    model_path = get_ov_cache_converted_models_dir() / model_id.split("/")[-1]
+    read_asr_model((model_id, model_path), pipeline_type=pipeline_type)
+    return model_path
+
+
+@functools.lru_cache()
+def forced_aligner_model_path():
+    override = os.environ.get("QWEN3_FORCED_ALIGNER_MODEL")
+    if override and pathlib.Path(override).exists():
+        return override
```
as-suvorov, line 1160: Why it's needed?

tests/python_tests/test_whisper_pipeline.py
```diff
@@ -1131,3 +1134,292 @@ def test_streamers(sample_from_dataset, pipelines_fixture, streamer_for_test):
...
+def forced_aligner_audio():
+    rng = np.random.default_rng(0)
+    return (rng.standard_normal(16000) * 0.01).astype(np.float32).tolist()
+
+
+def prepare_qwen3_asr_model(model_pipeline_pair):
+    model_id, pipeline_type = model_pipeline_pair[:2]
+    model_path = get_ov_cache_converted_models_dir() / model_id.split("/")[-1]
+    read_asr_model((model_id, model_path), pipeline_type=pipeline_type)
+    return model_path
```
as-suvorov, line 1153: Let's reuse existing approaches with pipeline instantiation. It's better to reuse pipeline across tests

tests/python_tests/test_whisper_pipeline.py
as-suvorov, line 1: Let's move forced aligner tests into dedicated test file.
This dedicated test file will only include test for public `ForcedAligner` API. There are a lot of tests for bad configuration, missing files, etc. Please move them to forced aligner test file. I'm not sure if we need to test every assert, if you think so, please prefer test parametrization instead of multiple tests.
In the current file let's test integration to `ASRPipeline` only. Let's add GenAI vs optimum-intel test for word timestamps

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
byekrang: right!

site/docs/use-cases/speech-recognition/_sections/_usage_options/index.mdx
byekrang: No. In the pending update, `forced_aligner` accepts only a pre-constructed `ASRForcedAligner` instance; the path-based property is removed. The local documentation has also been updated to show only the instance form. This change hasn’t been pushed yet.

src/cpp/include/openvino/genai/automatic_speech_recognition/pipeline.hpp
byekrang: Agreed on passing a constructed aligner instead of a path. In the pending revision I used `ov::Property<std::shared_ptr<ASRForcedAligner>>`. `ASRForcedAligner` is currently non-copyable/non-movable, so storing it directly in `ov::Any` as `ov::Property<ASRForcedAligner>` does not compile. The `shared_ptr` form also matches the existing pattern for injected runtime objects and keeps the same aligner instance alive for the pipeline. If you intended `ASRForcedAligner` itself to become a copyable handle so the property can be by value, I can adjust the ownership model accordingly.

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/forced_aligner_config.cpp
byekrang: Sure thing!

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
byekrang: Thank you!
Checked the sibling backends and they rely on OpenVINO for required-port validation, so I removed the manual input-port checks. I kept only the logits width vs `classify_num` check because that is a model/config semantic contract OpenVINO cannot validate. I also moved that check before `create_infer_request()`.

src/cpp/src/automatic_speech_recognition/models/qwen3-asr/qwen3_forced_aligner.cpp
byekrang: Good point! I took a closer look at this. Both the real and tiny forced-aligner exports currently return raw `f32` logits `[?, ?, 5000]`, and upstream Qwen also performs `argmax(dim=-1)` in the Python driver rather than inside the model.

Since I couldn't find a postprocessed export or an established output contract for that case today, would it be okay to keep the current raw-logits path for this PR? If exporter-side argmax is added later, we can add support for that contract then.

tests/python_tests/test_whisper_pipeline.py
byekrang: You're right, it doesn't seem necessary. The env var was only a local developer override to run the tests against a pre-converted real forced-aligner model; CI already uses the standard tiny-model cache/download path, and no test depends on the override

PR #4469, opened 2026-09-14
Support two outputs of PaKVReorder

mangguo321:
## Description

Fixes an EAGLE3 continuous-batching accuracy regression when CPU uses an F16 KV cache precision with F32 inference precision.

The EAGLE3 KV update helper previously concatenated the key and value `ScatterUpdate` outputs into one result. The paired OpenVINO `PaKVReorder` change now exposes key and value cache tensors as separate outputs. Update the GenAI helper and request binding to match that contract:

- Replace the per-layer `Concat(key_scatter, value_scatter) -> Result` with separate key and value `Result` nodes.
- Bind each helper output to its corresponding persistent KV cache tensor with `set_output_tensor()` before inference.

This makes K/V cache writeback explicit at the infer-request boundary. In particular, it prevents F16 external cache storage from being disconnected from the CPU-internal F32 buffer introduced by precision conversion before `PaKVReorder`.

This PR must be integrated together with the paired OpenVINO change that adds separate K/V outputs to `PaKVReorder`.
OpenVINO change: https://github.com/openvinotoolkit/openvino/pull/38079

CVS-191948

Fixes #(issue)

## Checklist:


rkazants (commented):
tests?

